PR #4, opened 2020-05-08
Add various low-level flash functions

bsmiles32:
This is an updated version of defunct PR https://github.com/n64decomp/majora/pull/7

This PR extracts flow-level flashram functions and provides a compilable C version of them. No attempt to make the generated assembly matching original code has been made so the file is full of NON_MATCHING macro.


Kenix3:
PR approval pending testing that the functions are functionally equivalent.

Our workflow is to only use the GLOBAL_ASM macro when the function is otherwise equivalent.

Kenix3:
PR on hold until someone can attempt matching on these functions.

PR #7, opened 2020-08-04
Cleanup some function identifications.

Rozelette:
Adds some functions that were missed during automatic function finding because they were unused. Removes some other "functions" because they were misidentified switch cases or the tail end of another function that Ghidra thought was a no-return.


Kenix3:
Because this PR changes the functions that are identified in the code it requires `make disasm` to be run. In the furue we should flag any PRs that require this build step.
We should also brainstorm some ideas such as `make` automatically calling `make diasasm` that would cause newer contributors to not run into errors when building newer commits of the code.

Kenix3:

PR #8, opened 2020-08-05
Add actor and scene enums

Rozelette:


Kenix3:
Looks good to me. Build is OK.

PR #10, opened 2020-08-08
Make the disasm target when appropiate instead of needing to do it manually

Rozelette:
Implemented by making disasm an intermediate target and having all assembly files need it as a prerequisite. Another option I considered was grouped explicit targets, a feature that allows one rule to generate multiple targets. However, this is a relatively new feature added in 4.3. At least with my Ubuntu installation, the package manager only has 4.2.1, so I did not think it was safe to assume that users would have this version of make.

A caveat of this change is that it also add disasm to most .c files, since some have non-matchings that depend on the split .asm to compile. Compilation could be speed up by only the .c files that need it to depend on disasm so that the others can be compiled while disassembling is occurring. Since disassembling does take all that long however, I choose not to do it at this time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,9 @@ DMADATA_FILES := $(DMADATA_FILES:build/decomp/ovl_%=)
 
 SRC_DIRS := $(shell find src -type d)
 
-S_FILES := $(wildcard asm/*)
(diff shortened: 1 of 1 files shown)


Kenix3:
Looks good to me.

PR #11, opened 2020-08-28
Decompile z_snap

Rozelette:
This file deals with some of the pictobox behavior. I'm calling it z_snap as I've seen mentions that that was what the pictobox was called internally, and the name fits alphabetically with the files around it.


Kenix3:
Looks good to me. Build is OK.

PR #12, opened 2020-08-28
Delete function type information

Rozelette:
This PR also removes the ability for the disassembler to output C headers. I choose to do this as removing the function type information breaks this feature, and it wasn't used in the first place.


Kenix3:
Looks good to me.

PR #13, opened 2020-08-31
Split headers to be more similar to OOT

Rozelette:
The main focus of this PR is to create headers similar to the ones already present in OOT and move our existing types into those headers. It does not try to get those headers to be an exact copy, so structs remain untouched for the most part and were simply moved from structs.h into their respective z64*.h. In addition, functions.h and variables.h were split into headers containing all the core code functions/variables, and headers for the ones found in overlays.

Future work will include copying some additional headers from OOT that include utilities such as macros and reorganizing and renaming structs to exactly match OOT where possible.


Kenix3:
Looks good to me.

PR #14, opened 2020-09-03
Bootstrap actors

Rozelette:
This PR creates a simple header and a .c file that includes all functions with GLOBAL_ASM for each actor, copying exactly what OOT has for them. Also restructures the files and changes some types to be closer to OOT. The definitions in functions_overlay.h and variables_overlay.h don't matter, so later we should consider just deleting them for actors. I included the script I used to automate most of this in tools, in case we need to do it again.


Kenix3:
Please rebase your `actor-setup` branch on `upstream/master`

Kenix3:
Looks good to me!

PR #15, opened 2020-09-06
Makefile rework. Add init and uncompressed targets.

Rozelette:
Changes:
- Reworked to be more similar to OOT's in features and organization
- Add `COMPARE` var
- Add `NON_MATCHING` var (note, none-matching builds currently do not compile)
- Add groundwork for using static recomp IDO
- Add target to make an uncompressed ROM (the intent for this is to use it in tools that parse information about the ROM, such as gfxdis, that don't work with a compressed ROM. It also builds much faster.)
- Add init target to do everything a newcomer would need to do after cloning the repo
- Add setup target to do first-time setup actions (note, this is simply for the init target to guarantee this all happens first)
- Add diff-init target to set up the ability to diff
- Add diff.sh to be a quicker way to diff a function (we discussed adding a make target, but make is not designed to pass arguments to targets, so a script was a better solution)


Kenix3:
Looks good to me. I like the uncompressed support that wasn't in our roadmap. Good work!

PR #17, opened 2020-09-15
Add some OOT tools

Rozelette:
Adds sym_info.py, first_diff.py, and the C tools: yaz0, makeromfs, elf2rom, mkldscript, and vtxdis. Most C tools are not used yet because they assume the rom is built from a spec file. Still TODO are to port over the overlay tools


Kenix3:
Looks good to me!

PR #18, opened 2020-09-15
Fix functions with switches to work in mips2c

Rozelette:
This PR encompasses several changes to make using mips2c possible on functions with switches. Changes:
- Rename jump tables to start with "jtbl_"
- Change generated label names
- Add .section declarations to data and rodata files
- Add splits for most overlay .rodata sections


m4xw (commented):
Review passes assuming we don't want them via files.txt

PR #27, opened 2020-09-23
en_tanron6 OK

fullgrowngaming:


Rozelette (approved):
LGTM

src/overlays/actors/ovl_En_Tanron6/z_en_tanron6.c
```diff
@@ -21,16 +23,39 @@ const ActorInit En_Tanron6_InitVars = {
...
+    this->actor.unkA0.damageChart = &D_80BE6170;
+    this->actor.unk1F = 6;
+    EnTanron6_SetupDoNothing(this);
+}
+
+void EnTanron6_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
 
-GLOBAL_ASM("asm/non_matchings/ovl_En_Tanron6_0x80BE6040/EnTanron6_Destroy.asm")
+void EnTanron6_SetupDoNothing(EnTanron6* this) {
```
Kenix3, line 45: Not a fan of this function name. EnTanron6_SetupActionFunc is better.

Kenix3 (changes requested):

fullgrowngaming:
I had it that originally, but the OoT people told me that was only used in cases where an actionfunc was actually passed into the "setupaction" function, and otherwise, it's named along these lines

Kenix3:
> 
> 
> I had it that originally, but the OoT people told me that was only used in cases where an actionfunc was actually passed into the "setupaction" function, and otherwise, it's named along these lines

It doesn't DoNothing, so the name is very bad. If the naming comes from the OOT people, then we need to have a conversation with them.

If there's no better name, then it should be named back to "func_80BE60BC" for now.

src/overlays/actors/ovl_En_Tanron6/z_en_tanron6.c
Kenix3: I made a mistake here. This doesn't accept a function pointer. It just sets a statically defined function to the actionFunc.

src/overlays/actors/ovl_En_Tanron6/z_en_tanron6.c
```diff
@@ -9,7 +9,9 @@ void EnTanron6_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnTanron6_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnTanron6_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-/*
+void EnTanron6_SetupDoNothing(EnTanron6* this);
+void func_80BE60D0(EnTanron6* this, GlobalContext* globalCtx);
```
Kenix3, line 13: This should be named EnTanron6_DoNothing

Kenix3 (changes requested):
I made a mistake in the initial review. I thought the actionFunc was being set to a function pointer argument rather than a statically defined function.

I do still think it's unclear that it sets up the "DoNothing" action, so I would suggest to name the "DoNothing" function at the same time.

Kenix3 (approved):

PR #30, opened 2020-09-23
Fix several issues with the makefile

Rozelette:
Changes:
- init now explicitly calls make on each target. This allows it to work with -j.
- setup now calls extract_rom.py, simplifying the setup process. Since extract_rom expects a path to the rom to extract, I gave a default which matches other decomp projects, with the option to override if desired.
- all now explicitly builds the uncompressed rom, then the compressed one. This fixes a problem when using -j where the compressed rom could finish building first, then fail and stop the uncompressed rom from finishing. Since the uncompressed rom was never built, you could not use first_diff.py to see what was wrong. This makes it slightly slower, but only by a few seconds in most cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,10 +50,10 @@ CFLAGS += -G 0 -non_shared -Xfullwarn -Xcpluscomm -Iinclude -Isrc -Wab,-r4300_mu
 
 # ROM image
 MM_BASEROM ?= baserom.z64
-ROM_NAME := rom
-ROM := $(ROM_NAME).z64
-UNCOMPRESSED_ROM := $(ROM_NAME)_uncompressed.z64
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -49,6 +49,7 @@ CFLAGS += -G 0 -non_shared -Xfullwarn -Xcpluscomm -Iinclude -Isrc -Wab,-r4300_mu
 #### Files ####
 
 # ROM image
+MM_BASEROM ?= baserom.z64
 ROM_NAME := rom
```
m4xw, line 53: might wanna make it ?= while at it?

m4xw (approved):

Kenix3 (approved):
Looks good to me!

PR #32, opened 2020-09-27
Fix ordering of relocations in overlay

Rozelette:
Fixing an issue that has been on my TODO list for awhile. Relocations are always in the order of .text, .data, then .rodata. Sometimes objects don't order the relocation sections that way, which causes our script to produce non-matching overlays. Fix it by just iterating over those sections in the expected order.


Kenix3:
Looks good to me.

PR #33, opened 2020-09-28
Arms_Hook OK

fullgrowngaming:


Kenix3 (approved):
Looks good to me.
Build is OK.
Pending +1 reviewer.

include/z64actor.h
```diff
@@ -179,22 +179,26 @@ typedef struct {
...
-    /* 0x147 */ s8 unk147;
-    /* 0x148 */ UNK_TYPE1 pad148[0x3];
+    /* 0x147 */ s8 itemActionParam;
+    /* 0x148 */ UNK_TYPE1 pad148[0x2];
+    /* 0x14A */ s8 heldItemActionParam;
     /* 0x14B */ u8 unk14B;
-    /* 0x14C */ UNK_TYPE1 pad14C[0x7];
+    /* 0x14C */ UNK_TYPE1 pad14C[0x5];
+    /* 0x151 */ u8 unk151;
     /* 0x153 */ u8 unk153;
```
Rozelette, line 188: These numbers don't line up. Is there another pad at 0x152?

linker_scripts/object_script.txt
```diff
@@ -32,3 +32,7 @@ D_0601F050 = 0x0601F050;
 D_06018DA0 = 0x06018DA0;
 D_0601EF10 = 0x0601EF10;
 D_06018C60 = 0x06018C60;
+
+/* arms_hook */
+D_0601D960 = 0x0601D960;
+D_040008D0 = 0x040008D0;
```
Rozelette, line 38: Group this with the other `D_04XXXXXX` symbols. Since segment 4 is gameplay_keep, there could be other overlays that use this symbol. It would be better to group gameplay_keep's symbols together to avoid duplication.

Rozelette (changes requested):
Just a couple of minor things, but besides that it looks good!

include/io/controller.h
```diff
@@ -3,6 +3,36 @@
 
 #include <PR/ultratypes.h>
 
+#define CONT_A      0x8000
+#define CONT_B      0x4000
+#define CONT_G	    0x2000
+#define CONT_START  0x1000
```
fig02, line 9: we recently reworked these defines (and all libultra related headers) in oot, not sure if the updated button defines want to be used here as well 

code looks good to me

fig02 (approved):

Rozelette (approved):
LGTM

PR #34, opened 2020-10-07
z_skelanime OK

fullgrowngaming:


fig02 (commented):
had a scroll through and it looks good for the most part.

the whole making a temp for the gfx pointer and then incrementing the temp + adding it at the end is all really weird. I hope that we can eventually find something that matches and seems a bit more sane

additionally just want to warn that the documentation on this file is planned to be reworked eventually in oot, since we've learned a bit more since this last round of documentation was done. so there might need to be updates here eventually too

fig02 (approved):

Kenix3:
Looks good to me.
Build is OK.

Kenix3 (approved):
Looks good to me.
Build is OK.

PR #36, opened 2020-10-12
en_ending_hero OK

fullgrowngaming:


src/overlays/actors/ovl_En_Ending_Hero/z_en_ending_hero.c
```diff
@@ -21,16 +23,73 @@ const ActorInit En_Ending_Hero_InitVars = {
...
+            this->unk242 = 0;
+            this->unk240 = (s16)randZeroOneScaled(60.0f) + 0x14;
+        }
+    }
+    this->actionFunc(this, globalCtx);
+    Actor_SetVelocityAndMoveYRotationAndGravity(&this->actor);
+    func_800B78B8(globalCtx, &this->actor, 20.0f, 20.0f, 50.0f, 0x1D);
+}
+
+UNK_PTR D_80C1E970[] = { 0x06007350, 0x06009590, 0x06009F90, 0x0600A790, 0x0600AB90 };
```
Kenix3, line 70: These look to be textures that come from the object file.
We don't currently have types for textures, but you should at least make a symbol for each of these segmented pointers if it's still OK.

src/overlays/actors/ovl_En_Ending_Hero/z_en_ending_hero.c
```diff
@@ -21,16 +23,73 @@ const ActorInit En_Ending_Hero_InitVars = {
...
+UNK_PTR D_80C1E984[] = { 0x06007750, 0x0600A390, 0x0600A490 };
```
Kenix3, line 71: These look to be textures that come from the object file.
We don't currently have types for textures, but you should at least make a symbol for each of these segmented pointers if it's still OK.

Kenix3 (changes requested):

src/overlays/actors/ovl_En_Ending_Hero/z_en_ending_hero.c
fullgrowngaming: I converted them to symbols, and also went ahead and moved the arrays into the function and made them static

Rozelette (approved):
LGTM

Kenix3 (approved):

PR #45, opened 2021-02-01
Added IDO recomp support

louist103:
Makefile stuff was basically just lifted out of OOT so if there are any issues let me know.
This was tested on WSL and got OK.  Someone using windows or MacOS should probably test it to.


Random06457:
The .exe weren't included in your commit (probably a `.gitignore` issue)

PR #47, opened 2021-02-14
en_okarina_effect OK

kyleburnette:
The build is OK, but I had to change some stuff around KankyoContext. Before merging this will probably require some thought, because I'm not sure if what I did is legit.


Kenix3:
Build is OK, looks good to me.

My only comment would be in general that we should try to correctly type variables that are set rather than remove them completely. In this case it's fine.

PR #48, opened 2021-02-24
Fixed ultratypes.h

kyleburnette:
It was broken before and required me to use int instead of s32. This fixes it though (copied from OoT)


Kenix3:

PR #49, opened 2021-02-24
pragma variant for GLOBAL_ASM

Thar0:
Much friendlier to editors and tools.


Kenix3:

PR #59, opened 2021-03-03
z_vr_box partially complete

Zelllll:
one function is missing since its massive and virtually impossible to decompile


Kenix3 (commented):
Looks good to me.
Pending build check.

Kenix3:
See https://github.com/zeldaret/mm/pull/73

PR #60, opened 2021-03-03
ovl_Door_Spiral OK and DISP macros fixed

Zelllll:


Zelllll:
I'm not dealing with any of these pr's anymore so anyone with edit access is free to do what they want with them

PR #62, opened 2021-03-05
En_Tubo_Trap (Flying Pot Enemy)

isghj5:
One function is non-matching: it wants to use a `V` register instead of an `T` register, nothing I have attempted or permuter has succeeded in fixing it, but its minor.


src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -21,26 +27,324 @@ const ActorInit En_Tubo_Trap_InitVars = {
...
+    0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00,
+};
+
+void EnTuboTrap_Init(Actor *thisx, GlobalContext *globalCtx) {
+    EnTuboTrap* this = THIS;
+
+    Actor_ProcessInitChain(thisx, &D_8093145C);
+    thisx->shape.rot.z = 0;
```
kyleburnette, line 51: thisx should be avoided for matching when possible, which is like 99.999% of the time (for example, this one should be `this->actor.shape`

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -21,26 +27,324 @@ const ActorInit En_Tubo_Trap_InitVars = {
...
+    thisx->world.rot.z = 0;
+    Actor_SetDrawParams(&thisx->shape, 0.0f,
+          (void (*)(struct Actor *actor,
+                    struct LightMapper *mapper,
+                    struct GlobalContext *ctxt)) func_800B3FC0, 1.8f);
```
kyleburnette, line 56: You can just put the function name on ones like this

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.h
```diff
@@ -5,11 +5,24 @@
...
+
 typedef struct EnTuboTrap {
     /* 0x000 */ Actor actor;
-    /* 0x144 */ char unk_144[0x60];
+    /* 0x0144 */ EnTuboTrapActionFunc actionFunc;
+    /* 0x015X */ f32 targetY;
+    /* 0x015X */ Vec3f originPos;
+    /* 0x0158 */ ColCylinder collider;
+
+    /* 0x144 */ //char unk_144[0x60];
```
kyleburnette, line 17: I believe this comment was left in by mistake

kyleburnette (commented):
If this were the OoT repo they'd say to remove like all of the comments in both the header and the source file. Not sure how the MM maintainers feel about them.

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
isghj5: Yeah I've been trying to avoid accidentally using thisx when it could be actor, it just slips past me sometimes because this and thisx are too alike visually. I'll take another pass

linker_scripts/code_script.txt
```diff
@@ -2061,7 +2061,7 @@ SECTIONS
     ovl_En_Tubo_Trap : AT(RomLocation)
     {
         build/src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.o(.text)
-        build/asm/overlays/ovl_En_Tubo_Trap_data.o(.data)
+        build/src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.o(.data)
         build/src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.o(.rodata)
         build/asm/overlays/ovl_En_Tubo_Trap_rodata.o(.rodata)
```
Rozelette, line 2066: Once you've migrated all the .data and .rodata, you should change this line to be
```suggestion
        build/src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap_overlay.o(.ovl)
```
This is the section that gives to relocations for the overlay.
Having `NON_MATCHING` _may_ cause this to not work, but it shouldn't. So if it gives you issues make a note of it since that probably means we need to fix things.

linker_scripts/object_script.txt
```diff
@@ -54,6 +54,10 @@ D_06009120 = 0x06009120;
 D_06005458 = 0x06005458;
 D_0600788C = 0x0600788C;
 
+/* z_en_tubo_trap */
+D_05018090 = 0x05018090;
+D_05017EA0 = 0x05017EA0;
+
```
Rozelette, line 60: These are in segment 5, so they're not actually in the `en_tubo_trap` actor. Move them to the top of the file, under the `0x04` addresses.
It's actually unclear what object they're from, since segment 5 could be 1 of 2. But since these are enemies, I would guess it's `gameplay_dangeon_keep`, so use that as the comment for what these are from.

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -8,8 +8,14 @@ void EnTuboTrap_Init(Actor* thisx, GlobalContext* globalCtx);
 void EnTuboTrap_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnTuboTrap_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnTuboTrap_Draw(Actor* thisx, GlobalContext* globalCtx);
+void func_80931004(EnTuboTrap *this, GlobalContext *globalCtx);
+void func_80931138(EnTuboTrap* this, GlobalContext* globalCtx);
+void func_809311C4(EnTuboTrap* this, GlobalContext* globalCtx);
+
+s32 D_80931410[] = { // Colliderinit
+    0x0A110900, 0x20010000, 0x00000000, 0xF7CFFFFF, 0x00040000, 0xF7CFFFFF, 0x00000000, 0x01010000, 0x000B001C, 0x00000000, 0x00000000, 
+};
```
Rozelette, line 17: This should be converted into a struct, specifically a `ColCylinderInit`.
There's a tool coming in #58 that should help, so you might want to wait for that (which btw with change the type name).

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -21,26 +27,317 @@ const ActorInit En_Tubo_Trap_InitVars = {
...
+    0x1C, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00,
+};
+
+void EnTuboTrap_Init(Actor *thisx, GlobalContext *globalCtx) {
+    EnTuboTrap* this = THIS;
+
+    Actor_ProcessInitChain(&this->actor, &D_8093145C);
```
Rozelette, line 50:
```suggestion
    Actor_ProcessInitChain(&this->actor, D_8093145C);
```

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -21,26 +27,317 @@ const ActorInit En_Tubo_Trap_InitVars = {
...
+// EnTuboTrap_HandleImpact
+void func_80930DDC(EnTuboTrap *this, GlobalContext *globalCtx) {
+    Actor *player = PLAYER;
+    Actor *player2 = PLAYER;
+    
+    // in oot func_800F0568 is Audio_PlaySoundAtPosition
+
+    if (((this->actor.bgCheckFlags & 0x20) != 0) && (this->actor.yDistToWater > 15.0f)) {
+        func_80930B60(this, globalCtx);
+        func_800F0568(globalCtx,  &this->actor.world, 0x28, (u16)0x2817);
```
Rozelette, line 204: Avoid unnecessary casts like this. Applies throughout the file.

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -21,26 +27,317 @@ const ActorInit En_Tubo_Trap_InitVars = {
...
+    }
+} // */
+#else
+  #pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Tubo_Trap_0x809307E0/func_80931004.asm")
+#endif
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Tubo_Trap_0x809307E0/EnTuboTrap_Destroy.asm")
+// levitate
+void func_80931138(EnTuboTrap *this, GlobalContext *globalCtx) {
+    this->actor.shape.rot.y += 0x1388;
```
Rozelette, line 286:
```suggestion
    this->actor.shape.rot.y += 5000;
```
Use the most natural number between hex and decimal. Sometimes it's a bit of a guess.

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -21,26 +27,317 @@ const ActorInit En_Tubo_Trap_InitVars = {
...
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Tubo_Trap_0x809307E0/func_80930B60.asm")
+    func_800B8EC8(&this->actor, 0x3037U); // play flying sfx
+    
+    if ((SQ(dX) + SQ(dY) + SQ(dZ) > 57600.0f)) { //240.0f ^ 2
+        Math_SmoothScaleMaxF(&this->actor.gravity, -3.0f, 0.2f, 0.5f);
+    }
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Tubo_Trap_0x809307E0/func_80930DDC.asm")
+    this->actor.shape.rot.y += 0x1388;
```
Rozelette, line 308:
```suggestion
    this->actor.shape.rot.y += 5000;
```

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.h
```diff
@@ -5,11 +5,20 @@
...
+    /* 0x0148 */ f32 targetY;
+    /* 0x014C */ Vec3f originPos;
+    /* 0x0158 */ ColCylinder collider;
 } EnTuboTrap; // size = 0x1A4
 
 extern const ActorInit En_Tubo_Trap_InitVars;
 
+extern u32 D_05018090; // pot break shard displaylist
+
+extern u32 D_05017EA0; // the pot displaylist
```
Rozelette, line 22: These should be of type `Gfx[]`

Also, since these are probably from `gameplay_dangeon_keep`, put them in `include\segment.h` instead. That file is a little messy, but the idea is that since multiple actors could be using those symbols, they should defined in one place. Eventually there will be a header file for each object.

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
```diff
@@ -21,26 +27,317 @@ const ActorInit En_Tubo_Trap_InitVars = {
...
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Tubo_Trap_0x809307E0/EnTuboTrap_Update.asm")
+    Collision_CylinderMoveToActor(&this->actor, &this->collider);
+    Collision_AddAC(globalCtx, &globalCtx->colCheckCtx, &this->collider.base);
+    Collision_AddAT(globalCtx, &globalCtx->colCheckCtx, &this->collider.base);
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Tubo_Trap_0x809307E0/EnTuboTrap_Draw.asm")
+void EnTuboTrap_Draw(Actor *thisx, GlobalContext *globalCtx) {
+    //  Gfx_DrawDListOpa with a display list
+    func_800BDFC0(globalCtx, &D_05017EA0);
```
Rozelette, line 342: Should be
```suggestion
    func_800BDFC0(globalCtx, D_05017EA0);
```
but only after `D_05017EA0` is a `Gfx[]`

Rozelette (changes requested):
Congrats on your first PR!

Overall it looks pretty good, most of my comments were for little things. I was pretty lax about style since we should be automatically running clang-format, and so eventually it'll run and fix it all, but you should keep it in mind in the future.

As for the comments throughout, I don't really mind. There will be a documentation pass later that'll clean it all up. As long as it's not distracting, like commented-out old code.

linker_scripts/code_script.txt
isghj5: Last I tried it didn't work and I assumed it was the non-matching, but I might have just attempted wrong, will re-attempt

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.h
isghj5: `extern Gfx[] D_05018090` wont build if placed in `include/segment.h` but `extern UNK_TYPE D_05018090` does.

is it maybe too early in the build process and Gfx isn't defined yet?

Rozelette: Since it's an array type, the `[]` goes after the name, like `extern Gfx D_05018090[]`

Rozelette:
Looks like this PR can't be built with the recent changes to master because of the spaces before the `#pragma GLOBAL_ASM`. I'll PR a fix sometime, but if you want this to be built before that just remove those spaces.

src/overlays/actors/ovl_En_Tubo_Trap/z_en_tubo_trap.c
isghj5: Kenix requested in my other PR that all rotation calculations should be left in hex, for now returning to hex.

Feel free to re-request the change if this overrides that.

Rozelette: I spoke to Kenix about this, and we agreed it can be decimal, so I'd like to re-request this. Angle calculations are probably almost always in hex, because that is a much more natural to work with them, but that doesn't mean the devs were consistent. So for angles, go with hex unless the decimal number is much rounder, like it is here. Sorry for the back and forth on this!

Rozelette (approved):
LGTM

Kenix3 (approved):
Looks good to me!

PR #63, opened 2021-03-09
Restructure global functions and variables headers

Thar0:
- Removed `functions_overlay.h` as there is no need for prototypes to be accessible outside of the overlay in general. For the few functions where external references are required, `functions.h` is more than sufficient.
 - Replaced `variables_overlay.h` with `initvars.h`, as the actor's initvars are the only actor variables that require external references.
 - Commented out many functions in `functions.h` as the incorrect prototypes make a first pass with mips_to_c full of unset register warnings and generally much more incorrect code output than it would have otherwise produces. Once functions are decompiled their own prototype as well as prototypes of functions it calls should be uncommented and corrected as needed.
 - Commented out many variables in `variables.h` for similar reasons as above. The workflow for this is similar: once a proper type is found through decompilation, a variable may be uncommented and appropriately typed.
 - The comments in both `functions.h` and `variables.h` were removed, they make find-replace renames unnecessarily difficult and offer little benefit, as the addresses may be found in the map file if needed.
 - The commented out functions and variables remain as they were with the speculated arguments, these may continue to be used as hints as to function arguments but should always be confirmed.
 - Once code rodata migration is taken care of, it would be possible to remove the entire `rodata` region of `variables.h`, as the symbols should never be accessed directly.


Rozelette (approved):
LGTM
Thanks or doing all of this work! I like all the changes, and agree with removing rodata once we are able.

Kenix3 (approved):
Looks good to me! Thanks for the submission.

PR #70, opened 2021-03-14
Generate dependencies for C files

Rozelette:
This PR adds a feature when `.c` files get a `.d` file generated that lists the `.h` and `.asm` files they depend on. For `.h` files, it uses `gcc`. For `.asm` files, a new python script rads the file for a `GLOBAL_ASM` and uses that function and it's base `.asm` as a dependencies. It only uses the first function it encounters as having all of them complicates the makefile, since it'll have to deal with recipes that create multiple outputs (the `disasm` target is bad enough already).

This allows us to move the `split_asm` part of `disasm` into it's own recipe. Which also allows make to parallelize it when building with `-j`. And it allows the removal of `files_with_nonmatching.txt`.

I also took the opportunity to silence some of the less useful commands to reduce the spam while building.

One issue is that splitting rodata files changes them, so we can't build their `.o`s until after the split is done. There is not an easy way to do this, so I'm having them be built after the `.c`'s `.o`s. I don't think this causes it to take any longer, however.

```
Time before, with -j:
real    1m12.826s
user    5m32.657s
sys     1m5.615s

Time after, with -j:
real    0m36.145s
user    6m14.328s
sys     0m59.698s

Time before, single thread:
real    3m38.645s
user    3m18.549s
sys     0m20.907s

Time after, single thread:
real    3m43.429s
user    3m22.129s
sys     0m22.174s
```


Kenix3:
This change looks good. Great job getting the time down some. I still think we should look at caching the disasm.

Going to review and merge this out of band to get Jenkins speed up in early.

PR #71, opened 2021-03-16
yaz0.c OK

Rozelette:
Had to change some flags in `preproccess.py` because this is our first file with Japanese text.


Kenix3 (approved):
Looks good to me!
+1 reviewer before merge

src/boot_O2_g3/yaz0.c
```diff
@@ -0,0 +1,150 @@
...
+    u32 bitIdx = 0;
+    u8* dstEnd;
+    u32 chunkHeader = 0;
+    u32 nibble;
+    u8* backPtr;
+    s32 chunkSize;
+    u32 off;
+    u32 magic;
+
+    magic = ((Yaz0Header*)src)->magic;
```
Thar0, line 79: Is this required for matching, or maybe warning suppression? OoT lets `Yaz0_DecompressImpl` take a `Yaz0Header*` instead of `u8*`. Why might we be adverse to doing the same?

Thar0 (commented):
LGTM, just one quick question.

src/boot_O2_g3/yaz0.c
Rozelette: Required for matching. I couldn't get it to match without reusing the `src` argument, which happens a couple of lines down.

Thar0: Ahh I see it now, got it.

Thar0 (approved):

PR #72, opened 2021-03-16
padutils OK

Rozelette:


Kenix3 (approved):
Looks good to me!
+1 reviewer before merge

include/z64.h
```diff
@@ -283,14 +283,6 @@ typedef enum IRQ_TYPE {
     IRQ_PRENMI_4 = 0x2A1
 } IRQ_TYPE;
 
-typedef struct {
-    /* 0x0 */ u16 buttons;
-    /* 0x2 */ s8 xAxis;
-    /* 0x3 */ s8 yAxis;
-    /* 0x4 */ s8 status;
-    /* 0x5 */ UNK_TYPE1 pad5[0x1];
-} InputInfo; // size = 0x6
```
fig02, line 292: curious on the removal of this and using `PadInput` instead. oot still uses this: 
```
typedef struct {
    /* 0x00 */ u16 button;
    /* 0x02 */ s8 stick_x;
    /* 0x03 */ s8 stick_y;
    /* 0x04 */ u8 errno;
} OSContPad; // size = 0x06

typedef struct {
    /* 0x00 */ OSContPad cur;
    /* 0x06 */ OSContPad prev;
    /* 0x0C */ OSContPad press; // X/Y store delta from last frame
    /* 0x12 */ OSContPad rel; // X/Y store adjusted
} Input; // size = 0x18
```
where these changes needed to match?

Thar0: I was going to comment on this also, `padutils` seems identical between MM and OoT so the `OSContPad` approach should also be fine here.

Thar0:
Totally optionally, this PR could also introduce the `CHECK_BTN` macros if desired:
`#define CHECK_BTN_ALL(state, combo) (~((state) | ~(combo)) == 0)`
`#define CHECK_BTN_ANY(state, combo) (((state) & (combo)) != 0)`

include/z64.h
Rozelette: I had done this file a few months ago and copied what OOT had then. Looks like it has been changed since then, so I'll do the same here.

Thar0 (approved):

PR #76, opened 2021-03-18
Improve the robustness of depend.py

Rozelette:
This fixes a couple of issues that people were having, such as when there was a space before the `#pragma` or having the path to the assembly file formatted oddly.


Kenix3 (approved):
Looks good to me. Great change and great idea!

PR #77, opened 2021-03-19
Restructure how buffers are set up

Rozelette:
This introduces absolute locations for the framebuffers and several other buffers that need to be in a specific location, not shifted if `code` or other files are shifted. Since the framebuffers for the two resolutions overlap, it's not possible to have them as sections in the load script, so I chose to have them as absolute locations instead.

There is also some cleanup, most notable of which is that the disassembler no longer has the `-u` option, which had it generate `undef.txt` each time. This lets us now delete symbols from it and slowly dwindle it down.


Kenix3 (approved):
This one looks great to me.

PR #79, opened 2021-03-22
Jenkins update

ethteck:
All building is now done on non-master agents and the progress information is stashed and then unstashed from master, which updates the public files.


Rozelette (approved):
LGTM

PR #80, opened 2021-03-22
z_lights fully matched

Rozelette:
This PR matchs the remaining functions for z_lights.c. It also updates it to have the same documentation as OOT.


Kenix3 (approved):
Looks good to me!

PR #82, opened 2021-03-23
Floats and Strings in asm

Thar0:
Also found several unreferenced stub functions.
Ice and Light Arrow rodata splits are now also `OK`, possibly others too but I would like to leave checking that for now.
In variables.txt I have added comments on a few somewhat problematic variables.


Rozelette (approved):
LGTM

Kenix3 (approved):
Looks good to me!

PR #85, opened 2021-03-25
z_lib fully matched

Rozelette:
Also updated the documentation to OOT and named the new MM functions


include/stdint.h
```diff
@@ -0,0 +1,29 @@
+#ifndef _STDINT_H_
+#define _STDINT_H_
+
+#include <ultra64.h>
+
+typedef long intptr_t;
+typedef unsigned long uintptr_t;
```
Thar0, line 7:
```suggestion
typedef s32 intptr_t;
typedef u32 uintptr_t;
```
Unfortunately IDO cares about `long` vs `int` for matching, so to be completely safe this should mirror the 32-bit types in `ultratypes.h`, at least for now.

Thar0 (commented):
The file itself looks good to me. Just some thoughts.

include/functions.h
```diff
@@ -1836,40 +1836,40 @@ void Math_Vec3f_Scale(Vec3f* vec, f32 scale);
...
+void func_801000A4(u16 sfxId);
+void func_801000CC(u16 sfxId);
+void func_801000F4(s32 a0, u16 a1);
+void Lib_Vec3f_TranslateAndRotateY(Vec3f* translation, s16 a, Vec3f* src, Vec3f* dst);
 void Lib_LerpRGB(Color_RGB8* a, Color_RGB8* b, f32 t, Color_RGB8* dst);
-f32 Lib_PushAwayVec3f(Vec3f* start, Vec3f* pusher, f32 distanceToApproach);
+f32 Math_Vec3f_StepTo(Vec3f* start, Vec3f* target, f32 speed);
 void Lib_Nop801004FC(void);
 void* Lib_PtrSegToVirt(void* ptr);
 void* Lib_PtrSegToVirtNull(void* ptr);
```
Thar0, line 1875: Additionally, this is solely my opinion but I believe a better naming scheme for these functions (as well as the `PtrSegToK0` functions just below these) would be names that directly mirror the macros. i.e. `Lib_SegmentedToVirtual` and `Lib_PhysicalToVirtual`

include/stdint.h
fig02: any reason this is needed on top of ultratypes?

m4xw: grafik

Kenix3:
Looks good to me!

PR #90, opened 2021-03-29
DoorAna: OK

isghj5:


include/functions.h
```diff
@@ -1654,7 +1654,7 @@ void func_800F0568(GlobalContext* ctxt, Vec3f* position, s32 param_3, s16 sfxId)
 void func_800F12D0(GlobalContext* ctxt);
 void func_800F1304(GlobalContext* ctxt, s16 param_2);
 void func_800F1374(s32 param_1, s16 param_2);
-// void func_800F13E8(void);
+s16 func_800F13E8(GlobalContext* ctxt, short param_2);
```
kyleburnette, line 1657: Short param should be s16 or u16 or whatever I think

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -9,7 +9,10 @@ void DoorAna_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void DoorAna_Update(Actor* thisx, GlobalContext* globalCtx);
 void DoorAna_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-/*
+void DoorAna_WaitClosed(DoorAna* thisx, GlobalContext* globalCtx);
+void DoorAna_WaitOpen(DoorAna* thisx, GlobalContext* globalCtx);
```
kyleburnette, line 13: WaitClosed and WaitOpen should have this as the param name I believe

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,166 @@ const ActorInit Door_Ana_InitVars = {
...
+    }
+}
+
+void DoorAna_WaitClosed(DoorAna* this, GlobalContext* globalCtx) {
+    s32 grottoIsOpen = 0;
+    u32 grottoType = GET_DOORANA_TYPE(this);
+
+    if ((grottoType) == DOORANA_TYPE_UNK) {
+        // in OOT decomp its marked as open with storms, but does not seem to open with storms in MM
+        if ((this->actor.xyzDistToPlayerSq < 40000.0f) && (func_800F13E8(globalCtx, 5) != 0)) {
```
kyleburnette, line 82: For (x != 0) type stuff I've been told to just omit the != 0 if it still matches, not sure about the current thoughts on that

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,166 @@ const ActorInit Door_Ana_InitVars = {
...
+            grottoIsOpen = 1;
+            this->actor.flags &= ~0x10; // always update OFF
+        }
+
+    } else {
+        if ((this->bombCollider.base.acFlags & AC_HIT) ) { // bomb collision
```
kyleburnette, line 88: Unnecessary extra parentheses

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,166 @@ const ActorInit Door_Ana_InitVars = {
...
+    s32 dooranaType = GET_DOORANA_TYPE(this);
+    s8 pad[4];
+    s32 entranceIndex;
+    f32 yDist;
+
+    if (Math_StepToF( &this->actor.scale, 0.01f, 0.001f) != 0) { 
+        if ((this->actor.targetMode != 0) && (globalCtx->unk18875 == 0) && (globalCtx->unk18B4A == 0) &&
+          (player->stateFlags1 & 0x80000000) && (player->unkAE7 == 0)) {
+
+            if ((dooranaType) == DOORANA_TYPE_ADJACENT) {
```
kyleburnette, line 118: parentheses

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,166 @@ const ActorInit Door_Ana_InitVars = {
...
+                // 300 uses scene exit addresses, not static grotto addresses, 
+                // eg. deku playground gets address in the NCT scene exit table, not in grotto table
+
+                entranceIndex = this->actor.params & 0x1F;
+                globalCtx->unk1887A = globalCtx->setupExitList[entranceIndex];
+                label_required_to_convince_compiler_to_not_use_delay_slot: ;
```
kyleburnette, line 124: This is almost certainly not necessary but I can understand leaving it in if it's too much trouble otherwise

isghj5: Sure looks required, doesn't match without it

required label

kyleburnette: It might be required as the code currently is, but I think the current thinking is that these fake labels are basically always fake, and that there is another solution 99.99% of the time

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,166 @@ const ActorInit Door_Ana_InitVars = {
...
+
+    if (Math_StepToF( &this->actor.scale, 0.01f, 0.001f) != 0) { 
+        if ((this->actor.targetMode != 0) && (globalCtx->unk18875 == 0) && (globalCtx->unk18B4A == 0) &&
+          (player->stateFlags1 & 0x80000000) && (player->unkAE7 == 0)) {
+
+            if (dooranaType == DOORANA_TYPE_ADJACENT) {
+                // 300 uses scene exit addresses, not static grotto addresses, 
+                // eg. deku playground gets address in the NCT scene exit table, not in grotto table
+
+                entranceIndex = this->actor.params & 0x1F;
```
Kenix3, line 122: Let's create a macro to access the params value.

Kenix3 (changes requested):

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
m4xw: I suggest a generic `lblUnk:;`

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,169 @@ const ActorInit Door_Ana_InitVars = {
...
+
+void DoorAna_Init(Actor* thisx, GlobalContext* globalCtx) {
+    DoorAna* this = THIS;
+    s32 grottoType = GET_DOORANA_TYPE(this);
+    this->actor.shape.rot.y = this->actor.shape.rot.z = 0;
+
+    if ((grottoType == DOORANA_TYPE_UNK) || (grottoType == DOORANA_TYPE_HIDDEN)) {
+        if (grottoType == DOORANA_TYPE_HIDDEN) {
+            Collider_InitAndSetCylinder(globalCtx, &this->bombCollider, &this->actor, &sCylinderInit);
+
```
Rozelette, line 52: Remove the newline before the else. There are several places in this file where this applies.

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,169 @@ const ActorInit Door_Ana_InitVars = {
...
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Door_Ana_0x808E01A0/DoorAna_Update.asm")
+void DoorAna_Update(Actor *thisx, GlobalContext *globalCtx) {
+    DoorAna* this = THIS;    
+    this->actionFunc(this, globalCtx);
+    this->actor.shape.rot.y = func_800DFCDC(globalCtx->cameraPtrs[globalCtx->activeCamera]) + 0x8000;
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Door_Ana_0x808E01A0/DoorAna_Draw.asm")
+void DoorAna_Draw(Actor *thisx, GlobalContext *globalCtx) {
+    Gfx* doorAnaDL = &D_05000C40;
```
Rozelette, line 190:
```suggestion
    Gfx* doorAnaDL = D_05000C40;
```

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
```diff
@@ -5,11 +5,30 @@
...
 typedef struct DoorAna {
     /* 0x000 */ Actor actor;
-    /* 0x144 */ char unk_144[0x50];
+    /* 0x144 */ ColliderCylinder bombCollider;
+    /* 0x190 */ DoorAnaActionFunc actionFunc;
 } DoorAna; // size = 0x194
 
 extern const ActorInit Door_Ana_InitVars;
 
+extern Gfx D_05000C40[];
```
Rozelette, line 18: Put this in `segment.h`

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,169 @@ const ActorInit Door_Ana_InitVars = {
...
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Door_Ana_0x808E01A0/func_808E02A4.asm")
+void DoorAna_GrabLink(DoorAna* this, GlobalContext* globalCtx) {
+    Actor* player;
+    s8 pad[2];
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Door_Ana_0x808E01A0/func_808E03B8.asm")
+    if (ActorCutscene_GetCurrentIndex() != this->actor.cutscene) {
+        if (ActorCutscene_GetCanPlayNext(this->actor.cutscene) != 0) {
```
Rozelette, line 169: This function returns a "bool", so remove the `!= 0` if possible

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,169 @@ const ActorInit Door_Ana_InitVars = {
...
+                func_80169E6C(globalCtx, 3, 0x4FF);
+
+                gSaveContext.extra.previousPlayerHeight = this->actor.world.pos.y;
+                gSaveContext.extra.previousPlayerYRot   = this->actor.home.rot.y;
+
+                // save the params lower byte for En_Torch to decide what item to use in the grotto chest
+                gSaveContext.extra.unk87 = GET_DOORANA_ITEMFLAGS(this);
+
+                // most grottos in the game use their zrotation as their entrance index, not params
+                if (DOORANA_TYPE_ROTATION_ENTRANCE(entranceIndex)) {
```
Rozelette, line 139: I don't think this is a good use of a macro, since all it does it test a local variable. It would be clearer if there was no macro, and the comment read something like "The params can only fit up to index 7. If the index was 0 (it could not fit in there), use the one stored in z rotation."

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,169 @@ const ActorInit Door_Ana_InitVars = {
...
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Door_Ana_0x808E01A0/func_808E01A0.asm")
+static ColliderCylinderInit sCylinderInit = { 
+    { COLTYPE_NONE, AT_NONE, AC_ON | AC_TYPE_PLAYER, OC1_NONE, OC2_NONE, COLSHAPE_CYLINDER, },
+    { ELEMTYPE_UNK2, { 0x00000000, 0x00, 0x00 }, { 0x00000008, 0x00, 0x00 }, TOUCH_NONE | TOUCH_SFX_NORMAL, BUMP_ON, OCELEM_NONE, },
+    { 50, 10, 0, { 0, 0, 0 } },
+};
+
+// entrances grottos can link to statically, type DOORANA_TYPE_ADJACENT uses scene exit addresses instead
+static u16 entrances[] = {
```
Rozelette, line 35: I feel that it would be helpful to have a `ENTRANCE` macro in `macros.h`, to pack these values together. Maybe something based on `Scene_CreateEntrance`, `#define ENTRANCE(sceneIndex, spawnIndex, offset) (((sceneIndex << 9) | (spawnIndex << 4)) | offset)`

Is there something already in OOT?

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
```diff
@@ -5,11 +5,30 @@
```
Rozelette, line 5: Some overall naming suggestion. I don't like "direct" and "adjacent", since it doesn't describe what the difference is: one uses a scene exit, and the other is for an internal list into the generic grottos. I would suggest:
`DOORANA_TYPE_ADJACENT` -> `DOORANA_TYPE_SCENE_EXIT`
`GET_DOORANA_DIRECT_ENTRANCE` -> `GET_DOORANA_GROTTO_TYPE`
`GET_DOORANA_ADJACENT_ENTRANCE` -> `GET_DOORANA_SCENE_EXIT`
`GET_DOORANA_ROTATION_ENTRANCE` -> `GET_DOORANA_GROTTO_TYPE_EX` (this does feel like an "ex" scenario lol)

include/z64.h
```diff
@@ -329,7 +329,11 @@ typedef struct {
     /* 0x010 */ s32 unk10;
     /* 0x014 */ UNK_TYPE1 pad14[0x2E];
     /* 0x042 */ s16 unk42;
-    /* 0x044 */ UNK_TYPE1 pad44[0x43];
+    /* 0x044 */ UNK_TYPE1 pad44[0x34];
+    /* 0x078 */ f32 previousPlayerHeight; //OvlDoorAna
```
Rozelette, line 333: Looking at OOT, this is `respawn[RESPAWN_MODE_RETURN]`. I'd prefer if these were left as `unkX` for now, or for the entire respawn struct be carried over.

Rozelette (changes requested):
Looks pretty good code-wise. Have some comments about naming and such

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -21,20 +24,169 @@ const ActorInit Door_Ana_InitVars = {
...
+    0x1A00, 0x1400, 0x1410, 0x1420, 0x1430, 0x1440, 0x1450, 0x1460,
+    0x1470, 0x1480, 0x1490, 0x14A0, 0x14B0, 0x14C0, 0x14D0, 0x0000, 0x0000, 0x0000,
```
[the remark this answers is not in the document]
isghj5: They are removable, and I can remove.

I left them in to show that there is space for three extra entrances that are usable, where as removing them hides that you can add more entrances without shifting the ROM, maybe not that useful in general but It was the kind of thing I might have wanted to find if I looked at this file.

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
Rozelette: Turns out OOT entrances worked a different way, so this would be a new macro for MM

PR #95, opened 2021-04-01
item_inbox OK

kyleburnette:


src/overlays/actors/ovl_Item_Inbox/z_item_inbox.c
```diff
@@ -21,14 +22,33 @@ const ActorInit Item_Inbox_InitVars = {
...
+    ItemInbox* this = THIS;
+
+    this->actionFunc = func_80945534;
+    Actor_SetScale(&this->actor, 0.2f);
+}
+
+void ItemInbox_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+void func_80945534(ItemInbox* this, GlobalContext* globalCtx) {
```
isghj5, line 36: Low Priority: This is fully decomped so function could be renamed?

The only actionfunc and checks a chest flag and dissapears (single frame, so shouldn't spawn either) if it's already collected? Maybe like `ItemInbox_WaitForDeath` or generic `ItemInbox_Idle` or something?

jpburnett (approved):
no complaints 👍

Kenix3 (approved):
Looks good to me.

PR #99, opened 2021-04-02
Scene decompilation

Rozelette:
This adds scene support. All scenes are now parsed by ZAPD and compiled from the C files ZAPD creates. Some important points:

- Cutscenes are not being parsed right now due since they aren't properly reverse-engineered yet. According, the cutscene macros weren't copied from OOT.
- The linker scripts now use macros so they are less verbose. This means that they are ran through the C preprocesser. As a PoC, I created a macro for actors and used it for one, but overhauling the linker script will need to wait for a future PR.
- I choose not to create automatic dependency (.d) files for the scenes. This is because if they were created, every time `z64.h` was changed, all scenes would have to be recompiled. I did not think this was necessary for now, but ideally once `z64.h` is split up into more nuanced headers it can be added back in.
- Scene extraction takes place as part of the default `make` target, instead of having it's target (e.g. `extract-assets`). This means it will happen after every time `make clean` is done. I was unsure if this would be the best option, or if it should be separate and make users explicitly say `make extract-assets` if they want to redo it, or to have it be part of `setup`.
- This PR restructures how we organize extracted files to be more easier to track progress for each one.
- progress.py now tracks how much of scenes, objects, and other files have been reconstructed. The csv now contains these additional stats.
- `make init` will need to be re-ran after this PR is merged.


Kenix3:
See my merge conflict resolved branch:
https://github.com/zeldaret/mm/pull/117

PR #102, opened 2021-04-09
ichaindis actual param names + permissions changes

kyleburnette:
Changed ichaindis to output actual names by default and added an --offset arg to output offset names if needed. Also changed permission on all the overlayhelper scripts.


Thar0 (approved):

Rozelette (approved):
LGTM

PR #105, opened 2021-04-14
(Mostly) Decompile PreRender.c

Zelllll:
a few of these functions I couldn't figure out, if anyone else could that would be neat.


Kenix3 (approved):
Looks good to me.

jpburnett (approved):
I see no issues

PR #117, opened 2021-04-29
Adds in scene support

Kenix3:


Kenix3:
Please also see Roz's original PR
https://github.com/zeldaret/mm/pull/99/files

PR #119, opened 2021-05-02
En_jc_mato matching

jpburnett:


.gitignore
```diff
@@ -28,6 +28,7 @@ tools/ido5.3_compiler/
 tools/ido7.1_compiler/
 tools/qemu-mips
 tools/ido_recomp/* binary
+ctx.c
```
kyleburnette, line 31: Really glad this was added. The amount of times I've committed the context by accident...

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
```diff
@@ -19,20 +28,111 @@ const ActorInit En_Jc_Mato_InitVars = {
     (ActorFunc)EnJcMato_Init,
     (ActorFunc)EnJcMato_Destroy,
     (ActorFunc)EnJcMato_Update,
-    (ActorFunc)EnJcMato_Draw
+    (ActorFunc)EnJcMato_Draw,
+};
+
+// was D_80B9E210
```
kyleburnette, line 34: Could probably be removed (not sure how others feel)

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
```diff
@@ -19,20 +28,111 @@ const ActorInit En_Jc_Mato_InitVars = {
...
+        return 0;
+    }
+}
+
+void func_80B9DFC8(EnJcMato* this) {
+    this->actionFunc = func_80B9DFDC;
+}
+
+void func_80B9DFDC(EnJcMato* this, GlobalContext* globalCtx) {
+    s16 phi_v1;
```
kyleburnette, line 82: This looks like a flag to track if unk1A0 has reached 0 or not yet, so I think a better name for this temp is probably possible. To me it looks like the target spawns with a timer (unk1A0) of 25, then once it gets hit, it counts down to 0, then it despawns. So unk1A0 would be a `despawnTimer` (or something), and phi_v1 would just be something that reflects that. That being said, I don't see the purpose of phi_v1 here. Not sure why they didn't just check for the value of unk1A0.

kyleburnette (changes requested):

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
jpburnett: Oops, yeah haha that is a left over comment

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
jpburnett: I had thought about it being a despawn timer, so I opened KZ and did the whole boat ride without hitting a target, but I never saw it despawn....could just be instantaneous though? I do agree that it looks to be a counter of some kind. I tried removing the phi temp, but it wouldn't match, I could try re-arranging things though.

kyleburnette: I think the timer doesn't start counting down until you hit the target. When you hit a target, it kinda flies backward for a bit and then despawns, so I'm guessing that after you hit it, it counts from 25 down to 0 before it disappears

jpburnett: I did not think about that! you could be right and the target despawns after 25 frames or something. I'll check :D

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
```diff
@@ -19,20 +28,110 @@ const ActorInit En_Jc_Mato_InitVars = {
...
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Jc_Mato_0x80B9DEE0/EnJcMato_Init.asm")
+    this->actionFunc(this, globalCtx);
+    if (!(gSaveContext.owl.unk4 & 1)) {
+        func_80B9DEE0(this, globalCtx);
+    }
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Jc_Mato_0x80B9DEE0/EnJcMato_Destroy.asm")
+extern Gfx D_06000390[];
+Vec3f D_80B9E25C = { 0.0f, -2500.0f, 0.0f };
```
Kenix3, line 126: This seems like a vector determining movement of the target.

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
```diff
@@ -19,20 +28,110 @@ const ActorInit En_Jc_Mato_InitVars = {
...
+};
+
+DamageTable EnJcMatoDamageTable = {
+    0x01, 0x01, 0x01, 0x01, 0x01, 0xF1, 0x01, 0x01, 0x01, 0x01, 0x01, 0xF1, 0xF1, 0xF1, 0x01, 0x01,
+    0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
 };
-*/
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Jc_Mato_0x80B9DEE0/func_80B9DEE0.asm")
+s32 func_80B9DEE0(EnJcMato* this, GlobalContext* globalCtx) {
```
Kenix3, line 59: `EnJcMato_CheckForHit`

Kenix3 (changes requested):
Just some naming suggestions. Thank you for the submission!

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
```diff
@@ -19,20 +28,110 @@ const ActorInit En_Jc_Mato_InitVars = {
...
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Jc_Mato_0x80B9DEE0/EnJcMato_Init.asm")
+    this->actionFunc(this, globalCtx);
+    if (!(gSaveContext.owl.unk4 & 1)) {
+        EnJcMato_CheckForHit(this, globalCtx);
+    }
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_En_Jc_Mato_0x80B9DEE0/EnJcMato_Destroy.asm")
+extern Gfx D_06000390[];
+Vec3f targetMovementScalar = { 0.0f, -2500.0f, 0.0f };
```
Kenix3, line 126: I might just call this movement.

Kenix3 (changes requested):

Kenix3 (approved):

PR #120, opened 2021-05-02
 ovl_Dm_Ravine (Link Riding Through Lost Woods Cutscene - Tree Trunk)

zbanks:
Hey - this is my first contribution to this project, so I picked a pretty simple overlay to get started. 

- I got the name "Link Riding Through Lost Woods Cutscene - Tree Trunk" from https://wiki.cloudmodding.com/mm/File_List/USA_GC - but I'm not sure how to verify it, or if it's descriptive enough? (Should I just call it "Unknown" instead?)
- There's a bit more info about what `RoomContext.unk7A` could be - but I think it's better to leave it `unk` until it's used in a few more places? (Also the code in `z_room.c` seems to suggest it really is an array and not 3 separate fields?)
- The unused function `DmRavine.actionFunc` is a bit weird. I'm sort of guessing it's an "action function" based on some other overlays that put a fn pointer after the base `Actor` struct. (Maybe it should stay "unknown" instead?)

### DIFF
--- a/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
+++ b/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
@@ -1,3 +1,9 @@
+/*
+ * File: z_dm_ravine.c
+ * Overlay: ovl_Dm_Ravine
+ * Description: Link Riding Through Lost Woods Cutscene - Tree Trunk
+ */
+
 #include "z_dm_ravine.h"
 
 #define FLAGS 0x00000030
@@ -6,29 +12,68 @@
 
 void DmRavine_Init(Actor* thisx, GlobalContext* globalCtx);
 void DmRavine_Destroy(Actor* thisx, GlobalContext* globalCtx);
+void DmRavine_Action(Actor* thisx, GlobalContext* globalCtx);
 void DmRavine_Update(Actor* thisx, GlobalContext* globalCtx);
 void DmRavine_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-/*
-const ActorInit Dm_Ravine_InitVars = {
-    ACTOR_DM_RAVINE,
-    ACTORCAT_ITEMACTION,
-    FLAGS,
-    OBJECT_KEIKOKU_DEMO,
-    sizeof(DmRavine),
-    (ActorFunc)DmRavine_Init,
-    (ActorFunc)DmRavine_Destroy,
-    (ActorFunc)DmRavine_Update,
-    (ActorFunc)DmRavine_Draw
-};
-*/
+const ActorInit Dm_Ravine_InitVars = { ACTOR_DM_RAVINE,
+                                       ACTORCAT_ITEMACTION,
+                                       FLAGS,
+                                       OBJECT_KEIKOKU_DEMO,
+                                       sizeof(DmRavine),
+                                       (ActorFunc)DmRavine_Init,
+                                       (ActorFunc)DmRavine_Destroy,
+                                       (ActorFunc)DmRavine_Update,
+                                       (ActorFunc)DmRavine_Draw };
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Init.asm")
+void DmRavine_Init(Actor* thisx, GlobalContext* globalCtx) {
+    u8 flag;
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Destroy.asm")
+    flag = gSaveContext.perm.weekEventReg[0];
+    if (((flag & 0x10) | cREG(0)) != 0) {
+        Actor_MarkForDeath(thisx);
+        return;
+    }
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/func_80A2E838.asm")
+    THIS->loaded = false;
+    globalCtx->roomContext.unk7A[0] = 1;
+    globalCtx->roomContext.unk7A[1] = 0;
+    THIS->state = 0;
+    Actor_SetScale(thisx, 1.0f);
+    THIS->actionFunc = &DmRavine_Action;
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Update.asm")
+void DmRavine_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Draw.asm")
+// This function is unused, but takes two word-sized arguments
+// `thisx` & `globalCtx` are guesses; but it could be something else
+void DmRavine_Action(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+void DmRavine_Update(Actor* thisx, GlobalContext* globalCtx) {
+    RoomContext* roomCtx;
+
+    switch (THIS->state) {
+        case 0:
+            return;
+        case 1:
+            THIS->loaded = true;
+            globalCtx->roomContext.unk7A[1]++;
+            if (globalCtx->roomContext.unk7A[1] > 254) {
+                globalCtx->roomContext.unk7A[1] = 254;
+                if (globalCtx->csCtx.frames > 700) {
+                    globalCtx->roomContext.unk7A[1] = 255;
+                    globalCtx->roomContext.unk7A[0] = 0;
+                    THIS->state++;
+                }
+            }
+            break;
+        case 2:
+            Actor_MarkForDeath(thisx);
+            break;
+    }
+}
+
+void DmRavine_Draw(Actor* thisx, GlobalContext* globalCtx) {
+}
--- a/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
+++ b/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
@@ -1,7 +1,7 @@
 /*
  * File: z_dm_ravine.c
  * Overlay: ovl_Dm_Ravine
- * Description: Link Riding Through Lost Woods Cutscene - Tree Trunk
+ * Description: Lost Woods Cutscene - Tree Trunk
  */
 
 #include "z_dm_ravine.h"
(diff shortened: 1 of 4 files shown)


src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -1,3 +1,9 @@
+/*
+ * File: z_dm_ravine.c
+ * Overlay: ovl_Dm_Ravine
+ * Description: Link Riding Through Lost Woods Cutscene - Tree Trunk
```
Zelllll, line 4: Just call it `Lost Woods Cutscene - Tree Trunk`

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -6,29 +12,68 @@
...
-*/
+const ActorInit Dm_Ravine_InitVars = { ACTOR_DM_RAVINE,
+                                       ACTORCAT_ITEMACTION,
+                                       FLAGS,
+                                       OBJECT_KEIKOKU_DEMO,
+                                       sizeof(DmRavine),
+                                       (ActorFunc)DmRavine_Init,
+                                       (ActorFunc)DmRavine_Destroy,
+                                       (ActorFunc)DmRavine_Update,
+                                       (ActorFunc)DmRavine_Draw };
```
Zelllll, line 27: Add a comma after `DmRavine_Draw ` and reformat

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -6,29 +12,68 @@
...
+void DmRavine_Init(Actor* thisx, GlobalContext* globalCtx) {
+    u8 flag;
+
+    flag = gSaveContext.perm.weekEventReg[0];
+    if (((flag & 0x10) | cREG(0)) != 0) {
+        Actor_MarkForDeath(thisx);
+        return;
+    }
+
+    THIS->loaded = false;
```
Zelllll, line 38: instead of THIS->.... make a temp before `u8 flag` that is `DmRavine* this = THIS;`, and then just use `this->...`

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -6,29 +12,68 @@
...
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Init.asm")
+void DmRavine_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Destroy.asm")
+// This function is unused, but takes two word-sized arguments
+// `thisx` & `globalCtx` are guesses; but it could be something else
+void DmRavine_Action(Actor* thisx, GlobalContext* globalCtx) {
```
Zelllll, line 51: This should simply be `DmRavine_Destroy`, without any comment. The function is not unused, it just does not have any action in this specific actor.

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -6,29 +12,68 @@
...
+                globalCtx->roomContext.unk7A[1] = 254;
+                if (globalCtx->csCtx.frames > 700) {
+                    globalCtx->roomContext.unk7A[1] = 255;
+                    globalCtx->roomContext.unk7A[0] = 0;
+                    THIS->state++;
+                }
+            }
+            break;
+        case 2:
+            Actor_MarkForDeath(thisx);
```
Zelllll, line 73: try using `&this->actor` here once you add `DmRavine* this = THIS;` here as well.

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.h
```diff
@@ -7,7 +7,10 @@ struct DmRavine;
 
 typedef struct DmRavine {
     /* 0x000 */ Actor actor;
-    /* 0x144 */ char unk_144[0x8];
+    /* 0x144 */ void (*actionFunc)(Actor*, GlobalContext*); // unused
+    /* 0x148 */ u8 state;
+    /* 0x149 */ u8 loaded;
```
Zelllll, line 12: I'm not sure if this is the best name but I'm also not entirely sure what this is doing to begin with

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.h
```diff
@@ -7,7 +7,10 @@ struct DmRavine;
 
 typedef struct DmRavine {
     /* 0x000 */ Actor actor;
-    /* 0x144 */ char unk_144[0x8];
+    /* 0x144 */ void (*actionFunc)(Actor*, GlobalContext*); // unused
```
Zelllll, line 10: Do `typedef void (*DmRavineActionFunc)(struct EnTest3*, struct GlobalContext*);` instead, and maybe make a comment like `// Although this is unused, it is set to the actor's destroy function for some reason`, which is what appears to be happening which is super weird lol

Zelllll (changes requested):
Nice job on your first file! Overall great work, but I left a few comments of things to address.

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.h
```diff
@@ -7,7 +7,10 @@ struct DmRavine;
...
+    /* 0x148 */ u8 state;
+    /* 0x149 */ u8 loaded;
+    /* 0x14a */ u8 pad_14a[2];
```
Zelllll, line 13: Make sure the hex is capital here

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
zbanks: The diff is a bit misleading: this function was previously `func_80A2E838`, not `DmRavine_Destroy` (but the contents of the functions happen to be identical)

Zelllll: So `DmRavine_Destroy` is still in the actor?

jpburnett: It would probably just be something like DmRavine_DoNothing(DmRavine* this, GlobalContext* globalCtx)

zbanks: Yeah

kyleburnette: I agree with this function being renamed DmRavine_DoNothing

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.h
```diff
@@ -5,9 +5,14 @@
 
 struct DmRavine;
 
+typedef void (*DmRavineActionFunc)(Actor*, GlobalContext*);
```
kyleburnette, line 8: Action func typedefs generally have the form (for example): `typedef void (*ArmsHookActionFunc)(struct ArmsHook*, GlobalContext*);`

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -19,16 +25,57 @@ const ActorInit Dm_Ravine_InitVars = {
...
+    this->actionFunc = &DmRavine_Action;
+}
+
+void DmRavine_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Destroy.asm")
+// This function is unused, but takes two word-sized arguments
+// `thisx` & `globalCtx` are guesses; but it could be something else
+void DmRavine_Action(Actor* thisx, GlobalContext* globalCtx) {
```
kyleburnette, line 52: My other comment on this seems to be gone, but I agree with this being `DmRavine_DoNothing(DmRavine* this, GlobalContext* globalCtx)`

kyleburnette (changes requested):

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -19,16 +25,57 @@ const ActorInit Dm_Ravine_InitVars = {
...
+        Actor_MarkForDeath(&this->actor);
+        return;
+    }
+
+    this->loaded = false;
+    globalCtx->roomContext.unk7A[0] = 1;
+    globalCtx->roomContext.unk7A[1] = 0;
+    this->state = 0;
+    Actor_SetScale(&this->actor, 1.0f);
+    this->actionFunc = &DmRavine_DoNothing;
```
kyleburnette, line 44: The & is unneeded here

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -6,10 +12,10 @@
 
 void DmRavine_Init(Actor* thisx, GlobalContext* globalCtx);
 void DmRavine_Destroy(Actor* thisx, GlobalContext* globalCtx);
+void DmRavine_DoNothing(DmRavine* thisx, GlobalContext* globalCtx);
```
kyleburnette, line 15: We generally use `this` instead of `thisx` in action funcs

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -19,16 +25,57 @@ const ActorInit Dm_Ravine_InitVars = {
...
+    Actor_SetScale(&this->actor, 1.0f);
+    this->actionFunc = &DmRavine_DoNothing;
+}
+
+void DmRavine_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Destroy.asm")
+// This function is unused, but takes two word-sized arguments
+// `thisx` & `globalCtx` are guesses; but it could be something else
```
kyleburnette, line 51: Empty "DoNothing" funcs are pretty common, so it's probably safe to remove these comments

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.h
```diff
@@ -5,9 +5,14 @@
...
 
+typedef void (*DmRavineActionFunc)(struct DmRavine*, GlobalContext*);
+
 typedef struct DmRavine {
     /* 0x000 */ Actor actor;
-    /* 0x144 */ char unk_144[0x8];
+    /* 0x144 */ DmRavineActionFunc actionFunc;
+    /* 0x148 */ u8 state;
+    /* 0x149 */ u8 loaded;
+    /* 0x14A */ u8 pad_14A[2];
```
kyleburnette, line 15: I meant to mention this in my first review, but I think I didn't - since this is such a small pad and it's at the end of the actor, in my experience, these are not usually necessary. I don't know the details, but the pad is usually handled automatically. I would try removing pad_14A and see if you still get OK. If it is required, it should be something like `UNK_TYPE1 unk14A[2];`

kyleburnette (changes requested):

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.h
Kenix3: It's actually redundant against the state. I would recommend the name `isLoaded`

Kenix3 (changes requested):

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -19,16 +25,55 @@ const ActorInit Dm_Ravine_InitVars = {
...
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/func_80A2E838.asm")
+void DmRavine_Update(Actor* thisx, GlobalContext* globalCtx) {
+    DmRavine* this = THIS;
+    RoomContext* roomCtx;
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Update.asm")
+    switch (this->state) {
+        case 0:
```
Kenix3, line 58: These states may make sense as an enum. State 0 is doing nothing. State 1 is waiting for the cutscene to hit a specific stage, then moving to state 2 which sets the actor to be removed.

Kenix3 (changes requested):

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
```diff
@@ -19,16 +25,55 @@ const ActorInit Dm_Ravine_InitVars = {
...
+void DmRavine_DoNothing(DmRavine* this, GlobalContext* globalCtx) {
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/func_80A2E838.asm")
+void DmRavine_Update(Actor* thisx, GlobalContext* globalCtx) {
+    DmRavine* this = THIS;
+    RoomContext* roomCtx;
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Dm_Ravine_0x80A2E7A0/DmRavine_Update.asm")
+    switch ((DmRavineState) this->state) {
```
Kenix3, line 57: Just a comment here, the cast is unnecessary.

Kenix3 (approved):

src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
zbanks: Modern compilers can emit a warning if you're missing an enum case w/ `-Wswitch-enum` -- I know we're not using them now, but if we ran the code through it in the future, it could catch other mistakes?
(I could just remove it though if it's really not convention)

zbanks: (The OOT project runs the code through GCC to check for warnings I think?)

Kenix3: It doesn't, but you make a good point. Good idea to keep it.

PR #121, opened 2021-05-03
Use decompiled overlays when building ROM

zbanks:
I brought up an issue on the Discord -- I'm not sure this initial PR is the best fix, but hopefully it can start a discussion?

The the Makefile on master right now is not incorporating the decompiled overlay objects back into the `.z64` ROM, so I needed to use this patch to check that my other PR matched.

---

It seems like the built code/assets in `code.elf` should not be copied into `build/baserom/...` (over the original ROM's files) but instead into a `build/decomp/...` tree or similar. (This is how it did worked before #117) `dmadata_table.txt` would also need to be updated to read from the correct location.

I'm not sure if there are other sections that also need to be copied?


zbanks:
I don't think this is ready to merge as-is, but I think it is ready for feedback. 

- `makerom_files.txt` & `makerom_uncompressed_files.txt` contained a lot of the same data in `dmadata_table.txt`, so I added a small python script to generate this information into `build/`.
- Segments are no longer dumped out of `code.elf` into `build/baserom/`, instead they are put in `build/binary/`. (I think "binary" is more explicit than "bin" - which imo would be ambiguous here)
- Previously, `linker_scripts/dmadata_script.txt` was checked in, but was generated by `dmadata.py`. I deleted it from git, and it is now generated into the `build/` folder.
    - I called the un-preprocessed version `dmadata_script.ld.pre`, I figured although `.ld.cpp` may technically be more accurate, it'd be way more confusing.
    - I think I'd like to go rename the other 2 linker scripts to `.ld` or `.ld.pre`, but that's not a blocker.
- I also introduced some sentinel files (`dep`). I ended up needing these to make incremental builds work smoothly? (Without them, there were a lot of steps that were getting re-triggered on every build.) 
    - If this style isn't welcome, I can try to fiddle with the Makefile more to try to avoid having them? 
    - This also avoids relying on GNU Make extensions, like order-only deps

zbanks:
I wanted to give an example of the current behavior on `master`, and what this PR tries to address.
Applying the following diff to `master` *should* cause the build to fail: it should change the output ROM -- but it doesn't.
```diff
diff --git a/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c b/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
index abea0ce..ee76bb6 100644
--- a/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
+++ b/src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c
@@ -6,7 +6,7 @@
 
 #include "z_dm_ravine.h"
 
-#define FLAGS 0x00000030
+#define FLAGS 0x12345678
 
 #define THIS ((DmRavine*)thisx)
```

On `master`, all of the overlay sections in the ROM are overwritten by the versions from the baserom. As a result, the decompiled overlays aren't being fully checked for matching.

`37510bc` is the smallest commit that can fix this (but is sort of a kludge - the file naming doesn't make sense). I don't love how large this PR is, but I think the changes help make the build a bit more stable. (My previous comment has more details)

AngheloAlf (approved):
I just tested this, and I can confirm the current master branch will build an OK rom even with dumb changes like the one zbanks posted.
This should be merged ASAP.

Kenix3:
Looks good to me!

AngheloAlf (approved):

PR #123, opened 2021-05-04
Ovl Bg_Ctower_Gear OK

hensldm:


src/overlays/actors/ovl_Bg_Ctower_Gear/z_bg_ctower_gear.c
```diff
@@ -21,18 +39,189 @@ const ActorInit Bg_Ctower_Gear_InitVars = {
...
+        for(i=0; i<4; i++) {
+            if ((u32) Rand_Next() >= 0x40000000) {
+                sp70.x = D_80AD3270[i].x - (Rand_ZeroOne() * 30.0f);
+                sp70.y = D_80AD3270[i].y;
+                sp70.z = D_80AD3270[i].z;
+                SysMatrix_MultiplyVector3fByState(&sp70, &sp7C);
+                sp7C.x += this->dyna.actor.world.pos.x + ((Rand_ZeroOne() * 20.0f) - 10.0f);
+                sp7C.y += this->dyna.actor.world.pos.y;
+                sp7C.z += this->dyna.actor.world.pos.z + ((Rand_ZeroOne() * 20.0f) - 10.0f);
+                EffectSsGSplash_Spawn(GlobalContext, &sp7C, NULL, NULL, 0, ((u32) Rand_Next() >> 25) + 340);
```
jpburnett, line 114: Based on the context of `EffectSsGSplash_Spawn`,  sp7C is a position vector, perhaps the position of the splash. sp70 could be the starting position of the cogs, I am not positive on that though.

jpburnett (commented):
:COGGERS:
nice work on this. just a couple thoughts on the vectors here.

src/overlays/actors/ovl_Bg_Ctower_Gear/z_bg_ctower_gear.c
hensldm: sp7C makes sense to be splash spawn pos, but still not sure on sp70. I don't think it is the starting position of the cog it get initialized to different values all in the same frame. Not to mention the x coordinate is randomized as well.

Kenix3 (approved):
looks good to me

PR #126, opened 2021-05-10
Port `actorfixer.py` to MM

AngheloAlf:
`actorfixer.py` is a small python script made to fix automatically problems produced by functions renames that could affect actors.

Currently, the only function it has is `Actor_GetSwitchFlag` -> `Flags_GetSwitch`

I suggest to everyone doing bigger function renames to add their renames to this script when PR their changes.

### DIFF
--- a/tools/actorfixer.py
+++ b/tools/actorfixer.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+
+import os
+import argparse
+
+# "old": "new"
+animdict ={
+    "Actor_GetSwitchFlag": "Flags_GetSwitch",
+}
+
+def replace_anim(file):
+    with open(file,'r',encoding = 'utf-8') as infile:
+        srcdata = infile.read()
+
+    funcs = list(animdict.keys())
+    fixes = 0
+    for func in funcs:
+        newfunc = animdict.get(func)
+        if(newfunc is None):
+            print("How did this happen?")
+            return -1
+        if(func in srcdata):
+            fixes += 1
+            print(func)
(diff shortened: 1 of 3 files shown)


Kenix3 (approved):
Looks good to me!

tools/actorfixer.py
```diff
@@ -0,0 +1,54 @@
...
+    if(fixes > 0):
+        print('Changed', fixes,'entr' + ('y' if fixes == 1 else 'ies') + ' in',file)
+        with open(file, 'w', encoding = 'utf-8', newline = '\n') as outfile:
+            outfile.write(srcdata)
+    return 1
+
+def replace_anim_all(repo):
+    for subdir, dirs, files in os.walk(repo + os.sep + 'src'):
+        for filename in files:
+            if(filename.endswith('.c')):
```
isghj5, line 36: thought/suggestion: might want to extend this to `.h` files in the future, in the event filetypes are changed with this script

Otherwise looks good to me

Zelllll (approved):

tools/actorfixer.py
AngheloAlf: I don't know if I understood your suggestion.
This script is aimed to fix the name of called functions which were renamed  by somebody else. Each one of those functions should be declared once in their respective `.h` file, so there shouldn't be a need to fix the `.h` files. Was this what you meant?

isghj5: Function names being changed and causing merge conflicts seems like the same issue caused by types being renamed (like BgMesh changed to CollisionMesh) so I figured the script could handle both but its fine

PR #127, opened 2021-05-11
Small env fixes: git submodules & python3

zbanks:
Using `/usr/bin/env python3` is preferred over `/usr/bin/python3` because it is compatible with virtualenvs, etc. It picks the version of `python3` currently on the `PATH`, rather than the system-wide default.

For the git submodules, I had two issues (at least on git 2.7.4 from Ubuntu 16.04, which is admittedly pretty old!)

- The ZAPD repo is currently pointing at a commit not on the master branch, which has to be specifically fetched. Adding `--remote` to the update command fixes this.
- The decomp-permuter has primary branch `main` not `master` like this repo, so this has to be specified in `.gitmodules` when using `--remote`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,7 @@
 [submodule "tools/decomp-permuter"]
 	path = tools/decomp-permuter
 	url = https://github.com/simonlindholm/decomp-permuter
+	branch = main
 [submodule "ZAPD"]
 	path = tools/ZAPD
 	url = https://github.com/zeldaret/ZAPD.git
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,10 @@ distclean: clean
 	rm -rf assets/src baserom/
 
 setup:
-	git submodule update --init --recursive --remote
+	# Initialize submodules, fetching commit in case it is not on the default branch
+	-git submodule update --init --recursive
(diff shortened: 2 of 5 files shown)


AngheloAlf:
I think this PR isn't building because it is updating ZAPD somehow, and the repo needs a few fixes to work properly with the latest version.
I was waiting for master branch to build again to PR a fix, so I'm going to do it soon.

zbanks:
Ah you're right; I don't actually want `--remote` on the `git submodule update...` - that is like a pull, not a fetch. I'll rework this.

zbanks:
Okay - fixed it up.

I couldn't find any prior art on how to solve this exact problem with git submodules online, but I think this is a pretty good solution. It's maybe a bit redundant if all of your submodule commits *are* on the default branch, but this part of the Makefile isn't frequently executed and it should just make it more robust.

I'm leaving the "`branch = main`" change in the `.gitmodules` even though it's not required to build -- it makes it easier for contributors to update the submodule.

I tested this with a fresh clone, and works on my machine (up until it tries to _build_ ZAPD, but that'll get fixed in [my other PR](https://github.com/zeldaret/ZAPD/pull/138)).

Kenix3 (approved):
Looks good to me.

PR #131, opened 2021-05-12
Decompile En_Test3 (Kafei); Unfinished

Zelllll:
I decided not to finish this actor but it has about 1000 lines already, hopefully eventually someone else is interested in it! I am making this PR now though because I documented a *ton* of player's struct and a lot of other things though which would be valuable to merge.


Zelllll:
this wont build ok and i dont know why, whatever

PR #132, opened 2021-05-12
Extract assets in `make setup`, don't extract them to `assets/src/` and build them in `make`

AngheloAlf:
- Extract the assets to `assets/` instead of `assets/src/`.
  - For `scenes`, they would be extracted to `assets/scenes/`
- Do the extraction in the `make setup` step instead of doing it in the build step.
- Port `extract_assets.py`. This is done to prevent increasing the complexity of the Makefile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,14 +31,9 @@ tools/ido_recomp/* binary
 ctx.c
 
 # Assets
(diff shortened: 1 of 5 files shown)


Kenix3 (approved):
Looks good to me!

PR #134, opened 2021-05-13
z_view matching

zbanks:
There was some existing `NON_MATCHING` sections, but they were matched once I converted them to use the macros.

There's definitely a pattern with `gDPPipeSync` + `View_WriteScissor` that looks like a macro, but there's enough variation that it's hard for me to pin down. I figured maybe it's worth waiting until more files are decompiled? (I didn't see any similar pattern in the OOT code.)

I could guess at some parts of the functions' names, but I don't know how to succinctly describe the differences between some of them. It might make more sense to wait until the call sites are decompiled?


tables/dmadata_table.txt
```diff
@@ -31,7 +31,7 @@
     ('build/baserom/assets/textures/nes_font_static', '', 0x1000, 0),
     ('build/baserom/assets/textures/message_data_static', '', 0x1000, 0),
     ('build/baserom/assets/textures/staff_message_data_static', '', 0x1000, 0),
-    ('build/baserom/code', 'build/comp/code.yaz0', 0x1000, 0),
+    ('build/binary/code', 'build/comp/code.yaz0', 0x1000, 0),
```
zbanks, line 34: This is actually a bug like #121 - where the recompiled `code` section wasn't being incorporated into the ROM.

src/code/z_view.c
```diff
@@ -279,137 +279,231 @@ void View_RenderView(View* view, s32 uParm2) {
...
+    guOrtho(projection, gScreenWidth * -0.5f, gScreenWidth * 0.5f, gScreenHeight * -0.5f, gScreenHeight * 0.5f,
+            view->zNear, view->zFar, view->scale);
+
+    view->projection = *projection;
+
+    gSPMatrix(gfx++, projection, G_MTX_LOAD | G_MTX_PROJECTION);
+    *gfxp = gfx;
 
+    return 1;
+}
```
zbanks, line 509: This was the trickiest function to get matching, I'm not super happy with it. 
I couldn't get it to match using the `OPEN_DISPS`/`CLOSE_DISPS` macros -- but maybe that's expected? Other functions which take a `Gfx**` also do not use these macros: probably the *caller* uses them?

src/code/z_view.c
```diff
@@ -279,137 +279,231 @@ void View_RenderView(View* view, s32 uParm2) {
...
 
     width = view->viewport.rightX - view->viewport.leftX;
     height = view->viewport.bottomY - view->viewport.topY;
     aspect = (f32)width / (f32)height;
 
     guPerspective(projection, &view->normal, view->fovy, aspect, view->zNear, view->zFar, view->scale);
-
     view->projection = *projection;
+    //! @bug: This cast of `projection` is invalid
+    View_StepQuake(view, (RSPMatrix*)projection);
```
zbanks, line 312: This is very weird. I think `View_StepQuake` definitely takes an `RSPMatrix *`. It calls `SysMatrix_FromRSPMatrix` & `SysMatrix_GetStateAsRSPMatrix` which aren't decompiled yet but make much more sense with `RSPMatrix` than `Mtx`.

If `view->projection` was an `RSPMatrix` instead of an `Mtx`, that would mean `guPerspective`, `guOrtho`, etc. would have to operate on `RSPMatrix` -- which doesn't seem to work out either?

Thar0: `RSPMatrix` and `Mtx` are actually the exact same type, but iirc some code required certain accesses that `Mtx` could not accommodate. The matrix types are a bit of a mess right now, these two types should probably be a union ideally, since they represent the exact same underlying data. This is likely best addressed in a separate PR however, since all usages would need cleaning up.

AngheloAlf (approved):

Kenix3 (approved):
Looks good to me. Make sure you run format.sh and resolve the merge conflict.

PR #135, opened 2021-05-13
ovl_Bg_Ladder: OK (Wooden Ladder)

zbanks:
- Formatted with `./format.sh`
- Verified the dlists in Z64Utils


src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -19,22 +37,129 @@ const ActorInit Bg_Ladder_InitVars = {
...
+    (ActorFunc)BgLadder_Draw,
 };
-*/
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Bg_Ladder_0x80A4D9F0/BgLadder_Init.asm")
+static InitChainEntry sInitChain[] = {
+    ICHAIN_VEC3F_DIV1000(scale, 100, ICHAIN_STOP),
+};
+
+static u32 sLadderDLists[] = {
```
isghj5, line 47: Looks like a list of Gfx not u32

That's all I saw, the draw function looks weird to me but I think that's just me.

Looks good!

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -19,22 +37,129 @@ const ActorInit Bg_Ladder_InitVars = {
...
+
+void BgLadder_Init(Actor* thisx, GlobalContext* globalCtx) {
+    BgLadder* this = THIS;
+    BgLadderSize size;
+
+    Actor_ProcessInitChain(&this->dyna.actor, sInitChain);
+
+    // params is treated as two bytes: switchFlag, size
+    this->switchFlag = (thisx->params >> 8) & 0xFF;
+    thisx->params &= 0xFF;
```
hensldm, line 62: Does `this->dyna.actor.params` not match? If `thisx` is required to match, a comment stating such would be nice.

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -19,22 +37,129 @@ const ActorInit Bg_Ladder_InitVars = {
...
+    } else {
+        Scene_SetRenderModeXlu(globalCtx, 1, 0x02);
+        gfx = POLY_XLU_DISP;
+    }
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Bg_Ladder_0x80A4D9F0/func_80A4DCCC.asm")
+    gSPDisplayList(&gfx[0], &sSetupDL[6 * 0x19]);
+    gDPSetEnvColor(&gfx[1], 255, 255, 255, this->alpha);
+    gSPMatrix(&gfx[2], Matrix_NewMtx(globalCtx->state.gfxCtx), G_MTX_LOAD);
+    gSPDisplayList(&gfx[3], sLadderDLists[this->dyna.actor.params]);
```
hensldm, line 156: Does it still match if you try using `gfx++` instead of indexing, then set `POLY_OPA_DISP` or `POLY_XLU_DISP` to gfx at the end?

hensldm (commented):
Just a few thoughts. Other then that looks good to me.

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
zbanks: It's weird, I could make 1 of the 3 lines into `this->dyna.actor.params`, but not all 3 (or 2). 
It seemed _worse_ to mix the two styles so I left them all `thisx->params`? I can add a comment though

zbanks:
I couldn't easily address some of the suggestions and keep it matching -- I agree the draw function in particular is super weird. 

(I also don't love the init fn either, the way `BcCheck3_BgActorInit` is in the middle of parsing `params`, but I tried a few ways and couldn't separate them out, with either `this` or `thisx`)

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -8,8 +13,21 @@ void BgLadder_Init(Actor* thisx, GlobalContext* globalCtx);
...
+
+extern Gfx object_ladder_12rung_dlist[];
+extern Gfx object_ladder_16rung_dlist[];
+extern Gfx object_ladder_20rung_dlist[];
+extern Gfx object_ladder_24rung_dlist[];
+
+extern CollisionHeader object_ladder_12rung_collision;
+extern CollisionHeader object_ladder_16rung_collision;
+extern CollisionHeader object_ladder_20rung_collision;
+extern CollisionHeader object_ladder_24rung_collision;
```
AngheloAlf, line 29: I recommend leaving the addresses, at least as a comment. Doing that is way easier to be able to know which symbol is which one when decompiling the objects.

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -19,22 +37,130 @@ const ActorInit Bg_Ladder_InitVars = {
...
+        this->alpha = 5;
+        func_800C62BC(globalCtx, &globalCtx->colCtx.dyna, this->dyna.bgId);
+        this->dyna.actor.draw = NULL;
+        this->action = BgLadder_ActionWait;
+    }
+}
+
+void BgLadder_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    BgLadder* this = THIS;
+    BgCheck_RemoveActorMesh(globalCtx, &globalCtx->colCtx.dyna, this->dyna.bgId);
```
AngheloAlf, line 96:
```suggestion
    BgLadder* this = THIS;

    BgCheck_RemoveActorMesh(globalCtx, &globalCtx->colCtx.dyna, this->dyna.bgId);
```

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -19,22 +37,130 @@ const ActorInit Bg_Ladder_InitVars = {
...
+    }
+}
+
+void BgLadder_ActionIdle(BgLadder* this, GlobalContext* globalCtx) {
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Bg_Ladder_0x80A4D9F0/BgLadder_Destroy.asm")
+void BgLadder_Update(Actor* thisx, GlobalContext* globalCtx) {
+    BgLadder* this = THIS;
+    this->action(this, globalCtx);
```
AngheloAlf, line 136:
```suggestion
    BgLadder* this = THIS;

    this->action(this, globalCtx);
```

AngheloAlf (commented):
Looks good. Just a few minor comments.

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
Kenix3: Agreed. We are going to be leaving object segment symbols as addresses until we are ready for object reconstruction.

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -19,22 +37,130 @@ const ActorInit Bg_Ladder_InitVars = {
...
+
+void BgLadder_Init(Actor* thisx, GlobalContext* globalCtx) {
+    BgLadder* this = THIS;
+    BgLadderSize size;
+
+    Actor_ProcessInitChain(&this->dyna.actor, sInitChain);
+
+    // `params` is treated as two bytes: switchFlag, size
+    // Has to be `thisx->params` to match
+    this->switchFlag = (thisx->params >> 8) & 0xFF;
```
Kenix3, line 62: Let's make macros for params access.

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
```diff
@@ -19,22 +37,130 @@ const ActorInit Bg_Ladder_InitVars = {
...
+    thisx->params &= 0xFF;
```
Kenix3, line 63: Would this match as `thisx->params = thisx->params & 0xFF`? Trying to see if we can convert this one to a macro as well.

Kenix3 (changes requested):
A few small things. Good job overall!
Don't forget to run format.sh now that the repository can be formatted.

src/overlays/actors/ovl_Bg_Ladder/z_bg_ladder.c
zbanks: Just checking, I should define `#define GET_BGLADDER_SWITCHFLAG(this) ...` & `#define GET_BGLADDER_SIZE(this) ...` in the header?

(If you don't like the naming scheme, I think there's only ~5 overlays that follow this, so now would be the time to change it I guess)

Kenix3: Yes, that's correct.

Kenix3 (approved):

PR #137, opened 2021-05-14
code_0x8017FEB0 OK

AngheloAlf:
~~I'm a bit unsure about `D_801E01E0`. It seems to be used only by `atan`. I don't know how to import that variable to code.~~

I also fixed the signature of `atan_flip`.

Update: 
- I renamed the functions and the file to match its OoT counterpart.
- I also added the renamed functions to `actorfixer.py`


AngheloAlf:
This is fixed now!

src/code/code_0x8017FEB0.c
```diff
@@ -1,12 +1,120 @@
...
+};
 
-#pragma GLOBAL_ASM("./asm/non_matchings/code/code_0x8017FEB0/atans.asm")
+u16 atans_first_8th(f32 opposite, f32 adjacent) {
+    return atan_first_8th_array[(s32)((opposite / adjacent) * 0x400)];
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/code/code_0x8017FEB0/atan.asm")
+s16 atans(f32 opposite, f32 adjacent) {
+    s32 phi_v1;
```
hensldm, line 66: `angle` seems like a better name then `phi_v1`.

hensldm (approved):
Looks good to me. Only saw one small thing.

Zelllll (changes requested):
If these functions exist in oot they should probably be using the same names, and everything should be camelCase. Also the `atan_first_8th_array` should be `static u16 sAtanFirstEighth[]` or something instead

AngheloAlf:
The equivalent file seems to be [sys_math_atan.c](https://github.com/zeldaret/oot/blob/master/src/code/sys_math_atan.c), so I'll rename the functions and the filename too

AngheloAlf:
Update: 
- I renamed the functions and the file to match its OoT counterpart.
- I also added the renamed functions to `actorfixer.py`

Zelllll (approved):
Looks good now

Kenix3 (approved):
Looks great to me! Don't forget to run format.sh on this.

PR #138, opened 2021-05-16
partially finish `code_0x8012EC80`, document all of gSaveContext

Zelllll:
I am not finishing `code_0x8012EC80` for now since it had some tricky functions to match, but this pr documents almost all of the variables in save context which should probably get merged in soon. Some of the variables seem to only be used in MM debug, like the OSTime's, but not entirely sure yet. I looked at almost every usage of save context in the debug rom and retail to try to figure out the variables, although currently some are sort of guesses, but as decomp evolves any potentially wrong types will fix themselves (like signed vs unsigned).


src/overlays/actors/ovl_En_Ginko_Man/z_en_ginko_man.c
```diff
@@ -257,17 +257,17 @@ void EnGinkoMan_DepositDialogue(EnGinkoMan *this, GlobalContext *globalCtx) { //
     case 32: // "Ah, yes...[Link], right?  If I remember, you're the little guy who deposited [rupees]."
     case 50: // "Your deposits total [rupees]."
         if (this->choiceDepositWithdrawl == GINKOMAN_CHOICE_DEPOSIT) {
-            if ((u32) (gSaveContext.perm.bankRupees & 0xFFFF) >= 0x1388) {
+            if ((u32) (gSaveContext.roomInf[127][0] & 0xFFFF) >= 0x1388) {
```
louist103, line 260: Decimal.

src/code/code_0x8012EC80.c
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    if (gSaveContext.roomInf[125][num] & gBitFlags[arg0 - (num << 5)]) {
+        return 1;
+    }
+
+    return 0;
+}
+
+/* static */ u16 D_801C2380[12][6] = {
```
louist103, line 232: I guess MM didn't get the new asm processor?

src/code/z_en_item00.c
```diff
@@ -1037,37 +1037,37 @@ void Item_DropCollectibleRandom(GlobalContext* globalCtx, Actor* fromActor, Vec3
...
         if (dropId == ITEM00_FLEXIBLE) {
-            if (gSaveContext.perm.unk24.currentLife < 0x11) {
+            if (gSaveContext.health < 0x11) {
                 Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_ELF, spawnPos->x, spawnPos->y + 40.0f, spawnPos->z, 0, 0, 0, 2);
                 func_800F0568(globalCtx, spawnPos, 0x28, 0x28E7);
                 return;
             }
 
-            if (gSaveContext.perm.unk24.currentLife < 0x31) {
+            if (gSaveContext.health < 0x31) {
```
louist103, line 1046: Health should be decimal IMO.

louist103 (changes requested):
Looks good to me but I probably missed a bunch of stuff since I'm not too familiar with MM's styling.

src/code/code_0x8012EC80.c
Zelllll: nope

src/code/z_en_item00.c
Zelllll: i don't see why 🤔 really there should just be a macro for health

AngheloAlf (commented):
One downside I see with this change is we "lose" a bit of information we currently have. 
For example, is nice to know that offset `0x0EDC` of the save context is `bankRupees`, but `roomInf[127][0]` hides that information we already had.
I think you should document the offsets of save context that we already have in the repo that are being deleted.

Zelllll:
Although it is unfortunate this is how it has to be. There are many portions of code, (which you can see I started on in this pr) that require this to be a 2d array like it is. Eventually macros for this would be good but I don't really think MM is at that stage right now so it doesn't make sense to do yet.

src/code/code_0x8012EC80.c
Kenix3: Why not try updating the submodule?

src/code/z_en_item00.c
Kenix3: Health should definitely be a decimal.

Kenix3:
> 
> 
> One downside I see with this change is we "lose" a bit of information we currently have.
> For example, is nice to know that offset `0x0EDC` of the save context is `bankRupees`, but `roomInf[127][0]` hides that information we already had.
> I think you should document the offsets of save context that we already have in the repo that are being deleted.

I agree. There is nothing preventing us from making macros. Losing information we already have would be bad in my opinion. Let's get some macros created for these.

src/code/z_en_item00.c
Zelllll: "should definitely be decimal" when the values dont make sense to begin with, a single heart is a value of 10, so neither hex or decimal is going to make sense until there is a macro for health

AngheloAlf: If a heart is a multiple of ten, then we should always have health in decimal imo.

Zelllll: Is this even my change to make though? Doesn't really have much to do with save context as it was this way already.

Zelllll:
hopefully someone can figure this out because like everything I PR it sits here forever and then gets completely broken by merging in master and doesn't match anymore, I thought we were supposed to be concerned about merge conflicts.

include/z64save.h
```diff
@@ -0,0 +1,199 @@
...
+    /* 0x0004 */ u8 equippedMask;   // "player_mask"
+    /* 0x0005 */ u8 unk_05;         // "opening_flag"
+    /* 0x0006 */ u8 unk_06;
+    /* 0x0007 */ u8 linkAge;   // "link_age"
+    /* 0x0008 */ s32 cutscene; // "day_time"
+    /* 0x000C */ u16 time;     // "zelda_time"
+    /* 0x000E */ u16 owlSaveLocation;
+    /* 0x0010 */ u32 isNight;        // "asahiru_fg"
+    /* 0x0014 */ u32 unk_14;         // "change_zelda_time"
+    /* 0x0018 */ u32 day;            // "totalday"
```
AngheloAlf, line 55:
```suggestion
    /* 0x0010 */ s32 isNight;        // "asahiru_fg"
    /* 0x0014 */ u32 unk_14;         // "change_zelda_time"
    /* 0x0018 */ s32 day;            // "totalday"
```

AngheloAlf (changes requested):

include/z64save.h
Zelllll: I don't think this is supposed to be signed.

AngheloAlf: https://github.com/zeldaret/mm/blob/cf3eb3b2e19bb60ec9d93b04a57127585e8ac6df/include/z64.h#L649

AngheloAlf: Out of curiosity, I checked out your branch, made the change I pointed out and now it builds OK. You should really try.

Zelllll: Oh huh strange, maybe that was a mistake when I made it a u32, let me try that

src/overlays/actors/ovl_Obj_Dinner/z_obj_dinner.c
```diff
@@ -26,7 +26,7 @@ extern Gfx D_060011E0[];
 void ObjDinner_Init(Actor* thisx, GlobalContext* globalCtx) {
     ObjDinner* this = THIS;
 
-    if (gSaveContext.perm.isNight != 1 || (CURRENT_DAY == 3 && gSaveContext.perm.weekEventReg[0x16] & 1)) {
+    if (gSaveContext.isNight != true || (CURRENT_DAY == 3 && gSaveContext.weekEventReg[0x16] & 1)) {
```
AngheloAlf, line 29: I wonder if this works :thinking: 
```suggestion
    if (!gSaveContext.isNight || (CURRENT_DAY == 3 && gSaveContext.weekEventReg[0x16] & 1)) {
```

src/code/z_en_item00.c
Kenix3: No, that's fair that it isn't your change.

Zelllll: Alright, I agree that doing something about the health thing is fine at some point, but it's out of the scope of the pr so I'm going to resolve this for now.

Kenix3 (approved):

PR #139, opened 2021-05-16
Hotfix: Fix running `make init` without `-jN`

AngheloAlf:
- Add `cp -r baserom/ build/baserom/` at the end of the `setup` target.
  - This is just a hacky way to fix the dependency bug we currently have. Hopefully somebody will figure out a proper way to fix this soon.
- Add an `assembly` target.
  - This should allow to not regenerate every assembly every time `make` is run, making possible to edit assembly files locally to do your own tests.
  - To regenerate the assemblies before compiling, just run `make assembly` before running `make`.
- Small cleaning of the `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,7 +209,7 @@ distclean: assetclean clean
 	$(RM) -rf baserom/ asm 
 
 ## Extraction step
-# The `cp -r baserom/ build/baserom/` is a temporary solution to the race condition/dependency bug we currently have.
+# TODO: The `cp -r baserom/ build/baserom/` is a temporary solution to the race condition/dependency bug we currently have.
 # It should properly fixed in the future.
 setup:
 	git submodule update --init --recursive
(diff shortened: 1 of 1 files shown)


hensldm:
This does indeed fix the `make init` issue, but id doesn't resolve the `make clean` `make` workflow.

hensldm:
Got an idea to address this. Give me a little time to test.

hensldm:
Here is my idea: https://github.com/hensldm/mm/blob/makehf/Makefile
Basically don't create build/baserom directories, instead use it as a target/dependency that just copies everything all in one pass.

AngheloAlf:
Sounds interesting. 
Did you test it? I would do it, but I have ran `make init`too many times today, and I'm a bit busy right now.

hensldm:
Yeah it works on my machine with or without `-jN`

AngheloAlf:
I added @hensldm to this PR. Thanks.

Kenix3 (approved):

PR #141, opened 2021-05-18
Format everything

AngheloAlf:
- Run the formatter in the whole repo.
- Explicitly specify the `clang-format` version in `format.sh` script. This should prevent different people formatting the same files in different ways.
  - The version is `clang-format-11`. To install it in a Debian/Ubuntu environment simply run
  ```bash
  apt install clang-format-11
  ```
- Add trailing commas to `InitVars`.


src/overlays/actors/ovl_En_M_Fire1/z_en_m_fire1.c
```diff
@@ -8,39 +8,37 @@ void EnMFire1_Init(Actor* thisx, GlobalContext* globalCtx);
...
-};
+const ActorInit En_M_Fire1_InitVars = { ACTOR_EN_M_FIRE1,
+                                        ACTORCAT_MISC,
+                                        FLAGS,
+                                        GAMEPLAY_KEEP,
+                                        sizeof(EnMFire1),
+                                        (ActorFunc)EnMFire1_Init,
+                                        (ActorFunc)EnMFire1_Destroy,
+                                        (ActorFunc)EnMFire1_Update,
+                                        NULL };
```
hensldm, line 19: This one was missed by your regex, because I removed the (ActorFunc) cast for the NULL. Sorry.

AngheloAlf: No problems. The good thing is you told us about this. Thanks you

format.sh
```diff
@@ -20,7 +20,7 @@ fi
 
 echo "Formatting C files. This will take a bit"
 echo "Running clang-format..."
-clang-format ${FORMAT_OPTS} src/**/*.c
+clang-format-11 ${FORMAT_OPTS} src/**/*.c
```
zbanks, line 23: Should we also format the `.h` files?

Kenix3:
See also https://github.com/zeldaret/mm/issues/148

format.sh
Kenix3: I don't think OOT does. In this case, I want to do the same thing.

Kenix3 (approved):

Issue #147, opened 2021-05-18
Reduce compile time warnings.


Kenix3:
We should fix up the code to remove as many compiler warnings as possible while still keeping OK.

Further, we should implement something like this PR for OOT until we are ready to elevate warnings to errors.
https://github.com/zeldaret/oot/pull/768

EllipticEllipsis:
I think we should perhaps be a little less stringent than this: in OoT there are several pretty horrible coding mistakes that cause warnings, that it would be possible to fix while maintaining OK, but OoT prefers not to since they are genuine mistakes in the original, not our own.

Kenix3:
> 
> 
> I think we should perhaps be a little less stringent than this: in OoT there are several pretty horrible coding mistakes that cause warnings, that it would be possible to fix while maintaining OK, but OoT prefers not to since they are genuine mistakes in the original, not our own.

These would probably not be OK? Either way, if it's OK, how can you prove the original code had the mistake?

For MM I would rather fix the defect, as long as it's OK.

Kenix3:
https://github.com/zeldaret/mm/pull/152

PR #150, opened 2021-05-18
Make Jenkins check if a PR will add new warnings

AngheloAlf:
Adds two new stages to Jenkins. The first one is used to check warnings produced by  `make setup, while the second one checks the warnings produced by `make all`. This way we will always know if the amount of warnings increases.

It is done by redirecting `stderr` of each command to a file, and then compare the lines amount to the current line count (stored in a file in `tools/warning_count/`).

If by any reason we need to increase the accepted amount of warnings in the repo, then we can just edit the corresponding warnings file in `tools/warning_count/`. The script `tools/warnings_count/update_current_warnings.sh` can do this automatically.

To test the amount of warnings locally, you can use the script `tools/warnings_count/create_new_warnings.sh` to create the new warnings files and compare them with the `tools/warnings_count/compare_warnings.py` python script.

This PR is being ported from: https://github.com/zeldaret/oot/pull/768

Related issue: #147


Jenkinsfile
```diff
@@ -10,9 +10,26 @@ pipeline {
                 sh 'cp /usr/local/etc/roms/mm.us.rev1.z64 baserom.mm.us.rev1.z64'
             }
         }
+        stage('Setup') {
+            steps {
+                sh 'make -j setup 2> tools/warnings_count/warnings_setup_new.txt'
+                sh 'cat tools/warnings_count/warnings_setup_new.txt'
```
zbanks, line 16: If this is a bash shell, it may be better to do `make -j setup 2> >(tee tools/warnings_count/warnings_setup_new.txt)` so that you don't have to wait until the second step to see the stderr. (Also I think if the first step *failed*, you wouldn't see stderr at all?)

AngheloAlf: `tee` prints whatever it gets passed to, while writing to a file?

zbanks: Yes

tools/warnings_count/compare_warnings.py
```diff
@@ -0,0 +1,33 @@
...
+    newLines = countFileLines(args.newwarnings)
+    if newLines > currentLines:
+        print()
+        print("There are more warnings now. Go fix them!")
+        print("\tCurrent warnings: " + str(currentLines))
+        print("\tNew warnings: " + str(newLines))
+        print()
+        with open(args.newwarnings) as f:
+            print("Warnings:\n\n" + f.read())
+        print()
```
zbanks, line 27: I think it'd be useful to print a comment here that describes how to bless these warnings, if the author decides that they're necessary.

"If these warnings need to be ignored to produce a matching build, run `tools/warnings_count/update_current_warnings.sh` and commit the updated files in `tools/warnings_count/`"

tools/warnings_count/create_new_warnings.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# This script can be used when you want to test locally the amount of warnings produced by your changes before doing a PR.
+
+make distclean
+make -j4 setup 2> tools/warnings_count/warnings_setup_new.txt
+make -j4 all 2> tools/warnings_count/warnings_build_new.txt
```
zbanks, line 7: These scripts will only work if they're run from the parent `mm/` directory. 
A common idiom is: `DIR="$(dirname "$(readlink -f "$0")")"`, you could then `cd "$DIR/../.."` to ensure all the paths are relative to `mm/`?

Also I'd recommend adding `set -e`: we shouldn't try the `make all` step if the `make setup` step failed.

```bash
#!/bin/bash
set -e

# This script can be used when you want to test locally the amount of warnings produced by your changes before doing a PR.

DIR="$(dirname "$(readlink -f "$0")")"
cd "$DIR/../.."

make distclean
make -j4 setup 2> tools/warnings_count/warnings_setup_new.txt
make -j4 all 2> tools/warnings_count/warnings_build_new.txt
```

zbanks: Should this script be called `check_new_warnings.sh` instead? 
Should it run `tools/warnings_count/compare_warnings.py` at the end?

zbanks (changes requested):

tools/warnings_count/create_new_warnings.sh
AngheloAlf: I like those suggestions, I'll add them

zbanks (approved):
lgtm!

Kenix3 (approved):

Jenkinsfile
```diff
@@ -10,9 +10,24 @@ pipeline {
                 sh 'cp /usr/local/etc/roms/mm.us.rev1.z64 baserom.mm.us.rev1.z64'
             }
         }
+        stage('Setup') {
+            steps {
+                sh 'make -j setup 2> >(tee tools/warnings_count/warnings_setup_new.txt)'
```
zbanks, line 15: I guess you can change to this?
```suggestion
                sh 'bash -c "make -j setup 2> >(tee tools/warnings_count/warnings_setup_new.txt)"'
```

zbanks (approved):
lgtm (for real now)

PR #151, opened 2021-05-18
`diff.py` symlink and `graphovl.py`

AngheloAlf:
- Adds a symlink to `tools/asm-differ/diff.py`.
  - This way `diff.py` can be used the same way it is used in OoT repo.
  - Remove `diff.sh`.
- Port `graphovl.py`.
  - `graphovl.py` is a tool that is able to create function calls graphs of actionFunc based actors.
  - Example: running `./tools/graphovl.py Arms_Hook` would create the image `graphs/Arms_hook.gv.png`.
  - `-s`/`--style` to select a color style for the output image. The styles are files stored in `tools/graphovl_styles/`. The default is `solarized_dark`. You can create your own styles too.
  - `--loners` flag to include functions that are not called or call any other overlay function.
  - This tool can be useful to document an actor in a more visual way.

I took `graphovl.py` from: https://github.com/zeldaret/oot/pull/783


diff.py
```diff
@@ -0,0 +1 @@
+tools/asm-differ/diff.py
```
Kenix3, line 1: I'm not completely sure we want to clutter up the main directory. The reason for the ./tools/ directory is to put stuff like that there.
I guess we did used to have a diff.sh there.

tools/graphovl.py
```diff
@@ -0,0 +1,356 @@
+#!/usr/bin/env python3
```
Kenix3, line 1: This should really be a subrepo rather than just including all of the files in a copy/paste manner.

Kenix3 (changes requested):

diff.py
AngheloAlf: This file is actually a symlink (like a Windows's shortcut). I don't know why Github shows it like this.

AngheloAlf: Also, the problem with `diff.sh` is it doesn't allow to change the arguments passed to `diff.py` without modifying the script.

tools/graphovl.py
AngheloAlf: Okey. 
What is prefered? subrepo or submodule?

diff.py
Kenix3: I know. I think this is fine.

tools/graphovl.py
AngheloAlf: `graphovl.py` is now a subrepo.

Kenix3 (approved):
Looks good to me!

hensldm (approved):
LGTM

PR #152, opened 2021-05-18
Fix all warnings raised by IDO

zbanks:
- Add warning 624 to the `woff` list to ignore it. This is the warning about "trailing commas," which we *purposefully* use in decls in order to guide `clang-format`'s binpacking.
- Fix all warnings raised by the IDO compiler. `make` now builds without raising *any* warnings.
- I didn't fully re-run `./format.sh` (I didn't want to cause unnecessary diffs) but I did run it on some of the files I changed.
- If this gets merged before #150, then we should be able to start off with no warnings :smile: 

I did this by changing `preprocess.py` to fail if the output of IDO had any warnings; but it sounded like on the discord people didn't want this moving forward. (And instead we'll adopt #150?)


include/functions.h
```diff
@@ -109,10 +109,10 @@ void FaultDrawer_SetFontColor(u16 color);
...
 void FaultDrawer_SetCursor(s32 x, s32 y);
 void FaultDrawer_FillScreen(void);
-FaultDrawer* FaultDrawer_FormatStringFunc(FaultDrawer* arg, const char* str, s32 count);
+FaultDrawer* FaultDrawer_FormatStringFunc(FaultDrawer* arg, char* str, s32 count);
 void FaultDrawer_VPrintf(char* str, char* args);
-void FaultDrawer_Printf(const char* fmt, ...);
-void FaultDrawer_DrawText(s32 x, s32 y, const char* fmt, ...);
+void FaultDrawer_Printf(char* fmt, ...);
+void FaultDrawer_DrawText(s32 x, s32 y, char* fmt, ...);
 void FaultDrawer_SetDrawerFB(void* fb, u16 w, u16 h);
```
zbanks, line 116: This is the one section I'm not super happy about, I'll keep trying to re-work it.

Kenix3 (approved):
Looks good to me.

zbanks:
Sorry that it's going to be hard to see what changed in the recent commits because of the reformatting. 

- I modified the Makefile to add `-woff=624,...` I had mistakenly forgotten it before. (The trailing comma warning)
- I cleaned up the printf stuff to keep more `const char *`s.

src/overlays/actors/ovl_En_Po_Fusen/z_en_po_fusen.c
```diff
@@ -72,10 +72,10 @@ void EnPoFusen_Init(Actor* thisx, GlobalContext* globalCtx) {
 
     if (0) {}
     this->collider.dim.worldSphere.radius = 40;
-    SkelAnime_InitSV(globalCtx, &this->anime, &D_060024F0, &D_06000040, &this->limbDrawTbl, &this->transitionDrawTbl,
-                     10);
+    SkelAnime_InitSV(globalCtx, &this->anime, (SkeletonHeader*)&D_060024F0, (AnimationHeader*)&D_06000040,
```
AngheloAlf, line 75: I feel like `D_060024F0` and `D_06000040` should be declared to their correct types instead of casting them like this.

src/overlays/actors/ovl_Obj_Boyo/z_obj_boyo.c
```diff
@@ -126,7 +126,7 @@ void ObjBoyo_Update(Actor* thisx, GlobalContext* globalCtx2) {
     Actor* target;
     s32 num;
 
-    target = ObjBoyo_GetCollidedActor(thisx, globalCtx, &num);
+    target = ObjBoyo_GetCollidedActor((ObjBoyo*)thisx, globalCtx, &num);
```
AngheloAlf, line 129:
```suggestion
    target = ObjBoyo_GetCollidedActor(this, globalCtx, &num);
```

zbanks: Oh weird, I thought I specifically tried this and it didn't match... but it matches now!

src/overlays/actors/ovl_En_Po_Fusen/z_en_po_fusen.c
zbanks: Ah yeah good catch. Turns out the types were just swapped.

AngheloAlf (approved):

Kenix3 (approved):

PR #156, opened 2021-05-19
Ovl_Obj_HsStump Ok but one function does not match

hensldm:
What ever I do I cannot get `func_80BDABCC` to match (it's the function that causes the stump to appear in ikana valley with the Garo Mask).


engineer124 (commented):
Looks great! Just a few small comments

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
```diff
@@ -23,18 +28,132 @@ const ActorInit Obj_HsStump_InitVars = {
...
+void ObjHsStump_Init(Actor* thisx, GlobalContext* globalCtx) {
+    ObjHsStump* this = THIS;
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Obj_HsStump_0x80BDAA30/func_80BDAB18.asm")
+    Actor_ProcessInitChain(&this->dyna.actor, D_80BDAF80);
+    this->isHidden = OBJHSSTUMP_GET_ISHIDDEN(thisx);
+    this->switchFlag = OBJHSSTUMP_GET_SWITCHFLAG(thisx); // Must be thisx to match
+    BcCheck3_BgActorInit(&this->dyna, 1);
+    BgCheck3_LoadMesh(globalCtx, &this->dyna, &D_060011B0);
+    if (this->isHidden != 0) {
```
engineer124, line 53: Could this be done as a switch to look cleaner?
```
switch (this->isHidden) {
    case 1:
    
    case 0:
```

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
```diff
@@ -23,18 +28,132 @@ const ActorInit Obj_HsStump_InitVars = {
...
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Obj_HsStump_0x80BDAA30/func_80BDAB6C.asm")
+void ObjHsStump_SetupIdle(ObjHsStump* this, GlobalContext* globalCtx) {
+    this->actionFunc = ObjHsStump_Idle;
+}
 
+void ObjHsStump_Idle(ObjHsStump* this, GlobalContext* globalCtx) {
+    if (this->isHidden == 1) {
+        if (Flags_GetSwitch(globalCtx, this->switchFlag)) {
```
engineer124, line 74: `if ((this->isHidden == 1) && Flags_GetSwitch(globalCtx, this->switchFlag) {`

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
```diff
@@ -23,18 +28,132 @@ const ActorInit Obj_HsStump_InitVars = {
...
+        this->rotAngle += 0x2000;
+    }
+    if (this->framesAppeared < 11) {
+        if (this->framesAppeared == 0) {
+            iceSmokePosOffset.x = 1.0f;
+            iceSmokePosOffset.y = 0.5f;
+            iceSmokePosOffset.z = 0.0f;
+            iceSmokeVelOffset.x = 1.0f;
+            iceSmokeVelOffset.y = 0.5f;
+            iceSmokeVelOffset.z = 0.0f;
```
engineer124, line 115: Could you try `iceSmokePosOffset.x = iceSmokeVecOffset.x = 1.0f`? Likewise for y and z

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
```diff
@@ -23,18 +28,132 @@ const ActorInit Obj_HsStump_InitVars = {
...
+            numDirections = 4;
+            angle = 360.0f / numDirections;
+            angleBAMS = angle * (0x10000 / 360.0f);
+
+            for (i = 0; i < numDirections; i++) {
+                offsetAngle = i * angleBAMS;
+                Lib_Vec3f_TranslateAndRotateY(&this->dyna.actor.world.pos, offsetAngle, &iceSmokePosOffset,
+                                              &iceSmokePos);
+                Lib_Vec3f_TranslateAndRotateY(&D_801D15B0, offsetAngle, &iceSmokeVelOffset, &iceSmokeVel);
+                EffectSsIceSmoke_Spawn(globalCtx, &iceSmokePos, &iceSmokeVel, &D_80BDAF84, 0x64);
```
engineer124, line 126: You could probably replace 0x64 with 100

engineer124 (commented):
Looks good! Just a couple of small minor comments.

ps. (sorry for the duplicate review, first time trying to give a GitHub Review)

hensldm:
In case anyone was curious, here is the asm diff for `func_80BDABCC `. The loop is further down and jumps back up to the green arrow
image

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
hensldm: 👍

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
hensldm: 👍

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
hensldm: 👍

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.c
```diff
@@ -23,18 +28,129 @@ const ActorInit Obj_HsStump_InitVars = {
...
+
+extern InitChainEntry D_80BDAF80[];
+/*
+static InitChainEntry sInitChain[] = {
+    ICHAIN_VEC3F_DIV1000(scale, 180, ICHAIN_STOP),
+};
+*/
+
+extern Vec3f D_80BDAF84;
+// static Vec3f iceSmokeAccel = {0.0f, 0.0f, 0.0f};
```
zbanks, line 43: So, my understanding is that you can't fully switch to the recompiled overlay in `linker_script/code_script.txt` util the last function matches. (The `.ovl` section won't match because it will be missing some relocation symbols).

Locally I was able to change just the `.data` section, which allowed me to uncomment all these variables.
```diff
diff --git a/linker_scripts/code_script.txt b/linker_scripts/code_script.txt
index a3afc53..ad8047c 100644
--- a/linker_scripts/code_script.txt
+++ b/linker_scripts/code_script.txt
@@ -8239,7 +8239,7 @@ SECTIONS
     ovl_Obj_HsStump : AT(RomLocation)
     {
         build/src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.o(.text)
-        build/asm/overlays/ovl_Obj_HsStump_data.o(.data)
+        build/src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.o(.data)
         build/src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.o(.rodata)
         build/asm/overlays/ovl_Obj_HsStump_rodata.o(.rodata)
     }
```

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.h
```diff
@@ -3,11 +3,21 @@
 
 #include <global.h>
 
+#define OBJHSSTUMP_GET_SWITCHFLAG(this) (((ObjHsStump*)this)->dyna.actor.params & 0x7F)
+#define OBJHSSTUMP_GET_ISHIDDEN(this) ((((ObjHsStump*)this)->dyna.actor.params >> 12) & 0xF)
```
zbanks, line 7: Is it better to have these macros take in the `Actor *` instead? Seems like that would be consistent with them requiring `thisx`?

```suggestion
#define OBJHSSTUMP_GET_SWITCHFLAG(actor) ((actor)->params & 0x7F)
#define OBJHSSTUMP_GET_ISHIDDEN(actor) ((actor)->params >> 12) & 0xF)
```


(I'm not sure about this, not sure if it should change or not)

zbanks (changes requested):
Looks great, two notes:
- I think you can uncomment `Obj_HsStump_InitVars`, `sInitChain`, and `iceSmokeAccel` if you replace the `.data` section (see my other comment)
- Maybe the params macros should expect an `Actor *` instead? (Really not sure)

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.h
hensldm: I think there are arguments for both sides. The reason I chose to cast as `ObjHsStump` even though `thisx` was passed in was that these macros don't really apply to any old actor. They only apply to `ObjHsStump` actors so the cast seems appropriate to specify this condition. Perhaps it is enough to that the names are prefixed with `ObjHsStump`. Any one have any strong opinions about this?

zbanks: Yeah makes sense. I'm fine either way.

One thing that I was worried about these macros is that the `(ObjHsStump *)` cast would hide some type errors? The compiler won't detect that `OBJHSSTUMP_GET_SWITCHFLAG(thisx->params)` is wrong (and would probably crash when ran). 
I guess this isn't a big deal for this project, because we have the matching checksum?

zbanks (approved):

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.h
Zelllll: Just make the macros take the params itself rather than the actor instance.

Kenix3: It makes sense to accept the derived Actor class, but currently no other macros for params do. Does the code still OK when using the updated macro?

hensldm: Code would still OK with these updated Macros
image

hensldm: I kind of like Zellll's suggestions of just passing the params in directly though. The we don't have to worry about passing in the correct pointer.

Kenix3: Let's get some more opinions, but I'd like to keep these consistent. We have three options:
- Pass in params.
- Pass in the base actor. (Current behaviour)
- Pass in the derived actor.

Zelllll: pass in the params, the point of the macro is to extract data out of the params, not out of the actor instance

linker_scripts/code_script.txt
```diff
@@ -8239,7 +8239,7 @@ SECTIONS
     ovl_Obj_HsStump : AT(RomLocation)
     {
         build/src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.o(.text)
-        build/asm/overlays/ovl_Obj_HsStump_data.o(.data)
+        build/src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.o(.data)
         build/src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.o(.rodata)
         build/asm/overlays/ovl_Obj_HsStump_rodata.o(.rodata)
```
Thar0, line 8244:
```suggestion
#ifdef NON_MATCHING
    build/src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump_overlay.o(.ovl)
#else
    build/asm/overlays/ovl_Obj_HsStump_rodata.o(.rodata)
#endif
```
In order to not break `NON_MATCHING` builds the generated relocation data should be used anyway when building with `NON_MATCHING`

src/overlays/actors/ovl_Obj_HsStump/z_obj_hsstump.h
Kenix3: There's no consensus yet.

We're going to go with the current paradigm for consistency: Accept the Actor* base struct pointer.

Kenix3 (changes requested):

Kenix3 (approved):

Issue #144, opened 2021-05-18
Add contributing documentation to the MM repo.


Kenix3:
- [x] readme.md - High level overview, building, and getting started.
   - [x] High level overview
   - [x] Preparing baserom
   - [x] How to set up the repo
   - [x] How to build the repo
   - [x] Day to day workflow (diff.py and maybe symbol rename)
- [x] contributing.md - Contributor's guidelines
   - [x] How to submit a PRs
   - [x] How to respond to reviewer comments
   - [x] Documentation rules
   - [x] Style and naming schemes
   - [x] Build warnings
   - [x] Updating trello cards
- [x] reviewing.md - What to do when reviewing code.
   - [x] When reviewing other people's code please follow the checklist.
   - [x] When someone addresses all of your comments you should approve the changes.
   - [x] If someone does not address your comments and expresses that a different way is better than yours, look for feedback from other contributors. The project lead will have final say in these situations, but the decisions will generally be guided by a consensus of contributors.
   - [x] Reviewer checklist. This is mostly the same as contributor submission checklist.
      - [x] `format.sh` was run.
      - [x] Files with NON_MATCHING functions are equivalent.
      - [x] `make` builds a matching ROM (jenkins)
      - [x] Any new compiler warnings that were added are required for matching (check the file that stores warnings)
      - [x] Comments and variables have correct spelling.
      - [x] `code` and `boot` segment files are documented. Overlays with documentation are complete.
         - [x] Overlays with documentation are required to have macros to define access to parameters if the parameter uses bitwise access. The params should have an enum otherwise.
      - [x] The following fields are declared in an Actor header file.
         - [x] Main Actor struct
         - [x] Extern'd initVar data.
         - [x] Types used in the actor struct. Specific example would be actionFunc typedefs.
         - [x] Param field macros and/or enums.
         - [x] We have proof that enum/struct/define is needed in another file.
      - [x] New variables & functions should follow standard naming conventions.
         - [x] Constants converted to whichever looks best in context: hexadecimal, decimal, or float
            - [x] Rotation angles should always be in hexadecimal
         - [x] Functions, structs, unions, enums, and typedefs are `TitleCase` (`DmRavine`)
            - [x] "Methods" for objects separate the system from the verb with an underscore (`DmRavine_Init`)
         - [x] Variable names are `camelCase` (`actionFunc`)
            - [x] Global variables start with `g` (`gSaveContext`)
            - [x] Static global variables start with `s` (`sSphereInit`)
[...]

PR #157, opened 2021-05-19
CONTRIBUTING.md

zbanks:
Moved from #142 after incorporating all of the great feedback. 

The PR template could either go in the root directory, `docs/` or `.github/`. I put it in the `.github/` because we don't have a `docs/` folder yet.


Kenix3:
https://github.com/zeldaret/mm/issues/144

.github/pull_request_template.md
```diff
@@ -0,0 +1,13 @@
+<!-- Leave this text intact in your PR -->
+
+Before opening this PR, ensure the following:
+
+- `./format.sh` was run to apply standard formatting.
+- `make` successfully builds a matching ROM.
+- No new compiler warnings during the build process
```
Kenix3, line 7:
```suggestion
- No new compiler warnings were introduced during the build process
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+
+Each PR needs a review from one reviewer, plus the project lead.
+
+The PR author marks the comments as resolved: the commenter may not have permissions to do so.
+Once all comments are addressed, it is courteous to ping the reviewer on either Discord or GitHub via re-requesting a review.
+
+After all the comments are addressed and at least one contributor has approved the review, the project lead can then merge the code.
+The project lead is also responsible for ensuring that all of these procedures are followed.
+
+Throughout the process, you (the PR author) should update the [Trello board](https://trello.com/b/ruxw9n6m/majoras-mask-decompilation) with the status of your PR.
```
Kenix3, line 147: Updating trello includes both putting it in the correct column, and updating it with appropriate labels.

Kenix3 (changes requested):

zbanks:
If anyone wants to see what the template looks like, I put this branch on my fork's master, so you can [preview it here](https://github.com/zbanks/mm/compare/master...zbanks:fix-warnings).

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+These files demonstrate the following:
+
+- Word order in names are from least-to-most specific (`DM_RAVINE_STATE_ACTIVE`, not `DM_RAVINE_ACTIVE_STATE`)
+- Functions, structs, unions, enums, and typedefs are `TitleCase` (`DmRavine`)
+    - "Methods" for objects separate the object from the verb with an underscore (`DmRavine_Init`)
+- Variable names are `camelCase` (`actionFunc`)
+    - Global variables start with `g` (`gSaveContext`)
+    - Static global variables start with `s` (`sSphereInit`)
+- Macros and enum constants are `SCREAMING_SNAKE_CASE` (`DM_RAVINE_STATE_ACTIVE`)
+- Auxiliary structs, enums, and constants are defined in the header file
```
AngheloAlf, line 78: There are some structs, enums or constants that are only related to the internal use of the an actor, so moving them to the header doesn't make much sense to me.

I would suggest only defining those stuff in a header if it does make sense using them _outside_ of that specific actor. For example, an enum which has the values used in `actor.params`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+- Trailing commas in array and struct definitions (see `EnJcMatoDamageTable`)
+- Constants converted to whichever looks best in context: hexadecimal, decimal, float, or boolean
```
AngheloAlf, line 80: May be worth noting we don't allow the use of the type `bool`, but we can use `true` and `false` if it makes sense.

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+    - Rotation angles should always be in hexadecimal
+- Define macros for bitwise access to `actor->params`
+
+### `NON_MATCHING`
```
AngheloAlf, line 84: We should at least mention the existence of `NON_EQUIVALENT` too.

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+Pull Requests (PRs)
+-------------------
+
+### Checklist
+
+Before opening a PR, walk through the following steps to ensure that your code conforms to the style guide & conventions.
+
+- `./format.sh` was run to apply standard formatting.
+- `make` successfully builds a matching ROM.
+- No new compiler warnings during the build process
```
AngheloAlf:
```suggestion
- No new compiler warnings during the build process.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+### Naming Scheme
+
+The following overlays are good examples of our naming conventions:
+
+- [`ovl_Dm_Ravine`](src/overlays/actors/ovl_Dm_Ravine/z_dm_ravine.c)
+- [`ovl_En_Jc_Mato`](src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c)
+
+These files demonstrate the following:
+
+- Word order in names are from least-to-most specific (`DM_RAVINE_STATE_ACTIVE`, not `DM_RAVINE_ACTIVE_STATE`)
```
Zelllll, line 71: Please, please do not suggest people to use the og names for enums, variables, etc

Zelllll (changes requested):

CONTRIBUTING.md
zbanks: What's the difference? 

I think explaining `NON_MATCHING` is probably good enough. I think the key point I wanted to convey here is that "if you get a function really close but you can't finish it and you're going to give up, consider `NON_MATCHING` instead of throwing it out."

CONTRIBUTING.md
zbanks: I'll just drop the mention of boolean entirely; `true`/`false` are pretty rare so far.

CONTRIBUTING.md
zbanks: I'd tend to agree, that would have been my preference, but I could have sworn someone had told me to make the state enum public in my #120 PR for Dm_Ravine. (I can't find a record of this though...).

But I'm happy to reword this either way. Do you (or does anyone else) have a succinct way of describing this? I'm having trouble.
I think normally I'd call them "public structs" or "structs that are part of the API" but I think those terms aren't very clear here because we don't use them anywhere else in the project.

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+Pull Requests (PRs)
+-------------------
+
+### Checklist
+
+Before opening a PR, walk through the following steps to ensure that your code conforms to the style guide & conventions.
+
+- `./format.sh` was run to apply standard formatting.
+- `make` successfully builds a matching ROM.
+- No new compiler warnings were introduced during the build process.
```
Kenix3, line 132: How would someone verify this?
tools/warnings_count/check_new_warnings.sh

Kenix3 (changes requested):

README.md
```diff
@@ -23,7 +23,7 @@ in an experimental and research phase and cannot currently be used traditionally
 source code base for general changes.
 ```
 
-This repo does not include all assets necessary for compiling the ROM. A prior copy of the game is required to extract the required assets.
+**This repo does not include any assets or code necessary for compiling the ROM. A prior copy of the game is required to extract the required assets.**
```
Kenix3, line 26:
```suggestion
**This repo does not include any assets necessary for compiling the ROM. A prior copy of the game is required to extract the required assets.**
```
This is the wording we decided on in Discord.

Kenix3 (changes requested):

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
+Contributing to the Majora's Mask Decompilation Project
+=======================================================
+
+Thanks for helping us reverse engineer *The Legend of Zelda: Majora's Mask* for the N64! 
+All contributions are welcome. This is a group effort, and even small contributions can make a difference. Some tasks also don't require much knowledge to get started.
+
+This document is meant to be a set of tips and guidelines for contributing to the project.
+For general information about the project, see [our readme](https://github.com/zeldaret/mm/blob/master/README.md).
+
+Most discussions happen on our [Discord Server](https://discord.zelda64.dev), where you are welcome to ask if you need help getting started, or if you have any questions regarding this project and other decompilation projects.
```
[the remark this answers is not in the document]
EllipticEllipsis: Isn't this a non-restrictive relative clause, and so conventionally is written with a comma?

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+- The `#resources` channel on the Discord contains many more links
+
+Getting Started
+---------------
+
+### What should I know to take part?
+
+Basic knowledge of C, particularly arrays and pointers, is extremely useful. Knowledge of MIPS is not required initially, you can pick it up as you work through some decompilation.
+
+You should be familiar with using git and GitHub. There are a number of tutorials available online, [such as this one](https://github.com/firstcontributions/first-contributions), which can help you get started.
```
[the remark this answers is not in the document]
EllipticEllipsis: And here, "such as this one" is distinct from the subsequent relative clause: the linked one is not the only tutorial that can help you get started.

CONTRIBUTING.md
```diff
@@ -0,0 +1,147 @@
...
+
+You should [join the Discord](https://discord.zelda64.dev/) to say hello and get suggestions on where to start on the `#mm-decomp` channel.
+
+We track who is working on what on the [Trello board](https://trello.com/b/ruxw9n6m/majoras-mask-decompilation).
+After joining the Discord, ask to be added to the Trello board so you can claim your actor and avoid duplicate work.
+
+Style Guide & Conventions
+-------------------------
+
+Most of the C formatting style is enforced by the `format.sh` script, which is based on `clang-format`.
```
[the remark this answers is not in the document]
EllipticEllipsis: Also a non-restrictive relative clause.

zbanks:
I think I got all of the fixes you suggested @Kenix3. (I removed the commas, I think it's readable either way)

I also changed the PR template a bit to make it a bit "denser", removing some of the whitespace & combined the two `<!-- ... -->` comments. I've been keeping it in sync with the "checklist" section in the main contributing.md file.

README.md
```diff
@@ -103,9 +103,21 @@ This means that something is wrong with the ROM's contents. Either the baserom f
 
 Running `make init` will also make the `./expected` directory and copy all of the files there, which will be useful when running the diff script. The diff script is useful in decompiling functions and can be ran with this command: `./tools/asm-differ/diff.py -wmo3 <insert_function_here>`
```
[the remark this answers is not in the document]
zbanks: I changed the format slightly to be consistent with the next note (`**Note**:`), and added an example `make -j4` (I think it can be confusing to convey that the `N` is not literal)

CONTRIBUTING.md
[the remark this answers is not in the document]
zbanks: I just removed the commas, I think it's reasonable either way.

Kenix3 (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,155 @@
...
+    - Rotation angles should always be in hexadecimal
+- Define macros for bitwise access to `actor->params`
+
+### `NON_MATCHING` & `NON_EQUIVALENT`
+
+If you make significant progress on decompiling a function, but can't get it to match perfectly, you can use a `NON_MATCHING` block to commit your code but have it excluded from the build, like so:
+
+```c
+#ifdef NON_MATCHING
+void CollisionCheck_SpawnWaterDroplets(GlobalContext* ctxt, Vec3f* v) {
```
Thar0, line 95:
```suggestion
void CollisionCheck_SpawnWaterDroplets(GlobalContext* globalCtx, Vec3f* v) {
```
Nitpick but we want to phase out all the variations on the global context variable as soon as possible, so might as well have it correct in the docs from the start, code will follow soon enough.

CONTRIBUTING.md
```diff
@@ -0,0 +1,155 @@
...
+However, the goal of this project is to produce a codebase that can be understood and modified.
+So, beyond producing matching C code, the next steps are *documenting* the code.
+
+Documenting is more than just adding comments. Documenting also includes:
+
+- Renaming functions, variables, and struct members
+- Using (or adding) constants, enums, and macros when possible
+- Explaining sections of code that are not straightforward
+
+Overlays are not required to be documented at this time, but `code/` and `boot/` are.
```
Thar0, line 120: I don't really agree that anything should be "required" to be documented. Just having decompiled code is a huge step forward and requiring people who might not have the time to sit down and fully understand the system they are decompiling at the time seems inappropriate. Not to mention, certain core systems that operate in close connection with other systems or make use of subsystems may be unreasonable to document properly at the time and will need to be revisited after all other relevant information is available to ensure the documentation is accurate.

.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+Before opening this PR, ensure the following:
+- `./format.sh` was run to apply standard formatting.
+- `make` successfully builds a matching ROM.
+- No new compiler warnings were introduced during the build process.
+    - Can be verified locally by running `tools/warnings_count/check_new_warnings.sh`
```
Thar0, line 5: Is there anywhere that mentions to consult in the discord server if it seems like the warnings may be necessary? I suppose people will naturally do this if they struggle with removing certain warnings, but it might be better to just have it written down to let them know that's encouraged. I think it would be better than someone potentially agonising over trying to remove a particularly tough warning.

Thar0 (commented):
Overall looks great, just a few things.

CONTRIBUTING.md
Kenix3: Good point that there are exceptions when one system calls on another system a lot. There should probably be a carve out exception for those situations. That being said, in most cases that documentation these days can be gleaned from looking at OOT code.

As far as someone not having time, to be honest, my solution to them would be to work on an actor where documentation is not required.

zbanks: Yeah this makes sense. I think avoiding saying "X is required" is a net positive, it sounds a bit rigid.

```suggestion
Overlays are not required to be documented at this time, but `code/` and `boot/` should be documented as much as possible.
```

@Kenix3 what do you think? Do you have any suggestions here?

Kenix3: I'm happy with that rewording.

.github/pull_request_template.md
zbanks: I added it to the contributing.md as a generic note, but not here? Do you think that's fine?
I think in practice people could run into issues at any step, not just the warnings. (e.g. if you don't have matching code)

Thar0: I agree it's not just limited to that one point so the generic statement is a good idea. I suppose what I was looking for in particular was a note or prompt in the PR template that if there is any particular issues preventing a PR they can be discussed, but if you feel as though that's already made clear enough then it's likely fine.

AngheloAlf (approved):

Kenix3 (approved):

isghj5 (approved):

Thar0 (approved):

PR #159, opened 2021-05-20
Misc fixes

AngheloAlf:
- Rename `SkeletonHeader` to `FlexSkeletonHeader`.
- Add a proper `SkeletonHeader` struct.
- Rename `SkelLimbEntry` to `StandardLimb`.
- Rename `AnimationRotationIndex` to `JointIndex`.
- Fix a bug in `extract_assets.py` where it was reextracting assets even when were 0 changes to the XMLs.
- Fix `first_diff.py` roms path.
- Add makefile's rules to compile blob files.
- Change `gSaveContext.perm.day` and `gSaveContext.perm.isNight` types from `u32` to `s32`
  - I noted every time we accessed `gSaveContext.perm.day`, we needed a cast to `s32`. Changing the type allows the removal of those casts.
- Add `CURRENT_DAY` macro.
  - I also noted that `gSaveContext.perm.day` is accessed doing a `% 5` operation to it, so I felt it was appropriate to make a macro for this common pattern (I'm doing a few objects that does this _a lot_, that's why I say it is common, I could be wrong).

Most of the renames are needed to have a proper ZAPD compatibility for when we start doing objects.


Kenix3 (approved):

src/overlays/actors/ovl_Obj_Tokei_Step/z_obj_tokei_step.c
```diff
@@ -201,7 +201,7 @@ void ObjTokeiStep_Init(Actor* thisx, GlobalContext* globalCtx) {
         BgCheck3_LoadMesh(globalCtx, &this->dyna, &D_06000968);
         ObjTokeiStep_InitSteps(this);
         ObjTokeiStep_SetupBeginOpen(this);
-    } else if (!((((s32)gSaveContext.perm.day % 5) != 3) || (gSaveContext.perm.time >= 0x4000)) ||
+    } else if (!((CURRENT_DAY != 3) || (gSaveContext.perm.time >= 0x4000)) ||
                (s32)gSaveContext.perm.day >= 4) {
```
hensldm, line 205: Should be able to remove this cast as well.

AngheloAlf: Good catch! Thanks

include/z64animation.h
```diff
@@ -24,37 +24,44 @@ typedef struct {
...
                                       // far model display list.
-} SkelLimbEntry;                      // Size = 0xC or 0x10
+} StandardLimb;                      // Size = 0xC or 0x10
 
+#if 0
 typedef struct {
-    /* 0x000 */ SkelLimbEntry* limbs[1]; // One or more limbs, index 0 is the root limb.
+    /* 0x000 */ StandardLimb* limbs; // One or more limbs, index 0 is the root limb.
 } Skeleton;                              // Size >= 4
+#endif
```
zbanks, line 33: Should this be deleted entirely instead of commented out?

hensldm (approved):
LGTM

src/code/z_skelanime.c
```diff
@@ -8,8 +8,9 @@ s32 func_801370B0(SkelAnime* skelAnime);
 s32 func_8013713C(SkelAnime* skelAnime);
 void SkelAnime_ChangeAnimDefaultRepeat(SkelAnime* skelAnime, AnimationHeader* animationSeg);
 void func_80137748(SkelAnime* skelAnime, Vec3f* pos, s16 angle);
-void SkelAnime_ChangeLinkAnim(GlobalContext* globalCtx, SkelAnime* skelAnime, LinkAnimetionEntry* linkAnimetionEntrySeg,
-                              f32 playbackSpeed, f32 frame, f32 frameCount, u8 animationMode, f32 transitionRate);
+void SkelAnime_ChangeLinkAnim(GlobalContext* globalCtx, SkelAnime* skelAnime,
+                              LinkAnimationHeader* LinkAnimationHeaderSeg, f32 playbackSpeed, f32 frame, f32 frameCount,
```
zbanks, line 12: Can the `LinkAnimationHeaderSeg` be called `linkAnimationHeaderSeg` everywhere instead?

zbanks (changes requested):

isghj5:
LGTM

include/z64animation.h
AngheloAlf: whoops, I added this for testing and forgot to remove it. Good catch!

zbanks (approved):

PR #161, opened 2021-05-22
ovl_Obj_Milk_Bin OK

AngheloAlf:
---

I made an enum for the params of this actor, but I'm not completely sure how to name them, so I gaved them generic names. Suggestions for those are welcome.

Btw, for some reason, `format.sh` modified files I didn't touch. dunno what happened there.


jpburnett (approved):
Once the warnings are fixed, I think this will look legen-dairy

src/overlays/actors/ovl_Obj_Milk_Bin/z_obj_milk_bin.c
```diff
@@ -21,12 +26,80 @@ const ActorInit Obj_Milk_Bin_InitVars = {
...
+
+    if (!(this->disableDraw & 1)) {
+        CollisionCheck_SetAC(globalCtx, &globalCtx->colCheckCtx, &this->collider.base);
+        CollisionCheck_SetOC(globalCtx, &globalCtx->colCheckCtx, &this->collider.base);
+    }
+}
 
-#pragma GLOBAL_ASM("./asm/non_matchings/overlays/ovl_Obj_Milk_Bin_0x80C08E40/ObjMilkBin_Destroy.asm")
+// gMilkBinMilkJarDL
+extern Gfx D_060004B0[];
```
zbanks, line 97: Really small nit: I would put any declarations like this at the top of the file, before any of the definitions?

zbanks (changes requested):
Looks great! :cow2: 

Only one small suggestion (placement of `extern Gfx ...`)

Note: It looks like `z_collision_check.c` was reformatted because it wasn't reformatted after the `s/ctxt/globalCtx/g` change (or was reformatted with a different version of `clang-format`) - so I think it makes sense to reformat it here. (Although no one else needs to reformat it in their PRs, to avoid merge conflicts)

src/overlays/actors/ovl_Obj_Milk_Bin/z_obj_milk_bin.c
AngheloAlf: Done :D

zbanks (approved):

Kenix3 (approved):

PR #210, opened 2021-07-15
C file for Fishing, some changes to collider scripts, a few minor cleanups

EllipticEllipsis:
---

- Added Fishing to the actor list, split it properly.
- Made some minor changes to `colliderinit.py` (it now outputs `D_address`s instead of `sCylinderInit` etc., and puts the likely eventual name on the line above: this should make it simpler to compile the C initially and then swap over later. (I wouldn't be too concerned about this change: it's intended to render the script obsolete shortly by importing the colliders, and most of the rest of the Init data, shortly anyway.)
- Improved `actor_symbols.py` to throw a clear exception rather than cryptic messages, especially when called in other scripts.
- Added a script to disassemble `CollisionCheckInfoInit`s
- Various minor cleanups in actors I saw while doing all of the above.


src/overlays/actors/ovl_En_Fishing/overlay.cfg
```diff
@@ -0,0 +1,2 @@
+ovl_En_Fishing
+z_en_fishing.c
```
Kenix3, line 2: missing trailing newline

Kenix3 (changes requested):

src/overlays/actors/ovl_En_Fishing/z_en_fishing.c
```diff
@@ -0,0 +1,122 @@
...
+#define THIS ((EnFishing*)thisx)
+
+void EnFishing_Init(Actor* thisx, GlobalContext* globalCtx);
+void EnFishing_Destroy(Actor* thisx, GlobalContext* globalCtx);
+void EnFishing_Update(Actor* thisx, GlobalContext* globalCtx);
+void EnFishing_Draw(Actor* thisx, GlobalContext* globalCtx);
+
+/*
+const ActorInit En_Fishing_InitVars = {
+    0x0079,
```
AngheloAlf, line 20:
```suggestion
    ACTOR_EN_FISHING,
```

tools/overlayhelpers/colliderinit.py
```diff
@@ -397,7 +404,7 @@ def GetColliderInit(address, type, num):
     if fileResult[0] is None:
         return("Invalid address")
 
-    print(hex(fileResult[1]), "in", fileResult[0].split(os.sep)[-1])
+    # print(hex(fileResult[1]), "in", fileResult[0].split(os.sep)[-1])
```
AngheloAlf, line 407: Is this line supposed to be commented?

EllipticEllipsis: Hm. I think I'll delete the previous comment (which is superseded by the error message in `actor_symbol`), but keep this line: it is a useful check that you're in the right file, if nothing else.

AngheloAlf (approved):

Kenix3 (approved):

PR #215, opened 2021-07-19
Remove header dependencies checks

AngheloAlf:
---

As we discussed in the discord server, it isn't a good idea to constantly check which C files should be recompiled when a header is  changed at this stage of this project since every header is included everywhere, so that check is now removed.

I also decided to remove most of the `@` at each command so we can see what's actually going in each step of the build process (instead of simply waiting of something that seems to be stuck...)


engineer124 (approved):
This included a solution that allowed https://github.com/zeldaret/mm/pull/214 to build properly. Great work!

PR #216, opened 2021-07-20
Automatically add Init data to actors

EllipticEllipsis:
---

**This PR needs to be checked very carefully**

Identified and imported the following types of init data in every actor but Player (for now)
- InitChain
- DamageTable
- CollisionCheckInfoInit
- Collider.*Init

Unfortunately the fancy new DamageTable format cannot be commented out, so I have used `#if 0 [...] #endif` instead of actual comments. VS Code is not bothered by this, and it builds OK, but it may be an issue for slightly less sophisticated editors.

Also filled out the struct of `ObjSyokudai` with its colliders, mostly as an illustration of what a struct with the common substructs added would look like.

There are over 470 changed files here. They in theory should be all the untouched actors apart from Player, but I can't rule out having missed a few, or overwritten some that I should not have. Therefore as many people need to look through this PR as possible, to make sure I haven't accidentally reverted any progress.


hensldm:
Really good work on this. I had a couple of thoughts that may or may not be worth the extra effort.

1. Since we know the data and address, how hard would it be to extern all the known data underneath the commented portion?

2. If we change the damage table by moving the damage type comment to other side of the comma, they could be single line comments allowing the whole data section to be a big multi-line comment block instead of the #if 0. For example `/* Deku Nut */ DMG_ENTRY(0, 0x1),` would become `DMG_ENTRY(0, 0x1), // Deku Nut`

Kenix3 (approved):

PR #236, opened 2021-07-29
En_Dnb

Kelebek1:
---


src/overlays/actors/ovl_En_Dnb/z_en_dnb.c
```diff
@@ -1,3 +1,9 @@
+/*
+ * File: z_en_dnb.c
+ * Overlay: ovl_En_Dnb
+ * Description: Dnb
```
EllipticEllipsis, line 4:
```suggestion
 * Description: 
```
Better to just leave this blank if you don't know IMO.

src/overlays/actors/ovl_En_Dnb/z_en_dnb.c
```diff
@@ -9,7 +15,24 @@ void EnDnb_Destroy(Actor* thisx, GlobalContext* globalCtx);
...
+extern Vec3s D_06004710[];
+extern Gfx D_06004638[];
+extern u64 D_0408F7E0[];
+extern u64 D_0408F3E0[];
+extern u64 D_0408EFE0[];
+extern u64 D_0408EBE0[];
+extern u64 D_0408E7E0[];
+extern u64 D_0408E3E0[];
+extern u64 D_0408DFE0[];
+extern u64 D_0408DBE0[];
```
EllipticEllipsis, line 32: Two things:
- It seems like 04 symbols are externed in variables.h, so I think they should be externed there instead of here?
- We don't type textures as u64 any more: we discovered some objects where this wouldn't match. For now we have the type `TexturePtr` until we decide what to do generally.

src/overlays/actors/ovl_En_Dnb/z_en_dnb.c
```diff
@@ -22,34 +45,273 @@ const ActorInit En_Dnb_InitVars = {
...
+    particle->unk_18 = D_801D15BC;
+    particle->unk_30 = 40.0f;
+    particle->unk_2C = 0.0f;
+    particle->unk_26 = arg1;
+    particle->unk_34 = (Rand_ZeroOne() * -2.0f) - 2.0f;
+}
+
+s32 func_80A500F8(EnDnb* this) {
+    static Vec3f D_80A50CB0 = { 0.0f, 0.0f, 1000.0f };
+
```
EllipticEllipsis, line 96:
```suggestion
```

src/overlays/actors/ovl_En_Dnb/z_en_dnb.h
```diff
@@ -5,9 +5,38 @@
...
+typedef struct {
+    /* 0x00 */ u8 isEnabled;
+    /* 0x01 */ u8 unk_01;
+    /* 0x02 */ u8 unk_02;
+    /* 0x04 */ f32 unk_04;
+    /* 0x08 */ f32 unk_08;
+    /* 0x0C */ Vec3f unk_0C;
+    /* 0x18 */ Vec3f unk_18;
+    /* 0x24 */ f32 unk_24;
+} EnDnbStruct; // size = 0x28
```
EllipticEllipsis, line 30:
```suggestion
} EnDnbUnkStruct; // size = 0x28
```

EllipticEllipsis (commented):
Just some minor things.

src/overlays/actors/ovl_En_Dnb/z_en_dnb.c
AngheloAlf: Agreed.

src/overlays/actors/ovl_En_Dnb/z_en_dnb.c
```diff
@@ -22,34 +45,273 @@ const ActorInit En_Dnb_InitVars = {
...
+        }
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dnb/func_80A4FDD0.s")
+    return count;
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dnb/func_80A4FEBC.s")
+s32 func_80A50950(EnDnbStruct* arg0, GlobalContext* globalCtx2) {
+    static u64* D_80A50CBC[] = {
```
AngheloAlf, line 270:
```suggestion
    static TexturePtr D_80A50CBC[] = {
```

src/overlays/actors/ovl_En_Dnb/z_en_dnb.c
Kelebek1: I can't see the change here, which line am I removing? I thought there's always a gap between static data and variables, and then another before the code.

EllipticEllipsis: No, point is there is usually no gap between the in-function static and the variables.

EllipticEllipsis:
This failed because there are arrays in other files that need to be changed from `u64` to `TexturePtr` as well, which created some new compiler warnings: you can look at the Jenkins output to see which.

Kenix3 (approved):

AngheloAlf (approved):

PR #238, opened 2021-07-30
En_Dno

Kelebek1:
---


src/overlays/actors/ovl_En_Dno/z_en_dno.c
```diff
@@ -36,94 +84,949 @@ const ActorInit En_Dno_InitVars = {
...
+void EnDno_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    EnDno* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dno/func_80A724B8.s")
+    SkelAnime_DrawSV(globalCtx, this->skelAnime.skeleton, this->skelAnime.limbDrawTbl, this->skelAnime.dListCount,
+                     func_80A7361C, func_80A73654, &this->actor);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dno/func_80A7256C.s")
+s32 func_80A7361C(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, Actor* thisx) {
```
EllipticEllipsis, line 932:
```suggestion
s32 EnDno_OverrideLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, Actor* thisx) {
```

src/overlays/actors/ovl_En_Dno/z_en_dno.c
```diff
@@ -36,94 +84,949 @@ const ActorInit En_Dno_InitVars = {
...
+    EnDno* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dno/func_80A72598.s")
+    *dList = NULL;
+    if (limbIndex == 9) {
+        rot->x += this->unk_466;
+    }
+    return 0;
```
EllipticEllipsis, line 939:
```suggestion
    return false;
```

src/overlays/actors/ovl_En_Dno/z_en_dno.c
```diff
@@ -36,94 +84,949 @@ const ActorInit En_Dno_InitVars = {
...
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dno/func_80A725E0.s")
+void func_80A73654(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx) {
```
EllipticEllipsis, line 942:
```suggestion
void EnDno_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx) {
```

src/overlays/actors/ovl_En_Dno/z_en_dno.c
```diff
@@ -36,94 +84,949 @@ const ActorInit En_Dno_InitVars = {
...
+        func_8012C28C(globalCtx->state.gfxCtx);
+        if (limbIndex == 13) {
+            Matrix_Scale(this->unk_454, this->unk_454, this->unk_454, 1);
+            SysMatrix_InsertXRotation_s(this->unk_45C, 1);
+        }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dno/func_80A72FAC.s")
+        gfxOpa = POLY_OPA_DISP;
+        gSPMatrix(gfxOpa, Matrix_NewMtx(globalCtx->state.gfxCtx), G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+        gSPDisplayList(gfxOpa + 1, *dList);
```
EllipticEllipsis, line 999:
```suggestion
        gSPDisplayList(&gfxOpa[1], *dList);
```
I mentioned this in another review, if this matches we should consider which style we prefer.

src/overlays/actors/ovl_En_Fsn/z_en_fsn.c
```diff
@@ -1003,7 +1003,7 @@ void EnFsn_GiveItem(EnFsn* this, GlobalContext* globalCtx) {
 }
 
 void EnFsn_SetupResumeInteraction(EnFsn* this, GlobalContext* globalCtx) {
-    if (((void)0, gSaveContext.inventory.questItems) & gBitFlags[18]) {
+    if (CHECK_QUEST_ITEM(18)) {
```
EllipticEllipsis, line 1006: We should decide on hex vs dec for these, although it'll hopefully be redundant once we have a proper enum anyway.

EllipticEllipsis (commented):
Wow this actor loves its do whiles. And what a huge PostLimbDraw.

include/z64actor.h
```diff
@@ -76,6 +76,31 @@ typedef struct {
     /* 0x3A */ s16 unk3A;
 } ActorEnTest20C; // size = 0x3C
 
+struct ActorUnkStruct;
```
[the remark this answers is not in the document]
Kenix3: They're usually given a name prefixed with the overlay name.

src/overlays/actors/ovl_En_Dno/z_en_dno.c
Kenix3: My preference is `&gfxOpa[1]`

src/overlays/actors/ovl_En_Fsn/z_en_fsn.c
Kenix3: Redundant when we have an enum. Just wait for that to standardize. However, we should use the macro when it matches.

EllipticEllipsis (approved):

Kenix3 (approved):

include/z64actor.h
```diff
@@ -76,6 +76,31 @@ typedef struct {
...
+    /* 0x00 */ Vec3f unk_00;
+    /* 0x0C */ UNK_TYPE1 unk_0C[0x4];
+    /* 0x10 */ s16 unk_10;
+    /* 0x12 */ UNK_TYPE1 unk_12[0xA];
+    /* 0x1C */ u8 unk_1C;
+    /* 0x1D */ u8 unk_1D;
+    /* 0x20 */ Vec3f unk_20;
+    /* 0x2C */ Vec3f unk_2C;
+    /* 0x38 */ UNK_TYPE1 unk_38[0x10];
+    /* 0x48 */ struct Actor* unk_48;
```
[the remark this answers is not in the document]
Kelebek1: Has to be here because it's defined before Actor. Can't build without this.

src/overlays/actors/ovl_En_Dno/z_en_dno.c
```diff
@@ -36,94 +84,949 @@ const ActorInit En_Dno_InitVars = {
...
+        ret = Math_ScaledStepToS(arg0, arg1, arg5);
+    }
+
+    return ret;
+}
+
+void func_80A714B4(EnDno* this, GlobalContext* globalCtx) {
+    Actor* actor = NULL;
+
+    do {
```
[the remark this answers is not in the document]
Kelebek1: I don't think so in this case. There an unknown loop count due to following the chain of actors, so it should be a while. And then the ending condition is actor being null, which being initialised to null at the start. I can't make it a basic while loop or it'll never run, because actor != null will immediately be false. I can't move the function call out of the loop and also then have it inside the loop, because that just duplicates the function call which isn't right.

I really don't see any way around it here, but if you can whip up some ideas I'm all ears.

AngheloAlf (approved):

AngheloAlf:
This PR failed

PR #89, opened 2021-03-28
Animation system updated, some more boot files decompiled (+6%), z_fcurve_data_skelanime decompiled (1 non-matching), some asm files split, etc

shawlucas:
I don't even recall everything I did, but basically its everything I have reserved on trello.

Also, it seems that preprocess.py runs faster when importing asm_processor as a module rather than calling it as subprocess. I'm not very good with python but I tried it (the same way that oot does it) and it was a bit faster (measure it with time command on linux). The makefile now explicitly calls python with "python3 -B" in order to prevent the creation of __pycache__ in the asm-processor folder.

Some of the early commits in this PR are from my previous PR, as I was working off that. Idk how that affects merging if at all (sorry im bad at git :( )

PR #248, opened 2021-08-04
sched.c OK

Thar0:
---

I had to fiddle with a lot of libultra headers while working on this. I hope to do a cleanup of some global headers in the near future, likely after #89 is in.
I'm not entirely sure on what level of documentation is expected so any feedback on that is much appreciated.


EllipticEllipsis (commented):
The actual code in this is a bit beyond me, but I'd say the level of documentation here is more than sufficient. I agree on the global headers stuff too: there's a number of adjustments that would be a good idea to make sooner rather than later.

src/code/sched.c
```diff
@@ -1,47 +1,602 @@
...
+ */
+void Sched_SendGfxCancelMsg(SchedContext* sched) {
+    osSendMesg(&sched->interruptQ, RSP_GFX_CANCEL_MSG, OS_MESG_BLOCK);
+}
+
+/**
+ * Fault Client for the scheduler. Reports information about the state of the scheduler
+ * and any current tasks in the crash debugger.
+ */
+void Sched_FaultClient(void* param1, void* param2) {
```
Kenix3, line 517:
```suggestion
void Sched_FaultClient(void* context, void* param2) {
```
or something like that? Why not type the argument?

Kenix3 (changes requested):
Great work! Just a couple of small things.

src/code/sched.c
Thar0: This is a fault client callback, the args are deliberately left generic much like the final arg of an override or post limb draw function in the skelanime system, since this callback can be passed any two arguments. Due to this, warnings are raised if I change the type to be more specific than `void*`, hence the cast. The casts in other callback functions with generic arguments have been known to affect codegen also, so I feel it's best to keep it consistent even when it doesn't matter.

Kenix3 (approved):

include/variables.h
```diff
@@ -3455,28 +3455,26 @@ extern StackEntry audioStackEntry;
...
-// extern UNK_TYPE1 D_801FBAC0;
+// extern UNK_TYPE1 sRSPOtherStartTime;
 // extern UNK_TYPE1 D_801FBAC4;
-// extern UNK_TYPE1 D_801FBAC8;
+// extern UNK_TYPE1 sRDPStartTime;
 // extern UNK_TYPE1 D_801FBACC;
-// extern UNK_TYPE1 D_801FBAD0;
-// extern UNK_TYPE1 D_801FBAD4;
+// extern UNK_TYPE1 gAudioSPDataPtr;
+// extern UNK_TYPE1 gAudioSPDataSize;
```
AngheloAlf, line 3468: Shouldn't this variables be either uncommented or removed?

AngheloAlf (approved):

PR #253, opened 2021-08-06
Color_RGBAu32

hensldm:
---

This was suggested in one of my earlier PRs, but wanted to wait for the NBS before implementing. Basically I added a new color struct that uses `u32` for each component, and used it to simplify the shopkeeper structs, along with unifying `cursorX`, `cursorY`, and `cursorZ` into a `Vec3f`. While I was in color.h I also noticed we had yet to change `RGB5A1` (deprecated by ZAPD) to `RGBA16` so went ahead and did that.


engineer124 (commented):
Nice. My only comment is that could you also rename two functions in `src/code/sys_initial_check.c`? They are:
`Check_WriteRGB5A1Pixel` and `Check_ClearRGB5A1`

AngheloAlf (approved):

EllipticEllipsis (approved):
Nice, this looks much better.

rylieb (approved):

Kenix3 (approved):

PR #257, opened 2021-08-07
sqrtf.c OK

EllipticEllipsis:
---

I must get out of the habit of doing such large PRs...


src/boot_O2/sqrtf.c
```diff
@@ -0,0 +1,9 @@
+#include "global.h"
+
+#ifndef __GNUC__
+#define __builtin_sqrtf sqrtf
```
AngheloAlf, line 4: If __GNUC__ is not defined would this produce an infinite recursion, right?

EllipticEllipsis: One would think so, but 
```c
f32 sqrtf(f32 f) {
    return sqrtf(f);
}
```
compiles fine, with no warnings and gives OK.

AngheloAlf: Really? I need an explanation...

Thar0: `sqrtf` is a compiler intrinsic, meaning when it's used inside a function it doesn't actually call the function `sqrtf`, a `sqrt.s` MIPS instruction gets inlined instead, so there is no recursion happening here despite how it appears.

EllipticEllipsis: Does that mean that this function (the one defined in this file) is actually pointless, then? Other functions will never call it, instead using the intrinsic?

AngheloAlf: That means that this function is never used? Every call to this function would trigger that compiler intrinsic, right?

AngheloAlf (approved):

Kenix3 (approved):

Kenix3:
Needs +1 reviewer for merge.

isghj5 (approved):

PR #259, opened 2021-08-07
Sort actor functions and extern'd symbols

Kelebek1:
---

This script gathers neccessary forward-declare functions and also sorts extern symbols, printing them to the console.

### DIFF
--- a/tools/sort_actor_funcs.py
+++ b/tools/sort_actor_funcs.py
@@ -3,7 +3,7 @@
 from pathlib import Path
 import argparse
 
-parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(description="Script to sort forward-declared functions and extern'd symbols in actors")
 parser.add_argument("file", type=str, help="Path to c file to sort")
 args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


tools/sort_actor_funcs.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+
+from pathlib import Path
+import argparse
+
+parser = argparse.ArgumentParser()
```
AngheloAlf, line 6: Don't forget to add the description here.

tools/sort_actor_funcs.py
```diff
@@ -0,0 +1,90 @@
...
+            extern_set.add(extern_name)
+        if end_func_line == 0:
+            end_func_line = i
+            start_extern_line = i
+        last_extern = i
+    i += 1
+end_extern_line = last_extern + 1
+
+assert start_func_line > 0 and end_func_line > 0
+assert start_extern_line > 0 and end_extern_line > 0
```
AngheloAlf, line 64: I pulled this branch to try it and it show me this error
```
$ ./tools/sort_actor_funcs.py src/overlays/actors/ovl_En_Ma_Yto/z_en_ma_yto.c 
Traceback (most recent call last):
  File "./tools/sort_actor_funcs.py", line 63, in <module>
    assert start_func_line > 0 and end_func_line > 0
AssertionError
```
I don't really understand what the problem is by just looking at the error. It would be nice if it had a more descriptive error message instead of a plain assert.

Kelebek1: It's trying to find where you've forward-declared functions already to remove them, needed if directly editing the file but since I removed that and just print them I could remove it.

AngheloAlf: Either removing it or adding a more clear error message would be awesome.

AngheloAlf (approved):

hensldm (approved):

Kenix3 (approved):

PR #260, opened 2021-08-07
Sort boot files into appropriate folders

Thar0:
---
Sorts the contents of the `boot/` folder into the appropriate folder for each file. Most of it was libultra.
The compiler flags should be correct for boot now, I wasn't sure about the `__os` files so I just went ahead and brought the decompiled versions over from OoT instead of remaining unsure about their flags. Interestingly, in OoT they're `-g` but in MM they're `-O2`.
Also split `sleep.c` from `printutils.c` and renamed `memcmp` to `bcmp`.
Also deleted c files that were still present for handwritten asm files.


Makefile
```diff
@@ -133,26 +133,39 @@ DEP_FILES := $(O_FILES:.o=.asmproc.d)
...
+build/src/libultra/%.o: CC := python3 tools/asm-processor/build.py $(CC_OLD) -- $(AS) $(ASFLAGS) --
+
+build/src/code/%.o: CC := python3 tools/asm-processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 
-CC := python3 tools/asm-processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+build/src/overlays/actors/%.o: CC := python3 tools/asm-processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+build/src/overlays/effects/%.o: CC := python3 tools/asm-processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+build/src/overlays/fbdemos/%.o: CC := python3 tools/asm-processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+build/src/overlays/gamestates/%.o: CC := python3 tools/asm-processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+build/src/overlays/kaleido_scope/%.o: CC := python3 tools/asm-processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
```
AngheloAlf, line 168: I'm curious, is there any reason to put those explicitly here?

Makefile
```diff
@@ -133,26 +133,39 @@ DEP_FILES := $(O_FILES:.o=.asmproc.d)
...
 build/src/boot_O2_g3/fault_drawer.o: CFLAGS += -trapuv
 
-build/src/libultra/%: CC := $(CC_OLD)
-build/src/libultra/io/%: CC := python3 tools/asm-processor/build.py $(CC_OLD) -- $(AS) $(ASFLAGS) --
-build/src/libultra/os/%: CC := python3 tools/asm-processor/build.py $(CC_OLD) -- $(AS) $(ASFLAGS) --
-build/src/libultra/voice/%: CC := python3 tools/asm-processor/build.py $(CC_OLD) -- $(AS) $(ASFLAGS) --
+build/src/libultra/libc/ll.o: OPTFLAGS := -O1
+build/src/libultra/libc/ll.o: MIPS_VERSION := -mips3 -32
+build/src/libultra/libc/llcvt.o: OPTFLAGS := -O1
+build/src/libultra/libc/llcvt.o: MIPS_VERSION := -mips3 -32
```
AngheloAlf, line 154: Any special reason to leave a gap between the libultra files?

Makefile
Thar0: This way it will be easier to remove the asm-processor dependence for individual folders once their contents no longer needs to be run through it.

Makefile
Thar0: I figured it was clearer organisation-wise this way, it's not necessary. I'm not especially attached to it so if it's not really important it can go if it's not very useful.

AngheloAlf: I see, no problem, it's fine this way.

AngheloAlf (approved):

Makefile
Kenix3: I prefer the gap.

Kenix3 (approved):
Merge is pending +1 reviewer approval.

src/boot_O2_g3/syncprintf.c
```diff
@@ -1,10 +1,11 @@
...
+void __osSyncVPrintf(const char* fmt, ...) {
 }
-void osSyncPrintf(char* str, ...) {
+
+void osSyncPrintf(const char* fmt, ...) {
 }
-void rmonPrintf(UNK_TYPE4 param_1, ...) {
+
+void rmonPrintf(const char* fmt, ...) {
 }
```
EllipticEllipsis, line 11: Two things about this file:
1. I assume this is that IDO stubbing we were discussing earlier?
2. Is the first function incorrectly-named, then?

Thar0: I think this is manual stubbing, for debug vs non-debug.
The first function appears to be incorrectly named, since `const char* fmt, va_list args` doesn't match. But looking at the debug sources, that "should" be what that function is, so it's weird. It's not called anywhere either, so we can't fall back to that for a good name either.

EllipticEllipsis: I suppose it doesn't really matter if it's not actually used, then. I think retail OoT has a similar file in boot.

EllipticEllipsis (approved):
Looks good as far as I can tell.

PR #272, opened 2021-08-11
ObjMakeoshihiki OK

gamestabled:
---

I needed to change some of the actor function prototypes in order to get `0xFFF1` to not become `-0xF`. I checked those functions out briefly with mips_to_c and it seems like it should be fine. Im not sure where the progress on z_actor.c is right now but maybe it should be checked if those functions can be matcheded with s32 instead of s16


include/functions.h
```diff
@@ -905,10 +905,10 @@ Actor* Actor_Spawn(ActorContext* actorCtx, GlobalContext* globalCtx, s16 index,
                    s16 rotZ, s16 sParm10);
 ActorInit* Actor_LoadOverlay(ActorContext* actorCtx, s16 index);
 Actor* Actor_SpawnAsChildAndCutscene(ActorContext* actorCtx, GlobalContext* globalCtx, s16 index, f32 x, f32 y, f32 z,
-                                     s16 rotX, s16 rotY, s16 rotZ, s16 variable, u32 cutscene, s32 param_12,
+                                     s16 rotX, s16 rotY, s16 rotZ, s32 variable, u32 cutscene, s32 param_12,
                                      Actor* parent);
 Actor* Actor_SpawnAsChild(ActorContext* actorCtx, Actor* parent, GlobalContext* globalCtx, s16 index, f32 x, f32 y,
-                          f32 z, s16 rotX, s16 rotY, s16 rotZ, s16 variable);
+                          f32 z, s16 rotX, s16 rotY, s16 rotZ, s32 variable);
```
isghj5, line 911: Huh? Variable for these is double the regular length?

That doesn't sound right, did you get a fake match earlier?

src/overlays/actors/ovl_Obj_Makeoshihiki/z_obj_makeoshihiki.c
```diff
@@ -20,12 +19,71 @@ const ActorInit Obj_Makeoshihiki_InitVars = {
     (ActorFunc)NULL,
 };
 
-#endif
+s32 ObjMakeoshihiki_GetPathIndex(ObjMakeoshihiki* this, GlobalContext* globalCtx) {
```
hensldm, line 22: I think this is too close a name to the macro `OBJMAKEOSHIHIKI_GET_PATHLISTINDEX` though they do different things. I feel something along the lines of `ObjMakeoshihiki_GetChildSpawnPointIndex` would be more appropriate.

src/overlays/actors/ovl_Obj_Makeoshihiki/z_obj_makeoshihiki.c
```diff
@@ -20,12 +19,71 @@ const ActorInit Obj_Makeoshihiki_InitVars = {
...
+    Vec3f pathPointF;
+
+    child = this->actor.child;
+    if (child != NULL) {
+        if (child->update == NULL) {
+            this->actor.child = NULL;
+        } else {
+            for (loopPathIndex = 0; loopPathIndex < this->pathCount; loopPathIndex++) {
+                Math_Vec3s_ToVec3f(&pathPointF, &this->pathPoints[loopPathIndex]);
+                if (Math3D_DistanceSquared(&this->actor.child->world.pos, &pathPointF) < 0.25f) {
```
hensldm, line 82:
```suggestion
                if (Math3D_DistanceSquared(&this->actor.child->world.pos, &pathPointF) < SQ(0.5f) {
```

include/functions.h
AngheloAlf: I'm working in decomping those, and I can confirm the `s32` is needed here.

src/overlays/actors/ovl_Obj_Makeoshihiki/z_obj_makeoshihiki.c
```diff
@@ -7,7 +7,6 @@
 void ObjMakeoshihiki_Init(Actor* thisx, GlobalContext* globalCtx);
```
AngheloAlf, line 7: Could you add the file description?
For example: https://github.com/zeldaret/mm/blob/9eb438860cf1bcfe3ee31b1f50c43114606d4f79/src/overlays/actors/ovl_En_Ma4/z_en_ma4.c#L1

isghj5 (approved):

jpburnett (approved):
Noice

AngheloAlf (approved):

Kenix3 (approved):
Looks good to me.

PR #278, opened 2021-08-20
Update asm-processor and diff.py

AngheloAlf:
---


Thar0 (changes requested):
As I've expressed before, I don't believe it's at all necessary to subrepo asm-processor. The only file we use from the repo that needs to be synced with upstream is the python3 script itself. Further, for organizational reasons I believe build.py should sit alongside asm-processor.py, which isn't feasible with a subrepo.

AngheloAlf:
Why build.py can't sit there if the folder is a subrepo?

Thar0:
It stands a high chance of being deleted when updates to the subrepo happen, and manually fixing that every time sounds worse than just manually copying the one file that we need. Not to mention, I believe subrepo folders really should just contain what's in the repo, it just doesn't seem to right to have things in those that aren't part of the associated upstream repo.

AngheloAlf:
I see. I'll undo it then

Thar0 (approved):

EllipticEllipsis (approved):

Kenix3 (approved):

PR #287, opened 2021-09-02
Two minor fixes

EllipticEllipsis:
---

Two very minor things:
- changed the `CC_CHECK` to use `-funsigned-char` since IDO apparently uses unsigned `char`s (we should think about if we want to avoid `char` altogether, but we don't really need to make that change yet since we are nowhere near supporting other compilers).
- Changed `sym_info.py` to print "VROM" and "VRAM" instead of "ROM" and "RAM", since technically it's always been VRAM, and the mapfile (which `sym_info.py` reads to get its data) only knows about VROM. I didn't change the internal variable names since anyone modifying the script hopefully knows enough to distinguish.
- Run the formatter, seems like a few recent PRs missed it.


AngheloAlf (approved):

jpburnett (approved):
lgtm

Kenix3 (approved):

PR #308, opened 2021-09-11
Improve check new warnings

EllipticEllipsis:
---

Changed check_new_warnings in two ways:
- stderr is now output to the terminal as well as redirected to file: this makes diagnosis of warnings easier should the build fail.
- colour is retained in terminal output, but any ANSI formatting codes are removed from the file, which I think gives maximum readability for both.
(This did require one slight tweak to the makefile as well, to force GCC to output colour.)


tools/warnings_count/check_new_warnings.sh
```diff
@@ -6,11 +6,29 @@ set -e
...
+    python3 tools/warnings_count/compare_warnings.py tools/warnings_count/warnings_$1_current.txt tools/warnings_count/warnings_$1_new.txt    
+}
+
 make distclean
-make setup 2> tools/warnings_count/warnings_setup_new.txt
-make disasm 2> tools/warnings_count/warnings_disasm_new.txt
-make all 2> tools/warnings_count/warnings_build_new.txt
+make_warnings setup setup
+make_warnings disasm disasm
+make_warnings all build
```
rylieb, line 31:
```suggestion
make diff-init

```
Not 100% sure if this is necessary or not, but I thought check_new_warnings.sh was running a full `make init`, and it seems like it is doing all the pieces of it except for `make diff-init`. I presume this means diff.py will not work after running it? Unless `make distclean` doesn't actually remove the diff.py stuff.

rylieb (approved):

tools/warnings_count/check_new_warnings.sh
EllipticEllipsis: `distclean` does remove `expected`. The point is not to run `init`, the point is to test all the stages that can produce warnings (that Jenkins tests for). Running `make diff-init` separately after will just copy everything to the expected folder (assuming a successful build), so I don't really see much point in adding more time on to what is already a very time-consuming script to run for most people, who are used to using `-jN`.

AngheloAlf (approved):

tools/warnings_count/check_new_warnings.sh
rylieb: Fair, that makes sense. Though I can't imagine a simply copy operation of ~100 MB of files takes hardly any time at all. But it does make sense to conserve the number of disk writes at the very least, especially if check_new_warnings is being run multiple times in a row. Probably a question for another PR, but I wonder if we couldn't just create symlinks in `expected` to the files in repo root to avoid having to copy them all the time.

EllipticEllipsis: The whole point of `expected` is to remain untouched and level with master (or a known good build) so that you can diff against it, though!

rylieb: Ah I see. I wonder if it would maybe make sense to have two makefile targets for building the roms. One that names the output file as what diff.py expects for a "known good build", and one that's named something else so that it can be used as "scratch" (run periodically to check matching). That might be a whole new build system PR at that point though and probably not worth it, just a musing.

EllipticEllipsis:
Made some substantial improvements to the script: it now has
- support for multithreading (pass `-j N`)
- warning before beginning
- the script's output is coloured and more descriptive
- a summary of the number of current, new and extra warnings

The only thing I sort of wish I could add is to check individual stages of the build, but this is not really possible since, e.g. the first two steps really have to be run to check the last step's warnings properly anyway.

AngheloAlf (approved):

Kenix3 (approved):

PR #310, opened 2021-09-14
`ovl_select`: ZELDA MAP SELECT matched and documented 

AngheloAlf:
---

Overall it was very similar to the OoT version, so I reused some stuff from it like functions names. It has some small differences:
- The stick can be used to browse the scenes too, and it goes faster than the control pad.
- OPT does something here! It changes in which entrance Link spawns in the selected scene.
- It has some unused labels for the selected form.


src/overlays/actors/ovl_En_Bigpo/z_en_bigpo.c
```diff
@@ -179,10 +179,22 @@ static u8 D_80B65078[] = {
...
+        -1000.0f,
+        1500.0f,
+        -2000.0f,
+    },
+    {
+        -1000.0f,
+        1500.0f,
+        2000.0f,
+    },
+};
```
EllipticEllipsis, line 197:
```suggestion
    {
        2000.0f,
        4000.0f,
        0.0f
    },
    {
        -1000.0f,
        1500.0f,
        -2000.0f
    },
    {
        -1000.0f,
        1500.0f,
        2000.0f
    },
};
```
and reformat.

src/overlays/actors/ovl_En_Bji_01/z_en_bji_01.c
```diff
@@ -304,7 +304,7 @@ void EnBji01_DialogueHandler(EnBji01* this, GlobalContext* globalCtx) {
 
 void func_809CD634(EnBji01* this, GlobalContext* globalCtx) {
     func_801A5BD0(0x6F);
-    func_801A89A8(0xE0000101);
+    Audio_QueueSeqCmd(0xE0000101);
```
EllipticEllipsis, line 307: @rylieb will be interested to see this function name, I think. I remember us discussing these magic numbers before.

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -1,40 +1,933 @@
...
+
+    GfxPrint_SetPos(printer, 4, 25);
+    GfxPrint_SetColor(printer, 255, 255, 55, 255);
+
+    switch (csIndex) {
+        case 0:
+            // clang-format off
+            // "Afternoon-jara"
+            gSaveContext.time = CLOCK_TIME(12, 0); stage = "\x8Dｵﾋﾙ\x8Cｼﾞｬﾗ";
+            // clang-format on
```
EllipticEllipsis, line 726: I'd note somewhere that '-jara' is a Kokiri speech quirk. Also, I assume this is same-line memes, and might be worth a comment to that effect somewhere.

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -1,40 +1,933 @@
...
+    if ((dREG(80) >= 0) && (dREG(80) < this->count)) {
+        this->currentScene = dREG(80);
+        this->topDisplayedScene = dREG(81);
+        this->pageDownIndex = dREG(82);
+    }
+
+    Game_SetFramerateDivisor(&this->state, 1);
+    gSaveContext.cutscene = 0;
+    gSaveContext.playerForm = PLAYER_FORM_HUMAN;
+    gSaveContext.linkAge = 0;
```
EllipticEllipsis, line 932: I've heard of 5 and 17, and MM has some new ones, but 0...

src/overlays/gamestates/ovl_select/z_select.h
```diff
@@ -3,12 +3,39 @@
...
+typedef struct SelectContext {
+    /* 0x000 */ GameState state;
+    /* 0x0A8 */ View view;
+    /* 0x210 */ s32 count;
+    /* 0x214 */ SceneSelectEntry* scenes;
+    /* 0x218 */ s32 currentScene;
+    /* 0x21C */ s32 pageDownIndex; // Index of pageDownStops
+    /* 0x220 */ s32 pageDownStops[7];
+    /* 0x23C */ char unk_23C[0xC];
+    /* 0x248 */ s32 opt;
```
EllipticEllipsis, line 28: I'd note that this is the name the menu uses and what its actual use is, since it's not obvious in the struct.

src/overlays/gamestates/ovl_select/z_select.c
AngheloAlf: I think that's how `linkAge` works. 
As a reference from OoT: <https://github.com/zeldaret/oot/blob/b9ad6aac524242a633927f8e0bfc426c6d4f87cb/src/overlays/gamestates/ovl_select/z_select.c#L192>

EllipticEllipsis: Wonder if it's used in any other ways in MM.

src/overlays/actors/ovl_En_Bji_01/z_en_bji_01.c
engineer124: How these are handled is an audio WIP, but to give an idea of what the general approach will look like, I'm thinking of following this approach with how it's done in OoT atm:
https://github.com/zeldaret/oot/blob/b9ad6aac524242a633927f8e0bfc426c6d4f87cb/src/code/code_800F9280.c#L16-L25

src/overlays/gamestates/ovl_select/z_select.c
AngheloAlf: I think we never put a comment about the same-line meme, the same way we don't put a comment about those random `if(1) {}` needed in the codebase

include/bgm.h
```diff
@@ -0,0 +1,6 @@
+#ifndef BGM_H
+#define BGM_H
+
+#define NA_BGM_STOP 0x100000FF
+
+#endif
```
engineer124, line 6: I think having this macro in is fine for now, but I suspect it might be replaced by a more general macro. I can't say for certain until I get to decomping `code_801A7B10.c` and have a better understanding of how these commands work. It may warrant further discussion, but it'll wait until `code_801A7B10.c` is further along

AngheloAlf: I took it from OoT (that's the whole header). Do you think I should remove it?

src/overlays/gamestates/ovl_select/z_select.c
EllipticEllipsis: True, but it's particularly bad in this file.

include/bgm.h
engineer124: No you can leave it in, it'll help with audio docs later on

engineer124: The OoT header `bgm.h` was last edited on the initial commit of the repo, that's pretty funny. But yeah not recommending any change here, only bringing this up as a setup for future discussions and to be aware that this is being thought of

include/functions.h
```diff
@@ -4013,7 +4013,7 @@ void func_801A7794(u32 param_1, u32 param_2, u32 param_3);
 // void func_801A7B10(void);
 // void func_801A7D04(void);
 // void func_801A7D84(void);
-void func_801A89A8(s32 arg0);
+void Audio_QueueSeqCmd(s32 arg0);
```
engineer124, line 4016:
```suggestion
void Audio_QueueSeqCmd(s32 cmd);
```

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -1,40 +1,934 @@
...
 
 #include "z_select.h"
+#include "alloca.h"
+#include "overlays/gamestates/ovl_title/z_title.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_select/D_80802390.s")
+void Select_LoadTitle(SelectContext* this) {
+    {
+        GameState* state = &this->state;
+        state->running = false;
```
engineer124, line 14:
```suggestion
        GameState* gameState = &this->gameState;
        gameState->running = false;
```
Because this is more internal, using just `state` might be fine. I think I still have a slight preference to staying consistent with using `gameState` instead of `state` throughout where applicable since `state` is still a little too general for me, but it's not a strong opinion and you can leave it if you'd like. This comment would apply through the whole file

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -1,40 +1,934 @@
...
+    gSaveContext.respawn[5].entranceIndex = 0xFF;
+    gSaveContext.respawn[6].entranceIndex = 0xFF;
+    gSaveContext.respawn[7].entranceIndex = 0xFF;
+    D_801BDBB0 = 0;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_select/D_80803588.s")
+    do {
+        GameState* state = &this->state;
+        state->running = false;
+    } while (0);
```
engineer124, line 61: Is there any way to change this into a single `if (1) {}`?

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -1,40 +1,934 @@
...
+
+void Select_PrintCutsceneSetting(SelectContext* this, GfxPrint* printer, u16 csIndex) {
+    const char* stage;
+    const char* day;
+
+    GfxPrint_SetPos(printer, 4, 25);
+    GfxPrint_SetColor(printer, 255, 255, 55, 255);
+
+    // "-jara" used in these strings is a Kokiri speech quirk word
+    switch (csIndex) {
```
engineer124, line 722: This `csIndex`, along with `gSaveContext.cutscene` is interesting to me, how does this cutscene index compare to the cutscene index used in `eventMgr` which deals with scene cutscenes and global cutscenes. The other `csIndex` deals with a range of `0-7F`, so I wonder if they're different. If they're the same, that makes things easier. But if they're different, which I suspect so, we should discuss in discord how to differentiate between them. But this is also outside the scope of this PR, just something we need to look into.

src/overlays/gamestates/ovl_select/z_select.h
```diff
@@ -3,12 +3,39 @@
...
+typedef struct SelectContext {
+    /* 0x000 */ GameState state;
+    /* 0x0A8 */ View view;
+    /* 0x210 */ s32 count;
+    /* 0x214 */ SceneSelectEntry* scenes;
+    /* 0x218 */ s32 currentScene;
+    /* 0x21C */ s32 pageDownIndex; // Index of pageDownStops
+    /* 0x220 */ s32 pageDownStops[7];
+    /* 0x23C */ char unk_23C[0xC];
+    /* 0x248 */ s32 opt; // This name is the one the menu uses to show this value. It can be used to change in which entrance the player spawns in the selected scene
```
engineer124, line 28: So is this like the `entranceIndex`? or `currentEntrance`? I'm not sure if I like the name `opt`, what does it stand for?

engineer124 (commented):
Really nice work with this! I see you selected your files wisely. Just some small comments

src/overlays/gamestates/ovl_select/z_select.c
AngheloAlf: I tried a few times. Nothing else seems to fix it

src/overlays/gamestates/ovl_select/z_select.h
AngheloAlf: I think this is more like `entranceIndex`.
`opt` probably is a shorthand from "option".
It comes from the menu itself. It prints `OPT:` on screen.
image

engineer124:
```suggestion
    /* 0x248 */ s32 opt; // (Option): This name is the one the menu uses to show this value. It can be used to change in which entrance the player spawns in the selected scene
```
Nice, makes sense. Then I'd either add that to the comment or rename `opt` to `option` personally, but I'll leave it up to you

engineer124 (approved):

EllipticEllipsis (approved):

include/alloca.h
```diff
@@ -0,0 +1,7 @@
+#ifndef ALLOCA_H
+#define ALLOCA_H
+
+void* alloca(u32);
+#define alloca  __builtin_alloca
```
Kenix3, line 5: This is likely to have problems with GCC build in the future. Doesn't necessarily need to be addressed right now, but a comment may help.

src/overlays/gamestates/ovl_select/z_select.h
```diff
@@ -3,12 +3,39 @@
...
+    /* 0x24C */ s32 topDisplayedScene; // The scene which is currently at the top of the screen
+    /* 0x250 */ char unk_250[0xC];
+    /* 0x25C */ s32 verticalInputAccumulator;
+    /* 0x260 */ s32 verticalInput;
+    /* 0x264 */ s32 timerUp;
+    /* 0x268 */ s32 timerDown;
+    /* 0x26C */ s32 lockUp;
+    /* 0x270 */ s32 lockDown;
+    /* 0x274 */ s32 unk_274; // unused
+    /* 0x278 */ char unk_278[0x08];
```
Kenix3, line 38:
```suggestion
    /* 0x278 */ UNK_TYPE1 unk_278[0x08];
```

Kenix3 (changes requested):

include/alloca.h
Kenix3: Looks like GCC supports it. Other compilers may not, but it should be fine.
https://github.com/zeldaret/mm/pull/310/files

Kenix3 (approved):

Kenix3:
Merge pending conflict resolution

PR #312, opened 2021-09-15
Initial object file xmls

Thar0:
---

Adds an automatically generated xml for every object file. I also went through and cleaned up some things manually. In particular I added things that weren't detected automatically to the keep files and the link objects. There are still a lot of things that were not caught, mostly textures referenced only from code and structures we as of yet know little to nothing about, and there may be the occasional mis-identification. Symbols from objects have not yet been introduced into code, this should be done on an individual basis, either when the object's contents are being documented or otherwise.


Makefile
```diff
@@ -282,8 +285,13 @@ build/src/libultra/libc/llcvt.o: src/libultra/libc/llcvt.c
 	@$(OBJDUMP) -d $@ > $(@:.o=.s)
 
 # Build C files from assets
+
+assets/%.c: assets/%.xml
+	$(ZAPD) bsf -eh -i $< -o $(dir $<)
+	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o build/$(@:.c=.o) $@
```
AngheloAlf, line 291: Is this necessary?

Makefile
```diff
@@ -282,8 +285,13 @@ build/src/libultra/libc/llcvt.o: src/libultra/libc/llcvt.c
...
+
 build/%.inc.c: %.png
-	$(ZAPD) btex -eh -tt $(lastword ,$(subst ., ,$(basename $<))) -i $< -o $@
+	$(ZAPD) btex -eh -tt $(subst .,,$(suffix $*)) -i $< -o $@
```
AngheloAlf, line 294: What is this doing?

tools/warnings_count/warnings_setup_current.txt
```diff
@@ -0,0 +1,9 @@
+WARNING: Intersection detected from 0x000968:0x000A10 (object_gi_reserve01_DL_000968), conflicts with 0x000990 (object_gi_reserve01_Blob_000990)
+WARNING: Intersection detected from 0x000990:0x000A10 (object_gi_reserve01_Blob_000990), conflicts with 0x0009B8 (object_gi_reserve01_Blob_0009B8)
+WARNING: Intersection detected from 0x0009B8:0x000A10 (object_gi_reserve01_Blob_0009B8), conflicts with 0x0009E0 (object_gi_reserve01_Blob_0009E0)
+WARNING: Intersection detected from 0x0009E0:0x000A10 (object_gi_reserve01_Blob_0009E0), conflicts with 0x000A08 (object_gi_reserve01_Blob_000A08)
+WARNING: Intersection detected from 0x0080F0:0x008108 (object_link_boy_DL_0080F0), conflicts with 0x008100 (object_link_boy_DL_008100)
+WARNING: Intersection detected from 0x01DB40:0x01DB88 (object_link_child_DL_01DB40), conflicts with 0x01DB60 (object_link_child_DL_01DB60)
+When processing file assets/xml/scenes/Z2_SECOM/Z2_SECOM.xml: in input binary file Z2_SECOM, offset 0x001920:
+	[97mTextureCyclingParams::DeclareReferences:[0m[95m warning: [0m[97mTexCycle declared here points to unknown texture at address 0x020047F0. Please declare the texture in the XML to use the proper symbol.
+[0m
```
AngheloAlf, line 9: There is no way to prevent those warnings? Those look very easy to fix tbh

Thar0: All but the last are due to ZAPD bugs.
All of the `object_gi_reserve01` are because of `SPBranchList` stuff, there are no real intersections there, ZAPD just doesn't know where to stop.
Ditto for the `object_link_boy` one.
Ditto for the `object_link_child` one.
The scene one I can look into fixing, I initially didn't since this is a PR for objects.

Makefile
Thar0: Presumably it does the exact same thing, extract the texture format from the filename. I updated it with the version used in OoT.

Makefile
Thar0: It re-extracts if the xml is touched, which is present in OoT, so I figured it would be good to bring over.

tools/warnings_count/warnings_setup_current.txt
EllipticEllipsis: The scene one is fixed in the ZAPD update PR anyway.

Makefile
AngheloAlf: But assets aren't being extracted by running `make`, those are being extracted by `extract_assets,py`.
I feel like that is something old that nobody has deleted in OoT

Thar0: The point is that this rule does extract by running `make`, if any xmls have changed since last setup. I think this can be useful in object decomp. As for whether it's old in OoT I have no idea, and someone from OoT will need to chime in on this and clarify the position of this make rule.

AngheloAlf: In all the time I worked in object decomp for OoT it never worked, I always needed to run `extract_assets.py`
Also, the current `extract_assets.py` only extract XML that has been changed since last extraction, so I don't see the point.

tools/warnings_count/warnings_setup_current.txt
```diff
@@ -0,0 +1,3 @@
...
```
AngheloAlf, line 3: Does this warning still persists?

tools/ZAPD/ZAPD/BuildInfo.h
```diff
@@ -0,0 +1 @@
+const char gBuildHash[] = "dc190122";
```
AngheloAlf, line 1: This file shouldn't be commited to the repo, it is autogenerated.

tools/ZAPD/ZAPD/ZAPD.vcxproj
```diff
@@ -305,6 +305,9 @@ python3 ZAPD/genbuildinfo.py</Command>
   <ItemGroup>
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
```
AngheloAlf, line 310: This looks like something went wrong in the merge
```suggestion
  <ItemGroup>
    <None Include="packages.config" />
  </ItemGroup>
```

tools/ZAPD/ZAPD/ZAPD.vcxproj.filters
```diff
@@ -506,4 +506,7 @@
   <ItemGroup>
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
```
AngheloAlf, line 511: Same here
```suggestion
  <ItemGroup>
    <None Include="packages.config" />
  </ItemGroup>
```

AngheloAlf (approved):

hensldm (approved):
Didn't go over each xml, but overall everything looks good.

Kenix3 (approved):

PR #315, opened 2021-09-17
Fix some misformatted files in current master

tom-overton:
---

When I was working on ArrowIce, I noticed that running `./format.sh` would change files outside of the ones I was touching. Turns out, even in `master` right now, there are some files that have some misformatted code. To make this PR, all I did was run `./format.sh` in `master`, then make a new branch and push it up.

### DIFF
--- a/src/overlays/actors/ovl_En_Bigpo/z_en_bigpo.c
+++ b/src/overlays/actors/ovl_En_Bigpo/z_en_bigpo.c
@@ -179,21 +179,9 @@ static u8 D_80B65078[] = {
 
 // used in limbdraw
 static Vec3f D_80B65084[] = {
-    {
-        2000.0f,
-        4000.0f,
-        0.0f,
-    },
-    {
-        -1000.0f,
-        1500.0f,
-        -2000.0f,
-    },
-    {
-        -1000.0f,
-        1500.0f,
-        2000.0f,
-    },
+    { 2000.0f, 4000.0f, 0.0f },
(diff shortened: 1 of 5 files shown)


src/overlays/actors/ovl_En_Bigpo/z_en_bigpo.c
```diff
@@ -179,10 +179,22 @@ static u8 D_80B65078[] = {
...
+    {
+        -1000.0f,
+        1500.0f,
+        -2000.0f,
+    },
+    {
+        -1000.0f,
+        1500.0f,
+        2000.0f,
+    },
```
EllipticEllipsis, line 196: Lose all the commas on the inner arrays here and reformat: we don't want it like this.

tom-overton: Can you elaborate further? I can't seem to delete any commas here without breaking the build

tom-overton: Oh I think I understand. You mean delete the final commas of the inner arrays, right?

EllipticEllipsis: Yes, those should all end up on one line. Test with or without the final comma on the outer one too, one may look better.

EllipticEllipsis (approved):
Much better.

AngheloAlf (approved):

Kenix3 (approved):

PR #334, opened 2021-09-25
`z_nmi_buff` OK, `code_8010C1B0` split and OK

EllipticEllipsis:
---

Found an extra unrelated function in the file, so separated it. Not sure what best to name it, so left a comment instead (thanks @Thar0 for the info). `gNMIBuffer` doesn't actually seem to be global.


AngheloAlf (approved):

jpburnett (approved):
lgtm

Kenix3 (approved):

PR #335, opened 2021-09-28
`flg_set.c` OK and documented

EllipticEllipsis:
---

Event editor, for all your flag editing needs. And an introduction to printing more than you can fit on the screen in one go. Unused on retail, but should be easy enough to plonk a call in another file.

Well that was an interesting file... little bit of everything. Linker padding in the rodata, bss entangled in another file, some serious reordering to rereorder, and had to fix all the printf calls for regalloc reasons. I can't really import the bss at the moment since it's between symbols used by z_effect, so left it externed for now; it's only a loop variable so I'm not that concerned. I made a padding file for the rodata which seems to work; although I don't really love it as a solution, it is pretty clean and self-contained. And remarkably, IDO put something `const` in rodata for once...

Documented both functions thoroughly, probably excessively, because of the usefulness of this file. There also remains the question of whether `FlagSetEntry` needs to go in `z64.h` since it's only used in this file, as far as I know, but it's there in OoT.


src/code/flg_set.c
```diff
@@ -1,9 +1,314 @@
...
+    { &gSaveContext.maskMaskBit[1], "mask_mask_bit[1]" },
+    { &gSaveContext.maskMaskBit[2], "mask_mask_bit[2]" },
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/flg_set/D_801DC9B8.s")
+    { NULL, NULL }, // used in the code to detect array end
+};
+
+static s32 entryIndex = 0;
+static u32 currentBit = 0;
+static s32 timer = 0;
```
AngheloAlf, line 134: Since those are static, should be prefixed with `s`
```suggestion
static s32 sEntryIndex = 0;
static u32 sCurrentBit = 0;
static s32 sTimer = 0;
```

tools/disasm/variables.txt
```diff
@@ -424,10 +424,10 @@
...
     0x801AE490:("D_801AE490","Color_RGBA8","[4]",0x10),
     0x801AE4A0:("particleOverlayTable","EffectSsOverlay","[39]",0x444),
-    0x801AE8F0:("D_801AE8F0","UNK_PTR","",0x4),
-    0x801AEC70:("D_801AEC70","UNK_TYPE4","",0x4),
-    0x801AEC74:("D_801AEC74","UNK_TYPE4","",0x4),
-    0x801AEC78:("D_801AEC78","UNK_TYPE4","",0x4),
+    0x801AE8F0:("entries","UNK_PTR","",0x4),
+    0x801AEC70:("entryIndex","UNK_TYPE4","",0x4),
+    0x801AEC74:("currentBit","UNK_TYPE4","",0x4),
+    0x801AEC78:("timer","UNK_TYPE4","",0x4),
```
AngheloAlf, line 430: Here too
```suggestion
    0x801AE8F0:("sEntries","UNK_PTR","",0x4),
    0x801AEC70:("sEntryIndex","UNK_TYPE4","",0x4),
    0x801AEC74:("sCurrentBit","UNK_TYPE4","",0x4),
    0x801AEC78:("sTimer","UNK_TYPE4","",0x4),
```

include/z64.h
```diff
@@ -1765,4 +1765,10 @@ typedef struct {
     /* 0x20 */ u16 fracPart[4][4];
 } MatrixInternal; // size = 0x40
 
+typedef struct {
+    /* 0x00 */ u8* value;
+    /* 0x04 */ const char* name;
+} FlagSetEntry; // size = 0x08
+
+
```
AngheloAlf, line 1773: You added an extra line here
```suggestion
} FlagSetEntry; // size = 0x08

```

AngheloAlf (commented):
More debugging stuff :eyes: 
Good work. Just some minor comments

rylieb (approved):

AngheloAlf (approved):
Jenkins is mad

src/code/flg_set.c
```diff
@@ -1,9 +1,314 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/flg_set/D_801DC9B8.s")
+    { NULL, NULL }, // used in the code to detect array end
+};
+
+static s32 sEntryIndex = 0;
+static u32 sCurrentBit = 0;
+static s32 sTimer = 0;
+
+void FlagSet_Update(GameState* gameState) {
+
```
m4xw, line 137: \n gone wild

src/code/flg_set.c
```diff
@@ -1,9 +1,314 @@
...
+    GlobalContext* globalCtx = (GlobalContext*)gameState;
+    Input* input = CONTROLLER1(globalCtx);
+
+    /* Intra-byte navigation */
+
```
m4xw, line 142: as above

src/code/flg_set.c
```diff
@@ -1,9 +1,314 @@
...
+            sTimer = 2;
+        }
+        if (CHECK_BTN_ALL(input->cur.button, BTN_DRIGHT)) {
+            sCurrentBit--;
+            sTimer = 2;
+        }
+    }
+
+    /* Navigation between bytes */
+
```
m4xw, line 165: another nit

src/code/flg_set.c
```diff
@@ -1,9 +1,314 @@
...
+            if (sEntryIndex > 100) {
+                sEntryIndex = 100;
+            }
+            sTimer = 2;
+        }
+    }
+
+    /* Other controls */
+
+    // A toggles the selected flag
```
m4xw, line 235: put them together, keep c style tho for the top

m4xw (approved):
nits

src/code/flg_set.c
EllipticEllipsis: The problem with removing the newlines is it makes it look like the comments only apply to the code block immediately after rather than a section. I don't think the way I've done it is that unusual, aside from mixing C- and C++-style: plenty of GCC headers do this, for example.

PR #339, opened 2021-10-01
Script to check global bss ordering

EllipticEllipsis:
---

Wrote a script to check for bss reordering by reading the built and expected mapfiles. Assuming a non-shifted rom, it will print all the bss in files where it is reordered, and an indication of if it correct or not. Will also report symbols that differ. Sample output:

image

Also prints a summary at the bottom:
image

### DIFF
--- a/tools/global_bss_check.py
+++ b/tools/global_bss_check.py
@@ -11,13 +11,7 @@ regex_fileDataEntry = re.compile(r"^\s+(?P<section>[^\s]+)\s+(?P<vram>0x[^\s]+)\
 regex_bssEntry = re.compile(r"^\s+(?P<vram>0x[^\s]+)\s+(?P<name>[^\s]+)$")
 regex_label = re.compile(r"^(?P<name>L[0-9A-F]{8})$")
 
-
-File = collections.namedtuple("File", ["name", "vram", "bssVariables"])
-FileEntry = collections.namedtuple("File", ["vram", "bssVariables"])
-Variable = collections.namedtuple("Variable", ["name", "vram"])
-# Compared = collections.namedtuple("Compared", ["expected", "build", "diff"])
-
-VarInfo = collections.namedtuple("Variable", ["file", "vram"])
+VarInfo = collections.namedtuple("VarInfo", ["file", "vram"])
 
 def parseMapFile(mapPath: str):
     with open(mapPath) as f:
@@ -74,18 +68,6 @@ def parseMapFile(mapPath: str):
                     if size > 0:
                         inFile = True
                         currentFile = name
-                        
-    # print(symbolsDict)
-    # resultFileDict = dict()
-
-    # for file in filesList:
-    #     bssCount = len(file.bssVariables)
-
-    #     # Filter out files with no bss
-    #     if bssCount == 0:
-    #         continue
-
-    #     resultFileDict[file.name] = FileEntry(file.vram, file.bssVariables)
 
     return symbolsDict
 
--- a/tools/global_bss_check.py
+++ b/tools/global_bss_check.py
@@ -1,0 +1,234 @@
+#!/usr/bin/env python3
+import colorama
+colorama.init()
+
+import argparse
+import os
+import re
+import collections
+
+regex_fileDataEntry = re.compile(r"^\s+(?P<section>[^\s]+)\s+(?P<vram>0x[^\s]+)\s+(?P<size>0x[^\s]+)\s+(?P<name>[^\s]+)$")
+regex_bssEntry = re.compile(r"^\s+(?P<vram>0x[^\s]+)\s+(?P<name>[^\s]+)$")
+regex_label = re.compile(r"^(?P<name>L[0-9A-F]{8})$")
+
+
+File = collections.namedtuple("File", ["name", "vram", "bssVariables"])
+FileEntry = collections.namedtuple("File", ["vram", "bssVariables"])
+Variable = collections.namedtuple("Variable", ["name", "vram"])
+# Compared = collections.namedtuple("Compared", ["expected", "build", "diff"])
+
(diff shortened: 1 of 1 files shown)


rylieb (approved):

tools/global_bss_check.py
```diff
@@ -0,0 +1,216 @@
...
+        print(colorama.Fore.RED + "  BAD" + colorama.Style.RESET_ALL)
+
+        for file in badFiles:
+            print(f"bss reordering in {file}", file=os.sys.stderr)
+        print("", file=os.sys.stderr)
+        
+        if not args.no_fun_allowed:
+            print(colorama.Fore.LIGHTWHITE_EX +
+            "  BSS is REORDERED!!\n"
+            "  Oh! MY GOD!!"
```
rylieb, line 178: If anyone was wondering, this message comes from a debugging screen the devs built in for when the RCP gets hung. I 100% approve of its inclusion in this PR 💯 

rcp

tools/global_bss_check.py
```diff
@@ -0,0 +1,216 @@
...
+    print("", file=os.sys.stderr)
+    print(colorama.Fore.GREEN + "  GOOD" + colorama.Style.RESET_ALL, file=os.sys.stderr)
+
+    if args.no_fun_allowed:
+        return 0
+
+    print("\n" + colorama.Fore.LIGHTWHITE_EX +
+    colorama.Back.RED + "                                  " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "         CONGRATURATIONS!         " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "    All global BSS is correct.    " + colorama.Back.RESET + "\n" +
```
rylieb, line 207: This one comes in from the last screen of the built-in crash debugger, which can be accessed by pressing +L, L, Start, R, CR simultaneously when the red underscore is in the top right of the screen from crashing. Speaking of which, there is a capitalization error here that is of the utmost importance to correct.

```suggestion
    colorama.Back.RED + "    All Global BSS is correct.    " + colorama.Back.RESET + "\n" +
```

debugger

engineer124 (approved):

tools/global_bss_check.py
```diff
@@ -0,0 +1,216 @@
...
+    return badFiles, missingFiles, comparedDict
+
+
+def printCsv(badFiles, missingFiles, comparedDict, printAll = True):
+    print("Symbol Name,Build Address,Build File,Expected Address,Expected File,Difference,GOOD/BAD/MISSING")
+
+    allGood = True
+    allFound = True
+    
+    problemFiles = set.union(badFiles, missingFiles)
```
AngheloAlf, line 122:
```suggestion
    problemFiles = badFiles | missingFiles
```
Or  this:
`problemFiles = badFiles.union(missingFiles)`
whichever you prefer

tools/global_bss_check.py
```diff
@@ -0,0 +1,216 @@
...
+            symbolGood += colorama.Fore.CYAN + " MOVED" + colorama.Fore.RESET
+        print(f"{symbol},{symbolInfo.buildAddress:X},{symbolInfo.buildFile},{symbolInfo.expectedAddress:X},{symbolInfo.expectedFile},{symbolInfo.diff:X},{symbolGood}")
+
+
+def main():
+    description = "Check that globally visible bss has not been reordered."
+    # TODO
+    epilog = """\
+    N.B. Since this script reads the map files, it can only see globally visible bss; in-function static bss must be examined with other tools.
+    """
```
AngheloAlf, line 150: I think this `TODO` isn't needed anymore, or is it?
```suggestion
    description = "Check that globally visible bss has not been reordered."
    epilog = """\
    N.B. Since this script reads the map files, it can only see globally visible bss; in-function static bss must be examined with other tools.
    """
```

tools/global_bss_check.py
```diff
@@ -0,0 +1,216 @@
...
+        print("", file=os.sys.stderr)
+
+        if not args.no_fun_allowed:
+            print(colorama.Fore.LIGHTWHITE_EX + "  Error, should (not) be in here " + colorama.Style.RESET_ALL, file=os.sys.stderr)
+            print("", file=os.sys.stderr)
+
+        print("Some files appear to be missing symbols. Have they been renamed or declared as static? You may need to remake 'expected'", file=os.sys.stderr)
+
+    if len(badFiles) + len(missingFiles) != 0:
+        return 1
```
AngheloAlf, line 196: The `main` function is retuning a value, but you seem to be ignoring. Is this intended?

tools/global_bss_check.py
```diff
@@ -0,0 +1,216 @@
...
+    colorama.Back.RED + "         CONGRATURATIONS!         " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "    All Global BSS is correct.    " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "             THANK YOU!           " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "      You are great decomper!     " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "                                  " + colorama.Style.RESET_ALL , file=os.sys.stderr)
+
+    return 0
+
+if __name__ == "__main__":
+    main()
```
AngheloAlf, line 215: #re: ignored main return value
I guess you wanted to mimic the C behaviour of the main return value. You would need to do something like this to make it work.
```suggestion
if __name__ == "__main__":
    ret = main()
    exit(ret)
```

tools/global_bss_check.py
EllipticEllipsis: Oops, looks like this isn't used anyway any more.

AngheloAlf (approved):

PR #341, opened 2021-10-01
sched.h

AngheloAlf:
---

As we discussed in discord, this PR aims to unbloat a bit some headers, in this case unbloat `z64.h` with the creation of `sched.h`
`sched.h` is still being included in `z64.h` because some stuff in that header needs it. Changing it to not include it would require a bigger change which is out of the scope of this PR.


include/PR/sched.h
```diff
@@ -0,0 +1,74 @@
+#ifndef SCHED_H
```
Thar0, line 1: The scheduler used in Zelda 64 is different to the scheduler of libultra. I don't think this should be in the PR directory.

include/PR/sched.h
```diff
@@ -0,0 +1,74 @@
...
+    /* 0x0C */ CfbInfo* framebuffer;
+    /* 0x10 */ OSTask   list;
+    /* 0x50 */ OSMesgQueue* msgQ;
+    /* 0x54 */ OSMesg   msg;
+} OSScTask; // size = 0x58
+
+typedef struct IrqMgrClient {
+    /* 0x0 */ struct IrqMgrClient* next;
+    /* 0x4 */ OSMesgQueue* queue;
+} IrqMgrClient; // size = 0x8
```
Thar0, line 50: This doesn't belong here. I think an `irqmgr.h` should be created and included here to contain this and other related structures like the `IrqMgr` structure itself.

Thar0 (changes requested):
This is a good change but I feel some of the specifics should be a little different.

include/PR/sched.h
```diff
@@ -0,0 +1,38 @@
...
+#define OS_SC_DP                0x0001
+#define OS_SC_SP                0x0002
+#define OS_SC_YIELD             0x0010
+#define OS_SC_YIELDED           0x0020
+
+typedef struct OSScTask {
+    /* 0x00 */ struct OSScTask* next;
+    /* 0x04 */ u32      state;
+    /* 0x08 */ u32      flags;
+    /* 0x0C */ void*    framebuffer; // CfbInfo*
```
hensldm, line 27: Out of curiosity, why is this being changed to `void*`?

AngheloAlf: This header is a libultra header,  so it should only contain general types provided by libultra.
`CfbInfo` is part of the Zelda64 specific scheduler (which uses parts of the libultra stuff).

Since libultra headers shouldn't be including Zelda64 specific headers, then it was just left as a `void*`

hensldm: That makes sense. With that in mind, I think I'd prefer to remove the `CfbInfo*` comment out of this header as well and remove all references to z64 specific stuff. Perhaps a macro would be better suited in the z64 scheduler header for this kind of documentation like 
`FRAMEBUFFER(task) ((CfbInfo*)(task)->framebuffer)`.

hensldm (approved):
Just one possible doc suggestion, though am perfectly fine with approving it as is.

include/PR/sched.h
AngheloAlf: That's actually a good idea. I'll change it

Kenix3:
I miscounted before, merge is still pending +1 contributor.

engineer124 (approved):

PR #346, opened 2021-10-03
En_Giant OK

tom-overton:
---

This is a really hard actor to document without being able to see its animations, since most of the actor's functionality is related to playing the right animation in the right circumstances. I documented what I could, like what the various `param` values mean.


include/functions.h
```diff
@@ -1738,7 +1738,7 @@ void Font_LoadOrderedFont(Font* font);
 // void func_800F50D4(void);
 void Kankyo_Init(GlobalContext* globalCtx, EnvironmentContext* envCtx);
 u32 func_800F5954(u8* param_1, u32 param_2, u32 param_3, u8 param_4, u8 param_5);
-// void func_800F5A8C(void);
+f32 func_800F5A8C(u16 param_1, u16 param_2, u16 param_3, GlobalContext* globalContext);
```
EllipticEllipsis, line 1741: Could you change these to use arg0, arg1, etc.? I'd like to standardise this header to only use arg over param

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+extern AnimationHeader D_06013FE8;
+extern AnimationHeader D_06015334;
+extern AnimationHeader D_06017944;
+
+static AnimationHeader* sAnimationTable[] = {
+    &D_06008394, &D_060096E4, &D_060102A4, &D_060116E4, &D_06012A38, &D_06013004, &D_06013FE8, &D_06015334,
+    &D_06017944, &D_0600A1C4, &D_0600D040, &D_0600DE84, &D_0600ACA4, &D_0600B784, &D_0600C5D4,
+};
+
+static TexturePtr sFaceTextures[] = { D_06005A80, D_06006280, D_06006A80 };
```
EllipticEllipsis, line 61: These can go in the Draw function, most likely. You can add some comments to remind you for when you do the object if that's helpful.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+            SkelAnime_ChangeAnim(&this->skelAnime, sAnimationTable[newAnimationId], 1.0f, 0.0f,
+                                 SkelAnime_GetFrameCount(&sAnimationTable[newAnimationId]->common), 2, 10.0f);
+        } else {
+            SkelAnime_ChangeAnimDefaultStop(&this->skelAnime, sAnimationTable[newAnimationId]);
+        }
+        this->animationId = newAnimationId;
+    }
+}
+
+s32 EnGiant_IsNotFreed(EnGiant* this) {
```
EllipticEllipsis, line 76: I think maybe use something like `GiantIsImprisoned` to avoid double negatives.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+        case GIANT_TYPE_CANYON_GIANTS_CHAMBER_AND_ENDING:
+        case GIANT_TYPE_CANYON_CLOCK_TOWER_FAILURE:
+            if (!CHECK_QUEST_ITEM(3)) {
+                return 1;
+            }
+            break;
+    }
+
+    return 0;
+}
```
EllipticEllipsis, line 113:
```suggestion
            if (!CHECK_QUEST_ITEM(0)) {
                return true;
            }
            break;
        case GIANT_TYPE_MOUNTAIN_TERMINA_FIELD:
        case GIANT_TYPE_MOUNTAIN_CLOCK_TOWER_SUCCESS:
        case GIANT_TYPE_MOUNTAIN_GIANTS_CHAMBER_AND_ENDING:
        case GIANT_TYPE_MOUNTAIN_CLOCK_TOWER_FAILURE:
            if (!CHECK_QUEST_ITEM(1)) {
                return true;
            }
            break;
        case GIANT_TYPE_OCEAN_TERMINA_FIELD:
        case GIANT_TYPE_OCEAN_CLOCK_TOWER_SUCCESS:
        case GIANT_TYPE_OCEAN_GIANTS_CHAMBER_AND_ENDING:
        case GIANT_TYPE_OCEAN_CLOCK_TOWER_FAILURE:
            if (!CHECK_QUEST_ITEM(2)) {
                return true;
            }
            break;
        case GIANT_TYPE_CANYON_TERMINA_FIELD:
        case GIANT_TYPE_CANYON_CLOCK_TOWER_SUCCESS:
        case GIANT_TYPE_CANYON_GIANTS_CHAMBER_AND_ENDING:
        case GIANT_TYPE_CANYON_CLOCK_TOWER_FAILURE:
            if (!CHECK_QUEST_ITEM(3)) {
                return true;
            }
            break;
    }

    return false;
}
```
This is returning a boolean. Love the case labels. You can add the quest item enums, no need to wait for @engineer124's PR, provided we agree on what they should be...

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+        case GIANT_TYPE_OCEAN_CLOCK_TOWER_SUCCESS:
+        case GIANT_TYPE_OCEAN_GIANTS_CHAMBER_AND_ENDING:
+            this->unk_24A = 0x1C8;
+            break;
+        default:
+            this->unk_24A = 0x1C5;
+            break;
+    }
+
+    if (type >= GIANT_TYPE_MOUNTAIN_CLOCK_TOWER_SUCCESS && type < GIANT_TYPE_MOUNTAIN_GIANTS_CHAMBER_AND_ENDING) {
```
EllipticEllipsis, line 153: This looks like it could do with some defines. I'd suggest making a define for the whole condition, but I suspect that wouldn't be very popular.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+        if (!(gSaveContext.weekEventReg[0x19] & 2)) {
```
EllipticEllipsis, line 154: @rylieb

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+        this->actor.draw = EnGiant_Draw;
+        this->actor.velocity.y = 0.0f;
+        this->actor.minVelocityY = 0.0f;
+        this->actor.gravity = 0.0f;
+        if (EnGiant_IsNotFreed(this)) {
+            Actor_MarkForDeath(&this->actor);
+        }
+    }
+
+    if (type >= GIANT_TYPE_MOUNTAIN_CLOCK_TOWER_SUCCESS) {
```
EllipticEllipsis, line 183: Ditto here on the defines.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+
+void EnGiant_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+/**
+ * The animations in sAnimationTable are organized such that looping animations
+ * appear immediately after their respective starting animations. The point of
+ * this function is to play the requested start animation if it has not been
+ * played yet and play the respetive looping animation otherwise.
+ **/
```
EllipticEllipsis, line 214:
```suggestion
 */
```

src/overlays/actors/ovl_En_Giant/z_en_giant.h
```diff
@@ -3,14 +3,75 @@
...
-    /* 0x0000 */ Actor actor;
-    /* 0x0144 */ char unk_144[0x154];
-    /* 0x0298 */ EnGiantActionFunc actionFunc;
+    /* 0x000 */ Actor actor;
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ Vec3s jointTable[16];
+    /* 0x1E8 */ Vec3s morphTable[16];
+    /* 0x248 */ s16 animationId;
+    /* 0x24A */ u16 unk_24A;
+    /* 0x24C */ u16 action;
```
EllipticEllipsis, line 68: I don't like the name of this variable, because at first glance it looks like it's related to the actionFunc. `csAction` or `npcAction` would be clearer IMO.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -26,42 +31,449 @@ const ActorInit En_Giant_InitVars = {
...
+            func_800B9010(&this->actor, NA_SE_EV_KYOJIN_SIGN - SFX_FLAG);
+        }
+    }
+}
+
+void EnGiant_UpdatePosition(EnGiant* this, GlobalContext* globalCtx, u32 actionIndex) {
+    CsCmdActorAction* actorAction = globalCtx->csCtx.npcActions[actionIndex];
+    f32 floatUnk10 = actorAction->unk10;
+    s32 pad[2];
+    f32 floatUnk1C = actorAction->unk1C;
```
EllipticEllipsis, line 352: Would be interesting to check the OoT cs system to see if the struct is the same, but you don't have to do it now.

EllipticEllipsis (commented):
Excellent job so far, looking forward to more documentation from the object. As such I haven't gone too much into trying to get every last thing named. I reckon you could also make an enum for the `action` struct variable.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
tom-overton: They can go in the Draw function so long as they remain `static`; making them non-`static` is not OK. I moved them there and added comments explaining what the three faces are for now.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
tom-overton: It's might be the same; I actually already checked this. Here's the OoT struct:
```c
typedef struct {
    /* 0x00 */ u16 action; // "dousa"
    /* 0x02 */ u16 startFrame;
    /* 0x04 */ u16 endFrame;
    union {
        /* 0x06 */ Vec3s rot;
        /* 0x06 */ Vec3us urot;
    };
    /* 0x0C */ Vec3i startPos;
    /* 0x18 */ Vec3i endPos;
    /* 0x24 */ Vec3i normal;
} CsCmdActorAction; // size = 0x30
```

The MM struct is the same size, and assuming everything is in the same position, this function is checking `startPos.y` and `endPos.y`, which would make sense. Should I port this definition into MM? Should I only bring over `action`, `startPos`, and `endPos`? Or should I do nothing, since I haven't decompiled any cutscene code and thus have no clue what this is really doing?

src/overlays/actors/ovl_En_Giant/z_en_giant.c
tom-overton: Do you mean grab the whole `QuestItem` enum and put it in z64item.h? I can do that, and I agree on what the values are, but I just wanted to make sure I'm doing the right thing and not stepping on another programmer's toes.

PR #356, opened 2021-10-15
Hopefully prevent misformatted files from ever entering master again.

rylieb:
---
I'm not really sure if the Jenkinsfile changes work, since I don't have a good way to test that ahead of time, but I've verified the bash files work as expected. Hopefully this PR initially fails Jenkins, and then succeeds it once the misformatted files PR gets merged.


rylieb:
I opened https://github.com/ethteck/decomp-jenkins/pull/4 which was merged. So now we just need all people running Jenkins agents to rebase against upstream and rebuild their docker images.

rylieb:
Converting to draft because we need to figure out how to get clang-format-11 into Jenkins. It doesn't exist in the distro's apt repo, so we need to come up with an alternative solution.

rylieb:
Okay, this is actually fixed in https://github.com/ethteck/decomp-jenkins/pull/7, so now we just need every agent-operator to rebase and rebuild the docker image before Jenkins will be able to check formatting.

rylieb:
unmarking as draft now that one of the jenkins agents has been updated. still waiting on a few more though.

PR #359, opened 2021-10-16
En_Elforg (Stray Fairy) OK and mostly documented

tom-overton:
---

This is a pretty mathematical actor, which is definitely outside my comfort zone. I tried my best to document what it was doing, but true documentation can't happen until Elfgrp is decompiled, since that's the "controller" actor for the Stray Fairies in the fountain.


jpburnett (approved):
Nice work on this 🧚

PR #344, opened 2021-10-03
Adding tools.md

jpburnett:
Before opening this PR, ensure the following:
- `./format.sh` was run to apply standard formatting.
- `make` successfully builds a matching ROM.
- No new compiler warnings were introduced during the build process.
    - Can be verified locally by running `tools/warnings_count/check_new_warnings.sh`
- New variables & functions should follow standard naming conventions.
- Comments and variables have correct spelling.
---

Getting back in to the swing of things I thought I would refresh myself on some of the tools. I filled in what I could / what I knew about the tools. This is also an effort to close #144 to clean things up. Feel free to let me know what is missing. This isn't a high priority to get it.

PR #362, opened 2021-10-17
Tutorial

EllipticEllipsis:
---

The long-awaited tutorial. Went through 3 small, simple actors (perhaps too small and simple in some cases). The main line follows EnRecepgirl, with several other examples taken from EnMs and ObjTree. I feel a bit bad about taking what were actually 3 very beginner-suitable actors out of the pool, as it were, but at the same time I didn't really want to leave them half-finished.

Format is pretty similar to the OoT tutorial: several documents covering the main actor decomp flow, and supplementary ones on specific features and other general useful stuff. I have occasionally linked the OoT tutorial where it has a better example of something available. Some stuff is not totally final, but I thought it was worth getting some more eyes on it now (and didn't want to delay it 3 months like the OoT one...).

Remaining work:
- Object decompilation, to be written once the automatic XMLs from #312 are in.
- If anyone would like to donate a nice Draw function with some colour macros, or even would like to write the bitpacking stuff themselves to improve their understanding, that would be a useful addition.
- Link up with the rest of the documentation, such as the tools doc from #344
- There is a fair amount of overlap between CONTRIBUTING.md and merging.md. We should decide what is better where.
- Glossary
- git document
- Maybe an overview of MIPS from a decomp PoV?
- other things listed in contents.md
- A good demonstration of using diff.py and the permuter, roughly the same as in the OoT tutorial.

This should finish off #144


tools/actorfixer.py
```diff
@@ -141,7 +141,7 @@
     "func_801376DC": "SkelAnime_CopyFrameTableFalse",
     "func_80137748": "SkelAnime_UpdateTranslation",
     "func_801378B8": "Animation_OnFrame",
-    "SkelAnime_CopyVec3s": "SkelAnime_CopyFrameTable",
+    "SkelAnime_CopyVec3s": "SkelAnime_Cop
```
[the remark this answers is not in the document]
EllipticEllipsis: No, that's probably VSCode's fault. And Python's. Definitely nothing to do with me.

AngheloAlf (approved):

src/overlays/actors/ovl_En_Recepgirl/z_en_recepgirl.c
```diff
@@ -28,38 +32,192 @@ const ActorInit En_Recepgirl_InitVars = {
...
+        if (this->skelAnime.animation == &D_0600A280) {
+            Animation_PlayLoop(&this->skelAnime, &D_06001384);
+        } else if (this->skelAnime.animation == &D_0600AD98) {
+            if (this->actor.textId == 0x2ADA) { // "The room on your left is The Mayor's office [...]" (meeting ongoing)
+                Animation_MorphToPlayOnce(&this->skelAnime, &D_06000968, 10.0f);
+            } else {
+                Animation_MorphToLoop(&this->skelAnime, &D_06009890, 10.0f);
+            }
+        } else if (this->actor.textId ==
+                   0x2ADA) { // "The room on your left is The Mayor's office [...]" (meeting ongoing)
```
hensldm, line 143: Got some strange formatting because of the long text comment. Could you perhaps shorten it, or move it a line up?

src/overlays/actors/ovl_En_Ms/z_en_ms.c
```diff
@@ -28,38 +32,159 @@ const ActorInit En_Ms_InitVars = {
...
+    }
+
+    if (func_800B84D0(&this->actor, globalCtx) != 0) {
+        this->actionFunc = EnMs_Talk;
+        return;
+    }
+
+    if ((this->actor.xzDistToPlayer < 90.0f) && (ABS_ALT(yawDiff) < 0x2000)) {
+        func_800B8614(&this->actor, globalCtx, 90.0f);
+    }
```
hensldm, line 100:
```suggestion
    if (func_800B84D0(&this->actor, globalCtx) != 0) {
        this->actionFunc = EnMs_Talk;
    } else if ((this->actor.xzDistToPlayer < 90.0f) && (ABS_ALT(yawDiff) < 0x2000)) {
        func_800B8614(&this->actor, globalCtx, 90.0f);
    }
```
Possible `else if` to remove the early return?

hensldm (commented):
Tutorial looks good. A couple things I noticed about the overlays.

src/overlays/actors/ovl_En_Recepgirl/z_en_recepgirl.c
EllipticEllipsis: Hmm, neither option looks very good. I agree the formatting isn't much good either, though.

EllipticEllipsis: Oh, I know what to do, and it'll make it more symmetrical as well: split the else and if.

EllipticEllipsis: Drat, there's another one further down this won't work for.

src/overlays/actors/ovl_En_Ms/z_en_ms.c
EllipticEllipsis: I remain in two minds about this. I think it's nice to do it where possible: saves indentation, makes it clear that setting the actionFunc is the last thing that happens. But we get cases where it must and can't be done.

docs/tutorial/beginning_decomp.md
```diff
@@ -0,0 +1,691 @@
...
+```C
+EnRecepgirl* this = THIS;
+```
+at the top of the function, where you can see from the top of the file that `THIS` just expands to `(EnRecepgirl*)thisx`. Since `Init` (and the other "main four" functions) will access these struct variables, it is a good idea to change the `Init` prototype to be wrong, to fool mips2c so that it will output the actual variable rather than just its offset in the struct.
+
+That's a lot of words for: change `Actor* thisx` in the `Init` prototype in the C file to `EnRecepgirl* this`:
+```C
+void EnRecepgirl_Init(EnRecepgirl *this, GlobalContext *globalCtx)
+```
+and remake the context file. (It's a good idea to do the same for all of the "main four", so we won't mention it again). *Once you have the function decompiled, remember to put the arguments back to how they should be in both the prototype and the function itself*
```
zbanks, line 216: This shouldn't be required anymore. The latest mips2c can detect this pattern and do it automatically.

Specifically, if:
- The function name starts with `Foo_`
- The first argument has type `Ptr *`
- There exists a struct called `Foo` in the context, whose first member is of type `Ptr`

Then, mips2c it automatically downcasts the first argument from a `Ptr *` into a `Foo *` & creates a `this`/`self` variable.

src/overlays/actors/ovl_Obj_Tree/z_obj_tree.c
```diff
@@ -71,32 +86,115 @@ static DamageTable D_80B9A59C = {
...
+extern Gfx D_06000680[];
+extern Gfx D_060007C8[];
+extern CollisionHeader D_06001B2C;
+
+void ObjTree_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    ObjTree* this = THIS;
+    CollisionHeader* colHeader = NULL;
+
+    if (this->dyna.actor.params & 0x8000) {
```
hensldm, line 100: Macro for the params access. Could even point it out in the tutorial as it is something we require.

src/overlays/actors/ovl_Obj_Tree/z_obj_tree.c
```diff
@@ -71,32 +86,115 @@ static DamageTable D_80B9A59C = {
...
+    this->swayAngle = 0;
+    this->swayVelocity = 0;
+    ObTree_SetupDoNothing(this);
+}
+
+void ObjTree_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    ObjTree* this = THIS;
+    s32 bgId;
+
+    if (!(this->dyna.actor.params & 0x8000)) {
```
hensldm, line 128: Macro

src/overlays/actors/ovl_Obj_Tree/z_obj_tree.c
```diff
@@ -71,32 +86,115 @@ static DamageTable D_80B9A59C = {
...
+        DynaPolyActor_Init(&this->dyna, 1);
+        CollisionHeader_GetVirtual(&D_06001B2C, &colHeader);
+        this->dyna.bgId = DynaPoly_SetBgActor(globalCtx, &globalCtx->colCtx.dyna, &this->dyna.actor, colHeader);
+    }
+
+    Collider_InitCylinder(globalCtx, &this->collider);
+    Collider_SetCylinder(globalCtx, &this->collider, &this->dyna.actor, &sCylinderInit);
+    CollisionCheck_SetInfo2(&this->dyna.actor.colChkInfo, &sDamageTable, &sColchkInfoInit);
+
+    if (this->dyna.actor.params & 0x8000) {
```
hensldm, line 114: Macro

src/overlays/actors/ovl_En_Ms/z_en_ms.c
hensldm: Personally I prefer the `else if`, but I'll leave it up to you.

src/overlays/actors/ovl_Obj_Tree/z_obj_tree.c
```diff
@@ -125,7 +125,7 @@ void ObjTree_Destroy(Actor* thisx, GlobalContext* globalCtx) {
     ObjTree* this = THIS;
     s32 bgId;
 
-    if (!(this->dyna.actor.params & 0x8000)) {
+    if (!(OBJTREE_ISLARGE(&this->dyna.actor))) {
```
hensldm, line 128:
```suggestion
    if (!OBJTREE_ISLARGE(&this->dyna.actor)) {
```
Parenthesis shouldn't be necessary.

docs/tutorial/documenting.md
```diff
@@ -502,13 +502,12 @@ void func_80C102D4(EnRecepgirl* this, GlobalContext* globalCtx) {
         if (this->skelAnime.animation == &D_0600A280) {
             Animation_ChangeDefaultRepeat(&this->skelAnime, &D_06001384);
         } else if (this->skelAnime.animation == &D_0600AD98) {
-            if (this->actor.textId == 0x2ADA) { // "The room on your left is The Mayor's office [...]" (meeting ongoing)
+            if (this->actor.textId == 0x2ADA) { // Mayor's office on is the left (meeting ongoing)
```
hensldm, line 505:
```suggestion
            if (this->actor.textId == 0x2ADA) { // Mayor's office is on the left (meeting ongoing)
```

docs/tutorial/documenting.md
```diff
@@ -502,13 +502,12 @@ void func_80C102D4(EnRecepgirl* this, GlobalContext* globalCtx) {
...
                 Animation_MorphToPlayOnce(&this->skelAnime, &D_06000968, 10.0f);
             } else {
                 Animation_ChangeTransitionRepeat(&this->skelAnime, &D_06009890, 10.0f);
             }
-        } else if (this->actor.textId ==
-                   0x2ADA) { // "The room on your left is The Mayor's office [...]" (meeting ongoing)
+        } else if (this->actor.textId == 0x2ADA) { // Mayor's office on is the left (meeting ongoing)
```
hensldm, line 510: same as above

docs/tutorial/documenting.md
```diff
@@ -523,10 +522,10 @@ void func_80C102D4(EnRecepgirl* this, GlobalContext* globalCtx) {
         if (this->actor.textId == 0x2AD9) { // "Welcome..."
             Actor_SetSwitchFlag(globalCtx, this->actor.params);
             Animation_MorphToPlayOnce(&this->skelAnime, &D_0600AD98, 10.0f);
-            if (gSaveContext.weekEventReg[63] & 0x80) { // meeting ended
-                this->actor.textId = 0x2ADF; // "The room on your left is The Mayor's office [...]" (meeting ended)
+            if (gSaveContext.weekEventReg[63] & 0x80) { // showed Couple's Mask to meeting
+                this->actor.textId = 0x2ADF; // Mayor's office on is the left (meeting ended)
```
hensldm, line 526: Same as above

docs/tutorial/documenting.md
```diff
@@ -523,10 +522,10 @@ void func_80C102D4(EnRecepgirl* this, GlobalContext* globalCtx) {
...
             } else {
-                this->actor.textId = 0x2ADA; // "The room on your left is The Mayor's office [...]" (meeting ongoing)
+                this->actor.textId = 0x2ADA; // Mayor's office on is the left (meeting ongoing)
```
hensldm, line 528: Same as above

docs/tutorial/documenting.md
```diff
@@ -537,7 +536,7 @@ void func_80C102D4(EnRecepgirl* this, GlobalContext* globalCtx) {
             if (this->actor.textId == 0x2ADD) { // "So..."
                 this->actor.textId = 0x2ADE;    // Mayor's office on is the left, drawing room on the right
```
hensldm, line 537: Same as above

docs/tutorial/documenting.md
```diff
@@ -537,7 +536,7 @@ void func_80C102D4(EnRecepgirl* this, GlobalContext* globalCtx) {
...
             } else if (this->actor.textId ==
-                       0x2ADA) {             // "The room on your left is The Mayor's office [...]" (meeting ongoing)
+                       0x2ADA) {             // Mayor's office on is the left (meeting ongoing)
```
hensldm, line 539: Same as above

hensldm (approved):
Looks good. Couple minor things, but not enough to stop an approval.

EllipticEllipsis:
Oops... should have read it before copypasting from the actor...

docs/tutorial/beginning_decomp.md
```diff
@@ -0,0 +1,673 @@
...
+
+```C
+EnRecepgirl* this = THIS;
+```
+
+Now everything points to the right place, even though the argument of the function seems inconsistent with the contents.
+
+(Again: this step is only necessary for the "main four" functions, and sometimes functions that are used by these: it relates to how such functions are used outside the actor.)
+
+While we are carrying out initial changes, you can also find-and-replace any instances of `(Actor *) this` by `&this->actor`. The function now looks like this:
```
zbanks, line 292: This section could also be edited to reflect the other change, sorry I didn't mark it before.
```suggestion
To allow the function to find the variables, we need another correction. Half of this has already been done at the top of the function, where we have

``C
#define THIS ((EnRecepgirl*)thisx)
``

To do the other half, use this macro to replace the declaration of `this` at the beginning of the function:

``C
EnRecepgirl* this = THIS;
``

While we are carrying out initial changes, you can also find-and-replace any instances of `(Actor *) this` by `&this->actor`. The function now looks like this:
```

This last bit *"you can also find-and-replace any instances of `(Actor *) this`"*, is probably not strictly necessary, but I don't think it hurts to leave in.

(NB: I can't use triple-backticks in a suggestion, so you'll have to make this edit yourself instead of using the "accept suggestion" feature)

EllipticEllipsis: Is this section necessary at all now?

zbanks: I don't think it's critical; I the only thing is that `mips2c` will output something like:
```c
EnRecepgirl* this = (EnRecepgirl *) thisx;
```
which can be rewritten to use `THIS`.

EllipticEllipsis: Ah right, yeah that probably is worth adding.

zbanks (approved):

Kenix3 (approved):

PR #375, opened 2021-10-24
En_Goroiwa (3 non-matching)

Kelebek1:
---

3 non-matching just due to stack, code is matching.


Kenix3 (approved):
Merge pending +1 contributor approval.

src/overlays/actors/ovl_En_Goroiwa/z_en_goroiwa.c
```diff
@@ -22,141 +57,1577 @@ const ActorInit En_Goroiwa_InitVars = {
...
+
+void func_80941274(EnGoroiwa* this, GlobalContext* globalCtx) {
+    Audio_PlaySoundAtPosition(globalCtx, &this->actor.world.pos, 30, NA_SE_EV_SNOWBALL_BROKEN);
+}
+
+void EnGoroiwa_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnGoroiwa* this = THIS;
+    f32 temp_f0;
+    s32 params = ENGOROIWA_GET_FF(&this->actor);
+    s32 sp28 = params * 8;
```
hensldm, line 974: I think since params are generally used as bit flags and not actual numbers, I would prefer this to be shifted left 3, instead of multiplied by 8. However don't feel too strongly about it.

hensldm (approved):
Just one very small nit that may just be me, so will approve.

src/overlays/actors/ovl_En_Goroiwa/z_en_goroiwa.c
Kelebek1: Does not match.

PR #377, opened 2021-10-26
Add an explanatory comment header for every actor that doesn't have one

tom-overton:
---

Elliptic and I talked about doing this on discord, so I just decided to get it done. This adds an explanatory comment header for every single actor that does not currently have one. It uses the description for the actor that is currently present on the spreadsheet; some of these descriptions may be slightly inaccurate or just flat-out wrong, but it's better than nothing.


Kenix3 (approved):
I didn't verify that the descriptions are right, but the changes look good.

PR #392, opened 2021-10-28
Match `SkelAnime_GetFrameDataLegacy`

AngheloAlf:
---

This is the only non matching left in z_skelanime.
I matched this one in OoT, and I just noted Lucas didn't update this function when he did his skelanime PR, so here we are.
The function is completely unused. It goes back to the `object_human` era of OoT, which is the only OoT object which still uses this old format.

This looks like a fake match for a loop unroll. If anybody wants to try to do a less fake match here's a scratch: https://decomp.me/scratch/KHaTp


isghj5 (approved):
Ew... good to know this can happen with loops at least.

Kenix3 (approved):

PR #395, opened 2021-10-30
`__osMalloc.c` OK

AngheloAlf:
---

Joining to the `boot` wagon 

I noticed this file wasn't completely decomped (2 functions were missing, bss and unreferenced strings wasn't imported), so I decided to look at it.
The functions weren't hard. I decided to cleanup the file a bit too


src/boot_O2/__osMalloc.c
```diff
@@ -59,15 +59,15 @@ void __osMallocAddBlock(Arena* arena, void* start, s32 size) {
     ArenaNode* lastNode;
 
     if (start != NULL) {
-        firstNode = (ArenaNode*)ALIGN16((u32)start);
-        diff = (s32)firstNode - (s32)start;
+        firstNode = (ArenaNode*)ALIGN16((uintptr_t)start);
+        diff = (intptr_t)firstNode - (intptr_t)start;
```
EllipticEllipsis, line 63: Sounds like we need `ptrdiff_t` as well.

src/boot_O2/__osMalloc.c
```diff
@@ -59,15 +59,15 @@ void __osMallocAddBlock(Arena* arena, void* start, s32 size) {
...
         size2 = (size - diff) & ~0xF;
```
EllipticEllipsis, line 64: Can this have a more descriptive name? What's its purpose?

src/boot_O2/__osMalloc.c
```diff
@@ -179,45 +186,91 @@ void __osFree(Arena* arena, void* ptr) {
...
-            next->prev = prev;
+    ArenaImpl_Unlock(arena);
+}
+
+void* __osRealloc(Arena* arena, void* ptr, size_t newSize) {
+    ArenaImpl_Lock(arena);
+
+    (void)"__osRealloc(%08x, %d)\n";
+
+    if (ptr == NULL) {
```
EllipticEllipsis, line 230: Interesting, I thought realloc usually required a prior malloc.

src/boot_O2/__osMalloc.c
```diff
@@ -179,45 +186,91 @@ void __osFree(Arena* arena, void* ptr) {
...
+        void* newPtr;
+        ArenaNode* node;
+
+        node = (uintptr_t)ptr - sizeof(ArenaNode);
+        newSize = ALIGN16(newSize);
+        if ((newSize != node->size) && (node->size < newSize)) {
+            ArenaNode* next = node->next;
+
+            diff = newSize - node->size;
+            if (((uintptr_t)next == ((uintptr_t)node + node->size + sizeof(ArenaNode))) && (next->isFree) &&
```
EllipticEllipsis, line 246: Since this is the fancy bit of realloc, could you annotate what this is doing a bit?

src/boot_O2/__osMalloc.c
```diff
@@ -245,4 +298,29 @@ void __osAnalyzeArena(Arena* arena, size_t* outMaxFree, size_t* outFree, size_t*
...
+
+    ArenaImpl_Lock(heap);
+
+    // "Checking the contents of the arena..."
+    (void)"アリーナの内容をチェックしています．．． (%08x)\n";
+
+    for (iter = heap->head; iter != NULL; iter = iter->next) {
+        if (iter->magic != NODE_MAGIC) {
+            // "Oops!!"
+            (void)"おおっと！！ (%08x %08x)\n";
```
EllipticEllipsis, line 313: Now where have I seen this style of error message before? 🤔

EllipticEllipsis (commented):
Code looks good, but I think this file could do with some more general documentation given its importance. I've pointed out a few places, but there's more I'm sure.

AngheloAlf:
This will be a draft until #396 is merged

src/boot_O2/__osMalloc.c
```diff
@@ -46,10 +46,10 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
 void __osMallocInit(Arena* arena, void* start, size_t size) {
-    bzero(arena, sizeof(*arena));
+    bzero(arena, sizeof(Arena));
     ArenaImpl_LockInit(arena);
     __osMallocAddBlock(arena, start, size);
-    arena->isInit = 1;
+    arena->isInit = true;
 }
 
 void __osMallocAddBlock(Arena* arena, void* start, s32 size) {
```
Kenix3, line 55:
```suggestion
void __osMallocAddBlock(Arena* arena, void* start, size_t size) {
```

Kenix3 (changes requested):

src/boot_O2/__osMalloc.c
AngheloAlf: Me too, but it looks like this is standard behaviour.

> If ptr is NULL, the behavior is the same as calling malloc(new_size). 

https://en.cppreference.com/w/c/memory/realloc

AngheloAlf:
Since the other PR was merged, I undrafted this PR and addressed the reviews.
I also added a `os_malloc.h` header to move those specific struct definitions out of `z64.h`

Kenix3 (approved):

hensldm (approved):
LGTM

Kenix3:
Merge pending @EllipticEllipsis or @Rozelette approval.

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,352 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
+
+/**
+ * @brief Reallocates the pointer \p ptr.
+ * The pointer \p ptr must be either a pointer previously allocated by `__osMalloc`, `__osMallocR` or `__osRealloc` and
+ * not freed yet, or a `NULL` pointer.
+ *
+ * If \p ptr is `NULL` a new pointer is allocated. See `__osMalloc` for more details.
+ * If \p newSize is 0, then the given pointer is freed and `NULL` is returned. See `__osFree` for more details.
+ * If \p newSize is bigger than the currently allocated allocated pointer, then the area of memory is expanded to a size
+ * big enough to fit the requested size.
```
EllipticEllipsis, line 318: This looks like an unordered list.

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,352 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
+ *
+ * The behaviour is undefined if the pointer \p ptr is not a memory region returned by one of the cited allocating
+ * functions. The behaviour is undefined if the pointer \p ptr doesn't correspond to the given \p arena. If the pointer
+ * is freed, then any access to the original freed pointer is undefined behaviour.
```
EllipticEllipsis, line 322: Might this also be clearer as an unordered list?

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,352 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
+ *
+ * @param[in, out] arena  The specific Arena to be used for the allocation.
+ * @param[in, out] ptr    The allocated memory block to deallocate.
+ * @param[in] newSize     The new requested size.
+ * @return void*          On success, the pointer to the reallocated area of memory. On failure, `NULL` is returned,
+ * also the original parameter \p ptr remains valid.
```
EllipticEllipsis, line 328:
```suggestion
 * and the original parameter \p ptr remains valid.
```

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,352 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
     ArenaImpl_Unlock(arena);
-}
 
-#pragma GLOBAL_ASM("asm/non_matchings/boot/__osMalloc/__osRealloc.s")
+    return ptr;
+}
 
-void __osAnalyzeArena(Arena* arena, size_t* outMaxFree, size_t* outFree, size_t* outAlloc) {
+/**
+ * @brief Gets the size of the bigger block which isn't allocated, the total free space and the total allocated space.
```
EllipticEllipsis, line 390:
```suggestion
 * @brief Gets the size of the biggest block which isn't allocated, the total free space and the total allocated space.
```

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,352 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
+ *
+ * @param[in, out] arena   The Arena which will be used to get the values from.
+ * @param[out] outMaxFree  The value of the bigger block which is not allocated.
```
EllipticEllipsis, line 393:
```suggestion
 * @param[out] outMaxFree  The value of the biggest block which is not allocated.
```

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,352 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
+ * @param[out] outFree     The accumulated free space.
+ * @param[out] outAlloc    The accumulated allocated space.
```
EllipticEllipsis, line 395: Maybe "total" instead of "accumulated"? Is there a subtlety I'm missing?

EllipticEllipsis (commented):
Wow, that's a lot of good documentation. I especially like the Doxygen bits, although I wonder if this is how to use it to its maximum potential. Just a few minor grammar/formatting things now.

src/boot_O2/__osMalloc.c
AngheloAlf: I blame the formater for messing up this

src/boot_O2/__osMalloc.c
```diff
@@ -124,13 +124,14 @@ u8 __osMallocIsInitalized(Arena* arena) {
...
  *
- * If there's not enough space in the given \p arena, this function will fail, returning `NULL`.
- * If \p size is zero, then an empty region of memory is returned. This region of memory must be freed to avoid memory
- * leaks.
+ *  - If there's not enough space in the given \p arena, this function will fail, returning `NULL`.
+ *  - If \p size is zero, then an empty region of memory is returned.
+ *  - This region of memory must be freed to avoid memory leaks.
  *
- * To avoid memory leaks, the returned pointer should be eventually deallocated using `__osFree` or `__osRealloc`.
+ * To avoid memory leaks, the returned pointer should be eventually deallocated using either `__osFree` or
```
EllipticEllipsis, line 133: Basically this is the same thing, right?

AngheloAlf: Yes...

EllipticEllipsis (approved):

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,353 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
 }
 
-void __osMallocInit(Arena* arena, void* start, size_t size) {
-    bzero(arena, sizeof(*arena));
+/**
+ * @brief Initializes \p arena using \p heap.
+ *
+ * @param arena  The Arena to initialize.
+ * @param heap   The heap to use.
+ * @param size   The size of the heap.
```
Thar0, line 57:
```suggestion
 * @brief Initializes \p arena to manage the memory region \p heap.
 *
 * @param arena  The Arena to initialize.
 * @param heap   The memory region to use as heap space.
 * @param size   The size of the heap.
```

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,353 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
-    arena->isInit = 1;
+
+    __osMallocAddBlock(arena, heap, size);
+    arena->isInit = true;
 }
 
-void __osMallocAddBlock(Arena* arena, void* start, s32 size) {
-    s32 diff;
-    s32 size2;
+void __osMallocAddBlock(Arena* arena, void* heap, size_t size) {
```
Thar0, line 68: This function initializes the first block, "add block" suggests to me that it's used whenever a new block is added, which is not the case. I'm not totally sure on a succinct name however.

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,353 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
+ * @param arena  The Arena to check.
+ * @return u8    `true` if the \p arena has been initialized. `false` otherwise.
+ */
 u8 __osMallocIsInitalized(Arena* arena) {
     return arena->isInit;
 }
 
+/**
+ * @brief Allocates at least \p size bytes of memory using the given \p arena.
+ * The block of memory will be allocated from the beginning of \p arena.
```
Thar0, line 127:
```suggestion
 * The block of memory will be allocated in the start of the first sufficiently large free block.
```
Maybe this is implied? I feel like it helps to be specific here

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,353 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
     }
+
     ArenaImpl_Unlock(arena);
 
     return alloc;
 }
 
+/**
+ * @brief Allocates at least \p size bytes of memory using the given \p arena.
+ * Unlike __osMalloc, the block of memory will be allocated at the end of the \p arena.
```
Thar0, line 192:
```suggestion
 * Unlike __osMalloc, the block of memory will be allocated from the end of the \p arena.
```
In a similar vein to above, "at" suggests to me at first glance that it would be placed at the very end of the arena, rather than the last sufficiently large free block.

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,353 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
     return alloc;
 }
 
+/**
+ * @brief Deallocates the pointer \p ptr previously allocated by `__osMalloc`, `__osMallocR` or `__osRealloc`.
+ * If the pointer \p ptr is `NULL` or it has been already been freed, then this function does nothing.
+ *
+ * - The behaviour is undefined if the pointer \p ptr is not a memory region returned by one of the cited allocating
+ * functions.
+ * - The behaviour is undefined if the pointer \p ptr doesn't correspond to the given \p arena.
```
Thar0, line 260: `the pointer \p ptr` feels unnecessarily verbose, does it really need to be stated explicitly that `ptr` is a pointer with each mention?

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,353 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
     ArenaImpl_Unlock(arena);
-}
 
-#pragma GLOBAL_ASM("asm/non_matchings/boot/__osMalloc/__osRealloc.s")
+    return ptr;
+}
 
-void __osAnalyzeArena(Arena* arena, size_t* outMaxFree, size_t* outFree, size_t* outAlloc) {
+/**
+ * @brief Gets the size of the biggest block which isn't allocated, the total free space and the total allocated space.
```
Thar0, line 391:
```suggestion
 * @brief Gets the size of the biggest block which isn't allocated, the total free space and the total allocated space.
```
```suggestion
 * @brief Gets the size of the largest free block, the total free space and the total allocated space.
```

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,353 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
+ *
+ * @param[in, out] arena   The Arena which will be used to get the values from.
+ * @param[out] outMaxFree  The value of the biggest block which is not allocated.
```
Thar0, line 394:
```suggestion
 * @param[out] outMaxFree  The size of the largest free block.
```

Thar0 (commented):
Some doc comments

src/boot_O2/__osMalloc.c
AngheloAlf: You are right.
But at the same time, this function _has_ the functionality to add a new block, so the name isn't technically wrong, it just happens that each an allocating function "adds" a block, it does it manually by spliting one of the already existing ones.

This function could theoretically be used to add blocks new blocks to completely unrelated heaps, it just happens that the game never does this.

What about `__osMallocAddHeap`?

AngheloAlf: Also, this name comes from the Animal Forest map.

I took a quick look at AF and I was able to find this function (ROM 0x008060, VRAM: 0x8002CC60 for anybody interested). It is exactly the same (and it matches without -g3. https://decomp.me/scratch/z9aiE).

There's a catch tho. I compared the other functions in AF, and looks like there are a few functions in the AF map missing from the AF rom, which shouldn't happen, right? Also, the VRAMs from the map doesn't match the VRAMs from the ROM itself (which may not be a huge deal? idk).

My conclusion is, what we have been calling `__osMalloc` all this time is not the real `__osMalloc` that the function map refers to. If this is true, I propose to rename all the functions in this file to something more ad hoc to our naming scheme (`SystemName_Func`). What do you think?

Thar0: Perhaps I'm wrong, but I'm pretty sure this cannot add a new block, only the first block, which is an important distinction. It looks to me that if this function were called more than once for a particular Arena, the arena would lose the ability to keep track of prior blocks in the other heap space. If that's correct, I suggest `__osMallocAddFirstBlock`. Otherwise, I think `__osMallocAddHeap` is good.

Regarding the AF map, the map comes from the port to Gamecube, so there are bound to be a number of differences when looking at the N64 version. It's likely AF on the Gamecube uses a modified `__osMalloc` but it should certainly have it's roots in the N64 version.
I see that `__osMallocAddBlock` is the original name as you pointed out, which is surely why it was named that way in the first place. I don't believe we should stick to it though if it's not representative of the purpose of the function, which I don't think it really is. Perhaps it would be nice however to comment it like
`// Original name: __osMallocAddBlock`

Thar0: Ah I see what you mean now, I failed to notice that it does in fact link with existing blocks via `ArenaImpl_GetLastBlock`. In that case, I think `__osMallocAddHeap` is suitable.

AngheloAlf: Sounds good. I'll rename it to `__osMallocAddHeap`

src/boot_O2/__osMalloc.c
```diff
@@ -45,181 +49,356 @@ ArenaNode* ArenaImpl_GetLastBlock(Arena* arena) {
...
-    arena->isInit = 1;
+
+    __osMallocAddHeap(arena, heap, size);
+    arena->isInit = true;
 }
 
-void __osMallocAddBlock(Arena* arena, void* start, s32 size) {
-    s32 diff;
-    s32 size2;
+// Original name: __osMallocAddHeap
```
hensldm, line 68:
```suggestion
// Original name: __osMallocAddBlock
```

Thar0 (approved):

PR #397, opened 2021-10-30
`motor.c` OK

AngheloAlf:
---

This file was already decomped, but it was not used in the `spec` (it was using rumblepak.c)

- I renamed the functions to match their original names.
- I made a `motor.h` header and moved the prototypes there.
- I deleted `viextendvstart.c` because it was unused


src/libultra/io/motor.c
```diff
@@ -1,12 +1,18 @@
+#include "ultra64/motor.h"
+
+#include "io/controller.h"
+
```
EllipticEllipsis, line 4: Why the blank lines?

EllipticEllipsis (commented):
Just a header thing. And do we really need both ultra64.h and global.h in here?

src/libultra/io/motor.c
AngheloAlf: ultra64.h and global.h were already here.
I think ultra64.h could be removed, but global.h is needed, because it needs the functions prototypes from another files. I'll try changing it to functions.h

Kenix3 (approved):

src/libultra/io/motor.c
```diff
@@ -112,7 +117,7 @@ s32 osProbeRumblePak(OSMesgQueue* ctrlrqueue, OSPfs* pfs, u32 channel) {
         return 0xB;
     }
     if ((pfs->status & PFS_MOTOR_INITIALIZED) == 0) {
```
EllipticEllipsis, line 119:
```suggestion
    if (!(pfs->status & PFS_MOTOR_INITIALIZED)) {
```

EllipticEllipsis (approved):

hensldm (approved):

PR #403, opened 2021-11-01
Updates readme to install dos2unix.

Kenix3:
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,18 @@ Under Debian / Ubuntu (which we recommend using), you can install them with the 
 
 ```bash
 sudo apt update
-sudo apt install make git build-essential binutils-mips-linux-gnu python3 python3-pip libpng-dev
+sudo apt install make git build-essential binutils-mips-linux-gnu python3 python3-pip libpng-dev dos2unix
+```
+
+Our tools have the following package requirements:
+
+* dos2unix
+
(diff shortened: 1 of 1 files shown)


AngheloAlf:
We have more tools that have their own requirements, like diff.py, graphovl.py, etc
Should we list all of them in the readme?
Maybe tools.md could be a better place to list that kind of stuff

EllipticEllipsis:
The ones for essentials like diff.py should probably be listed here. And IMO it should be specified in what circumstances one would want to install dos2unix: most of us never need to touch fixle, so there's no point in us installing dos2unix.

Kenix3:
> 
> 
> We have more tools that have their own requirements, like diff.py, graphovl.py, etc Should we list all of them in the readme? Maybe tools.md could be a better place to list that kind of stuff

Yes, we should list them all. I don't have a list available, but this was meant as a conversation starter so we can make a list.

Kenix3:
the asm differ has these requirements:
```
python3 -m pip install --user colorama watchdog python-Levenshtein cxxfilt
```

AngheloAlf:
I tried installing the `cxxfilt` package, but apt couldn't find it. Is it really needed?

Kenix3:
Going to close this PR for now. We'd probably want a list of all dependencies for all tools.

PR #404, opened 2021-11-01
EnvFlags OK

hensldm:
---

Copy paste of OOT's code_8006C3A0. I think I would prefer the system to be `EnvFlags` not `Flags` since those are the only flags this is touching, however decided to just use what OOT already had. Would love other's opinions though.


AngheloAlf (approved):
`EnvFlags` sounds good to me. Would be nice to hear more feedback about it tho

Kenix3 (commented):
I agree with the EnvFlags system rename. Also, we should rename the code file.

hensldm:
Okay, renamed the system to EnvFlags and the file to z_env_flags.c.

Kenix3 (approved):
I wonder how hard it would be to document a flag enum. Itd require reading the asm most likely.

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -92,7 +92,7 @@ void DoorAna_WaitClosed(DoorAna* this, GlobalContext* globalCtx) {
 
     if (grottoType == DOORANA_TYPE_UNK) {
         // in OOT decomp its marked as open with storms, but does not seem to open with storms in MM
-        if ((this->actor.xyzDistToPlayerSq < 40000.0f) && (func_800F13E8(globalCtx, 5))) {
+        if ((this->actor.xyzDistToPlayerSq < 40000.0f) && (EnvFlags_Get(globalCtx, 5))) {
```
isghj5, line 95: oooo interesting, so that door_ana variant code is dependent on these flags?

isghj5 (approved):

PR #407, opened 2021-11-02
Moves func_800CB000 to z_bg_item, decompiles func_800CB000, and documents z_bg_item and a few DynaPolyActor struct members.

Kenix3:
---


src/code/z_bg_item.c
```diff
@@ -1,99 +1,144 @@
...
+    dynaActor->flags = flags;
+    dynaActor->stateFlags = DYNAPOLY_STATE_NONE;
 }
 
-void BgCheck3_LoadMesh(GlobalContext* globalCtx, DynaPolyActor* actor, CollisionHeader* meshHeader) {
-    CollisionHeader* header;
+void DynaPoly_LoadMesh(GlobalContext* globalCtx, DynaPolyActor* dynaActor, CollisionHeader* meshHeader) {
+    CollisionHeader* header = NULL;
 
     header = NULL;
```
hensldm, line 21:
```suggestion
```
No longer needed as you set header to NULL above

src/code/z_bg_item.c
```diff
@@ -1,99 +1,144 @@
...
-void func_800CAE9C(DynaPolyActor* actor) {
-    actor->unk_158 |= 2;
+void DynaPoly_SetRidingMovingState(DynaPolyActor* dynaActor) {
+    dynaActor->stateFlags |= DYNAPOLY_STATE_RIDING_MOVING;
 }
 
-void func_800CAEB0(CollisionContext* colCtx, s32 index) {
-    DynaPolyActor* actor;
+void DynaPoly_SetRidingMovingStateByIndex(CollisionContext* colCtx, s32 index) {
+    DynaPolyActor* dynaActor;
```
hensldm, line 39:
```suggestion
    DynaPolyActor* dynaActor = BgCheck_GetActorOfMesh(colCtx, index);
```
Could set dynaActor right away.

src/code/z_bg_item.c
```diff
@@ -1,99 +1,144 @@
...
 
-    actor = BgCheck_GetActorOfMesh(colCtx, index);
-    if (actor != (DynaPolyActor*)0x0) {
-        func_800CAE9C(actor);
+    dynaActor = BgCheck_GetActorOfMesh(colCtx, index);
```
hensldm, line 41:
```suggestion
```
With the above suggestion this would no longer be needed.

src/code/z_bg_item.c
```diff
@@ -1,99 +1,144 @@
...
-void func_800CAEE0(DynaPolyActor* actor) {
-    actor->unk_158 |= 4;
+void DynaPoly_SetRidingRotatingState(DynaPolyActor* dynaActor) {
+    dynaActor->stateFlags |= DYNAPOLY_STATE_RIDING_ROTATING;
 }
 
-void func_800CAEF4(CollisionContext* colCtx, s32 index) {
-    DynaPolyActor* actor;
+void DynaPoly_SetRidingRotatingStateByIndex(CollisionContext* colCtx, s32 index) {
+    DynaPolyActor* dynaActor;
```
hensldm, line 52:
```suggestion
    DynaPolyActor* dynaActor = BgCheck_GetActorOfMesh(colCtx, index);
```

src/code/z_bg_item.c
```diff
@@ -1,99 +1,144 @@
...
 
-    actor = BgCheck_GetActorOfMesh(colCtx, index);
-    if (actor != (DynaPolyActor*)0x0) {
-        func_800CAEE0(actor);
+    dynaActor = BgCheck_GetActorOfMesh(colCtx, index);
```
hensldm, line 54:
```suggestion
```

hensldm (changes requested):
A few things I noticed.

src/code/z_bg_collect.c
```diff
@@ -90,12 +90,12 @@ u32 BgCheck2_UpdateActorAttachedToMesh(CollisionContext* colCtx, s32 index, Acto
         return 0;
     }
 
-    if ((meshActor->unk154 & 1) != 0) {
+    if ((meshActor->flags & 1) != 0) {
```
AngheloAlf, line 93: For bitwise checks like this you can drop the `!= 0` 
```suggestion
    if (meshActor->flags & 1) {
```

src/code/z_bg_collect.c
```diff
@@ -90,12 +90,12 @@ u32 BgCheck2_UpdateActorAttachedToMesh(CollisionContext* colCtx, s32 index, Acto
...
         BgCheck2_UpdateActorPosition(colCtx, index, actor);
         wasUpdated = 1;
     }
 
-    if ((meshActor->unk154 & 2) != 0) {
+    if ((meshActor->flags & 2) != 0) {
```
AngheloAlf, line 98: Same here
```suggestion
    if (meshActor->flags & 2) {
```

src/code/z_bg_item.c
```diff
@@ -1,99 +1,144 @@
...
 }
 
-s32 func_800CAF4C(DynaPolyActor* actor) {
-    if (actor->unk_158 & 1) {
+s32 DynaPoly_IsInRidingFallingState(DynaPolyActor* dynaActor) {
+    if (dynaActor->stateFlags & DYNAPOLY_STATE_RIDING_FALLING) {
         return 1;
     } else {
         return 0;
     }
```
AngheloAlf, line 73: You can replace those 1s and 0s to `true` and `false` now we know those functions are supposed to be boolean checks
```suggestion
        return true;
    } else {
        return false;
    }
```

AngheloAlf (commented):
Overall looks good. Just a few comments

src/code/z_bg_collect.c
Kenix3: Technically out of scope as the only changes to this file were automated. Updated anyway. z_bg_collect will need a documentation pass eventually.

src/code/z_bg_collect.c
Kenix3: Technically out of scope as the only changes to this file were automated. Updated anyway. z_bg_collect will need a documentation pass eventually.

AngheloAlf (approved):

hensldm (approved):
LGTM

src/code/z_bg_item.c
```diff
@@ -1,99 +1,138 @@
...
-    actor->unk154 = param_2;
-    actor->unk_158 = 0;
+#define DYNAPOLY_STATE_NONE 0
+#define DYNAPOLY_STATE_RIDING_FALLING 1 << 0
+#define DYNAPOLY_STATE_RIDING_MOVING 1 << 1
+#define DYNAPOLY_STATE_RIDING_ROTATING 1 << 2
+#define DYNAPOLY_STATE_SWITCH_PRESSED 1 << 3
+#define DYNAPOLY_STATE_HEAVY_SWITCH_PRESSED 1 << 4
+
+void DynaPoly_Init(DynaPolyActor* dynaActor, s32 flags) {
```
Thar0, line 10: In oot we know some things about the values `flags` takes, see the `DynaPolyMoveFlag` enum. Maybe you want to wait until more is confirmed though?

include/functions.h
```diff
@@ -1120,22 +1120,22 @@ void BgCheck2_UpdateActorPosition(CollisionContext* colCtx, s32 index, Actor* ac
...
-void func_800CAEF4(CollisionContext* colCtx, s32 index);
-void func_800CAF24(DynaPolyActor* actor);
-void func_800CAF38(DynaPolyActor* actor);
-s32 func_800CAF4C(DynaPolyActor* actor);
-s32 func_800CAF70(DynaPolyActor* actor);
-s32 func_800CAF94(DynaPolyActor* actor);
-s32 func_800CAFB8(DynaPolyActor* actor);
-s32 func_800CAFDC(DynaPolyActor* actor);
-// void func_800CB000(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE2 param_5);
+void DynaPoly_Init(DynaPolyActor* dynaActor, s32 flags);
```
Thar0, line 1123: I'm unsure about simply `DynaPoly` for these functions. In oot, we use `DynaPolyActor` for these as they are external functions for the `DynaPolyActor` structure, and we use `DynaPoly` in `z_bgcheck.c` for functions that manage the dynapoly system as a whole.

Thar0 (changes requested):
Some thoughts

src/code/z_bg_item.c
Kenix3: I'm confident in the use cases for these in both OOT and MM and find the names to be better. I didn't see really anything referring to enemies for example.

include/functions.h
Kenix3: These manage the actor, so I'd have to no issues with that. It is a bit too long, though. Wonder if it'd work to merge the two systems.

src/code/z_bg_item.c
mzxrules: I'm actually not sure what DynaPolyMoveFlag does in OoT. When decomping bgcheck in oot I simply used the existing enumeration since it makes modifications easier to do later on.

Thar0: I don't see where values for the `flags` field have been named, only `stateFlags`, could you point to where if you have? And if you have, existing usages of `DynaPoly_Init` would need their second arg updating to use the defines/enum.

Kenix3: Ahh I misunderstood, my bad. Yeah I didn't look into what the flags meant at all because it's not really used by the file. Figured it was out of scope for this documentation.

engineer124 (approved):
lgtm

PR #412, opened 2021-11-04
cartrominit OK

EllipticEllipsis:
---

This one actually had a flag change compared to OoT, which I noted.


src/libultra/io/cartrominit.c
```diff
@@ -1,3 +1,63 @@
...
+
+    D_800980D0 = 0;
+    CartRomHandle.type = DEVICE_TYPE_CART;
+    CartRomHandle.baseAddress = 0xB0000000;
+    CartRomHandle.domain = 0;
+    CartRomHandle.speed = 0;
+    bzero(&CartRomHandle.transferInfo, sizeof(__OSTranxInfo));
+
+    /* Uses `status & PI_STATUS_ERROR` in OoT */
+    while (status = HW_REG(PI_STATUS_REG, u32), status & (PI_STATUS_BUSY | PI_STATUS_IOBUSY)) {
```
Thar0, line 31: This is a mistake in oot. `PI_STATUS_ERROR` is currently wrongly defined in oot, and if it were defined correctly it would not be usable here, so I don't think it's worth mentioning.

src/libultra/io/cartrominit.c
```diff
@@ -1,3 +1,63 @@
...
+    HW_REG(PI_BSD_DOM1_LAT_REG, u32) = 0xFF;
+    HW_REG(PI_BSD_DOM1_PGS_REG, u32) = 0;
+    HW_REG(PI_BSD_DOM1_RLS_REG, u32) = 3;
+    HW_REG(PI_BSD_DOM1_PWD_REG, u32) = 0xFF;
+
+    a = HW_REG(CartRomHandle.baseAddress, u32);
+    CartRomHandle.latency = a & 0xFF;
+    CartRomHandle.pageSize = (a >> 0x10) & 0xF;
+    CartRomHandle.relDuration = (a >> 0x14) & 0xF;
+    CartRomHandle.pulse = (a >> 8) & 0xFF;
```
Thar0, line 49: This reads the initial domain configuration from the first four bytes of the rom header. I think `a` could be named `initialConfig` or the like to reflect this.

src/libultra/io/cartrominit.c
```diff
@@ -1,3 +1,63 @@
...
+OSPiHandle* osCartRomInit(void) {
+    register u32 a;
+    register s32 status;
+    register u32 prevInt;
+    register u32 lastLatency;
+    register u32 lastPageSize;
+    register u32 lastRelDuration;
+    register u32 lastPulse;
+
+    static u32 D_800980D0 = 1;
```
Thar0, line 14: This is a boolean, I suggest something like `sCartRomNeedsInit` since if it's false it skips the initialization and just returns the handle right away. I tried to find the original name from the 2.0I dev cd but I was unsuccessful.

Thar0 (changes requested):

src/libultra/io/cartrominit.c
EllipticEllipsis: That makes sense. Should it be s32 if it's a bool? Matches as either.

src/libultra/io/cartrominit.c
EllipticEllipsis: It's still different from OoT, though... is it a libultra version difference?

Thar0: `PI_STATUS_ERROR` in oot is currently 0x3, so 0x2 | 0x1, or `PI_STATUS_BUSY | PI_STATUS_IOBUSY`, which is the same as here. Unless I've misunderstood and you're referring to something else?

src/libultra/io/cartrominit.c
Thar0: iirc we like to keep bools s32 when we can? I'm not really sure.

src/libultra/io/cartrominit.c
EllipticEllipsis: I copypasted the OoT file, which used 0x4 here instead of 0x3, so seems to be a genuine code difference.

src/libultra/io/cartrominit.c
EllipticEllipsis: I'll make it s32 then.

src/libultra/io/cartrominit.c
Thar0: OoT definitely uses 0x3 here too is what I'm trying to say. It's just our definition if `PI_STATUS_ERROR` is not the same as in OoT since it's wrong there currently (but I fixed it in the handwritten asm PR)

EllipticEllipsis: Oh right, I get it now.

Kenix3 (approved):
Merge pending Anghelo approval and +1 contributor approval

AngheloAlf (approved):

Thar0 (changes requested):
I don't think my suggestions here were ever pushed?

EllipticEllipsis:
Odd, they show up now.

jpburnett (approved):

PR #414, opened 2021-11-04
Add ability to warnings script to just check building (not setup or disasm)

hensldm:
---

Since I feel most warnings will show up in the actual build (not setup or disasm), changed the warnings script to just check for build warnings by default. I have also added an option to still check warnings from start to finish for those cases where it still matters.

### DIFF
--- a/tools/warnings_count/check_new_warnings.sh
+++ b/tools/warnings_count/check_new_warnings.sh
@@ -109,9 +109,6 @@ else
     make_warnings uncompressed build
 fi
 
-echo "
-$(tput ${TPUTTERM} setaf 3)(lots of make output ${TPUTTERM} here...) 
-$RST"
 
 if [[ $full ]]; then
     $COMPARE_WARNINGS setup
--- a/tools/warnings_count/check_new_warnings.sh
+++ b/tools/warnings_count/check_new_warnings.sh
@@ -32,17 +32,29 @@ Check for new warnings created.
 
 Optional arguments:
     -h    Display this message and exit.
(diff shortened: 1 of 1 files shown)


AngheloAlf (approved):

isghj5 (approved):

tools/warnings_count/check_new_warnings.sh
```diff
@@ -90,16 +99,23 @@ make_warnings () {
...
+    make_warnings disasm disasm
+    make_warnings all build
+else
+    make clean
+    make_warnings uncompressed build
+fi
 
 echo "
 $(tput ${TPUTTERM} setaf 3)(lots of make output ${TPUTTERM} here...) 
 $RST"
```
EllipticEllipsis, line 114:
```suggestion
```
Huh, that should probably not be there any more... that was for testing.

EllipticEllipsis (commented):
I'm not sure how I feel about partial being the default... it's not that the disassembly stage itself would generate warnings, but that rearranging the data can cause them in the build part.

(Indeed, I'm not sure that the disassembly *can* generate any warnings...)

hensldm:
The main reason why I feel just build is fine as the default, is that this is not used in any automated process. Instead it is a tool for decompers to find any warnings that would break Jenkins builds, and again I feel just build warnings will be the most common. Thus it makes sense to me to have the one that takes considerably less time but still catches most of the warnings be the default. However still have the option for a full build as needed.

Kenix3 (approved):

PR #415, opened 2021-11-04
`z_skin` and `z_skin_awb` OK and documented a bit

AngheloAlf:
---

I decided to decomp both files together in the same PR because both depend on the same structs and are very interconnected.

The code wasn't hard to match, mostly OoT copy pastes with minor changes and scope memes.

I tried to document as much as I could, but most of the system relies on matrix transformations which I don't completely understand. This means like half of the functions remain unnamed.

I renamed a few structs used by this system, and added some typedefs to prevent ZAPD compatibility issues.

Since this is a very self-contained system, I decided to put everything skin-related in the new `z64skin.h` header.

To be honest, the documentation/naming is not that great. I'm open to new names for anything I named in this PR.


include/variables.h
```diff
@@ -1644,8 +1644,8 @@ extern UNK_PTR D_801D1540;
 // extern f32 sFactorialTbl[13];
 extern Vec3f D_801D15B0;
 extern Vec3s D_801D15BC;
-extern Mtx D_801D1DE0;
-extern MtxF D_801D1E20;
+extern Mtx gMtxClear;
+extern MtxF gMtxFClear;
```
engineer124, line 1648: fyi this will conflict with my name of `gIdentityMtx`/`gIdentityMtxF` in Bigslime, so we should probably resolve it at some point

include/z64skin.h
```diff
@@ -0,0 +1,104 @@
...
+    /* 0x004 */ Vtx* buf[2];
+} SkinAvb; // size = 0xC
+
+typedef struct {
+    /* 0x000 */ SkeletonHeader* skeletonHeader;
+    /* 0x004 */ MtxF mtx;
+    /* 0x044 */ s32 avbCount;
+    /* 0x048 */ SkinAvb* avbTbl;
+    /* 0x04C */ SkelAnime skelAnime;
+} PSkinAwb; // size = 0x90
```
engineer124, line 76: What does `avb` stand for? Also what does `awb` stand for? And what does the extra `P` stand for?

engineer124 (commented):
Great work! Just some very minor comments/questions

include/z64skin.h
AngheloAlf: Those come from OoT debug strings. I don't think we know their meaning
https://github.com/zeldaret/oot/blob/a497f33bda7686e3aa948898000ecd3a861c5704/src/code/z_skin_awb.c#L44

include/variables.h
AngheloAlf: From what I remember, we didn't get a satisfactory resolution about the naming of those variables.
I think we should keep the `gMtxClear` and `gMtxFClear` to keep the parity with OoT at least for now.
Tell me if you have any other idea

AngheloAlf: After the discussion in the Discord, I'll rename those two to `gIdentityMtx`/`gIdentityMtxF`

include/functions.h
```diff
@@ -2584,23 +2584,26 @@ void SkelAnime_UpdateTranslation(SkelAnime* skelAnime, Vec3f* pos, s16 angle);
...
+void func_801382C4(Actor* actor, GlobalContext* globalCtx, PSkinAwb* skin, SkinPostLimbDraw postLimbDraw, SkinOverrideLimbDraw arg4, s32 arg5, s32 arg6, s32 arg7);
+void func_80138300(PSkinAwb* skin, s32 joint, Vec3f* arg2, Vec3f* arg3);
+void func_8013835C(PSkinAwb* arg0, s32 arg1, s32 arg2, Vec3f* arg3);
+
+void func_80138410(PSkinAwb* skin);
+//void Skin_InitAnimatedLimb(GameState* gameState, PSkinAwb* skin, s32 limbIndex);
+void Skin_Init(GameState* gameState, PSkinAwb* skin, SkeletonHeader* skeletonHeader, AnimationHeader* animationHeader);
+void Skin_Free(GameState* gameState, PSkinAwb* skin);
+//s32 func_801387D4(PSkinAwb* skin, SkinLimb** skeleton, MtxF* mf, u8 parentIndex, u8 limbIndex);
+s32 func_801388E4(PSkinAwb* skin, MtxF* arg1, Actor* actor, s32 arg3);
```
hensldm, line 2605: Can you match up the argument names with the actual function definitions in the c files?

src/code/z_skin.c
```diff
@@ -1,23 +1,262 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin/func_80137B34.s")
+void func_80137970(MtxF* mtx, SkinVtx* skinVertices, Struct_800A598C* arg2, Vtx* verticesDst, Vec3f* arg4) {
+    SkinVtx* skinVtxEntry;
+    Vtx* vtx;
+    Vec3f sp5C;
+    Vec3f sp50;
+    Vec3f sp44;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin/func_80137EBC.s")
+    sp5C.x = mtx->mf[3][0];
```
hensldm, line 14: This is more of a general question since I haven't used `MtxF` all that often. I noticed it has a union of a 2D array, and a struct with the indices named into the 2D array. Which one should we be using, index as a 2D array (what you have here), or use the indices themselves (for example `mf[3][0]` is the same as `zx` I think)? Also do we know if which one we uses affects codegen?

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,203 @@
...
+            (((SkinLimb*)Lib_SegmentedToVirtual(skeleton[i]))->segment == NULL)) {
+            avbEntry->index = 0;
+
+            avbEntry->buf[0] = NULL;
+            avbEntry->buf[1] = NULL;
+        } else {
+            SkinAnimatedLimbData* animatedLimbData =
+                Lib_SegmentedToVirtual((((SkinLimb*)Lib_SegmentedToVirtual(skeleton[i]))->segment));
+
+            { s32 tmp; }
```
hensldm, line 72:
```suggestion
        } else {
            s32 tmp;
            
            SkinAnimatedLimbData* animatedLimbData =
                Lib_SegmentedToVirtual((((SkinLimb*)Lib_SegmentedToVirtual(skeleton[i]))->segment));
```
Does this still match?

hensldm (commented):
A few cosmetic touch ups.

src/code/z_skin_awb.c
AngheloAlf: No, the temp in the middle is required :c

include/z64skin.h
```diff
@@ -0,0 +1,85 @@
...
+} Struct_800A598C; // size = 0x10
+
+typedef struct {
+    /* 0x00 */ u16 vtxCount;
+    /* 0x02 */ u16 unk_2; // count of unk_4
+    /* 0x04 */ Struct_800A598C* unk_4;
+    /* 0x08 */ Gfx* dlist;
+} SkinAnimatedLimbData; // size = 0xC
+
+// ZAPD compatibility typedefs
```
hensldm, line 46: Could you add a TODO to remove these when ZAPD is compatible?

hensldm (approved):
One more request, but good enough to approve.

include/z64skin.h
AngheloAlf:
```suggestion
// ZAPD compatibility typedefs
// TODO: Remove when ZAPD adds support for them
```

src/code/z_skin.c
```diff
@@ -1,23 +1,272 @@
...
+    if (postDraw != NULL) {
+        postDraw(actor, globalCtx, skin);
+    }
+
+close_disps:;
+    CLOSE_DISPS(gfxCtx);
+}
+
+// allows specifying postDraw and setTranslation
+void func_80138228(Actor* actor, GlobalContext* globalCtx, Skin* skin, SkinPostDraw postDraw, s32 setTranslation) {
```
engineer124, line 235:
```suggestion
/**
 * Allows specifying postDraw and setTranslation
 */
void func_80138228(Actor* actor, GlobalContext* globalCtx, Skin* skin, SkinPostDraw postDraw, s32 setTranslation) {
```
I think this is more the style for explanations of what a function does? (applies to the functions below as well)

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,203 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_awb/func_80138424.s")
+/**
+ * Initialises the Vtx buffers used for limb at index `limbIndex`
+ */
+void Skin_InitAnimatedLimb(GameState* gameState, Skin* skin, s32 limbIndex) {
+    s32 i;
+    SkinLimb** skeleton = Lib_SegmentedToVirtual(skin->skeletonHeader->segment);
+    SkinAnimatedLimbData* animatedLimbData =
+        Lib_SegmentedToVirtual(((SkinLimb*)Lib_SegmentedToVirtual(skeleton[limbIndex]))->segment);
+    SkinLimbModif* temp_v0 = Lib_SegmentedToVirtual(animatedLimbData->limbModifications);
```
engineer124, line 18:
```suggestion
    SkinLimbModif* modifications = Lib_SegmentedToVirtual(animatedLimbData->limbModifications);
```
Also, `limbModifications` is a case of using the full word `modifications` instead of `motif`, so it's currently inconsistent

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,203 @@
...
+
+/**
+ * Initializes a skin skeleton to looping animation, dynamically allocating the frame tables,
+ * and dynamically allocating and initializing the Vtx and SkinLimbVtx buffers for its animated limbs
+ */
+void Skin_Init(GameState* gameState, Skin* skin, SkeletonHeader* skeletonHeader, AnimationHeader* animationHeader) {
+    s32 limbCount;
+    s32 i;
+    SkinLimb** skeleton;
+    SkeletonHeader* virtSkelHeader = Lib_SegmentedToVirtual(skeletonHeader);
```
engineer124, line 50:
```suggestion
    SkeletonHeader* skelHeader = Lib_SegmentedToVirtual(skeletonHeader);
```
Is the `virt` part necessary? `Lib_SegmentedToVirtual` is very common but this is the first time I've seen that prefix used. But I guess it's to differentiate between this temp and the argument `skeletonHeader`?

engineer124 (commented):
Great work with the new docs! Just some small nitpicky stuff

include/z64skin.h
```diff
@@ -0,0 +1,83 @@
...
+    /* 0x044 */ s32 limbCount;
+    /* 0x048 */ SkinLimbVtx* vtxTable; // double buffered list of vertices for each limb
+    /* 0x04C */ SkelAnime skelAnime;
+} Skin; // size = 0x90
+
+typedef void (*SkinPostDraw)(struct Actor*, struct GlobalContext*, Skin*);
+typedef s32 (*SkinOverrideLimbDraw)(struct Actor*, struct GlobalContext*, s32, Skin*);
+
+#define SKIN_DRAW_FLAG_CUSTOM_TRANSFORMS (1 << 0)
+#define SKIN_DRAW_FLAG_CUSTOM_MATRIX     (1 << 1)
```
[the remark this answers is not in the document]
fig02: they get used as args in the draw calls for drawFlags

src/code/z_skin.c
fig02: at least in oot these are meant to be temporary and aid with actually naming the functions

include/z64skin.h
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Describes a position displacement and a scale to be applied to a limb at index `limbIndex`
+ */
+typedef struct {
+    /* 0x00 */ u8 limbIndex;
+    /* 0x02 */ s16 x;
+    /* 0x04 */ s16 y;
+    /* 0x06 */ s16 z;
+    /* 0x08 */ u8 scale;
+} SkinTransformation; // size = 0xA
```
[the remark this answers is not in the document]
fig02: they should match what would be returned if you were to call sizeof on the struct. And that result depends on struct alignment. In this case I believe that would be 0xA cause the largest size in the struct is 2 byte aligned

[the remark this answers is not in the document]
EllipticEllipsis: It should be the actual padded size, which in this case is a multiple of 2 because it contains s16s

include/z64skin.h
```diff
@@ -0,0 +1,83 @@
...
+    /* 0x02 */ u16 transformCount;
+    /* 0x04 */ u16 transformIndex; // index of limbTransformations
+    /* 0x08 */ SkinVertex* skinVertices;
+    /* 0x0C */ SkinTransformation* limbTransformations;
+} SkinLimbModif; // size = 0x10
+
+typedef struct {
+    /* 0x00 */ u16 totalVtxCount; // total vertex count for all modif entries
+    /* 0x02 */ u16 limbModifCount;
+    /* 0x04 */ SkinLimbModif* limbModifications;
```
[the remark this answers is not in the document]
fig02: in comments I agree it makes sense, but the full word in the type name would make things pretty long. I think having the context from the struct member helps you work out the type name if you dont get it right away

src/code/z_skin_awb.c
EllipticEllipsis: It's pretty unusual to do this to a skeleton at all.

fig02:
some of these cleanups would be good to have over in oot too, so whenever this gets merged we can coordinate that

src/code/z_skin_awb.c
AngheloAlf: yeah, the point is to differentiate it from `skeletonHeader`. I'm not sure if I like `skelHeader` since it is too similar to `skeletonHeader`. Maybe rename the parameter one (`skeletonHeader`) to `skeletonHeaderSeg`?

fig02: I think the arg name is more deserving of "skeletonHeader" since its what users will be interacting with when using the function. virtSkelHeader seems fine to me honestly even though unusual

fig02: skelHeader is also fine if its close to the arg name imo, since they are literally the same thing. doesnt make the code any less confusing

src/code/z_skin.c
```diff
@@ -1,23 +1,272 @@
...
+        transformCount = modif->transformCount;
+
+        if (transformCount == 1) {
+            spAC.x = limbTransformations[0].x;
+            spAC.y = limbTransformations[0].y;
+            spAC.z = limbTransformations[0].z;
+
+            SkinMatrix_Vec3fMtxFMultXYZ(&gSkinLimbMatrices[limbTransformations[0].limbIndex], &spAC, &spDC);
+        } else if (arg3) {
+            transformationEntry = &limbTransformations[modif->transformIndex];
```
EllipticEllipsis, line 94: So it seems like
- if there's one, it will do that
- if not, and arg3, it will do the one corresponding to the index
- otherwise, it'll do all of them, in order?

src/code/z_skin.c
```diff
@@ -1,23 +1,272 @@
...
+
+    gSPSegment(POLY_OPA_DISP++, 0x08, vtxEntry->buf[vtxEntry->index]);
+
+    vtxEntry->index = (vtxEntry->index == 0);
+
+    CLOSE_DISPS(gfxCtx);
+}
+
+/**
+ * Draw a limb of type SKIN_LIMB_TYPE_ANIMATED, of the skeleton `skin` at index `limbIndex`
```
EllipticEllipsis, line 133: So the distinction is that the modifications can be used to change the vertices/shape on this one, whereas the NORMAL ones just draw in the usual way?

src/code/z_skin.c
```diff
@@ -1,23 +1,272 @@
...
+// allows specifying OverrideLimbDraw, postDraw and setTranslation
+void func_80138258(Actor* actor, GlobalContext* globalCtx, Skin* skin, SkinPostDraw postDraw,
+                   SkinOverrideLimbDraw overrideLimbDraw, s32 setTranslation) {
+    Skin_DrawImpl(actor, globalCtx, skin, postDraw, overrideLimbDraw, setTranslation, false, 0);
+}
+
+// allows specifying OverrideLimbDraw, postDraw, setTranslation, and arg6
+void func_8013828C(Actor* actor, GlobalContext* globalCtx, Skin* skin, SkinPostDraw postDraw,
+                   SkinOverrideLimbDraw overrideLimbDraw, s32 setTranslation, s32 arg6) {
+    Skin_DrawImpl(actor, globalCtx, skin, postDraw, overrideLimbDraw, setTranslation, arg6, 0);
```
EllipticEllipsis, line 248: I think we can probably name `arg6` and its counterpart `arg3` as something like `useIndexedTransformOnly` (or something less bad).

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,204 @@
 #include "global.h"
+#include "z64skin.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_awb/func_80138410.s")
+void func_80138410(Skin* skin) {
```
EllipticEllipsis, line 4: ClearSomething?

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,204 @@
...
+                Vtx* vtx = &vertices[skinVtxEntry->index];
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_awb/func_801388E4.s")
+                skinVtxEntry++;
+
+                vtx->n.flag = 0;
+                vtx->n.tc[0] = skinVtxEntry[-1].s;
+                vtx->n.tc[1] = skinVtxEntry[-1].t;
+                vtx->n.a = skinVtxEntry[-1].alpha;
+            }
```
EllipticEllipsis, line 38: What an odd way to do this.

src/code/z_skin_awb.c
EllipticEllipsis: Having looked at the rest of the files involved here, I think it'd be better to have a consistent convention for all the virtualisation that occurs.

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,204 @@
...
+    } else {
+        mtx = &limbMatrices[(s32)parentIndex];
+    }
+
+    SkinMatrix_MtxFMtxFMult(mtx, &limbMatrices[limbIndex], &sp28);
+    SkinMatrix_MtxFCopy(&sp28, &limbMatrices[limbIndex]);
+
+    if (limb->child != LIMB_DONE) {
+        ret = func_801387D4(skin, skeleton, limbMatrices, limbIndex, limb->child);
+        if (ret) { // func_801387D4 only returns false
```
EllipticEllipsis, line 130:
```suggestion
        if (ret) { // func_801387D4 can only return false
```
and elsewhere.

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,204 @@
...
+    s32 pad;
+    SkinLimb* limb = Lib_SegmentedToVirtual(skeleton[limbIndex]);
+    MtxF* mtx;
+    s32 ret;
+    MtxF sp28;
+
+    if (parentIndex == LIMB_DONE) {
+        SkinMatrix_GetClear(&mtx);
+    } else {
+        mtx = &limbMatrices[(s32)parentIndex];
```
EllipticEllipsis, line 122: Cast required? Or argument is definitely a u8?

EllipticEllipsis (commented):
Good job on this, I have a few mostly documentation-related comments. As you say, it's going to need another pass from someone who understands how the matrix stuff fits together with the actual animation format.

include/z64skin.h
```diff
@@ -0,0 +1,83 @@
+#ifndef Z64_SKIN_H
+#define Z64_SKIN_H
+
+#include "z64animation.h"
+
+/**
+ * Holds a compact version of a vertex used in the Skin system
```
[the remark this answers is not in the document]
fig02: it doesnt contain all elements of a Vtx, only some of them

include/z64skin.h
```diff
@@ -0,0 +1,83 @@
...
+
+#define SKIN_LIMB_TYPE_ANIMATED 4
+#define SKIN_LIMB_TYPE_NORMAL 11
+
+typedef struct {
+    /* 0x00 */ Vec3s jointPos; // Root is position in model space, children are relative to parent
+    /* 0x06 */ u8 child;
+    /* 0x07 */ u8 sibling;
+    /* 0x08 */ s32 segmentType; // Type of data contained in segment
+    /* 0x0C */ void* segment; // Gfx* if segmentType is SKIN_LIMB_TYPE_NORMAL, SkinAnimatedLimbData if segmentType is SKIN_LIMB_TYPE_NORMAL, NULL otherwise
```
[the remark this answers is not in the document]
fig02: yes, that is why its typed as void. What it actually points to is context dependent

src/code/z_skin.c
```diff
@@ -1,23 +1,272 @@
 #include "global.h"
+#include "z64skin.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin/func_80137970.s")
+MtxF gSkinLimbMatrices[60];
```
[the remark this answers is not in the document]
fig02: it doesnt really get referenced anywhere in the system, no. Its an arbitrary number that tries to encapsulate how big a skin skeleton can be. If a skin skeleton has over 60 limbs that would be an issue. But I dont think a define would help with anything there

src/code/z_skin.c
fig02: thats the idea, but my main issue with naming this still is the why-- I dont understand what `modif->transformIndex` even is and why you would want to single one specific one out. Which is why this variable is still unnamed over in oot

src/code/z_skin.c
fig02: similar to my previous comment, I dont think its appropriate to name this until we understand the purpose of `modif->transformIndex` (or unk_04 in oot) and why its indexing something specific to begin with. Having that understanding will help make a name that sounds less mechanical/verbose

include/z64skin.h
```diff
@@ -0,0 +1,83 @@
...
+    /* 0x6 */ s16 z;
+    /* 0x8 */ u8 scale;
+} SkinTransformation; // size = 0xA
+
+typedef struct {
+    /* 0x00 */ u16 vtxCount; // number of vertices in this modif entry
+    /* 0x02 */ u16 transformCount;
+    /* 0x04 */ u16 transformIndex; // index of limbTransformations
+    /* 0x08 */ SkinVertex* skinVertices;
+    /* 0x0C */ SkinTransformation* limbTransformations;
```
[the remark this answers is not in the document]
fig02: as mentioned in other comments, I dont think we really understand what `transformationIndex` is, and is why its still unnamed in oot.

src/code/z_skin_awb.c
fig02: fwiw I removed these comments from the oot version-- I feel like the fact that it only returns false in practice doesnt matter for the intent of what the check is doing here. It doesnt feel important to me to single out

src/code/z_skin_awb.c
fig02: this function doesnt exist in oot-- what context does it get used in?

src/code/z_skin.c
```diff
@@ -1,23 +1,272 @@
...
+            }
+        }
+
+        Skin_UpdateVertices(&gSkinLimbMatrices[limbTransformations[modif->transformIndex].limbIndex], skinVertices,
+                            modif, vtxBuf, &spDC);
+    }
+
+    gSPSegment(POLY_OPA_DISP++, 0x08, vtxEntry->buf[vtxEntry->index]);
+
+    vtxEntry->index = (vtxEntry->index == 0);
```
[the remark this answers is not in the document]
fig02: this doesnt seem too crazy, its just alternating the index every frame. Id prob have written a ternary personally to make it more clear, but its not the worst lol

src/code/z_skin_awb.c
fig02: im guessing the way its done in oot didnt match? that is way more sensible

include/z64skin.h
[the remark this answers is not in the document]
EllipticEllipsis: I mean the `Gfx* if segmentType is SKIN_LIMB_TYPE_NORMAL, SkinAnimatedLimbData if segmentType is SKIN_LIMB_TYPE_NORMAL` part.

src/code/z_skin.c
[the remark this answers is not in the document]
EllipticEllipsis: `^=`

include/z64skin.h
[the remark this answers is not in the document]
fig02: yes, that is what im referring to as well. `segment` is either of type Gfx* or SkinAnimatedLimbData* depending on what type of limb it is.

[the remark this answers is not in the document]
fig02: oh, I see the problem. The second one should say ANIMATED, not NORMAL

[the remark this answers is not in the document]
AngheloAlf: Yeah, I forgot to fix that here :sweat_smile:

src/code/z_skin.c
AngheloAlf: Yes. 
`NORMAL` holds a simple dlist, while `ANIMATED` holds a pointer to all the weird structures that allow the shape/vertices to change in runtime

src/code/z_skin_awb.c
AngheloAlf: It is only used in `EnHorse_Init`
https://github.com/zeldaret/mm/pull/413/files#diff-1dac57939127037837e51b46d879fe422dfed7fb223d76d7b76fb4c6c4c12eb5R716

I'm not really sure why they decided to add it, it feels a bit unnecessary.

src/code/z_skin_awb.c
AngheloAlf: Yeah, the way it is done in OoT doesn't match for some reason. I hope this is not yet another `-g3` meme

src/code/z_skin_awb.c
AngheloAlf: The cast is required

src/code/z_skin_awb.c
```diff
@@ -1,13 +1,204 @@
...
+                vtx->n.a = skinVtxEntry[-1].alpha;
+            }
+        }
+    }
+}
+
+/**
+ * Initialises a skin skeleton to looping animation, dynamically allocating the frame tables,
+ * and dynamically allocating and initializing the Vtx and SkinLimbVtx buffers for its animated limbs
+ */
```
engineer124, line 46:
```suggestion
/**
 * Initializes a skin skeleton to looping animation, dynamically allocating the frame tables,
 * and dynamically allocating and initialising the Vtx and SkinLimbVtx buffers for its animated limbs
 */
```
`Initialise` with an `s` and a `z` in the same sentence. Nice. I don't have a preference, as long as its consistent

engineer124 (approved):

include/z64skin.h
```diff
@@ -29,16 +33,16 @@ typedef struct {
 } SkinTransformation; // size = 0xA
 
 typedef struct {
-    /* 0x00 */ u16 vtxCount; // number of vertices in this modif entry
-    /* 0x02 */ u16 transformCount;
-    /* 0x04 */ u16 transformIndex; // index of limbTransformations
+    /* 0x00 */ u16 vtxCount; // number of skinVertices in this modif entry
+    /* 0x02 */ u16 transformCount; // count of limbTransformations
```
EllipticEllipsis, line 37: I wonder if this is a bit redundant?

Makefile
```diff
@@ -291,7 +291,7 @@ build/asm/%.o: asm/%.s
 build/assets/%.o: assets/%.c
 	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $<
 	$(OBJCOPY) -O binary $@ $@.bin
-	$(RM_MDEBUG)
+	@$(RM_MDEBUG)
```
EllipticEllipsis, line 294: I think this should stay in the output: unlike the objcopy and odjdump, it directly affects something that goes into the compiled ELF.

include/z64skin.h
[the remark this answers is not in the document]
EllipticEllipsis: I still think it would be better to be more specific about which parts.

src/code/z_skin_awb.c
EllipticEllipsis: I think something like init or setup or clear, then.

AngheloAlf: We already have `Skin_Init`.
I guess I'll use `Skin_Setup`

EllipticEllipsis (approved):

Kenix3 (approved):

PR #416, opened 2021-11-05
aisetfreq OK, some libultra improvements

EllipticEllipsis:
---

- Matched `aisetfreq.c`
- Add `HW_REG` macros to all the `(u32*)0xA...` magic numbers (apart from the CIC ones, which I think we're stuck with, at least for now)
- Add `rdp.h` with some `HW_REG`-style defines
- Eliminate a few other magic numbers I spotted

Some of these files appear to have unimported data, e.g. `vi.c`. I didn't import it to attempt to limit the scope of this PR, that already outgrew what I expected it to do.


include/ultra64/rdp.h
```diff
@@ -0,0 +1,52 @@
+#ifndef ULTRA64_RDP_H
+#define ULTRA64_RDP_H
+
+/* DP Command Registers */
+
+#define DPC_REG_BASE 0xA4100000
```
Thar0, line 6: These should all use their physical addresses, i.e.
```suggestion
#define DPC_REG_BASE 0x04100000
```
`HW_REG` will OR these with 0xA0000000 to form the uncached kseg0 address

src/libultra/io/pirawdma.c
```diff
@@ -1,23 +1,23 @@
 #include "global.h"
 
 s32 __osPiRawStartDma(s32 direction, u32 devAddr, void* dramAddr, size_t size) {
-    register int stat;
+    register int status;
```
Thar0, line 4: Personally I think the code looks cleaner like this but I don't think we have a set standard:
```suggestion
    register int status = HW_REG(PI_STATUS_REG, u32);
```

src/libultra/io/si.c
```diff
@@ -1,11 +1,11 @@
 #include "global.h"
 
-int __osSiDeviceBusy() {
-    register u32 status;
-    status = *(u32*)0xA4800018;
+s32 __osSiDeviceBusy() {
+    register u32 status = HW_REG(SI_STATUS_REG, u32);
+
     if (status & 3) {
```
Thar0, line 6:
```suggestion
    if (status & (SI_STATUS_DMA_BUSY | SI_STATUS_IO_READ_BUSY)) {
```

src/libultra/io/sirawdma.c
```diff
@@ -1,24 +1,24 @@
...
 s32 __osSiRawStartDma(s32 direction, void* dramAddr) {
-    if ((*(u32*)0xA4800018 & 0x3) != 0) {
+    if (HW_REG(SI_STATUS_REG, u32) & (SI_STATUS_DMA_BUSY | SI_STATUS_IO_READ_BUSY)) {
         return -1;
     }
 
-    if (direction == 1) {
-        osWritebackDCache(dramAddr, 64);
+    if (direction == OS_WRITE) {
+        osWritebackDCache(dramAddr, 0x40);
```
Thar0, line 9: `0x40` is the PIF RAM size. I don't think we have a define for this right now but maybe worthwhile.

Thar0 (changes requested):

src/libultra/io/pirawdma.c
```diff
@@ -1,23 +1,23 @@
 #include "global.h"
 
 s32 __osPiRawStartDma(s32 direction, u32 devAddr, void* dramAddr, size_t size) {
```
Kenix3, line 3:
```suggestion
s32 __osPiRawStartDma(s32 direction, uintptr_t devAddr, void* dramAddr, size_t size) {
```
I know you didn't touch this line, but I noticed it needs fixed.

Kenix3 (changes requested):

src/libultra/io/pirawdma.c
EllipticEllipsis: What makes the return of this a size_t? I know @Thar0 mentioned there's a lot of stuff in libultra that returns 0 for success and -1 for error.

EllipticEllipsis: Indeed, this can't be a size_t because of that, surely? If anything it would be an ssize_t

Thar0: It definitely isn't a size_t or similar. In other works where a function returns 0 on success or some negative number as an error indicator, the return type is simply int. Since we don't use int anywhere I think s32 is most appropriate.

src/libultra/io/sirawdma.c
EllipticEllipsis: Where should such a define go?

Thar0: hardware.h next to PIF_RAM_START seems like a good place

EllipticEllipsis: So PIF_RAM_SIZE?

Thar0: Yeah sounds good

hensldm (approved):

Kenix3 (approved):

PR #426, opened 2021-11-06
guS2DInitBg OK

EllipticEllipsis:
---

Also added the required gbi header. For now I put it in z64.h, but it's likely going to be used so little that it's probably better to be included filewise when we sort the headers out.


Thar0 (approved):

kyleburnette (approved):

AngheloAlf (approved):
I think I would prefer if we included this header filewise

Kenix3 (approved):

PR #427, opened 2021-11-06
viswapcontext OK

EllipticEllipsis:
---

Not sure if we want to keep the __ on structs like `__OSViContext`, but that can be decided later.

### DIFF
--- a/src/libultra/io/viswapcontext.c
+++ b/src/libultra/io/viswapcontext.c
@@ -1,3 +1,66 @@
 #include "global.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/boot/viswapcontext/__osViSwapContext.s")
+void __osViSwapContext(void) {
+    register OSViMode* viMode;
+    register __OSViContext* viNext;
+    u32 origin;
+    u32 hStart;
+    u32 vstart;
+    u32 sp34;
+    u32 field;
+    register u32 s2;
+
+    field = 0;
+    viNext = __osViNext;
+    viMode = viNext->modep;
+    field = HW_REG(VI_V_CURRENT_LINE_REG, u32) & 1;
+    s2 = osVirtualToPhysical(viNext->buffer);
+    origin = (viMode->fldRegs[field].origin) + s2;
+    if (viNext->state & 2) {
+        viNext->x.scale |= viMode->comRegs.xScale & ~0xFFF;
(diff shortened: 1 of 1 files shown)


Thar0 (approved):

kyleburnette (approved):

AngheloAlf (approved):

src/libultra/io/viswapcontext.c
```diff
@@ -1,3 +1,66 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/boot/viswapcontext/__osViSwapContext.s")
+void __osViSwapContext(void) {
+    register OSViMode* viMode;
+    register __OSViContext* viNext;
+    u32 origin;
+    u32 hStart;
+    u32 vstart;
+    u32 sp34;
+    u32 field;
+    register u32 s2;
```
Kenix3, line 11:
```suggestion
    uintptr_t origin;
    uintptr_t hStart;
    uintptr_t vstart;
    u32 sp34;
    u32 field;
    register uintptr_t s2;
```

Kenix3 (changes requested):

src/libultra/io/viswapcontext.c
EllipticEllipsis: Are the top 3 actually pointers? They certainly don't look like it in the struct that this refers to: for example, from 
```c
OSViMode osViModePalLan1 = {
[...]
        0x800300,     // hStart
[...]
          0x280,    // origin
[...]
          0x5F0239, // vStart
      } },
};
```
they just seem to be magic numbers.

Kenix3: They look like pointers just in the context of this file. Maybe someone with more knowledge of libultra can chime in, like @Thar0

EllipticEllipsis: We discussed this a while ago in Discord. More research showed that `vstart` and `hstart` were not pointers, but that the use of `origin` was inconsistent, but since it's set in a hardware register, I think there ended up being agreement it should stay as `u32`? `s2` technically should be, but it's also part of a hardware register, so I'm not convinced either way, really. Again, @Thar0 may have a better idea of how we should handle this to be consistent.

Kenix3 (approved):

PR #428, opened 2021-11-06
contchannelreset OK

EllipticEllipsis:
---

Thanks to @Thar0 for this one. It has one of those weird loops that looks innocuous when matching, but generates a lot of extra code for different cases,
```c
    if (arg1 > 0) {
        temp_a2 = arg1 & 3;
        if (temp_a2 != 0) {
            do {
                temp_v1 = phi_v1 + 1;
                phi_v0->ramarray = 0;
                temp_v0_2 = phi_v0 + 1;
                phi_v0 = temp_v0_2;
                phi_v1 = temp_v1;
                phi_v0_2 = temp_v0_2;
                phi_v1_2 = temp_v1;
                phi_v0_3 = temp_v0_2;
            } while (temp_a2 != temp_v1);
            if (temp_v1 != arg1) {
                goto loop_4;
            }
        } else {
            do {
loop_4:
                temp_v1_2 = phi_v1_2 + 4;
                phi_v0_2->unk1 = 0;
                phi_v0_2->unk2 = 0;
                phi_v0_2->unk3 = 0;
                temp_v0_3 = phi_v0_2 + 4;
                temp_v0_3->unk-4 = 0;
                phi_v0_2 = temp_v0_3;
                phi_v1_2 = temp_v1_2;
                phi_v0_3 = temp_v0_3;
            } while (temp_v1_2 != arg1);
        }
    }
```
kind of thing.

Another interesting fact about this function: it's only used in osVoiceInit, so it's not in most other games.


Thar0 (approved):

kyleburnette (approved):

AngheloAlf (approved):

src/libultra/io/contchannelreset.c
```diff
@@ -1,3 +1,27 @@
 #include "global.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/boot/contchannelreset/__osContChannelReset.s")
+s32 __osContChannelReset(OSMesgQueue* mq, s32 count) {
```
Kenix3, line 3:
```suggestion
s32 __osContChannelReset(OSMesgQueue* mq, size_t count) {
```
seems like this is a size of memory to blank out

Kenix3 (changes requested):

src/libultra/io/contchannelreset.c
EllipticEllipsis: Turns out this also has to be `s32` from its use elsewhere. It's also actually the `channel`, rather than just a bare memory number, so `s32` is a better type after all: it's all over the voice stuff.

Kenix3 (approved):

PR #432, opened 2021-11-06
sys_ucode OK

EllipticEllipsis:
---

Ended up changing more files than I expected for this very small one. I think all the appropriate types are correct now, at least.


AngheloAlf (approved):

include/ultra64/sptask.h
```diff
@@ -80,25 +81,25 @@ typedef struct {
     /* 0x04 */ u32 flags;
 
     /* 0x08 */ u64* ucodeBoot;
-    /* 0x0C */ u32 ucodeBootSize;
+    /* 0x0C */ size_t ucodeBootSize;
```
Thar0, line 84: I am slightly concerned about these not being explicitly `u32`, since microcode running on the RSP will access these fields with `lw`, so if a 64-bit build were ever made and this field becomes 64-bit, it would break unexpectedly. I'm not sure if this is a real concern or not but I thought I'd mention it anyway.

Kenix3: Good question. I'd say size_t is the correct type, especially since we typedef'd it to 32bits. I would say the hardware register size shouldn't factor in to the types we use beyond matching.

Kenix3 (approved):

src/code/sys_ucode.c
```diff
@@ -1,9 +1,20 @@
 #include "global.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_ucode/SysUcode_GetUCodeBoot.s")
+u64* D_801D1E60 = D_801E2160;
+u64* D_801D1E64 = D_801E3790;
```
Kenix3, line 4: Are these the start and end offsets of the ucode? If so we can document this data.

PR #434, opened 2021-11-07
Fault and Fault_Drawer (3 MATCHING and 1 NON_MATCHING)

hensldm:
---

Matched the last two functions in fault_drawer as well as one in fault (however that one requires in function bss, so going to keep it NON_MATCHING for now). Also got the last function in fault very close, but still having regalloc issues (scratch: https://decomp.me/scratch/x8y6X).


src/boot_O2_g3/fault_drawer.c
```diff
@@ -142,12 +172,57 @@ void FaultDrawer_FillScreen() {
...
+                    FaultDrawer_FillScreen();
+                }
+                sFaultDrawContext->cursorY = sFaultDrawContext->yStart;
+            }
+        }
+    }
+
+    osWritebackDCacheAll();
+
+    return arg;
```
isghj5, line 219: ... It returns its own un-modified parameter?

src/boot_O2_g3/fault_drawer.c
```diff
@@ -16,7 +18,7 @@ FaultDrawer sFaultDrawerDefault = {
     GPACK_RGBA5551(0, 0, 0, 0),         // backColor
     22,                                 // cursorX
     16,                                 // cursorY
-    (u32*)sFaultDrawerFont,             // font
+    sFaultDrawerFont,                   // font
     8,                                  // charW
     8,                                  // charH
     0,                                  // charWPad
```
EllipticEllipsis, line 24: Could you label the colours too? Most likely they're the same as the [ASCII terminal ones](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors), i.e.
Black
Red
Green
Yellow
Blue
Magenta
Cyan
White
and two greys.

EllipticEllipsis (commented):
Nice work, just some minor suggestions.

src/boot_O2_g3/fault.c
```diff
@@ -751,8 +750,10 @@ void Fault_ProcessClients(void) {
 }
 
 #ifdef NON_MATCHING
-// regalloc and ordering differences around the two bool variables (faultCustomOptions and faultCopyToLog)
+// needs in-function static bss
 void Fault_SetOptionsFromController3(void) {
+    static u32 faultCustomOptions;
+
     Input* input3;
```
Thar0, line 757: We typically don't separate `static` and local variables
```suggestion
    static u32 faultCustomOptions;
    Input* input3;
```

src/boot_O2_g3/fault.c
```diff
@@ -762,21 +763,23 @@ void Fault_SetOptionsFromController3(void) {
...
 
     if (faultCustomOptions) {
         graphPC = sGraphThread.context.pc;
         graphRA = sGraphThread.context.ra;
         graphSP = sGraphThread.context.sp;
-        if (CHECK_BTN_ALL(input3->press.button, BTN_R)) {
+        if (CHECK_BTN_ALL(input3->cur.button, BTN_R)) {
+            static u32 faultCopyToLog;
+
             faultCopyToLog = !faultCopyToLog;
```
Thar0, line 776: imo it looks cleaner this way
```suggestion
            static u32 faultCopyToLog = !faultCopyToLog;
```

src/boot_O2_g3/fault_drawer.c
```diff
@@ -74,7 +77,36 @@ void FaultDrawer_DrawRecImpl(s32 xStart, s32 yStart, s32 xEnd, s32 yEnd, u16 col
...
+    const u32* dataPtr = &sFaultDrawContext->font[(((c / 8) * 16) + ((c & 4) >> 2))];
+    u16* fb = sFaultDrawContext->fb + (sFaultDrawContext->w * cursorY) + cursorX;
+
+    if ((sFaultDrawContext->xStart <= cursorX) &&
+        ((sFaultDrawContext->charW + cursorX - 1) <= sFaultDrawContext->xEnd) &&
+        (sFaultDrawContext->yStart <= cursorY) &&
+        ((sFaultDrawContext->charH + cursorY - 1) <= sFaultDrawContext->yEnd)) {
+        for (y = 0; y < sFaultDrawContext->charH; y++) {
+            u32 mask = 0x10000000 << shift;
+            data = *dataPtr;
```
Thar0, line 96:
```suggestion
            u32 mask = 0x10000000 << shift;

            data = *dataPtr;
```

Thar0 (changes requested):
Little things

src/boot_O2_g3/fault.c
hensldm: Apparently this is not allowed.
image

Thar0: Oh right, I misread what was going on here, sorry. Now that I read it properly I think it's best left as-is.

AngheloAlf (approved):

hensldm:
Was able to match `Fault_FindNextStackCall`, so the only thing left in fault is the NON_MATCHING which matches but needs in function static bss.

engineer124 (approved):
lgtm

EllipticEllipsis (approved):
I'd have probably written the greys as light and dark, but I'm not particularly bothered since it's obvious from looking at the colour that they're different.

hensldm:
Oh that is a good idea to distinguish the grays. Just added that real quick.

Kenix3 (approved):

PR #437, opened 2021-11-07
sys_initial_check OK, documented, error message files debinarised

EllipticEllipsis:
---

The functions in this file are run very early in Main, and check for the Expansion Pak and thhe correct TV type. If neither are found to be correct, error messages are written directly to a display buffer at the end of normal (i.e. non-expanded) RDRAM.

Most of the functions were straightforward to decompile, with just a couple of stupid ones. The interesting part is the rather strange way in which this file deals with RAM: it has to use direct accesses to fixed addresses since no heap is available, but it also uses symbols based on the error messages files' sizes, which required declaring them with particular segments, and adding these to `segment_symbols.h`. Since I was changing the spec for these anyway, I made XMLs for them and put them in the `misc` folder, since they don't really fit into a category with anything else.

... and I ended up touching a lot of other stuff while documenting, to fix various pointers and defines that I noticed weren't being used correctly.


src/code/sys_initial_check.c
```diff
@@ -1,17 +1,114 @@
...
+ */
+void Check_ExpansionPak(void) {
+    // Expansion pak installed
+    if (osMemSize >= 0x800000) {
+        return;
+    }
+
+    Check_DrawExpansionPakErrorMessage();
+    osDestroyThread(NULL);
+    while (true) {}
```
AngheloAlf, line 96: For infinite loops like this one you can do this. 
```suggestion
    while (true);
```
I'm not saying it is better or not, up to you if you like it

AngheloAlf (approved):

src/code/sys_initial_check.c
EllipticEllipsis: Wow C allows a lot of terrible syntax.

Kenix3 (approved):
Merge pending +1 contributor approval

hensldm (approved):

src/code/sys_initial_check.c
```diff
@@ -1,17 +1,114 @@
...
+ */
 #include "global.h"
+#include "misc/locerrmsg/locerrmsg.h"
+#include "misc/memerrmsg/memerrmsg.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_initial_check/Check_WriteRGBA16Pixel.s")
+// Address with enough room after to load either of the error message image files before the fault screen buffer at the
+// end of RDRAM
+#define CHECK_ERRMSG_STATIC_SEGMENT \
+    (u8*)(FAULT_FB_ADDRESS - sizeof(gExpansionPakNotInstalledErrorTex) - sizeof(gSeeInstructionBookletErrorTex))
```
Thar0, line 16: I can't really think of a way to make this friendlier towards modifications of these files..
My first thought was `MAX(sizeof(...) + ..., sizeof(...))` for both files, to ensure the larger of the two is always chosen, but even that isn't great since if either file were modified more `sizeof`s would need adding to the expression anyway. My next thought would be to have ZAPD output a `SIZEOF_<FILE>` define, consisting of the sum of the `sizeof`s of the file's contents, though I'm unsure how much use such a thing would get.
Anyway, something to think about. I don't think waiting on a solution to this is important right now.

src/code/sys_initial_check.c
```diff
@@ -1,17 +1,114 @@
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_initial_check/Check_WriteI4Pixel.s")
+void Check_WriteRGBA16Pixel(u16* buffer, u32 x, u32 y, u32 value) {
+    if (value & 1) {
```
Thar0, line 19: This checks if the alpha value of the pixel is not zero. I sort of feel like it might be good to wrap this in an explicit `!= 0`, since it's not exactly a "flag" necessarily. But of course, it doesn't really matter (and runs the risk of being haphazardly reverted in cleanup passes) so I don't feel too strongly on it.
```suggestion
    if ((value & 1) != 0) {
```

Thar0 (approved):
Overall looks good. I have two comments but as you'll see they don't stop me from approving it as-is.

src/code/sys_initial_check.c
EllipticEllipsis: I think we're just going to have to use a define for the filesize and ifdef for PAL or something: they should have used the segment size, but didn't.

src/code/sys_initial_check.c
EllipticEllipsis: I'll add a note. Or we could have something like a define for `RGBA16_OPAQUE`?

EllipticEllipsis:
Okay, I addressed the two comments... let's see if it will build.

PR #446, opened 2021-11-10
m2ctx: Optionally return Player* for Player accesses

Kelebek1:
---

This PR adds the ability to alter m2ctx output, right now just to edit actorLists[2] to return a Player* under a `--fix` or `-f` flag.

### DIFF
--- a/tools/m2ctx.py
+++ b/tools/m2ctx.py
@@ -7,6 +7,46 @@ script_dir = os.path.dirname(os.path.realpath(__file__))
 root_dir = script_dir + "/../"
 src_dir = root_dir + "src/"
 
+# Read through the processes context and replace whatever
+def replace_things(output):
+    output = output.splitlines()
+
+    i = 0
+    while i < len(output):
+        line = output[i]
+
+        ############### actorLists[2].first -> Player* ###############
+        if "typedef struct ActorListEntry " in line:
+            actorListText = ""
+            i += 1
+            while not output[i].startswith("}"):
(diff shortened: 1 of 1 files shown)


Kenix3 (approved):
Merge pending Anghelo approval +1 contributor approval.

tools/m2ctx.py
```diff
@@ -7,6 +7,46 @@
 root_dir = script_dir + "/../"
 src_dir = root_dir + "src/"
 
+# Read through the processes context and replace whatever
+def replace_things(output):
```
AngheloAlf, line 11: I prefer if the name of this function would be a bit less generic. Something like `custom_replacements` for example?

tools/m2ctx.py
```diff
@@ -42,6 +82,8 @@ def main():
                                      description="Creates a ctx.c file for mips2c. "
                                      "Output will be saved as oot/ctx.c")
     parser.add_argument('filepath', help="path of c file to be processed")
+    parser.add_argument("--fix", "-f", dest="fix", action="store_true", default=False,
```
AngheloAlf, line 85: `--fix` sounds a bit too generic to me. Also, some people may argue this is _breaking_ stuff instead of fixing it. I suggest something like `--custom` instead
```suggestion
    parser.add_argument("--custom", "-c", dest="fix", action="store_true", default=False, 
```

AngheloAlf (approved):

jpburnett (approved):

PR #449, opened 2021-11-11
Add multiprocessing to disassembly script

thaumavorio:
---
This PR adds multiprocessing to the disassembly script. On my machine, this brings the runtime down from 2 minutes on a single core to 45 seconds when run with 8 processes.

I made my best attempt to avoid changing the code's structure. Therefore there is a bit of a hacky solution for dealing with the multiple sets of symbols (see `update_symbols_from_dict`). If anyone has a better idea that does not disrupt the current state of things too much, please let me know.

I have tested this with both a single process and multiple. In each case, it outputs `asm` and `data` directories whose contents are equivalent to those output by the current version of the script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,7 +230,7 @@ setup:
 ## Assembly generation
 disasm:
 	$(RM) -rf asm data
-	python3 tools/disasm/disasm.py
+	python3 tools/disasm/disasm.py -j $(N_THREADS)
 
 diff-init: uncompressed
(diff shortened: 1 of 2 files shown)


Kenix3 (approved):
Great work on this! It'll be nice to have disasm sped up.

louist103 (approved):

AngheloAlf (approved):

PR #455, opened 2021-11-12
pause (Frame Advance) OK

hensldm:
---

Matched z_pause.c, which turned out to be FrameAdvance init and update, so renamed file to z_frame_advance.c


src/code/z_frame_advance.c
```diff
@@ -0,0 +1,25 @@
+#include "global.h"
+
+void FrameAdvance_Init(FrameAdvanceContext* frameAdvCtx) {
+    frameAdvCtx->timer = 0;
+    frameAdvCtx->enabled = false;
+}
+
+/**
+ * Frame advance allows you to advance through the game one frame at a time on command.
+ * To advance a frame, hold Z and press R on the spceified controller.
```
Kelebek1, line 10: specified

AngheloAlf (changes requested):
Could you rename this file to `z_frame_advance` in `tools/disasm/files.txt` too?

engineer124 (approved):
lgtm

AngheloAlf (approved):

src/code/z_frame_advance.c
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Frame advance allows you to advance through the game one frame at a time on command.
+ * To advance a frame, hold Z and press R on the specified controller.
+ * Holding Z and R will advance a frame every half second.
+ *
+ * This function returns true when frame advance is not active (game will run normally)
+ *
+ * Note: There is no way to enable it
+ */
```
EllipticEllipsis, line 16: Since this comment really applies to the whole system, it seems better to put it at the top of the file in a file header.

EllipticEllipsis (changes requested):
Are these functions completely unused? I'd explain what is required to be able to use this somewhere: you can check how OoT has it set up, for example.

On a more general note, I don't like renaming this file from what is far more likely to be a correct name: that these are in alphabetical order is very useful information for understanding how `code` is laid out and how we can expect it to be laid out in other versions, and renaming it loses that information. I seriously doubt that anyone would expect more than a very rough idea of what a file does from its name anyway: that's the whole point of having top-of-file descriptions.

(Note that my view on this is specific to files and to doing it *now*: I am all in favour of renaming functions etc, but I do not think this is a good time to start renaming files that have a file name that reflects their position like this.)

Kenix3 (commented):
I agree with Elliptic on this case. It's too early to rename files and knowing the order can be helpful. I think we should set the name back to z_pause.

hensldm:
Ok, reverted the file name change for now (though added a note on the sheet that this is frame advance). Also moved the comment section as a header, and added a note of what code could possibly be added to enable it.

src/code/z_pause.c
```diff
@@ -1,19 +1,29 @@
+/**
```
EllipticEllipsis, line 1: I think I'd like to start this off with `File:` and `Description:` lines like in overlay files... we do need to discuss how we'd like to format them (should the initial description be a synopsis and then a fuller description given under a different header, for example?

hensldm: Added `File:` and `Description:` lines to the header.

EllipticEllipsis (approved):

Kenix3 (approved):

PR #465, opened 2021-11-16
Refactor disasm for speedups

Kelebek1:
---

This was aimed at speeding up disasm.py further, some key features:

- Introduced file pruning by reading the spec, and skipping already decompiled files. Currently this prunes about 34% of total files. This will help speed up disasm in the future. Full disassembly can still be done by passing `--full`.
- Added finding rodata symbols to multiprocessing
- Refactored .text disassembly to be mostly done inside the symbol lookup pass, so only needing to loop over the raw data and interact with mips_isa once. A second pass does purely symbol lookups later. Also optimised the .text instruction parsing a bit to save the instructions and not decode them many times.
- Flattened out boot and code segments, so a single thread isn't stuck with the entirety of those segments anymore. Sections were split too, though this may have been overkill. I did this before the flattening though.

Times for me (post Anghelo/zbank's PRs):
Current master `-j12`:
```
real    0m33.031s
user    2m47.124s
sys     0m2.226s
```

New `-j12` (~34% pruned currently in master, the new default):
```
real    0m19.630s
user    1m20.494s
sys     0m4.945s
```
Only disassembling boot (~92% pruned):
```
real    0m1.714s
user    0m2.143s
sys     0m0.671s
```


tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
-dwords = set()         # doublewords
-multiply_referenced_rodata = set() # rodata with more than 1 reference outside of the same function cannot be migrated
-
-files = set()          # vram start of file
-
-vrom_variables = list() # (name,addr)
+jtbls = set()  # vram of jump tables
+jtbl_labels = (
+    set()
+)  # vram of branch labels via jtbls, higher output priority than regular branch labels
```
AngheloAlf, line 187: I think this is a bit ugly. What about this instead?
```suggestion
# vram of branch labels via jtbls, higher output priority than regular branch labels
jtbl_labels = set()
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
-    *MIPS_FP_LOAD_STORE_INSNS
+    *MIPS_FP_LOAD_STORE_INSNS,
 ]
 
-functions = set()      # vram of functions
-data_labels = set()    # vram of data labels, TODO this + functions can merge into something more general eventually
+functions = set()  # vram of functions
+data_labels = (
+    set()
+)  # vram of data labels, TODO this + functions can merge into something more general eventually
```
AngheloAlf, line 182:
```suggestion
# vram of data labels, TODO this + functions can merge into something more general eventually
data_labels = set()
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
+floats = set()  # vram of floats
+doubles = set()  # vram of doubles
+prospective_strings = set()  # vram of possible strings
+strings = set()  # vram of confirmed strings
+symbols = (
+    {}
+)  # lui/addiu pairs : {addr of %hi : full symbol} AND {addr of %lo : full symbol} , twice the space complexity but much less time complexity
+constants = (
+    {}
+)  # lui/ori pairs : {addr of %hi : full constant} AND {addr of %lo : full constant} , twice the space complexity but much less time complexity
```
AngheloAlf, line 197:
```suggestion
# lui/addiu pairs : {addr of %hi : full symbol} AND {addr of %lo : full symbol} , twice the space complexity but much less time complexity
symbols = {}
# lui/ori pairs : {addr of %hi : full constant} AND {addr of %lo : full constant} , twice the space complexity but much less time complexity
constants = {}
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
+dwords = set()  # doublewords
+multiply_referenced_rodata = (
+    set()
+)  # rodata with more than 1 reference outside of the same function cannot be migrated
```
AngheloAlf, line 201:
```suggestion
# rodata with more than 1 reference outside of the same function cannot be migrated
multiply_referenced_rodata = set()
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
+
+files_text = {}
+
 
 def proper_name(symbol, in_data=False, is_symbol=True):
     # hacks
-    if symbol == 0x809C46F0: # ovl_En_Encount4 fake symbol at the very end of the data section
+    if (
+        symbol == 0x809C46F0
+    ):  # ovl_En_Encount4 fake symbol at the very end of the data section
```
AngheloAlf, line 219:
```suggestion
    # ovl_En_Encount4 fake symbol at the very end of the data section
    if symbol == 0x809C46F0:
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
-    elif symbol == 0x801EF66D: # z_message_nes constant-folding stray fairy array
+    elif symbol == 0x801EF66D:  # z_message_nes constant-folding stray fairy array
         return variables_ast[0x801EF670][0] + f" - 0x{0x801EF670 - 0x801EF66D:X}"
-    elif symbol == 0x80A09740: # boss_07 symbol with large addend folded into %lo
+    elif symbol == 0x80A09740:  # boss_07 symbol with large addend folded into %lo
         return variables_ast[0x80A09A60][0] + f" - 0x{0x80A09A60 - 0x80A09740:X}"
-    elif symbol == 0x80B80248: # bg_ikana_mirror symbol with large addend folded into %lo
+    elif (
+        symbol == 0x80B80248
+    ):  # bg_ikana_mirror symbol with large addend folded into %lo
```
AngheloAlf, line 227:
```suggestion
    # bg_ikana_mirror symbol with large addend folded into %lo
    elif symbol == 0x80B80248:
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
         return variables_ast[0x80B801A8][0] + f" + 0x{0x80B80248 - 0x80B801A8:X}"
-    elif symbol == 0x8084D2FC: # player symbol with very large addend folded into %lo, since we don't know the real symbol just use the first data symbol for now
+    elif (
+        symbol == 0x8084D2FC
+    ):  # player symbol with very large addend folded into %lo, since we don't know the real symbol just use the first data symbol for now
```
AngheloAlf, line 231:
```suggestion
    # player symbol with very large addend folded into %lo, since we don't know the real symbol just use the first data symbol for now
    elif symbol == 0x8084D2FC:
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
         return variables_ast[0x8085B9F0][0] + f" - 0x{0x8085B9F0 - 0x8084D2FC:X}"
-    elif symbol == 0x001ABAB0 or symbol == 0x001E3BB0: # OS_K0_TO_PHYSICAL on rspS2DEX text and data symbols
+    elif (
+        symbol == 0x001ABAB0 or symbol == 0x001E3BB0
+    ):  # OS_K0_TO_PHYSICAL on rspS2DEX text and data symbols
```
AngheloAlf, line 235:
```suggestion
    # OS_K0_TO_PHYSICAL on rspS2DEX text and data symbols
    elif symbol == 0x001ABAB0 or symbol == 0x001E3BB0:  
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
         return variables_ast[symbol + 0x80000000][0] + " - 0x80000000"
-    elif symbol == 0x00AC0480: # do_action_static + 0x480, this is the only rom segment that has a constant offset folded into it so just hack it
+    elif (
+        symbol == 0x00AC0480
+    ):  # do_action_static + 0x480, this is the only rom segment that has a constant offset folded into it so just hack it
```
AngheloAlf, line 239:
```suggestion
    # do_action_static + 0x480, this is the only rom segment that has a constant offset folded into it so just hack it
    elif symbol == 0x00AC0480:
```

tools/disasm/disasm.py
```diff
@@ -2,126 +2,260 @@
...
-        if symbol in [addr for name,addr in vrom_variables if "SegmentRomStart" in name]:
-            return [name for name,addr in vrom_variables if "SegmentRomStart" in name and addr == symbol][0]
+        if symbol in [
+            addr for name, addr in vrom_variables if "SegmentRomStart" in name
+        ]:
+            return [
+                name
+                for name, addr in vrom_variables
+                if "SegmentRomStart" in name and addr == symbol
+            ][0]
```
AngheloAlf, line 256: This bit looks a bit wasteful, since it is doing the same operation twice
I think using a temp would benefit performance a bit
Something like:
```suggestion
        filteredVromSymbols = { addr: name for name, addr in vrom_variables if "SegmentRomStart" in name}
        if symbol in filteredVromSymbols:
            return filteredVromSymbols[symbol]
```

tools/disasm/disasm.py
```diff
@@ -263,7 +440,11 @@ def str_round(string, precision):
         return int_part + "." + frac_part
 
     def to_binary(flt_str):
-        return as_dword(struct.pack(">d", float(flt_str))) if is_double else as_word(struct.pack(">f", float(flt_str)))
+        return (
+            as_dword(struct.pack(">d", float(flt_str)))
+            if is_double
+            else as_word(struct.pack(">f", float(flt_str)))
+        )
```
AngheloAlf, line 447: I think redability would be benefited if one liners like this wouldn't be used in functions like this one

```suggestion
        if is_double:
            return as_dword(struct.pack(">d", float(flt_str)))
        return as_word(struct.pack(">f", float(flt_str)))
```

AngheloAlf (commented):
Mostly formatting suggestions

tools/disasm/disasm.py
Kelebek1: It's not quite the same thing twice. First it checks if if the name is in the list, then it returns if name is in and the addr == symbol. I dunno what this code does really so I don't know if that will break things.

Makefile
```diff
@@ -10,13 +10,19 @@ NON_MATCHING ?= 0
 ORIG_COMPILER ?= 0
 # Keep .mdebug section in build
 KEEP_MDEBUG ?= 0
+# Disassembles the full ROM instead of a subset
```
louist103, line 13:
```suggestion
# Disassembles the full ROM instead of a only required files
```
Maybe something like that?

tools/disasm/variables.txt
```diff
@@ -10997,9 +10997,6 @@
     0x80A0C55C:("D_80A0C55C","f32","",0x4),
     0x80A0C560:("D_80A0C560","f32","",0x4),
     0x80A0C7C0:("En_Boj_05_InitVars","UNK_TYPE1","",0x1),
-    0x80A0C7E0:("enBoj05OverlayInfo","OverlayRelocationSection","",0x14),
-    0x80A0C7F4:("enBoj05OverlayRelocations","u32","[4]",0x10),
-    0x80A0C80C:("enBoj05OverlayInfoOffset","u32","",0x4),
```
isghj5, line 11002: These modifications were intentional right? This feels like a different PR

Kelebek1: Yeah this is intentional, these are wrong symbols that shouldn't be in there. Anghelo figured it out and said I should remove them.

tools/disasm/disasm.py
```diff
@@ -2,126 +2,250 @@
...
+        name, _, type, _, file_list = f
+
+        # Find the beginseg for this file
+        i = 0
+        while i < len(spec):
+            if spec[i].startswith("beginseg") and f'"{name}"' in spec[i + 1]:
+                break
+            i += 1
+
+        # For every fil;e within this segment, look through the seg for lines with this file's name
```
hensldm, line 48:
```suggestion
        # For every file within this segment, look through the seg for lines with this file's name
```

hensldm (changes requested):
Small typo

Makefile
hensldm:
```suggestion
# Disassembles the full ROM instead of just non-decomped files
```
perhaps? Would prefer something describing what the actual subset is.

tools/disasm/disasm.py
```diff
@@ -2,126 +2,250 @@
...
+
+        seg_start = i
+        new_files = {}
+        included = False
+        saved_off = 0
+        for offset, file in file_list.items():
+            if file == "[PADDING]":
+                continue
+
+            # why are these not named?
```
hensldm: Is this an actual question you have?

hensldm (changes requested):
Could you also fix the conflicts? I would love to check the PR out to fully test, but am afraid testing would not be beneficial if things change too much due to conflict fixing.

AngheloAlf:
One thing I noted this changed is the comment with the offset seems to be reset to zero in data and rodata. 
Could you fix that back to what it was?

image

hensldm (approved):
Been doing some testing and works fine for me.

mzxrules:
> One thing I noted this changed is the comment with the offset seems to be reset to zero in data and rodata. Could you fix that back to what it was?

Not sure why you'd want the offset relative to the linked data section over the offset relative to the data section for the obj file here.

Kelebek1:
> > One thing I noted this changed is the comment with the offset seems to be reset to zero in data and rodata. Could you fix that back to what it was?
> 
> Not sure why you'd want the offset relative to the linked data section over the offset relative to the data section for the obj file here.

They weren't object-relative before, but section-relative, so the first sym in any .data will be 0000, and the first sym in the same object's rodata was also 0000, which in retrospect was probably pretty useless. 

I've changed it now to actually be object-relative like you said since that seemed pretty useful.

AngheloAlf (approved):

Kenix3 (approved):

louist103 (approved):

PR #477, opened 2021-11-25
Improve disassembler's RAM usage

AngheloAlf:
---

I noted the disassembler was munching _a lot_ of RAM during the disassembly step. After investigating it, I noted the problem came from a huge `vars_cache` variable which had all the possibles addends a variable could have used in the assembly.

From what I can imagine, this is just a solution to the "Python doesn't have sorted maps" problem (other languages like C++ does). With this in mind I decided to use the `sortedcontainers` module for this, which provides a sorted dictionary which I think it is the right tool for the job for this case.

I tried to do this change in the less invasive way possible.

As a comparison:

| Tool used                                    | Current master                            | This PR                                | %         |
|----------------------------------------------|-------------------------------------------|----------------------------------------|-----------|
| `valgrind --tool=massif --pages-as-heap=yes` | `1440.386 MiB` (or `1,510,354,944 bytes`) | `349.843 MiB` (or `366,837,760 bytes`) | `−75.71%` | 

This PR doesn't improve much the time it takes to run the disassembler (in my tests it reduced around 5 seconds which could be the time taken to create all the possible addends or just statistical noise), but if somebody was hitting their swap when running the disassembler, then it could be an improvement to them.

Lastly, as mentioned this adds a new dependency (`sortedcontainers`). I added it to the requirements.txt file, so installing it should be as easy as running `python3 -m pip install -r requirements.txt` again


EllipticEllipsis (approved):
I tested this when my RAM was quite full, and I noticed a pretty significant difference:
master:
```
real    15m35.123s
user    22m29.899s
sys     3m10.885s
```

this branch:
```
real    5m59.244s
user    12m56.114s
sys     1m6.079s
```
Admittedly this is only one data point apiece, but I did try to make the conditions as similar as possible (same programs open, looking at same files, dropping caches before both runs).

Even without such a noticable improvement, it seems like an inherently good idea to reduce RAM usage anyway. (I also tested with a decent amount of free RAM and it was pretty much identical, so it's at worst the same speed) And for the way that this seems to use and store the symbols, to the best of my knowledge this is the appropriate data structure to use.

Overall, this would be of significant benefit to me and anyone else who is RAM-limited, so I'd really like if we could get this in as soon as possible.

Also, this is failing because of the missing `sortedcontainers` dependency. Not sure what the best way to get that into Jenkins is.

Thar0 (approved):
LGTM, this is a much appreciated improvement

AngheloAlf:
Superseded by #478

PR #481, opened 2021-11-28
z_vr_box_draw OK

shawlucas:
---

mostly the same as oot. main draw function is shorter


include/functions.h
```diff
@@ -2759,9 +2759,9 @@ void func_80143148(SkyboxContext* skyboxCtx, s32 arg1);
 void func_801431E8(GameState* gameState, SkyboxContext* skyboxCtx, s16 skyType);    
 void func_80143324(GlobalContext* globalCtx, SkyboxContext* skyboxCtx, s16 skyType);
 void func_801434E4(GameState* gameState, SkyboxContext* skyboxCtx, s16 skyType);    
-// void func_801435A0(void);
-// void func_80143624(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE1 param_5, UNK_TYPE1 param_6, UNK_TYPE1 param_7);
-// void func_80143668(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE4 param_5, UNK_TYPE4 param_6, UNK_TYPE4 param_7);
+// void SkyboxDraw_UpdateMatrix(void);
+// void SkyboxDraw_SetColors(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE1 param_5, UNK_TYPE1 param_6, UNK_TYPE1 param_7);
+// void SkyboxDraw_Draw(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE4 param_5, UNK_TYPE4 param_6, UNK_TYPE4 param_7);
```
EllipticEllipsis, line 2764: If you know what these functions are now, and they are used in other files (as I think is the case here?), you should fill in the correct prototypes for them. Otherwise, remove them.

spec
```diff
@@ -555,7 +555,6 @@ beginseg
     include "build/src/code/z_vr_box.o"
     include "build/src/code/z_vr_box_draw.o"
     include "build/data/code/z_vr_box_draw.data.o"
```
EllipticEllipsis, line 557: What's the story on this data?

src/code/z_vr_box_draw.c
```diff
@@ -1,9 +1,67 @@
...
+    gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, skyboxCtx->primR, skyboxCtx->primG, skyboxCtx->primB, blend);
+    gDPSetEnvColor(POLY_OPA_DISP++, skyboxCtx->envR, skyboxCtx->envG, skyboxCtx->envB, 0);
+
+    gSPDisplayList(POLY_OPA_DISP++, &skyboxCtx->dListBuf[0]);
+    gSPDisplayList(POLY_OPA_DISP++, &skyboxCtx->dListBuf[300]);
+    gSPDisplayList(POLY_OPA_DISP++, &skyboxCtx->dListBuf[600]);
+    gSPDisplayList(POLY_OPA_DISP++, &skyboxCtx->dListBuf[900]);
+    gSPDisplayList(POLY_OPA_DISP++, &skyboxCtx->dListBuf[1200]);
+
+    if (skyboxId == 5) {
```
EllipticEllipsis, line 57: Is this just an index into an array or something? A particular face on the cube? A particular skybox? It may need a define or enum.

EllipticEllipsis (commented):
Nice job, a couple of documentation comments, and one question about the data.

spec
shawlucas: Looks like z_vr_box_draw actually has no data, and the data labeled as z_vr_box_draw data is actually part of z_sram_NES data. fixing that now

src/code/z_vr_box_draw.c
shawlucas: There is an enum `SkyboxId` from oot but I don't think all the values are the same, so I just added the value for 5 in the enum so far, because that is the enum value for the cutscene map skybox

EllipticEllipsis (approved):
Could name the final function something like `SkyboxDraw_Noop`, but I'm not fussed either way.

AngheloAlf (approved):

hensldm (approved):
LGTM

Kenix3 (approved):
Merge pending conflict resolution.

Kenix3 (approved):
Merge pending conflict resolution.

PR #488, opened 2021-11-29
z_draw

Kelebek1:
---

Adds data to z_eff_footmark, because the vertexes were inside z_draw when they weren't meant to be, so as part of moving it out I just XML'd it.


EllipticEllipsis:
Superseded by #900 .

PR #507, opened 2021-12-08
ZAPD fixes in `sys_initial_check`, update subrepos

EllipticEllipsis:
---

Fixed sys_initial_check to use manual defines rather than relying on ZAPD for texture sizes (opened a ZAPD issue for outputting them too: zeldaret/ZAPD#212 ). Happy for feedback on the defines' names or usage; there's a thread in #mm-decomp if anyone has suggestions that don't fit in a review.

Also did some subrepo updates. New stuff in ZAPD:
- zeldaret/ZAPD#205 (the one that caused this to break in the first place 😓 )
- zeldaret/ZAPD#189 (prettier and more helpful user-facing warnings)
- zeldaret/ZAPD#208 (github will no longer complain about security of old libpng)
- zeldaret/ZAPD#209 (internal cleanup)

Of these, 189 is the most interesting for the end user.


AngheloAlf (commented):
Since ZAPD now has a new syntax for warnings and a better system to use them, it would be nice to have a way to expose those flags in the `extract_assets.py` script, or at least a way to enable the unaccounted warning which seems to be the only one which is off by defect

EllipticEllipsis:
Updated `extract_assets.py` to be able to pass other arguments to ZAPD. Kept `-u` around since it's probably the most common option to want to add, and it's much faster than using `-ZWunaccounted`.

AngheloAlf (approved):

hensldm (approved):

EllipticEllipsis:
Added the latest ZAPD fixes,
- zeldaret/ZAPD#210
- zeldaret/ZAPD#217

Kenix3 (approved):

PR #513, opened 2021-12-15
Building on Macs

EllipticEllipsis:
---

Various fixes to allow the repo to build on Macs. These include
- Fix multiprocessing in `disasm.py` and `extract_assets.py` (thanks to @AngheloAlf for finding the fix for that one)
- Since these are both fixed, allow `nproc` to run on other platforms: Mac will install it as part of the coreutils requirement, and most likely Windows will have to too.
- gitignoring more stuff
- Add a temporary fix to the tools makefile for z64compress; I'll have a word with the devs of that about the issue and see how they want to handle it.
- Correct an incorrect prototype of *one function* in `functions.h` that somehow GCC didn't notice.
- Add documentation for building on Mac.

This does *not* fix the ZAPD problems: those are covered in the ZAPD update in #507, so I'm drafting this until that is merged.


extract_assets.py
```diff
@@ -69,7 +69,6 @@ def initializeWorker(abort, unaccounted: bool, extractedAssetsTracker: dict, man
 def main():
     parser = argparse.ArgumentParser(description="baserom asset extractor")
     parser.add_argument("-s", "--single", help="asset path relative to assets/, e.g. objects/gameplay_keep")
-    parser.add_argument("-t", "--threads", help="Number of cpu cores to extract with.")
```
AngheloAlf, line 72: Why did you remove this?

Makefile
```diff
@@ -235,12 +230,12 @@ setup:
...
 	python3 tools/fixbaserom.py
 	python3 tools/extract_baserom.py
-	python3 extract_assets.py -t $(N_THREADS)
+	python3 extract_assets.py
 
 ## Assembly generation
 disasm:
 	$(RM) -rf asm data
-	python3 tools/disasm/disasm.py -j $(N_THREADS)
+	python3 tools/disasm/disasm.py -j ${N_THREADS}
```
AngheloAlf, line 238: What's the reason to change `$()` to `${}`?

README.md
```diff
@@ -23,26 +23,55 @@ in an experimental and research phase and cannot currently be used traditionally
...
+```bash
+brew update
+brew install coreutils make python3 libpng
+```
+
+(The repository expects Homebrew-installed programs to be either linked correctly in `$PATH` etc. or in their default locations.)
+
+You'll also need to [build and install mips-linux-binutils](docs/BUILDING_BINUTILS_MACOS.md).
+
+The makefile should alias Apple's out-of-date `make` to `gmake` automatically, but if you have issues, try using `gmake` in place of `make`.
```
AngheloAlf, line 71: From what I understand, the default `make` command in Mac machines isn't out of date, it is just another variation/flavour of Makefile. For this project we use GNU Makefile specifically, that's why Mac users require to use `gmake` instead.

README.md
```diff
@@ -111,29 +140,23 @@ This means that something is wrong with the ROM's contents. Either the baserom f
 Running `make init` will also make the `./expected` directory and copy all of the files there, which will be useful when running the diff script. The diff script is useful in decompiling functions and can be ran with this command: `./tools/asm-differ/diff.py -wmo3 <insert_function_here>`
 
 **Note**: to speed up the build, you can either:
-* Pass `-jN` to `make setup` and `make`, where N is the number of threads to use in the build, e.g. `make -j4`. The generally-accepted wisdom is to use the number of virtual cores your computer has.
-* Pass `-j` to `make setup` and `make`, to use as many threads as possible, but beware that this can use too much memory on lower-end systems.
+* Pass `-jN` to `make setup` and `make`, where N is the number of threads to use in the build, e.g. `make -j4`. The generally-accepted wisdom is to use the number of virtual cores your computer has, which is the output of `nproc` (which should be installed as part of `coreutils`).
+* Pass `-j` to `make setup` and `make`, to use as many threads as possible, but beware that this can use too much memory on lower-end systems, and even high-end systems tend to be fastest with roughly `-j$(nproc)` threads.
```
AngheloAlf, line 144: I don't think the new stuff you added in this 2 lines add much to the building process instructions. I suggest to just remove them.

AngheloAlf (changes requested):

extract_assets.py
EllipticEllipsis: I copied most of the OoT script over so as to only maintain one. This had the side effect that all the `-t`-related code is gone.

Makefile
EllipticEllipsis: Looks like a copypaste error. I'll fix it and make sure it still works.

README.md
EllipticEllipsis: No, the reason it's different is because Apple refuse to use anything GPLv3-licensed, so they have by default the last GPLv2 version of make:
```
$ make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
$ gmake -v
GNU Make 4.3
Built for arm-apple-darwin21.1.0
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

README.md
EllipticEllipsis: The first is useful because it tells people how to find out how many virtual cores they might have, since often it's different from the physical number, and they may not even know how to check that.

We wrote the second comment based on what we knew at the time. Now we have more information (from @rylieb and my testing on higher-end machines), we know that discouraging use of `-j` is worth it on a wider range of machines. Indeed, if it was up to me, I would just remove the second bullet entirely.

extract_assets.py
AngheloAlf: Could you add it back?

README.md
AngheloAlf: Yeah, I think the second bullet should go.

extract_assets.py
EllipticEllipsis: How? I don't understand how you want it to work.

AngheloAlf: The same way it is working currently.
Use the value of args,threads for `numCores` instead of using `cpu_count`

include/functions.h
```diff
@@ -97,7 +97,7 @@ void FaultDrawer_SetCharPad(s8 padW, s8 padH);
 void FaultDrawer_SetCursor(s32 x, s32 y);
 void FaultDrawer_FillScreen(void);
 void* FaultDrawer_FormatStringFunc(void* arg, const char* str, size_t count);
-void FaultDrawer_VPrintf(const char* str, char* args);
+void FaultDrawer_VPrintf(const char* str, va_list args);
```
hensldm, line 100:
```suggestion
void FaultDrawer_VPrintf(const char* fmt, va_list ap);
```
Might as well just match it up with the function definition in `fault_drawer.c`. Don't know how I missed this when I PR'd it.

AngheloAlf (approved):

hensldm (approved):
Code wise looks good. However I do not have a MAC to actually test, someone else would have to do that.

jpburnett (approved):
Tested this on my mac and it worked.

README.md
```diff
@@ -23,26 +23,55 @@ in an experimental and research phase and cannot currently be used traditionally
...
+
+### MacOS
+
+For macOS, use Homebrew to install the following dependencies:
+
+* coreutils
+* make
+* python3
+* libpng
+
```
jpburnett, line 59:
```suggestion
* libpng
* clang-format@11
* bash
```
From working on my mac I have found that it would be helpful to add these here since they will also be necessary for submitting PRs

README.md
```diff
@@ -23,26 +23,55 @@ in an experimental and research phase and cannot currently be used traditionally
...
+You can install them with the following commands:
+
+```bash
+brew update
+brew install coreutils make python3 libpng
```
jpburnett, line 64:
```suggestion
brew install coreutils make python3 libpng clang-format@11 bash
```
see comment above

jpburnett (changes requested):
I think these will be helpful additions, there is also a tweak to `format.sh` that I think will be beneficial that I can put in discord (or just commit to your branch, whichever you prefer)

Kenix3 (approved):
Code changes look fine. I do not have Mac to test with, so this is a rubber stamp approval.

EllipticEllipsis:
Updated ZAPD to current. Includes the following:
- zeldaret/ZAPD#218
- zeldaret/ZAPD#214
- zeldaret/ZAPD#226
- zeldaret/ZAPD#220
- zeldaret/ZAPD#228
- zeldaret/ZAPD#223
- zeldaret/ZAPD#216
Provided it builds, this should be completely ready now.

AngheloAlf (approved):

PR #516, opened 2021-12-18
Update z64compress

Thar0:
---
Also includes an updated wrapper script.

 - The `--mb` argument is now optional, and defaults to the largest multiple of 8mb that fits the whole compressed rom. Matches with this default.
 - Resolves issues with `-march` `-mtune`


AngheloAlf (approved):

EllipticEllipsis (approved):

Makefile
```diff
@@ -209,7 +209,7 @@ $(ROM): $(ELF)
 	$(ELF2ROM) -cic 6105 $< $@
 
 $(ROMC): uncompressed
-	python3 tools/z64compress_wrapper.py --mb 32 --matching --threads $(N_THREADS) $(ROM) $@ $(ELF) build/$(SPEC)
+	python3 tools/z64compress_wrapper.py --matching --threads $(N_THREADS) $(ROM) $@ $(ELF) build/$(SPEC)
```
hensldm, line 212: Would we want to remove the `matching` flag for `NON-MATCHING` builds? I don't know enough about `z64compress` to know what differences it would make.

hensldm (approved):
Just one small question, but not enough to stop an approval.

Makefile
Thar0: All that removing this option does is allow z64compress to cut some corners to run a little faster, such as not padding the rom with the pattern `00010203...FF...` and skip some unnecessary alignment of rom segments. I don't mind adding something to not use this flag if `NON_MATCHING` is set, it would probably be nice to have.

Kenix3 (approved):

PR #517, opened 2021-12-18
Document object_gmo

tom-overton:
---

This is the first object I've ever documented, so let me know if I messed anything up. It's a pretty straightforward object though.


AngheloAlf (approved):
Nice work on the very first documented object :partying_face:

EllipticEllipsis (approved):

assets/xml/objects/object_gmo.xml
```diff
@@ -1,13 +1,15 @@
...
-        <Animation Name="object_gmo_Anim_000018" Offset="0x18" />
-        <DList Name="object_gmo_DL_000B70" Offset="0xB70" />
-        <DList Name="object_gmo_DL_000E38" Offset="0xE38" />
-        <Texture Name="object_gmo_TLUT_000EC8" OutName="tlut_000EC8" Format="rgba16" Width="16" Height="16" Offset="0xEC8" />
-        <Texture Name="object_gmo_Tex_0010C8" OutName="tex_0010C8" Format="ci8" Width="32" Height="32" Offset="0x10C8" />
-        <!-- <Blob Name="object_gmo_Blob_0014C8" Size="0xC00" Offset="0x14C8" /> -->
-        <Limb Name="object_gmo_Standardlimb_0020C8" Type="Standard" Offset="0x20C8" />
-        <Limb Name="object_gmo_Standardlimb_0020D4" Type="Standard" Offset="0x20D4" />
-        <Skeleton Name="object_gmo_Skel_0020E8" Type="Normal" LimbType="Standard" Offset="0x20E8" />
+        <Animation Name="gNejironUnusedAnim" Offset="0x18" />
```
hensldm, line 3: Even if it unused, can we figure out what it is in z64Utils? Would prefer a more descriptive name if we can.

hensldm (approved):
Looks good. Just one question.

assets/xml/objects/object_gmo.xml
tom-overton: Unfortunately, we can't really tell what it is. It's only two frames long, and the Nejiron is completely still during it. There is an animation in MM3D called "anim/gm_default.csab" that *may* be the original name for this, but it's hard to tell because that animation *also* is completely idle. I can put a comment about the potential original name.

AngheloAlf: What about `gNejironIdleAnim`?

tom-overton: Sure, that works.

Kenix3 (approved):

PR #520, opened 2021-12-19
Effect_soft_sprite_old_init and particleOverlayTable OK

hensldm:
---

Imported the data for effect_soft_sprite_old_init which allows matching for the two non-matchings left. As part of this also split off the particleOverlayTable into its data file, as well as imported it into `src/code/z_effect_soft_sprite_dlftbls.c`.


include/variables.h
```diff
@@ -384,7 +384,7 @@ extern UNK_PTR D_801AE260[3];
 // extern UNK_TYPE4 D_801AE2DC;
 extern F3DVertex sEffectShieldParticleVertices[4];
 extern EffInfo sEffInfoTable[5];
-extern EffectSsInfo EffectSS2Info;
+extern EffectSsInfo sEffectSsInfo;
```
EllipticEllipsis, line 387: If these are static, they should be able to be commented out here.

src/code/z_effect_soft_sprite.c
```diff
@@ -27,17 +29,17 @@ void EffectSS_Clear(GlobalContext* globalCtx) {
...
-    EffectSS2Info.data_table = NULL;
-    EffectSS2Info.searchIndex = 0;
-    EffectSS2Info.size = 0;
+    sEffectSsInfo.data_table = NULL;
+    sEffectSsInfo.searchIndex = 0;
+    sEffectSsInfo.size = 0;
 
-    // This code is completely useless, as data_table was just set to NULL and size to 0
-    for (effectsSs = EffectSS2Info.data_table; effectsSs < EffectSS2Info.data_table + EffectSS2Info.size; effectsSs++) {
+    //! @bug: This code is completely useless, as data_table was just set to NULL and size to 0
```
EllipticEllipsis, line 36: Good job IDO.

src/code/z_effect_soft_sprite_old_init.c
```diff
@@ -364,24 +361,18 @@ void EffectSsGSpk_SpawnNoAccel(GlobalContext* globalCtx, Actor* actor, Vec3f* po
...
-extern Color_RGBA8 D_801AE3D8;
-extern Color_RGBA8 D_801AE3DC;
-
 // unused
 void EffectSsGSpk_SpawnRandColor(GlobalContext* globalCtx, Actor* actor, Vec3f* pos, Vec3f* velocity, Vec3f* accel,
                                  s16 scale, s16 scaleStep) {
-    Color_RGBA8 primColor = D_801AE3D8;
-    Color_RGBA8 envColor = D_801AE3DC;
+    Color_RGBA8 primColor = { 255, 255, 150, 255 };
+    Color_RGBA8 envColor = { 255, 0, 0, 0 };
```
EllipticEllipsis, line 375: So they forgot to make these ones static? Oops.

EllipticEllipsis (approved):
One minor thing in variables.h, otherwise looks good.

src/code/z_effect_soft_sprite_dlftbls.c
```diff
@@ -0,0 +1,56 @@
+#include "global.h"
+#include "initvars.h"
+
+#define EFFECT_SS_OVERLAY(name)                                                                         \
+    {                                                                                                   \
+        (u32) _ovl_##name##SegmentRomStart, (u32)_ovl_##name##SegmentRomEnd, _ovl_##name##SegmentStart, \
+            _ovl_##name##SegmentEnd, NULL, &name##_InitVars, 1,                                         \
```
AngheloAlf, line 7: I recommend using the macros for the Segments instead.
Also, is the `u32` cast required? Would this match with `uintptr_t` instead?
```suggestion
        SEGMENT_ROM_START(ovl_##name), SEGMENT_ROM_END(ovl_##name), SEGMENT_START(ovl_##name), \
            SEGMENT_END(ovl_##name), NULL, &name##_InitVars, 1,                                         \
```

AngheloAlf (commented):
Nice work. Just one suggestion

src/code/z_effect_soft_sprite_dlftbls.c
hensldm: Ah those are better.
The casts are not needed.

AngheloAlf (approved):

engineer124 (commented):
Nice work! Just some small comments

src/code/z_effect_soft_sprite_dlftbls.c
```diff
@@ -0,0 +1,56 @@
...
+#include "initvars.h"
+
+#define EFFECT_SS_OVERLAY(name)                                                                \
+    {                                                                                          \
+        SEGMENT_ROM_START(ovl_##name), SEGMENT_ROM_END(ovl_##name), SEGMENT_START(ovl_##name), \
+            SEGMENT_END(ovl_##name), NULL, &name##_InitVars, 1,                                \
+    }
+
+#define EFFECT_SS_OVERLAY_INTERNAL(name) \
+    { 0, 0, NULL, NULL, NULL, &name##_InitVars, 1 }
```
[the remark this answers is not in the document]
EllipticEllipsis: Dynamically Linked/Loaded File, i.e. overlay.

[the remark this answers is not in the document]
EllipticEllipsis: Or possibly "Fragment", given the file in code called load fragment. But that's even less clear.

src/code/z_effect_soft_sprite.c
hensldm: I don't think it's redundant. Because the table was just set to `NULL`, the individual effects don't get destroyed properly, i.e. `EffectSS_Delete` doesn't get called for each effect left in the table.

engineer124: Hmmm I see, then yeah that does kinda lean towards being a bug then. It'd be good to add that to the description of the bug to make clearer why it's a bug instead of useless/redundant

hensldm: Updated the comment to be clearer.

engineer124 (approved):

Kenix3 (approved):

PR #521, opened 2021-12-20
Document object_giant and finish documenting En_Giant

tom-overton:
---

A much more complicated object than my other PR. I took some pointers from OoT XMLs on how to name the limbs and how to structure the XML in general. I didn't bother breaking apart all the body vertices into separate limb vertices, but I could if it's desirable.


src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -440,21 +425,16 @@ void EnGiant_PostLimbDrawXlu(GlobalContext* globalCtx, s32 limbIndex, Gfx** dLis
...
-    if (limbIndex == 1) {
-        Matrix_CopyCurrentState(&this->unk_254);
+    if (limbIndex == EN_GIANT_LIMB_HEAD) {
+        Matrix_CopyCurrentState(&this->headDrawMtxF);
     }
 }
 
 void EnGiant_Draw(Actor* thisx, GlobalContext* globalCtx) {
     s32 pad;
     EnGiant* this = THIS;
```
EllipticEllipsis, line 436:
```suggestion
```
Don't need a blank line between the static and the ordinary variables.

assets/xml/objects/object_giant.xml
```diff
@@ -1,63 +1,90 @@
...
+
+        <!-- Giant Body Vertices. Contains everything but the beard. -->
+        <Array Name="gGiantBodyVtx" Count="444" Offset="0x2180">
+            <Vtx/>
+        </Array>
+
+        <!-- Giant Unused Beard Vertices. An exact duplicate of the beard vertices that are actually used. -->
+        <Array Name="gGiantUnusedBeardVtx" Count="9" Offset="0x3D40">
+            <Vtx/>
+        </Array>
```
EllipticEllipsis, line 14: I'm not sure about only doing some of the Vtx... is there a reason to single this one out?

EllipticEllipsis (commented):
It's amazing how much more readable it becomes with the object stuff in as well. A few things I spotted, mostly minor.

assets/xml/objects/object_giant.xml
```diff
@@ -1,63 +1,90 @@
...
+        <!-- Giant Limb DisplayLists -->
+        <DList Name="gGiantHeadDL" Offset="0x3DD0" />
+        <DList Name="gGiantRightThighDL" Offset="0x40C0" />
+        <DList Name="gGiantRightShinDL" Offset="0x4200" />
+        <DList Name="gGiantRightFootDL" Offset="0x4438" />
+        <DList Name="gGiantLeftThighDL" Offset="0x4560" />
+        <DList Name="gGiantLeftShinDL" Offset="0x46A0" />
+        <DList Name="gGiantLeftFootDL" Offset="0x48D8" />
+        <DList Name="gGiantRightShoulderDL" Offset="0x4A00" />
+        <DList Name="gGiantRightUpperArmDL" Offset="0x4B10" />
```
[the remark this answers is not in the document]
tom-overton: Honestly easier to show this with pictures:
image
image

The shoulder looks a bit weird, but it's definitely separate from the rest of the arm.

src/overlays/actors/ovl_En_Giant/z_en_giant.c
```diff
@@ -352,20 +337,20 @@ void EnGiant_PlaySound(EnGiant* this) {
 
 void EnGiant_UpdatePosition(EnGiant* this, GlobalContext* globalCtx, u32 actionIndex) {
     CsCmdActorAction* actorAction = globalCtx->csCtx.npcActions[actionIndex];
-    f32 floatUnk10 = actorAction->unk10;
+    f32 startPosY = actorAction->unk10;
     s32 pad[2];
-    f32 floatUnk1C = actorAction->unk1C;
-    f32 functionTemp;
+    f32 endPosY = actorAction->unk1C;
+    f32 lerpScale;
```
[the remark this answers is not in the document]
tom-overton: Well, I just called it that because `func_800F5A8C` is unnamed in this branch (it's named in the z_demo PR). The `lerp` part will be redundant once the function is named, though, so let me just change it to `scale`.

assets/xml/objects/object_giant.xml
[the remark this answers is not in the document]
EllipticEllipsis: Ah yeah, they're so weird-looking... I agree, usually people like to use shoulder for the upper arm, so just wanted to make sure.

assets/xml/objects/object_giant.xml
tom-overton: Because it was this blob in the original XML:
```xml
        <!-- <Blob Name="object_giant_Blob_003D40" Size="0x90" Offset="0x3D40" /> -->
```
I didn't want to leave the blob as-is, so I defined it. But then having *only* that Vtx looked weird, so I defined the body and (used) beard Vtx too. But the body Vtx could be broken up further into individual limbs and pieces of limbs...not sure what the best solution for this is, because documenting every single set of Vtx sounds like a pain.

EllipticEllipsis: If the body vtx are all in one big lump (and ZAPD has extracted it as such), I think something like `gGiantBodyVtx` would be fine. Or maybe `model`, but that would require discussion.

tom-overton: Yeah, ZAPD extracted them as one big chunk; all I did is give them a name. Z64Utils breaks them down into 51 separate batches of vertices, but they're all adjacent, so I guess ZAPD either didn't notice or didn't care and just grabbed the entire thing as if it were one big chunk.

EllipticEllipsis: This is to do with how they're used: most flex skeletons seem to put them in one big lump and access them as an array. Non-flex skeletons are rather worse.

EllipticEllipsis (approved):

hensldm (approved):
LGTM, though there is currently a merge conflict.

tom-overton:
Merge conflict fixed, should be able to remove the "Waiting-for-author" tag

Kenix3 (approved):

PR #522, opened 2021-12-21
Three Makefile tweaks

EllipticEllipsis:
---

Made a few efficiency changes to the Makefile:
- Only copy the baserom files that are actually needed (58 instead of 1532).
- Remove use of `iconv` on the data before it's assembled: there is no data that needs this any more.
- Remove the optimisation flags from the assets: these won't do anything to data, but `-g3` at least will needlessly bloat the symbol tables in the files and increase linking time.

I found this saved approximately 40 seconds over building on master:

master:
```
real    11m42.878s
user    9m39.837s
sys     1m40.324s
```
With these three changes:
```
real    11m1.702s
user    9m4.837s
sys     1m31.141s
```
(tested under similar conditions, dropping caches between, so I think this effect is real.)


AngheloAlf (approved):

jpburnett (approved):
the need for speed is real 🏃 💨

Kenix3 (approved):

PR #530, opened 2021-12-24
Document Object Ctower Rot

hensldm:
---

Another opportunity to cleanup a couple of my first actors :).


src/overlays/actors/ovl_Bg_Ctower_Gear/z_bg_ctower_gear.c
```diff
@@ -198,18 +178,19 @@ void BgCtowerGear_UpdateOrgan(Actor* thisx, GlobalContext* globalCtx) {
     }
 }
 
-// Using BgCtowerGear *this = THIS causes regalloc issues
 void BgCtowerGear_Draw(Actor* thisx, GlobalContext* globalCtx) {
-    func_800BDFC0(globalCtx, D_80AD32E8[BGCTOWERGEAR_GET_TYPE(thisx)]);
+    static Gfx* sDLists[] = { gClockTowerCeilingCogDL, gClockTowerCenterCogDL, gClockTowerWaterWheelDL };
+
+    func_800BDFC0(globalCtx, sDLists[BGCTOWERGEAR_GET_TYPE(thisx)]);
```
AngheloAlf, line 184: I'm not a huge fan of having in-function static data if it isn't required to match. I would personally prefer to put it outside of the function, but up to you.

src/overlays/actors/ovl_Bg_Ctower_Rot/z_bg_ctower_rot.h
```diff
@@ -9,9 +9,9 @@ typedef void (*BgCtowerRotActionFunc)(struct BgCtowerRot*, GlobalContext*);
 
 typedef enum {
     /* 0 */ CORRIDOR,
-    /* 1 */ MAIN_DOOR, // Main door controls cutscene flow (updating Link fields and playing sounds). It is the one on
-                       // the left when entering from the Lost Woods
-    /* 2 */ DOOR,
+    /* 1 */ STONE_DOOR_MAIN, // Main door controls cutscene flow (updating Link fields and playing sounds). 
+                            // It is the one on the left when entering from the Lost Woods
+    /* 2 */ STONE_DOOR,
```
AngheloAlf, line 14: What about prefixing those with something like `BGCTOWER_` or similar?

AngheloAlf (commented):
Nice work with this object and the cleanup!

src/overlays/actors/ovl_Bg_Ctower_Rot/z_bg_ctower_rot.h
EllipticEllipsis: Especially since they're in the header.

hensldm: Ah, I did it for `Gear`, but must have forgot for `Rot`. Prefixed with `BGCTOWERROT_`.

AngheloAlf (approved):

hensldm:
Just want to note that I've seen some comments about what object files should have, so have periodically added them to this PR. Namely using tlut instead of palette, adding a file header, using `empty#DL` instead of `emptyDL#`.

jpburnett (approved):
Nice documentation!

Kenix3 (approved):
Great formatting in your xml!

PR #533, opened 2021-12-25
En_Ig

Kelebek1:
---


src/overlays/actors/ovl_En_Ig/z_en_ig.c
```diff
@@ -31,105 +70,951 @@ const ActorInit En_Ig_InitVars = {
...
+    Actor* temp_v0;
+
+    while (true) {
+        foundActor = func_ActorCategoryIterateById(globalCtx, foundActor, actorCat, actorId);
+
+        if (foundActor == NULL) {
+            break;
+        }
+
+        if ((this != (EnIg*)foundActor) && (foundActor->update != NULL)) {
```
AngheloAlf, line 134: I think I would prefer to do this the other way around
```suggestion
        if ((&this->actor != foundActor) && (foundActor->update != NULL)) {
```

AngheloAlf (approved):
Just a minor thing, but I'll approve anyways

src/overlays/actors/ovl_En_Ig/z_en_ig.c
```diff
@@ -31,105 +70,951 @@ const ActorInit En_Ig_InitVars = {
...
+        if (this->unk_274 != NULL) {
+            sp4C = Lib_SegmentedToVirtual(this->unk_274->points);
+            Math_Vec3s_ToVec3f(&sp40, &sp4C[0]);
+            Math_Vec3s_ToVec3f(&sp34, &sp4C[1]);
+            Math_Vec3f_Copy(&this->unk_2B0, &sp40);
+            Math_Vec3f_Copy(&this->unk_2BC, &sp34);
+            this->actor.world.rot.y = Math_Vec3f_Yaw(&sp40, &sp34);
+            Math_Vec3f_Copy(&this->actor.world.pos, &sp40);
+
+            if (ABS_ALT((s16)(this->actor.world.rot.y - sp50->actor.shape.rot.y)) <= 0x4000) {
```
hensldm, line 533:
```suggestion
            if (ABS_ALT(BINANG_SUB(this->actor.world.rot.y, sp50->actor.shape.rot.y)) <= 0x4000) {
```

hensldm (approved):
Small cosmetic suggestion.

src/overlays/actors/ovl_En_Ig/z_en_ig.c
Kenix3: +1 for this

Kenix3 (changes requested):

src/overlays/actors/ovl_En_Ig/z_en_ig.c
Kelebek1: It's non-equivalent, it flips the compare and can't be fixed by moving them. I was stuck on this with a different actor ages ago and Roman found this.

src/overlays/actors/ovl_En_Ig/z_en_ig.c
```diff
@@ -31,105 +70,951 @@ const ActorInit En_Ig_InitVars = {
...
+
+        case 8:
+            phi_a1 = 15;
+            break;
+
+        default:
+            phi_a1 = -1;
+            break;
+    }
+    return (EnDoor*)SubS_FindDoor(globalCtx, phi_a1);
```
hensldm, line 168:
```suggestion
    return SubS_FindDoor(globalCtx, phi_a1);
```
I was able to find a solution so that `SubS_FindDoor` actually returns an `EnDoor` now.

Kenix3 (approved):

PR #543, opened 2021-12-27
Update recomp

EllipticEllipsis:
---

Updated recomp to use the latest `per-function` branch of https://github.com/Emill/ido-static-recomp , compiled with Clang as this produced better speed than GCC when I tested it. 

Performance improvements seem to be hard to measure: 
- I ran 3 of old/new on WSL and found anywhere from a 10% to 40% speed improvement.
- On Mac the improvement seems to be consistently higher, roughly 50% faster.

I have not attempted to update the Windows binaries since I don't have Cygwin, git-for-Windows, or any of that; happy to take PRs for other architectures or improvements if we want to update the others in this PR.

Same as the OoT PR, zeldaret/oot#1085.


AngheloAlf (approved):
Faster builds :eyes:

hensldm (approved):

isghj5 (approved):

Kenix3 (approved):

PR #544, opened 2021-12-28
Document Deku Flower objects and other Obj_Etcetera-related cleanup

tom-overton:
---

A rare instance of the original animation names actually being helpful; it's much easier to see what the animations are supposed to convey when you see how they were named. I went ahead and did some cleanup on Obj_Etcetera (the Deku Flower actor) too in order to better match what names I gave in the object.


assets/xml/objects/gameplay_keep.xml
```diff
@@ -766,59 +766,59 @@
...
-        <Limb Name="gameplay_keep_Standardlimb_012778" Type="Standard" Offset="0x12778" />
-        <Limb Name="gameplay_keep_Standardlimb_012784" Type="Standard" Offset="0x12784" />
-        <Limb Name="gameplay_keep_Standardlimb_012790" Type="Standard" Offset="0x12790" />
-        <Limb Name="gameplay_keep_Standardlimb_01279C" Type="Standard" Offset="0x1279C" />
-        <Limb Name="gameplay_keep_Standardlimb_0127A8" Type="Standard" Offset="0x127A8" />
-        <Limb Name="gameplay_keep_Standardlimb_0127B4" Type="Standard" Offset="0x127B4" />
-        <Skeleton Name="gameplay_keep_Skel_0127E8" Type="Flex" LimbType="Standard" Offset="0x127E8" />
+        <Collision Name="gPinkDekuFlowerCol" Offset="0xE710" />
+        <Animation Name="gDekuFlowerBounceAnim" Offset="0xEB7C" /> <!-- Original name is "dnp_boyoyon" ("bounce") -->
+        <DList Name="gPinkDekuFlowerStaticDL" Offset="0xED80" />
```
EllipticEllipsis, line 771: I have a bit of a thing about `static` now, wonder if we can't use another word. In what situations does this actually get used?

assets/xml/objects/gameplay_keep.xml
```diff
@@ -766,59 +766,59 @@
...
-        <Limb Name="gameplay_keep_Standardlimb_012760" Type="Standard" Offset="0x12760" />
-        <Limb Name="gameplay_keep_Standardlimb_01276C" Type="Standard" Offset="0x1276C" />
-        <Limb Name="gameplay_keep_Standardlimb_012778" Type="Standard" Offset="0x12778" />
-        <Limb Name="gameplay_keep_Standardlimb_012784" Type="Standard" Offset="0x12784" />
-        <Limb Name="gameplay_keep_Standardlimb_012790" Type="Standard" Offset="0x12790" />
-        <Limb Name="gameplay_keep_Standardlimb_01279C" Type="Standard" Offset="0x1279C" />
-        <Limb Name="gameplay_keep_Standardlimb_0127A8" Type="Standard" Offset="0x127A8" />
-        <Limb Name="gameplay_keep_Standardlimb_0127B4" Type="Standard" Offset="0x127B4" />
-        <Skeleton Name="gameplay_keep_Skel_0127E8" Type="Flex" LimbType="Standard" Offset="0x127E8" />
+        <Collision Name="gPinkDekuFlowerCol" Offset="0xE710" />
```
EllipticEllipsis, line 769: Hmm, I wonder if we should have newlines where the things aren't related like you did at the bottom of this block. This xml is already huge.

src/overlays/actors/ovl_Obj_Etcetera/z_obj_etcetera.c
```diff
@@ -105,6 +106,11 @@ void ObjEtcetera_Destroy(Actor* thisx, GlobalContext* globalCtx) {
     Collider_DestroyCylinder(globalCtx, &this->collider);
 }
 
+/**
+ * This function will make the flower oscillate on the X and Z axes in most situations
+ * where something interacts with it. When the player launches out of the flower, the
+ * oscillation is handled by ObjEtcetera_DoBounceOscillation instead
```
EllipticEllipsis, line 112:
```suggestion
 * oscillation is handled by ObjEtcetera_DoBounceOscillation instead.
```

EllipticEllipsis (commented):
Nice job, just some very minor things.

EllipticEllipsis:
And apparently the type of something is wrong... do you need to delete the &s on the dlists?

assets/xml/objects/gameplay_keep.xml
tom-overton: There are two different ways to draw the Deku Flower. One is as an animated bunch of limbs (`ObjEtcetera_DrawAnimated`) and one is as a single, static, non-moving Deku Flower (`ObjEtcetera_DrawIdle`). The `static` DLs are used in the latter case. I suppose I could use `Idle` instead?

assets/xml/objects/gameplay_keep.xml
tom-overton: Huh, I don't see any newlines on my end. Do you want me to insert a newline (and maybe a comment) at the start and end of the Deku Flower section?

assets/xml/objects/gameplay_keep.xml
EllipticEllipsis: Ah, I see, it switches the draws when it wants. That should be documented somewhere explicitly, I suppose, maybe above the draw functions? Idle would make sense given what you've called the associated functions, but it might not be descriptive enough.

tom-overton: Yeah, I'll just use `Idle` for now and add some documentation.

assets/xml/objects/gameplay_keep.xml
EllipticEllipsis: Oh, it's GitHub's pagewidth playing with the xml tags. Yes, I think that would be a good idea, and head the section with a comment.

src/overlays/actors/ovl_Obj_Etcetera/z_obj_etcetera.c
```diff
@@ -208,20 +221,21 @@ void ObjEtcetera_DoIntenseOscillation(ObjEtcetera* this, GlobalContext* globalCt
...
+    scaleTemp = dyna->actor.scale.x + this->bounceOscillationScale;
     Actor_SetScale(&this->dyna.actor, scaleTemp);
     this->dyna.actor.scale.y = 2.0f * scaleTemp;
 }
 
 void ObjEtcetera_Setup(ObjEtcetera* this, GlobalContext* globalCtx) {
     CollisionHeader* colHeader = NULL;
     s32 type;
-    CollisionHeader* collisionHeaders[] = { &D_0400E710, &D_0400E710, &D_040118D8, &D_040118D8 };
+    CollisionHeader* collisionHeaders[] = { &gPinkDekuFlowerCol, &gPinkDekuFlowerCol, &gGoldDekuFlowerCol,
```
hensldm, line 237: With the longer names, this would probably be better formatted with a trailing comma.

hensldm (approved):
Small formatting suggestion

EllipticEllipsis (approved):
I particularly like the new documentation, good job.

Kenix3 (approved):

PR #547, opened 2021-12-29
EnMag OK and documented

EllipticEllipsis:
---

All functions decompiled, all used variables named, some macros made, added a lot of documentation of various functions. Inspiration for some names was taken from OoT, but this one is rather more complicated than OoT's, so I had to improvise a fair amount.

Very interesting and unusual actor, I learnt a lot about graphics macros decompiling this one. And it's been a long time since I did an actor and documented it this thoroughly, but I thought it was worth doing so since it's so unusual and prominent. I did my best on the graphics stuff although I'm still rather ignorant about it. Overall I went into more detail than I usually do since this actor is delightfully easy to test: sometimes you don't even need to do any inputs!

I'm not sure about the placement of the various macros I introduced. 
- The custom gbi one will almost certainly be used elsewhere, so I put it in macros.h like OoT. It's also important to note I had to remove the do while to get it to match here.
- Following conversations in Discord it sounds like the timed linear stepping ones may not be useful elsewhere, so I left them in the C file for now.
- `TMEM_SIZE` I put in hardware.h, it could fit equally well in rdp.h, or even somewhere else.
- The texture size macros will eventually be made and replaced by ZAPD; I put them above the function that used them rather than cluttering the top of the file further: it may be a reasonable idea to move the TIMED_STEP ones down too.
- The texture coordinate ones I think are best in the context of the function that uses them.

There is some rather eccentric use of temps vs statics in here too. The in-function static is required to match in the draw function. The Updates don't need it but it was little enough that I thought I might as well put the rest in too.

I'm open to better name suggestions too: some of the ones I ended up mith are stupidly long, but seem necessary to distinguish things easily.


AngheloAlf (approved):
Nice work!

src/overlays/actors/ovl_En_Mag/z_en_mag.c
```diff
@@ -28,28 +97,878 @@ const ActorInit En_Mag_InitVars = {
     (ActorFunc)EnMag_Draw,
 };
 
-#endif
+void EnMag_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnMag* this = (EnMag*)thisx;
```
hensldm, line 101:
```suggestion
    EnMag* this = THIS;
```

src/overlays/actors/ovl_En_Mag/z_en_mag.c
```diff
@@ -28,28 +97,878 @@ const ActorInit En_Mag_InitVars = {
...
+ * Controls the actions performed using a switch and the `state` struct variable rather than action functions. Most of
+ * these are to do with fading various parts in and out.
+ */
+void EnMag_Update(Actor* thisx, GlobalContext* globalCtx) {
+    static s16 sAppearEffectPrimGreenTargets[] = { 255, 155 };
+    static s16 sAppearEffectEnvRedTargets[] = { 255, 0 };
+    static s16 sAppearEffectEnvBlueTargets[] = { 0, 155 };
+    s16 step;
+    s32 pad[2];
+    EnMag* this = (EnMag*)thisx;
```
hensldm, line 234:
```suggestion
    EnMag* this = THIS;
```

src/overlays/actors/ovl_En_Mag/z_en_mag.c
```diff
@@ -28,28 +97,878 @@ const ActorInit En_Mag_InitVars = {
...
+ * Draws an i8 texture.
+ *
+ * @param gfxp Pointer to current displaylist.
+ * @param texture Texture to draw.
+ * @param texWidth Width of the texture.
+ * @param texHeight Height of the texture.
+ * @param rectLeft X coordinate of the top-left of the draw position.
+ * @param rectTop Y coordinate of the top-left of the draw position.
+ */
+void EnMag_DrawTextureI8(Gfx** gfxp, void* texture, s16 texWidth, s16 texHeight, s16 rectLeft, s16 rectTop) {
```
hensldm, line 462:
```suggestion
void EnMag_DrawTextureI8(Gfx** gfxp, TexturePtr texture, s16 texWidth, s16 texHeight, s16 rectLeft, s16 rectTop) {
```
More of a thought/question rather than suggestion. Applies elsewhere.

hensldm (approved):
Just a few cosmetic things I saw.

src/overlays/actors/ovl_En_Mag/z_en_mag.c
EllipticEllipsis: I think that makes sense. The functions where it matters treat them as a byte array anyway.

Kenix3 (approved):
Amazing work on this;

PR #558, opened 2021-12-31
Document objects used with En_Fall (fall, fall2, lodmoon, moonston)

tom-overton:
---

This PR completely documents object_fall, object_lodmoon, and object_moonston. It also partially documents object_fall2; the parts used by the moon's warp beam are left undocumented because they use the c_keyframe system, and I have no clue how that works at all. In addition to documenting objects, two minor changes were made to En_Fall itself:
- Change `FireBall` to `Fireball`, since it's one word
- Move the flags related to it crashing to the .c file, since no other code file needs to know or care about them.


assets/xml/objects/object_fall.xml
```diff
@@ -1,31 +1,31 @@
...
-        <DList Name="object_fall_DL_000198" Offset="0x198" />
-        <DList Name="object_fall_DL_000220" Offset="0x220" />
-        <Texture Name="object_fall_Tex_000260" OutName="tex_000260" Format="i8" Width="16" Height="16" Offset="0x260" />
-        <DList Name="object_fall_DL_000420" Offset="0x420" />
-        <DList Name="object_fall_DL_000428" Offset="0x428" />
-        <DList Name="object_fall_DL_000490" Offset="0x490" />
-        <DList Name="object_fall_DL_000498" Offset="0x498" />
-        <Array Name="object_fall_Vtx_0004C0" Count="209" Offset="0x4C0">
+        <DList Name="gMoonEmptyDL1" Offset="0x190" /> <!-- Was likely going to be used with gMoonDebrisModel1DL before they made all debris use the same material -->
+        <DList Name="gMoonDebrisMaterialDL" Offset="0x198" />
```
EllipticEllipsis, line 4: In what sense is the word being used here?

assets/xml/objects/object_fall.xml
```diff
@@ -1,31 +1,31 @@
...
+        <Texture Name="gMoonFireballFlecksTex" OutName="moon_fireball_flecks" Format="i8" Width="64" Height="64" Offset="0x1688" />
+        <Texture Name="gMoonFireballFlamesTex" OutName="moon_fireball_flames" Format="i4" Width="64" Height="64" Offset="0x2688" />
+        <Texture Name="gMoonFireballMaskTex" OutName="moon_fireball_mask" Format="i4" Width="64" Height="64" Offset="0x2E88" />
+        <DList Name="gMoonFireRingDL" Offset="0x3C30" />
+        <Texture Name="gMoonFireRingFlamesTex" OutName="moon_fire_ring_flames" Format="i8" Width="32" Height="64" Offset="0x3E28" />
+        <Texture Name="gMoonFireRingMaskTex" OutName="moon_fire_ring_mask" Format="i8" Width="32" Height="64" Offset="0x4628" />
+        <TextureAnimation Name="gMoonFireRingTexAnim" Offset="0x4E38" />
+        <TextureAnimation Name="gMoonUnusedFireballTexAnim" Offset="0x4E58" /> <!-- Almost certainly scrapped for just doing the TwoTexScrolls manually in EnFall_Fireball_Draw -->
+        <DList Name="gMoonDL" Offset="0x77F0" />
+        <Texture Name="gMoonFarSideTLUT" OutName="moon_far_side_palette" Format="rgba16" Width="4" Height="4" Offset="0x82F8" />
```
EllipticEllipsis, line 24: I believe we use tlut for the OutName too.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -24,6 +24,9 @@
```
EllipticEllipsis, line 24: Since you name the functions by the subtypes, I think it would be a good idea to include the names you use in the file description at the top; basically make it a description list rather than an unordered one, to use the html terms.

assets/xml/objects/object_lodmoon.xml
```diff
@@ -1,12 +1,12 @@
 ﻿<Root>
     <File Name="object_lodmoon" Segment="6">
```
EllipticEllipsis, line 2: Since it's not immediately obvious what this object is, could you put a comment with a description before the contents of the `File` tag? Actually I'm quite tempted to ask for that across the board while we've only got a few done.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -762,62 +765,62 @@ void EnFall_LodMoon_DrawWithLerp(Actor* thisx, GlobalContext* globalCtx) {
...
+                    (s32)(((1.0f - this->fireballIntensity) * 160.0f) + (255.0f * this->fireballIntensity)),
+                    (s32)((70.0f * (1.0f - this->fireballIntensity)) + (255.0f * this->fireballIntensity)),
+                    (s32)(70.0f * (1.0f - this->fireballIntensity)), 255);
     gDPSetEnvColor(
-        POLY_XLU_DISP++, (s32)(((1.0f - this->fireBallIntensity) * 50.0f) + (200.0f * this->fireBallIntensity)),
-        (s32)(20.0f * (1.0f - this->fireBallIntensity)), (s32)(20.0f * (1.0f - this->fireBallIntensity)), 255);
+        POLY_XLU_DISP++, (s32)(((1.0f - this->fireballIntensity) * 50.0f) + (200.0f * this->fireballIntensity)),
+        (s32)(20.0f * (1.0f - this->fireballIntensity)), (s32)(20.0f * (1.0f - this->fireballIntensity)), 255);
 
     // Glowing sphere of fire
```
EllipticEllipsis, line 786: I think it's fairer to say that these are used on this particular component: currently it sounds like this is the whole thing, which of course it is not.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -516,65 +519,65 @@ void EnFall_FireBall_SetPerVertexAlpha(f32 fireBallAlpha) {
...
+        this->actor.draw = EnFall_Fireball_Draw;
+        if (this->flags & FLAG_FIRE_BALL_INTENSIFIES) {
+            this->fireballIntensity += 0.01f;
+            if (this->fireballIntensity > 1.0f) {
+                this->fireballIntensity = 1.0f;
             }
         }
         func_800EDF24(&this->actor, globalCtx, func_800EE200(globalCtx, 0x1C2));
 
         switch (globalCtx->csCtx.npcActions[func_800EE200(globalCtx, 0x1C2)]->unk0) {
```
EllipticEllipsis, line 556: Here and elsewhere, I think it would be good to have blank lines between the `break` and the next case, to make it a bit easier to read.

If you really want to go all the way, you could note by any defaults which cases they cover, providing it's only a few.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -693,7 +696,7 @@ void EnFall_Moon_Draw(Actor* thisx, GlobalContext* globalCtx) {
     Matrix_MultiplyVector3fByState(sFocusOffset, &this->actor.focus.pos);
     primColor = (this->eyeGlowIntensity * 200.0f) + 40.0f;
```
EllipticEllipsis, line 697: I think a space between the matrix and the primcolor set would make this a bit clearer. Applies elsewhere.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -727,9 +730,9 @@ void EnFall_LodMoon_Draw(Actor* thisx, GlobalContext* globalCtx) {
     gSPLoadGeometryMode(POLY_OPA_DISP++, G_ZBUFFER | G_SHADE | G_CULL_BACK | G_SHADING_SMOOTH);
     primColor = (this->eyeGlowIntensity * 200.0f) + 40.0f;
     gDPSetPrimColor(POLY_OPA_DISP++, 0, 0x80, primColor, primColor, primColor, 255);
-    gSPDisplayList(POLY_OPA_DISP++, object_lodmoon_DL_0010E0);
+    gSPDisplayList(POLY_OPA_DISP++, gLodmoonEyesDL);
     gSPLoadGeometryMode(POLY_OPA_DISP++, G_ZBUFFER | G_SHADE | G_CULL_BACK | G_LIGHTING | G_SHADING_SMOOTH);
-    gSPDisplayList(POLY_OPA_DISP++, object_lodmoon_DL_001158);
+    gSPDisplayList(POLY_OPA_DISP++, gLodmoonMoonDL);
     POLY_OPA_DISP = func_801660B8(globalCtx, POLY_OPA_DISP);
```
EllipticEllipsis, line 736: Same here - this block's a bit intense.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -762,62 +765,62 @@ void EnFall_LodMoon_DrawWithLerp(Actor* thisx, GlobalContext* globalCtx) {
...
     f32 scale = this->scale * 0.06f;
     s32 i;
 
     OPEN_DISPS(globalCtx->state.gfxCtx);
     func_8012C28C(globalCtx->state.gfxCtx);
-    gSPDisplayList(POLY_OPA_DISP++, object_fall_DL_000198);
+    gSPDisplayList(POLY_OPA_DISP++, gMoonDebrisMaterialDL);
     gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 255, 255, 255, 255);
     gDPSetEnvColor(POLY_OPA_DISP++, 0, 0, 0, 255);
     for (i = 0; i < ARRAY_COUNT(debrisParticles); i++) {
```
EllipticEllipsis, line 814: I think I'd put a space before the for loop.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -832,13 +835,13 @@ void EnFall_FireRing_Draw(Actor* thisx, GlobalContext* globalCtx) {
             this->fireRingAlpha = 1.0f;
         }
         OPEN_DISPS(globalCtx->state.gfxCtx);
```
EllipticEllipsis, line 837: I like spaces between the DISPS macros and their content.

EllipticEllipsis (commented):
Great job, my comments are mostly going to be a final pass over the actor. There's a few I can't get into the lines shown:
- L357: 0.033333335f is 1/30.0f, I think. There's another one later.
- L473: `EnFall_DoNothing` is only used by the moon's tear version, so I think can be named as such.
- L745ff: It's just moved to be at most 9000 units away, along the same line of sight, I don't really feel like this is worth calling LERP.

assets/xml/objects/object_fall.xml
tom-overton: This is where we have to make a decision, because we are split exactly 50/50 in the codebase right now. Bigslime uses _tlut, and gmo uses _palette. All my PRs use _palette too. I got mine from OoT, which upon closer inspection is tilted in favor of "_tlut". 83 uses of "_tlut" vs. 21 uses of "_pal".

assets/xml/objects/object_fall.xml
tom-overton: Do you mean the word "material"? It came from here: https://github.com/zeldaret/oot/pull/985
"Material" means "this DL sets the texture and other stuff, but has no geometry" and "model" means "this DL has the geometry".

assets/xml/objects/object_fall.xml
tom-overton: As we discussed in discord, we decided to use _tlut for everything.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
tom-overton: Can you clarify what you mean by "these"? I'm not sure I'm quite following what you mean here.

tom-overton:
RE: the lerp stuff, we talked about it here: https://discord.com/channels/688807550715560050/688851976746041391/911742721448878161
It's definitely doing lerp calculations, even if I can't use the macro because of float registers. I'm not sure what else I'd call it.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
EllipticEllipsis: I mean the texscrolls: they're part of drawing these bits, but not all of them.

tom-overton: Oh, I think I get what you mean. You're basically saying that these comments make it look like the TwoTexScrolls are actually *drawing* the "glowing sphere of fire", when it's actually just part of drawing it, right? Let me update the comments to be a little more specific.

EllipticEllipsis: Yeah, it could be interpreted as being the whole drawing operation rather than just part of it, exactly.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -759,67 +807,74 @@ void EnFall_LodMoon_DrawWithLerp(Actor* thisx, GlobalContext* globalCtx) {
...
+                    (s32)((70.0f * (1.0f - this->fireballIntensity)) + (255.0f * this->fireballIntensity)),
+                    (s32)(70.0f * (1.0f - this->fireballIntensity)), 255);
     gDPSetEnvColor(
-        POLY_XLU_DISP++, (s32)(((1.0f - this->fireBallIntensity) * 50.0f) + (200.0f * this->fireBallIntensity)),
-        (s32)(20.0f * (1.0f - this->fireBallIntensity)), (s32)(20.0f * (1.0f - this->fireBallIntensity)), 255);
+        POLY_XLU_DISP++, (s32)(((1.0f - this->fireballIntensity) * 50.0f) + (200.0f * this->fireballIntensity)),
+        (s32)(20.0f * (1.0f - this->fireballIntensity)), (s32)(20.0f * (1.0f - this->fireballIntensity)), 255);
 
-    // Glowing sphere of fire
+    // Handles the texture scrolling for the glowing sphere of fire
```
EllipticEllipsis, line 835:
```suggestion
    // For the glowing sphere of fire
```
I think this would probably be sufficient.

src/overlays/actors/ovl_En_Fall/z_en_fall.c
```diff
@@ -393,14 +425,17 @@ void EnFall_StoppedClosedMouthMoon_PerformCutsceneActions(EnFall* this, GlobalCo
...
                         Audio_PlayActorSound2(&this->actor, NA_SE_EN_MOON_SCREAM1);
                         break;
+
                     case 590:
                         Audio_PlayActorSound2(&this->actor, NA_SE_EV_MOON_CRY);
                         break;
+
                     case 737:
                         Audio_PlayActorSound2(&this->actor, NA_SE_EV_SLIP_MOON);
                         break;
```
EllipticEllipsis, line 441: I think it's probably okay to not have newlines in switches where the cases are this small, but perhaps let's just standardise to always.

tom-overton: I'll just leave them as-is for now. I don't think they help readability in this case, but they don't hurt either. Unless someone disagrees with that, I'll just keep them the way they are to standardize things.

EllipticEllipsis (approved):

jpburnett (approved):
Nice work! 🌕

Kenix3 (approved):
Great formatting on your XMLs.

PR #559, opened 2021-12-31
Document object_boss03

tom-overton:
---

Don't merge this until @AngheloAlf approves it, since they have Gyorg's actor reserved.

This documents the entirety of object_boss03, which constitutes a variety of different things. There's also a ton of unused junk in there, which I have done my best to name, but please let me know if anything could be improved.

I also updated Tanron3 in a variety of ways:
- Used the new object symbols
- Removed some stuff from the .h files that other actors don't need to care about
- Added an enum for limbs
- Used that limb enum to correct some variable names. Whoops!
- Other fixes to enhance stability and create a more pleasant decomp experience


assets/xml/objects/object_boss03.xml
```diff
@@ -1,93 +1,135 @@
 ﻿<Root>
+    <!-- 
+        Gyorg, his title card, the small fishes he summons, the seaweed in his boss arena,
+        and the bubbles in his intro cutscene. Also contains an unused copy of Majora's Wrath's
+        whip assets, along with an unused duplicate of the small fishes.
+    -->
     <File Name="object_boss03" Segment="6">
```
AngheloAlf, line 7: Nice work with the object.
There's a detail tho. Lately we are trying to preserve the order of the offsets of each extracted asset.
So, could you move those back to the original order this file had?

AngheloAlf (commented):
Looks good.
Just one thing

assets/xml/objects/object_boss03.xml
tom-overton: Done, but can you please point to where this discussion happened? This is literally the first I've heard of it. I went and updated all my other asset PRs to preserve the order of the original assets, but some of those PRs were already approved, giving me the impression that this practice was supported.

AngheloAlf: I couldn't find where the discussion happened (I think it happened somewhere in OoT). 
The main reason for this is objects having file splits themselves (those can be seen by the `possiblePadding` ZAPD generates). There's the intention to split those files, and ZAPD is currently able to do that, but we haven't decided yet if we want to do that.

If you are curious of how an splitted object file looks like, you could take a look at the object_human PR in OoT: https://github.com/zeldaret/oot/pull/1004/files

AngheloAlf (approved):

assets/xml/objects/object_boss03.xml
tom-overton: Ahh, I see. I've been trying to follow MM conversations as much as possible, so I thought it strange that I had somehow missed something; turns out, it was an OoT discussion. And yeah, that makes total sense. Thanks for clarifying!

assets/xml/objects/object_boss03.xml
```diff
@@ -119,18 +119,18 @@
...
-        <Limb Name="gGyorgSmallFishHeadLimb" Type="Standard" Offset="0xD9F0" />
+        <Limb Name="gGyorgSmallFishRootLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_ROOT" Offset="0xD990" />
+        <Limb Name="gGyorgSmallFishBodyRootLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_BODY_ROOT" Offset="0xD99C" />
+        <Limb Name="gGyorgSmallFishTrunkRootLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_TRUNK_ROOT" Offset="0xD9A8" />
+        <Limb Name="gGyorgSmallFishTailFinLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_TAIL_FIN" Offset="0xD9B4" />
+        <Limb Name="gGyorgSmallFishTrunkLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_TRUNK" Offset="0xD9C0" />
+        <Limb Name="gGyorgSmallFishLeftFinLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_LEFT_FIN" Offset="0xD9CC" />
+        <Limb Name="gGyorgSmallFishDorsalFinLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_DORSAL_FIN" Offset="0xD9D8" />
+        <Limb Name="gGyorgSmallFishRightFinLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_RIGHT_FIN" Offset="0xD9E4" />
+        <Limb Name="gGyorgSmallFishHeadLimb" Type="Standard" EnumName="EN_TANRON3_LIMB_HEAD" Offset="0xD9F0" />
```
AngheloAlf, line 130: As a personal preference I think it would be better to name the limbs after the object, not the actor.
So, either `BOSS03_LIMB_X` (in the general case) or `GYORG_SMALLFISH_LIMB_X` since that's how you had named the other limbs

AngheloAlf (approved):

hensldm (approved):

Kenix3 (approved):
Another great example of xml formatting.

PR #574, opened 2022-01-05
Document the boulder fragments in gameplay_keep and clean up En_Baguo

tom-overton:
---

This PR brings Nejiron up to "modern" standards since it's a bit crusty (early `return`s, has some enums only the actor cares about in the header file, wrong define used to compare against `atFlags`, etc). It also documents the boulder fragment DList in gameplay_keep, which I believe only Nejiron uses. Goroiwa used it in OoT, but in MM the fragments are contained in object_goroiwa instead. From MM3D, this thing is apparently called "ddan_hahen/model/stone3_model".


AngheloAlf (approved):
Jenkins is mad because of misformatted files

tom-overton:
Fixed, sorry about that!

AngheloAlf (approved):

jpburnett (approved):

Kenix3 (approved):

PR #576, opened 2022-01-05
Clean up z_door_ana

mzxrules:
---
Improves DoorAna_WaitOpen code by introducing a second scoped temp var, and fixes name issues that were never addressed.


AngheloAlf (approved):

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
```diff
@@ -63,11 +63,11 @@ void DoorAna_SetupAction(DoorAna* this, DoorAnaActionFunc actionFunction) {
 
 void DoorAna_Init(Actor* thisx, GlobalContext* globalCtx) {
     DoorAna* this = THIS;
-    s32 grottoType = GET_DOORANA_TYPE(this);
+    s32 grottoType = DOORANA_GET_TYPE(this);
```
hensldm, line 66:
```suggestion
    s32 grottoType = DOORANA_GET_TYPE(this);
    
```

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
```diff
@@ -15,18 +15,17 @@ typedef struct DoorAna {
...
+#define DOORANA_TYPE_HIDDEN_BOMB        0x200
+#define DOORANA_TYPE_VISIBLE_SCENE_EXIT 0x300 // Destination dependent on scene exit list
+#define DOORANA_TYPE_BITRANGE           0x300
+
+#define DOORANA_GET_TYPE(this)(this->actor.params & 0x300)
+
+#define DOORANA_GET_ENTRANCE(this) ((((this)->actor.params >> 12) & 0x7) - 1) 
+#define DOORANA_GET_EX_ENTRANCE(this) (this->actor.home.rot.z + 1)
+#define DOORANA_GET_EXIT_INDEX(this) ((this)->actor.params & 0x1F)
+#define DOORANA_GET_ITEMFLAGS(this) ((this)->actor.params & 0xFF)
```
hensldm, line 29:
```suggestion
#define DOORANA_GET_TYPE(thisx)((thisx)->params & 0x300)
#define DOORANA_GET_ENTRANCE(thisx) ((((thisx)->params >> 12) & 0x7) - 1) 
#define DOORANA_GET_EX_ENTRANCE(thisx) ((thisx)->home.rot.z + 1)
#define DOORANA_GET_EXIT_INDEX(thisx) ((thisx)->params & 0x1F)
#define DOORANA_GET_ITEMFLAGS(thisx) ((thisx)->params & 0xFF) 
```
The extra newline seems strange. Also should change these macros to take the base actor pointer.

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
```diff
@@ -15,18 +15,17 @@ typedef struct DoorAna {
...
-#define DOORANA_TYPE_ROTATION_ENTRANCE(directEntrance)(directEntrance < 0)
-#define GET_DOORANA_DIRECT_ENTRANCE(this) ((((this)->actor.params >> 0xC) & 0x7) - 1) 
-#define GET_DOORANA_ADJACENT_ENTRANCE(this) ((this)->actor.params & 0x1F) 
-#define GET_DOORANA_ROTATION_ENTRANCE(this) (this->actor.home.rot.z + 1) 
-#define GET_DOORANA_ITEMFLAGS(this) ((this)->actor.params & 0xFF) 
+#define DOORANA_TYPE_VISIBLE            0x000
+#define DOORANA_TYPE_HIDDEN_STORMS      0x100 // Unused, leftover from OoT 
+#define DOORANA_TYPE_HIDDEN_BOMB        0x200
+#define DOORANA_TYPE_VISIBLE_SCENE_EXIT 0x300 // Destination dependent on scene exit list
+#define DOORANA_TYPE_BITRANGE           0x300
```
hensldm, line 22:
```suggestion
#define DOORANA_TYPE_HIDDEN           DOORANA_TYPE_HIDDEN_STORMS | DOORANA_TYPE_HIDDEN_BOMB
```
Based on its usage, I think this makes more sense.

hensldm (commented):
A few things

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
mzxrules: strongly disagree. `DOORANA_TYPE` stores an enumeration, not a set of bitflags, and 0x0300 maps to `DOORANA_TYPE_VISIBLE_SCENE_EXIT` which shouldn't be called `DOORANA_TYPE_HIDDEN`.

The true purpose of the code that uses `DOORANA_TYPE_BITRANGE` is to force the type to become `DOORANA_TYPE_VISIBLE` in a fast but unconventional way.

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
mzxrules: why should it take the base actor?

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
hensldm: I agree. The 0x300 should map to `DOORANA_TYPE_VISIBLE_SCENE_EXIT` and should stay. My suggestion was meant to just apply to the extra 0x300  `DOORANA_TYPE_BITRANGE`. 

I feel that renaming `DOORANA_TYPE_BITRANGE` to `DOORANA_TYPE_HIDDEN` and using the or of the hidden types shows that is changing to `DOORANA_TYPE_VISIBLE` in clear way. Just using the word bitrange gives no information at all. Though looking at it now, oring the hidden types, probably doesn't make sense since as you said the types really should be an enum not bit flags. 

I still think `DOORANA_TYPE_BITRANGE` should be renamed to `DOORANA_TYPE_HIDDEN`, possibly with a small comment that 0x300 encapsulates all the hidden types.

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
hensldm: Currently our style is for these macro params to take the base actor pointer, which most, if not all, other actors already follow.

src/overlays/actors/ovl_Door_Ana/z_door_ana.c
hensldm: This seems a little unclear, as I didn't grab the next line, but this should just add a newline to seperate variable declarations from code. I.e, it would become 
```c
    DoorAna* this = THIS;
    s32 grottoType = DOORANA_GET_TYPE(this);

    this->actor.shape.rot.y = this->actor.shape.rot.z = 0;
```

src/overlays/actors/ovl_Door_Ana/z_door_ana.h
mzxrules: I opted for a different implementation that should clear things up more.

`DOORANA_TYPE_BITRANGE` is meant to be used as the bitmask for the "Type" property within the grotto's params, so I've renamed it as such and added it to DOORANA_GET_TYPE(), as well as implemented a `DOORANA_SET_TYPE` macro.

hensldm (approved):

jpburnett (approved):

Kenix3 (approved):

PR #577, opened 2022-01-05
contramread OK

EllipticEllipsis:
---

I did not rename the structs or variables in the header, but adapted the function to work with the current ones, since @Thar0 will be cleaning up libultra once it's all decompiled anyway. (And after this and contramwrite, it's just voice left)

### DIFF
--- a/include/variables.h
+++ b/include/variables.h
@@ -79,7 +79,7 @@ extern float D_80097F90;
 // extern UNK_TYPE1 D_80097FA0;
 // extern UNK_TYPE1 D_80097FA4;
 // extern UNK_TYPE1 D_80097FA5;
-// extern UNK_TYPE4 D_80097FB0;
+extern UNK_TYPE4 __osPfsLastChannel;
 extern OSViMode osViModeNtscLan1;
 extern OSViMode osViModeMpalLan1;
 // extern __OSViContext D_80098060[2];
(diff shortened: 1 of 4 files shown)


include/variables.h
```diff
@@ -79,7 +79,7 @@ extern float D_80097F90;
 // extern UNK_TYPE1 D_80097FA0;
 // extern UNK_TYPE1 D_80097FA4;
 // extern UNK_TYPE1 D_80097FA5;
-// extern UNK_TYPE4 D_80097FB0;
+extern UNK_TYPE4 __osPfsLastChannel;
```
AngheloAlf, line 82: You know the type of this variable
```suggestion
extern s32 __osPfsLastChannel;
```

src/libultra/io/contramread.c
```diff
@@ -1,3 +1,63 @@
 #include "global.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/boot/contramread/__osContRamRead.s")
+#define BLOCKSIZE 32
+
+s32 __osPfsLastChannel = -1;
+
+s32 __osContRamRead(OSMesgQueue* ctrlrqueue, s32 channel, u16 addr, u8* data) {
```
AngheloAlf, line 7: Any reason why use `u8*` for `data` instead of `void*`?

EllipticEllipsis: It's used for a crc thing in the other file, which takes a byte array, so I'll stick with u8 and rename it to buffer.

AngheloAlf (approved):

hensldm (approved):

Kenix3 (approved):
Nice job!

PR #604, opened 2022-01-13
Document object_rd, document En_Rd and En_Railgibud, and clean up En_Talk_Gibud

tom-overton:
---

This PR touches basically every Redead/Gibdo-esque thing in this game. It documents the object that Rd, Railgibud, and Talk_Gibud all share. It also documents Rd and Railgibud, which were previously completely undocumented. Lastly, it cleans up Talk_Gibud and makes its consistent with the other Redead/Gibdo variations.

En_Rd is quite a bit different in implementation compared to Railgibud and Talk_Gibud; it's more of a copy-paste job from Ocarina of Time. Whenever possible, I tried to make sure that all three shared the same names for functions that accomplish the same thing. For Railgibud and Talk_Gibud, which are *very* similar, I tried to make sure that even the actual function bodies were as close as possible while still maintaining matching; it's clear that one of them started life as a copypasta of the other.


AngheloAlf (approved):
Nice work in the documentation!

hensldm (approved):

Kenix3 (approved):
Top notch work on this documentation.

PR #605, opened 2022-01-13
Add comment to Arrow_Ice xml and update DL names

tom-overton:
---

This PR does two simple things:
- Adds a comment to the top of the ovl_Arrow_Ice XML
- Updates the names of the display lists to use the "material"/"model" naming scheme like OoT now does from: https://github.com/zeldaret/oot/pull/985

### DIFF
--- a/assets/xml/overlays/ovl_Arrow_Ice.xml
+++ b/assets/xml/overlays/ovl_Arrow_Ice.xml
@@ -1,11 +1,12 @@
 <Root>
+    <!-- Ice Arrow assets -->
     <File Name="ovl_Arrow_Ice" BaseAddress="0x80922430" RangeStart="0x980" RangeEnd="0x1DA0">
         <Texture Name="sIceArrow1Tex" OutName="ice_tex_1" Format="ia8" Width="32" Height="64" Offset="0x980"/>
         <Texture Name="sIceArrow2Tex" OutName="ice_tex_2" Format="ia8" Width="32" Height="64" Offset="0x1180"/>
         <Array Name="sIceArrowVtx" Count="43" Offset="0x1980">
             <Vtx/>
         </Array>
-        <DList Name="sIceArrowDL" Offset="0x1C30"/>
-        <DList Name="sIceArrowVtxDL" Offset="0x1CE0"/>
+        <DList Name="sMaterialDL" Offset="0x1C30"/>
+        <DList Name="sModelDL" Offset="0x1CE0"/>
     </File>
 </Root>
--- a/assets/xml/overlays/ovl_Arrow_Ice.xml
+++ b/assets/xml/overlays/ovl_Arrow_Ice.xml
@@ -1,12 +1,12 @@
 <Root>
     <!-- Ice Arrow assets -->
     <File Name="ovl_Arrow_Ice" BaseAddress="0x80922430" RangeStart="0x980" RangeEnd="0x1DA0">
-        <Texture Name="sIceArrow1Tex" OutName="ice_tex_1" Format="ia8" Width="32" Height="64" Offset="0x980"/>
-        <Texture Name="sIceArrow2Tex" OutName="ice_tex_2" Format="ia8" Width="32" Height="64" Offset="0x1180"/>
-        <Array Name="sIceArrowVtx" Count="43" Offset="0x1980">
+        <Texture Name="gIceArrow1Tex" OutName="ice_arrow_tex_1" Format="ia8" Width="32" Height="64" Offset="0x980"/>
+        <Texture Name="gIceArrow2Tex" OutName="ice_arrow_tex_2" Format="ia8" Width="32" Height="64" Offset="0x1180"/>
+        <Array Name="gIceArrowVtx" Count="43" Offset="0x1980">
             <Vtx/>
         </Array>
-        <DList Name="sMaterialDL" Offset="0x1C30"/>
-        <DList Name="sModelDL" Offset="0x1CE0"/>
+        <DList Name="gIceArrowMaterialDL" Offset="0x1C30"/>
+        <DList Name="gIceArrowModelDL" Offset="0x1CE0"/>
     </File>
 </Root>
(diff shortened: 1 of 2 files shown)


assets/xml/overlays/ovl_Arrow_Ice.xml
```diff
@@ -1,11 +1,12 @@
...
     <File Name="ovl_Arrow_Ice" BaseAddress="0x80922430" RangeStart="0x980" RangeEnd="0x1DA0">
         <Texture Name="sIceArrow1Tex" OutName="ice_tex_1" Format="ia8" Width="32" Height="64" Offset="0x980"/>
         <Texture Name="sIceArrow2Tex" OutName="ice_tex_2" Format="ia8" Width="32" Height="64" Offset="0x1180"/>
         <Array Name="sIceArrowVtx" Count="43" Offset="0x1980">
             <Vtx/>
         </Array>
-        <DList Name="sIceArrowDL" Offset="0x1C30"/>
-        <DList Name="sIceArrowVtxDL" Offset="0x1CE0"/>
+        <DList Name="sMaterialDL" Offset="0x1C30"/>
+        <DList Name="sModelDL" Offset="0x1CE0"/>
```
AngheloAlf, line 10: I personally don't like dropping the overlay's name from the variable's names.
Also, this kind of naming doesn't follow our naming scheme for extracted assets.

tom-overton: Since this asset is identical to OoT, I would really strongly prefer to use the same name for it as OoT, which is what I used. Is the problem that it isn't something like `sIceArrowMaterialDL` and `sIceArrowModelDL`?

AngheloAlf: Yeah, that naming scheme is preferred.
The problem with dropping the overlay name from the variable is it could cause name collisions with other variables since this can be a common name. OoT works around that problem by marking assets extracted from overlays as `static` which we don't do, mainly because we don't think marking extracted assets as static makes much sense

tom-overton: Ah, got it. In this instance, I'll relent and update the names. But I still really strongly believe that assets (or, at the very least, DLists that get used in an overlay) that are the same between the two games should be named the same thing. We know from OoT3D/MM3D file names that that's what Nintendo/Grezzo did (check OoT3D's zelda_dog.zar vs. MM3D's zelda2_dog.gar.lzs for example), and even if they *didn't* do that, it just makes good sense to do it ourselves. But this PR is neither the time or place to debate this sort of thing, so I'll switch to `gIceArrowMaterialDL` and `gIceArrowModelDL`.

AngheloAlf (approved):

assets/xml/overlays/ovl_Arrow_Ice.xml
```diff
@@ -1,11 +1,12 @@
 <Root>
+    <!-- Ice Arrow assets -->
     <File Name="ovl_Arrow_Ice" BaseAddress="0x80922430" RangeStart="0x980" RangeEnd="0x1DA0">
-        <Texture Name="sIceArrow1Tex" OutName="ice_tex_1" Format="ia8" Width="32" Height="64" Offset="0x980"/>
-        <Texture Name="sIceArrow2Tex" OutName="ice_tex_2" Format="ia8" Width="32" Height="64" Offset="0x1180"/>
-        <Array Name="sIceArrowVtx" Count="43" Offset="0x1980">
+        <Texture Name="gIceArrow1Tex" OutName="ice_arrow_tex_1" Format="ia8" Width="32" Height="64" Offset="0x980"/>
+        <Texture Name="gIceArrow2Tex" OutName="ice_arrow_tex_2" Format="ia8" Width="32" Height="64" Offset="0x1180"/>
```
engineer124, line 5: What differentiates these two textures? Is there something more meaningful other than `1` and `2` that could be used? Or are both textures similar and generic?

assets/xml/overlays/ovl_Arrow_Ice.xml
```diff
@@ -1,11 +1,12 @@
...
+        <Array Name="gIceArrowVtx" Count="43" Offset="0x1980">
             <Vtx/>
         </Array>
-        <DList Name="sIceArrowDL" Offset="0x1C30"/>
-        <DList Name="sIceArrowVtxDL" Offset="0x1CE0"/>
+        <DList Name="gIceArrowMaterialDL" Offset="0x1C30"/>
+        <DList Name="gIceArrowModelDL" Offset="0x1CE0"/>
```
engineer124, line 10: Just to add to the conversation about the names being the same between OoT and MM. I too believe it's better when both names between the games are the same. But if there's a better way to name something, then using the better name in MM and seeing if OoT will change it later is also a valid option. I think having the prefix `gIceArrow` would be better for both games.

assets/xml/overlays/ovl_Arrow_Ice.xml
tom-overton: Good question. Below are three screenshots: the original Ice Arrow, the Ice Arrow with ice_arrow_tex_1 turned into a smiley face, and the Ice Arrow with ice_arrow_tex_2 turned into a smiley face:
image
image
image

So it looks like the first texture is probably what most would call "the ice arrow texture"; it contains the things that sort of look like ice crystals. The second texture is basically an intensity mask on top of the first one. So maybe `gIceArrowTex` and `gIceArrowMaskTex`?

engineer124: Nice, that seems reasonable to me. Do you happen to have what the actual png `gIceArrowMaskTex` file looks like? You should be able to grab it after ZAPD extraction

tom-overton: Yeah, sure. This is `gIceArrowTex`:
ice_arrow_tex_1 ia8

And this is `gIceArrowMaskTex`:
ice_arrow_tex_2 ia8

assets/xml/overlays/ovl_Arrow_Ice.xml
tom-overton: Well, for OoT, if it's static then it should probably be `sIceArrow` as the prefix, no? Which sort of goes back to the question of "why do OoT and MM handle in-overlay assets differently?" But that's something way beyond the scope of this tiny PR.

hensldm (approved):

Kenix3 (approved):

PR #614, opened 2022-01-15
Clean up En_Giant header

tom-overton:
---

For some reason, I missed cleaning up the header in the last round of Giant cleanup. I only noticed that it was wrong when it was pointed out in the review for Demoheishi. This PR moves things out of the header and into the C file if they aren't strictly needed in the header. Note that I'm not actually sure if `GiantCsActionIndex` belongs in the header or in the C file; would cutscene data eventually use it? But for now, I just moved it to the C file, since nobody *at the moment* is using it but Giant, and it's ambiguous where it belongs.


EllipticEllipsis (commented):
I think my only issue is that I'd rather have the enums and defines next to one another. And possibly below the function prototypes rather than above, although we should discuss.

tom-overton:
I put the defines next to the `FLAGS` define at the top so that all defines were together; the `FLAGS` define is above the function prototypes (and should probably stay that way). So we could either split up the `FLAGS` define from the rest of the defines, or we can move the enums above the function prototypes. The latter is what Door_Spiral does.

tom-overton:
I decided to move the defines down to be with the enums. It still looks a *bit* strange to have these defines separated from the `FLAGS` and `THIS` defines, but it's nice to have all the giant-related things in one place. If we prefer the old way I did it, I can always revert.

AngheloAlf (approved):

hensldm (approved):

Kenix3 (approved):

PR #623, opened 2022-01-17
Door_Shutter

Kelebek1:
---


src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -27,64 +50,664 @@ const ActorInit Door_Shutter_InitVars = {
...
+
+        if (this->unk_164 == 7) {
+            Matrix_InsertTranslation(0.0f, 64.96f, 0.0f, MTXMODE_APPLY);
+            Matrix_InsertZRotation_s(this->actor.home.rot.z, MTXMODE_APPLY);
+            Matrix_InsertTranslation(0.0f, -64.96f, 0.0f, MTXMODE_APPLY);
+        }
+
+        if (sp44->unk_04 != 0) {
+            TransitionActorEntry* transitionEntry =
+                &globalCtx->doorCtx.transitionActorList[(u16)this->actor.params >> 0xA];
```
AngheloAlf, line 679: Make a macro for this too

jpburnett (approved):

include/macros.h
```diff
@@ -74,6 +74,8 @@
 #define CUR_FORM_EQUIP(button) (gSaveContext.equips.buttonItems[CUR_FORM][button])
 #define CHECK_QUEST_ITEM(item) (((void)0, gSaveContext.inventory.questItems) & gBitFlags[item])
 #define REMOVE_QUEST_ITEM(item) (gSaveContext.inventory.questItems = (((void)0, gSaveContext.inventory.questItems) & (-1 - gBitFlags[item])))
+#define CHECK_DUNGEON_ITEM(item) (gSaveContext.inventory.dungeonItems[(void)0, gSaveContext.mapIndex] & gBitFlags[item])
```
AngheloAlf, line 77: This caught my curiosity.
OoT defines this macro as
```
#define CHECK_DUNGEON_ITEM(item, dungeonIndex) (gSaveContext.inventory.dungeonItems[dungeonIndex] & gBitFlags[item])
```
(requiring the `dungeonIndex` parameter explicitly).
I tried to change the OoT version if this macro to this one, but it didn't match. The specific file I had problems with was `z_map_exp`.
I think it would be better to use the OoT version of this macro for now, at least until we have the files that use this macro decomped so we can properly test it here (we know some macros changed between OoT and MM).

Kenix3 (approved):

PR #633, opened 2022-01-22
Add autogenerated limb enum info in the object's XMLs

AngheloAlf:
---

Adds autogenerated limb enum info to the XMLs, so ZAPD can generate enums for almost every extracted skeleton.
The autogenerated names have the format `OBJECTNAME_LIMB_XX`, where `XX` is the index of the enum in hex. For objects with multiples skeletons, I named them like `OBJECTNAME_Y_LIMB_XX`, where `Y` is just a number for the skeleton from 1 onwards.
This PR does not add autogenerated limb enum info for limbs that are already named. It didn't felt right to give an autogenerated name to those.


assets/xml/objects/object_cne.xml
```diff
@@ -16,7 +16,7 @@
         <!-- <Limb Name="object_cne_Standardlimb_000090" Type="Standard" Offset="0x90" /> -->
         <!-- <Limb Name="object_cne_Standardlimb_00009C" Type="Standard" Offset="0x9C" /> -->
         <!-- <Limb Name="object_cne_Standardlimb_0000A8" Type="Standard" Offset="0xA8" /> -->
-        <!-- <Skeleton Name="object_cne_Skel_0000F0" Type="Flex" LimbType="Standard" Offset="0xF0" /> -->
+        <!-- <Skeleton Name="object_cne_Skel_0000F0" Type="Flex" LimbType="Standard" LimbNone="_LIMB_NONE" LimbMax="_LIMB_MAX" EnumName="_Limbs" Offset="0xF0" /> -->
```
tom-overton, line 19: Is this intended? I'm not sure what ZAPD would do with this assuming this was uncommented. I suppose it doesn't *really* matter since, whenever the unaligned textures issue is solved, someone is going to have to submit a PR to uncomment things in this file, and in doing so they can fix this enum.

AngheloAlf: No, it isn't.
I guess I can fix this one manually

tom-overton (approved):

assets/xml/objects/object_ah.xml
```diff
@@ -37,22 +37,22 @@
...
+        <Limb Name="object_ah_Standardlimb_009DD0" Type="Standard" EnumName="OBJECT_AH_LIMB_09" Offset="0x9DD0" />
+        <Limb Name="object_ah_Standardlimb_009DDC" Type="Standard" EnumName="OBJECT_AH_LIMB_0A" Offset="0x9DDC" />
+        <Limb Name="object_ah_Standardlimb_009DE8" Type="Standard" EnumName="OBJECT_AH_LIMB_0B" Offset="0x9DE8" />
+        <Limb Name="object_ah_Standardlimb_009DF4" Type="Standard" EnumName="OBJECT_AH_LIMB_0C" Offset="0x9DF4" />
+        <Limb Name="object_ah_Standardlimb_009E00" Type="Standard" EnumName="OBJECT_AH_LIMB_0D" Offset="0x9E00" />
+        <Limb Name="object_ah_Standardlimb_009E0C" Type="Standard" EnumName="OBJECT_AH_LIMB_0E" Offset="0x9E0C" />
+        <Limb Name="object_ah_Standardlimb_009E18" Type="Standard" EnumName="OBJECT_AH_LIMB_0F" Offset="0x9E18" />
+        <Limb Name="object_ah_Standardlimb_009E24" Type="Standard" EnumName="OBJECT_AH_LIMB_10" Offset="0x9E24" />
+        <Skeleton Name="object_ah_Skel_009E70" Type="Flex" LimbType="Standard" LimbNone="OBJECT_AH_LIMB_NONE" LimbMax="OBJECT_AH_LIMB_MAX" EnumName="object_ah_Limbs" Offset="0x9E70" />
```
hensldm, line 56: Should the `EnumName` be pascalcase to match style, i.e ObjectAhLimbs?

AngheloAlf: Yeah, definitely. I didn't bother too much with this since I was just automating it.
I think I could change it if it is preferred that way for the automated ones

AngheloAlf:
> Oh for some reason I thought the Enum would just be the limb variable name but upscaled to snake case, I didn't realize we have to name the limb and the enum value per limb

Yeah, it was requested to work this way instead of autogenerating the enum name.
iirc there were mainly two reasons for this.
- There could be some weird and annoying edge cases that could be hard to properly handle in the ZAPD side.
- In case a limb name gets renamed, it is easier this way because it would be just a global search and replace

assets/xml/objects/object_ah.xml
hensldm: Sorry forgot to respond to this. I would personally prefer that even the automated match the correct style, but if its a lot of extra work probably better to just not bother with it, since they are going to be changed anyways.

hensldm (approved):

EllipticEllipsis (approved):
I still have misgivings about doing it this way, but I don't see any reason not to approve (no doubt it'll have some conflicts to clear when the time comes, though)

Kenix3 (approved):

PR #634, opened 2022-01-22
Mir_Ray (1 non-matching)

Kelebek1:
---

MirRay_ReflectedBeam: https://decomp.me/scratch/yZ6UP


AngheloAlf (approved):

jpburnett (approved):
lgtm

src/overlays/actors/ovl_Mir_Ray/z_mir_ray.c
```diff
@@ -28,64 +52,666 @@ const ActorInit Mir_Ray_InitVars = {
...
+        OC2_NONE,
+        COLSHAPE_JNTSPH,
+    },
+    1,
+    sJntSphElementsInit,
 };
 
-// static InitChainEntry sInitChain[] = {
-static InitChainEntry D_808E3DB8[] = {
+MirRayDataEntry sMirRayData[] = {
```
EllipticEllipsis, line 104: This will format much better without the trailing commas, I'm sure: it certainly did in OoT. Also, can you add a
```suggestion
// Note: this data is unchanged from OoT and hardcoded for Spirit Temple, so the actor will not work at all without modification.
MirRayDataEntry sMirRayData[] = {
```
(This isn't really a bug, just hardcoding not being dealt with properly.)

EllipticEllipsis (commented):
Looks good. Since this is basically documented already, it's probably worth adding a note now about why it's broken and unused (see suggestion).

EllipticEllipsis (approved):

Kenix3 (approved):

PR #641, opened 2022-02-05
Ovl_En_Cne_01 Ok and Documented

hensldm:
---

As part of this PR:

- Decomped and documented EnCne01 (Unused NPC from Hyrule Market)
   - Any naming suggestions for `func_809CB4A0` would be appreciated, though might need to wait for documentation of `func_800B85E0`.
- Documented object_cne (ZAPD supports unaligned textures now :) )
- Split off the EnHy specific stuff in z64.h, functions.h, and variables.h into their own z_en_hy.h header.
- Cleaned up EnHy a bit.


src/overlays/actors/ovl_En_Cne_01/z_en_cne_01.c
```diff
@@ -5,6 +5,7 @@
  */
 
 #include "z_en_cne_01.h"
+#include "objects/object_cne/object_cne.h"
 
 #define FLAGS 0x00000019
```
jpburnett, line 10:
```suggestion
#define FLAGS (ACTOR_FLAG_10 | ACTOR_FLAG_8 | ACTOR_FLAG_1)
```
I think we are starting to use this way

src/overlays/actors/ovl_En_Cne_01/z_en_cne_01.c
```diff
@@ -33,85 +34,295 @@ const ActorInit En_Cne_01_InitVars = {
...
+s32 EnCne01_OverrideLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot,
+                             Actor* thisx) {
+    EnCne01* this = THIS;
+    s8 bodyPart;
+    Vec3f zeroVec = { 0.0f, 0.0f, 0.0f };
 
-extern UNK_TYPE D_060000F0;
-extern UNK_TYPE D_06001300;
+    bodyPart = gEnHyBodyParts[limbIndex];
+    if (bodyPart >= 0) {
```
jpburnett, line 243: Could this be `bodyPart >= CNE_LIMB_NONE`?

jpburnett (commented):
Very nice, just a few minor thoughts.

src/overlays/actors/ovl_En_Cne_01/z_en_cne_01.c
hensldm: It doesn't quite make sense, as these are bodyParts not limbs, though they do have a link to each limb based on `gEnHyBodyParts`. The bodyParts could probably have their own enum, but that should probably be discussed based on all the actors that use them.

jpburnett (approved):

EllipticEllipsis (approved):
Good job on this, it looks pretty tricky to document this much more as-is. Just the one significant comment.

src/overlays/actors/ovl_En_Cne_01/z_en_cne_01.c
```diff
@@ -33,85 +34,295 @@ const ActorInit En_Cne_01_InitVars = {
...
+    return isTalking;
+}
+
+s32 func_809CB4A0(EnCne01* this, GlobalContext* globalCtx) {
+    s16 x;
+    s16 y;
+
+    Actor_GetScreenPos(globalCtx, &this->enHy.actor, &x, &y);
+    //! @bug: Both x and y conditionals are always true, || should be an &&
+    if (!this->enHy.waitingOnInit && ((x >= 0) || (x < 320)) && ((y >= 0) || (y < 240))) {
```
isghj5, line 140:
```suggestion
    if (!this->enHy.waitingOnInit && ((x >= 0) || (x < SCREEN_WIDTH)) && ((y >= 0) || (y < SCREEN_HEIGHT))) {
```
probably needs formater pass

isghj5 (commented):
So annoying this actor has to use two objects to have a working animation...

undefined_syms.txt
```diff
@@ -1437,6 +1413,7 @@ D_06003DC8 = 0x06003DC8;
 
 // ovl_En_Fishing
 
+D_0600007C = 0x0600007C;
 D_0600CFE0 = 0x0600CFE0;
 D_06011058 = 0x06011058;
```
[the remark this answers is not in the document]
hensldm: So both en_hy and en_fishing use something from there object files at offset `0x7C` (`object_aob_Anim_00007C` for en_hy and `gFishingFishAnim` for en_fishing), ie the segment pointer `0x0600007C`. The reason I had to add this back for fishing, even though it is in the xml, is that `gFishingFishAnim` is used in a NON-MATCHING function so the asm will use `D_0600007C`.

include/z_en_hy.h
```diff
@@ -0,0 +1,92 @@
+#ifndef Z_EN_HY_H
```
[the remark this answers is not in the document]
hensldm: I'd prefer en_hy being first (`z_en_hy_code`) as it fits better alphabetically in the spec.

PR #667, opened 2022-02-15
EnBba01 Ok and Documented

hensldm:
---

Here is the other actor that makes use of the `EnHy` system. Note that I also decided to append the word `code` to the `en_hy` files to hopefully help differentiate it from overlays.

Note: The skeleton `EnBba01` uses actually doesn't match most of the assumptions made by the `EnHy` system. I've noted this in a bug comment, but let me know if I should elaborate more on it.


assets/xml/objects/object_bba.xml
```diff
@@ -1,56 +1,82 @@
...
+
+         <!-- Animations -->
+        <Animation Name="gBbaIdleAnim" Offset="0x6550" />
+        <Animation Name="gBbaWalkingHoldingBagAnim" Offset="0x6B10" />
+
+        <!-- TLUT -->
+        <Texture Name="gBbaTLUT" OutName="bba_tlut" Format="rgba16" Width="16" Height="16" Offset="0x6B20" />
+
+        <!-- Unused Textures -->
+        <!-- These look to just be exact copies of the used textures-->
```
EllipticEllipsis, line 54: You could probably test this in a hex editor if you want to be sure.

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
```diff
@@ -33,84 +34,296 @@ const ActorInit En_Bba_01_InitVars = {
...
+    //! @bug: Both x and y conditionals are always true, || should be an &&
+    if (!this->enHy.waitingOnInit && ((x >= 0) || (x < SCREEN_WIDTH)) && ((y >= 0) || (y < SCREEN_HEIGHT))) {
+        func_800B85E0(&this->enHy.actor, globalCtx, 30.0f, EXCH_ITEM_2E);
+    }
+    return true;
+}
+
+void EnBba01_FinishInit(EnHy* this, GlobalContext* globalCtx) {
+    //! @bug: gBbaSkel does not match EnHy's skeleton assumptions.
+    //! Main offender is that gBbaSkel has 18 limbs, while EnHy expects 16
```
EllipticEllipsis, line 149: Are there more details worth putting here?

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
```diff
@@ -33,84 +34,296 @@ const ActorInit En_Bba_01_InitVars = {
...
+    }
+}
+
+void EnBba01_FaceFoward(EnHy* this, GlobalContext* globalCtx) {
+    this->actor.shape.rot = this->actor.world.rot;
+}
+
+void EnBba01_Talk(EnHy* this, GlobalContext* globalCtx) {
+    s16 yaw;
+    u8 talkstate;
```
EllipticEllipsis, line 174:
```suggestion
    u8 talkState;
```

EllipticEllipsis (commented):
Good job on this, couple of minor things.

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
hensldm: I could also mention that since the expected limbs is smaller, the joint and morph tables are too small so will overflow. 

The other kind of buggy aspect is that seen below in the Draw functions, the limbs don't match with what you expect. Thus for example, instead of her head and torso following Link, it is instead her right arm and bag that follow him. Though this is probably more because the whole actor seems to have been written with OOT's object_bba in mind, and this never got updated when the model was.

EllipticEllipsis: Yeah, all of those sound like appropriate to add in various places - I'd even make a general description under the comment block since this is so mis-implemented: sounds like it requires significant modification to use properly.

assets/xml/objects/object_bba.xml
hensldm: Confirmed with diffing the `inc.c`. They are the exact same.

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
hensldm: Adjusted the bug comment to now mention the overflow, also added a note to the header about the other buggy limb behavior.

EllipticEllipsis (approved):

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
```diff
@@ -33,84 +42,296 @@ const ActorInit En_Bba_01_InitVars = {
...
+    this->actor.shape.rot = this->actor.world.rot;
+}
+
+void EnBba01_Talk(EnHy* this, GlobalContext* globalCtx) {
+    s16 yaw;
+    u8 talkState;
+
+    Math_SmoothStepToS(&this->actor.shape.rot.y, this->actor.yawTowardsPlayer, 4, 0xFA0, 1);
+    talkState = Message_GetState(&globalCtx->msgCtx);
+    this->inMsgState3 = (talkState == 3) ? true : false;
```
isghj5, line 186: I'm guessing because ternary follows logic, this cannot be replaced with just the expression? (IDO isn't smart enough to know they are the same)
```
this->inMsgState3 = (talkState == 3);
```

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
```diff
@@ -33,84 +42,296 @@ const ActorInit En_Bba_01_InitVars = {
...
+                           ARRAY_COUNT(this->enHy.limbRotTableY));
+    EnHy_UpdateCollider(&this->enHy, globalCtx);
+}
+
+s32 EnBba01_TestIsTalking(EnBba01* this, GlobalContext* globalCtx) {
+    s32 isTalking = false;
+
+    if (Actor_ProcessTalkRequest(&this->enHy.actor, &globalCtx->state)) {
+        isTalking = true;
+        this->enHy.textId = 0x10B9;
```
isghj5, line 133:
```suggestion
        this->enHy.textId = 0x10B9; // invalid textId, produces empty textbox
```
Optional: I like textIds to be documented, in this case I spawned her and she produces an empty textbox that ends with a blue square conclusion icon, nothing interesting.

[Zelda Decomp Discord Video Link] https://cdn.discordapp.com/attachments/688852030403379277/935281291472035860/weird_en_bba_shadow.mp4

I just don't want people checking this actor out of curiosity to see this textId thinking it might have beta dialogue and go on a goose chase for an empty textbox.

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
```diff
@@ -33,84 +42,296 @@ const ActorInit En_Bba_01_InitVars = {
...
+    /* Thrown object  */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* Zora punch     */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* Spin attack    */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* Sword beam     */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* Normal Roll    */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* UNK_DMG_0x1B   */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* UNK_DMG_0x1C   */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* Unblockable    */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* UNK_DMG_0x1E   */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
+    /* Powder Keg     */ DMG_ENTRY(0, BBA01_DMG_EFF_NONE),
```
isghj5, line 99: Opinion: Kinda feel like we should have one generic `DMG_EFF_NONE` for all actors to use otherwise by the time we document every damage table we'll have > 400 copies of zero

isghj5 (approved):
Good to see this actor finally PR'd, just yesterday I went looking for the shadow code thinking it was already merged (since I saw the code in your branch weeks ago)

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
hensldm: Yeah, different codegen. If you take look at my SubS part 5 PR, there are two more functions that require this same structure, so it seems at least one dev used this style.

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
```diff
@@ -33,84 +42,296 @@ const ActorInit En_Bba_01_InitVars = {
...
+void EnBba01_Talk(EnHy* this, GlobalContext* globalCtx) {
+    s16 yaw;
+    u8 talkState;
+
+    Math_SmoothStepToS(&this->actor.shape.rot.y, this->actor.yawTowardsPlayer, 4, 0xFA0, 1);
+    talkState = Message_GetState(&globalCtx->msgCtx);
+    this->inMsgState3 = (talkState == 3) ? true : false;
+
+    switch (talkState) {
+        case 0:
```
Kenix3, line 189: Talk state could probably use an enum.

Kenix3 (approved):
Any ideas on the talk state having an enum? Not sure we need to hold up merge for that.

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
EllipticEllipsis: I wouldn't even bother making an enum/defines for this sort of trivial damagetable.

src/overlays/actors/ovl_En_Bba_01/z_en_bba_01.c
EllipticEllipsis: I don't think we can ascertain that until we know more about z_message.

PR #668, opened 2022-02-16
En_guard_nuts OK and Partially documented

jpburnett:
---

Fun actor to work with. I attempted to document what I could. object_dnk is documented, I am working on documenting look_nuts, so in terms of conflicts, this would probably be best to get in first, if I end up submitting look_nuts in the near future.


src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.c
```diff
@@ -28,42 +36,338 @@ const ActorInit En_Guard_Nuts_InitVars = {
...
+    if (D_80ABBE20 == 1) {
+        if (this->animIndex != WALK_ANIM) {
+            EnGuardNuts_ChangeAnim(this, WALK_ANIM);
+        }
+        phi_a1 = (this->actor.home.rot.y + 0x8000);
+    }
+    if (fabsf((f32)(this->actor.shape.rot.y - phi_a1)) < 100.0f) {
+        this->actor.shape.rot.y = phi_a1;
+        if ((D_80ABBE20 == 1) && (this->animIndex != WALK_ANIM_2)) {
+            EnGuardNuts_ChangeAnim(this, WALK_ANIM_2);
```
isghj5, line 172: I don't... recall the guards at the front of the palace ever walking anywhere, are they reusing the walk animation without moving for something (surprise? head shaking? anger?) is this unused code? assuming the animation gets used in look nuts so we know its the real walking animation.

src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.h
```diff
@@ -2,16 +2,38 @@
...
 
 #include "global.h"
+#include "objects/object_dnk/object_dnk.h"
 
 struct EnGuardNuts;
 
 typedef void (*EnGuardNutsActionFunc)(struct EnGuardNuts*, GlobalContext*);
 
+typedef enum {
+    /* 0 */ WAIT_ANIM,
```
isghj5, line 12: I would probably rename WAIT_ANIM to better define what it was, I was sure it was the animation for when they were hiding under the ground until I opened z64util and saw its the head cocked sideways watching you animation.

Maybe `HEAD_TILT_ANIM`? You could instead leave it with its current name but add `// Head tilted above ground animation.`?

src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.c
```diff
@@ -28,42 +36,338 @@ const ActorInit En_Guard_Nuts_InitVars = {
...
+    ActorShape_Init(&this->actor.shape, 0.0f, ActorShadow_DrawCircle, 20.0f);
+    SkelAnime_Init(globalCtx, &this->skelAnime, &gDekuPalaceGuardSkel, &gDekuPalaceGuardWaitAnim, this->jointTable,
+                   this->morphTable, OBJECT_DNK_LIMB_MAX);
+    this->actor.colChkInfo.mass = MASS_IMMOVABLE;
+    this->actor.targetMode = 1;
+    Collider_InitAndSetCylinder(globalCtx, &this->collider, &this->actor, &sCylinderInit);
+    Actor_SetScale(&this->actor, 0.015f);
+    Math_Vec3f_Copy(&this->unk228, &this->actor.world.pos);
+    this->unk21E = D_80ABBE1C;
+    D_80ABBE1C++;
```
isghj5, line 94: Hmm... looks like its counting how many guard_nuts are loaded, saving their number in order they were initialized... but its unused (by guard nuts anyway, maybe another actor would look for it)

src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.c
```diff
@@ -28,42 +36,338 @@ const ActorInit En_Guard_Nuts_InitVars = {
...
-extern ColliderCylinderInit D_80ABBDF0;
+void EnGuardNuts_Update(Actor* thisx, GlobalContext* globalCtx) {
+    EnGuardNuts* this = THIS;
+    s32 pad;
 
-extern UNK_TYPE D_06002700;
-extern UNK_TYPE D_06002A08;
+    if (this->blinkTimer == 0) {
+        this->eyeState++;
+        if (this->eyeState >= 3) {
```
isghj5, line 312: You don't have to do this, but these two back to back can often be combined into
`if (++this->eyeState >= 3) { `

Just thought I'd point that out

src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.c
jpburnett: When you finish talking with them, they rotate 180 degrees "walking" in place, that is where the walking animation comes in to play.

src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.c
```diff
@@ -28,42 +36,338 @@ const ActorInit En_Guard_Nuts_InitVars = {
...
+            this->textId = 7;
+        } else if (gSaveContext.weekEventReg[12] & 0x40) {
+            // come to see the monkey again?
+            this->textId = 4;
+        }
+    } else {
+        // No place for an outsider
+        this->textId = 3;
+    }
+    this->actor.textId = sTextIDs[this->textId];
```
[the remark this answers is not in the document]
jpburnett: Yeah I agree, even I had got confused but a better name escaped me at the moment. I do like this suggestion though, thanks!

src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.c
```diff
@@ -28,42 +36,338 @@ const ActorInit En_Guard_Nuts_InitVars = {
...
+
+    EnGuardNuts_ChangeAnim(this, DIG_ANIM);
+    Math_Vec3f_Copy(&digPos, &this->actor.world.pos);
+    digPos.y = this->actor.floorHeight;
+    EffectSsHahen_SpawnBurst(globalCtx, &digPos, 4.0f, 0, 10, 3, 15, -1, 10, NULL);
+    this->unk23A.y = 0;
+    this->actor.flags |= ACTOR_FLAG_8000000;
+    this->unk23A.x = this->unk23A.y;
+    Actor_PlaySfxAtPos(&this->actor, NA_SE_EN_NUTS_DOWN);
+    Actor_PlaySfxAtPos(&this->actor, NA_SE_EN_NUTS_UP);
```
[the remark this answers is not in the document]
jpburnett: Looking at it closer, my best guess is that it gives more of a sense of two burrowing (even though both guards play it)...so basically they might have wanted it to be noisy? haha

AngheloAlf (approved):

src/overlays/actors/ovl_En_Guard_Nuts/z_en_guard_nuts.c
isghj5: Oh uh... I meant more along the lines of the .data value being the count, and the saved value of unk21E being an identifier.

Like, in order:
[The First Actor] will init and the starting value of (0) is saved to that actor, then data is incremented to (1)
[The Second Actor] will init and the starting value will be (1) from the last actor, the unk21E will be (1) because it was incremented from the last actor, and then we increment D_80ABBE1C to 2, which will be the identifier of the next actor (except there isn't a third actor)

I mean obviously its not used, so its intentional purpose isn't really known as it wasn't fully implemented in the final game, but it looks like it was a way to specify one guard compared to the other from load order rather than params. Maybe they wanted the two guards to talk to each other and wanted to identify tweedle dee from tweedle dumb but changed their mind

isghj5 (approved):

Kenix3 (approved):

PR #679, opened 2022-02-19
En_Bom_Chu (Bombchu) OK and documented

tom-overton:
---

I deliberately took a ton of pointers from [OoT's version of this actor](https://github.com/zeldaret/oot/blob/master/src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c), since they are quite similar. Still, it's clear they did some refactoring, and when they did I did my best to name things sensibly.


undefined_syms.txt
```diff
@@ -384,7 +384,6 @@ D_040128BC = 0x040128BC;
 D_04014560 = 0x04014560;
 D_04015DB0 = 0x04015DB0;
 D_04015FA0 = 0x04015FA0;
-D_04016360 = 0x04016360;
```
EllipticEllipsis, line 387: Generally it's better to retain these in case other code needs them; we can wipe them all when gameplay_keep is completely identified.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -32,50 +37,539 @@ const ActorInit En_Bom_Chu_InitVars = {
...
+    0, 0, 0, { 250, 0, 0, 250 }, { 200, 0, 0, 130 }, { 150, 0, 0, 100 }, { 100, 0, 0, 50 }, 16,
+    0, 0, 0, { 0, 0, 0, 0 },     { 0, 0, 0, 0 },
 };
 
-#endif
+static Vec3f sBlureP1Model = { 0.0f, 7.0f, -6.0f };
+
+static Vec3f sBlureP2LeftModel = { 12.0f, 0.0f, -5.0f };
+
+static Vec3f sBlureP2RightModel = { -12.0f, 0.0f, -5.0f };
```
EllipticEllipsis, line 74: These look like they were inlined, and if you're going to make them static, you might as well.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -32,50 +37,539 @@ const ActorInit En_Bom_Chu_InitVars = {
...
-static InitChainEntry D_808F890C[] = {
+static InitChainEntry sInitChain[] = {
     ICHAIN_U8(targetMode, 2, ICHAIN_CONTINUE),
-    ICHAIN_VEC3F_DIV1000(scale, 10, ICHAIN_STOP),
+    ICHAIN_VEC3F_DIV1000(scale, 1000 * BOMBCHU_SCALE, ICHAIN_STOP),
+};
+
+static EffectBlureInit2 sBlureInit = {
+    0, 0, 0, { 250, 0, 0, 250 }, { 200, 0, 0, 130 }, { 150, 0, 0, 100 }, { 100, 0, 0, 50 }, 16,
+    0, 0, 0, { 0, 0, 0, 0 },     { 0, 0, 0, 0 },
```
EllipticEllipsis, line 67: I'm rather surprised the formatter's okay with this.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -32,50 +37,539 @@ const ActorInit En_Bom_Chu_InitVars = {
...
+    Vec3f normal;
+    Vec3f vec;
+    f32 angle;
+    f32 magnitude;
+    f32 normDotUp;
+
+    this->actor.floorPoly = floorPoly;
+
+    // This NULL check means if the player releases a Bombchu
+    // out of bounds, the game will not crash, unlike OoT.
```
EllipticEllipsis, line 114: Is there a reason for the lines being so short? I'd have thought the formatter would put a line break at the appropriate width.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -32,50 +37,539 @@ const ActorInit En_Bom_Chu_InitVars = {
...
+void EnBomChu_WaitForDeath(EnBomChu* this, GlobalContext* globalCtx) {
+    if (this->timer == 0) {
+        Actor_MarkForDeath(&this->actor);
+    }
+}
+
+/**
+ * Transform coordinates from model space to world space, according to current orientation.
+ * `posModel` is expected to already be at world scale (1/100 compared to model scale)
+ */
```
EllipticEllipsis, line 368: This isn't really correct, since model coordinates have the scale factor in them. It's more like the actor's special coordinates (z forward, x left, y up) to world. I think we settled on "actor coordinates" in the z_actor docs?

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -32,50 +37,539 @@ const ActorInit En_Bom_Chu_InitVars = {
...
+        this->axisForwards.x = (this->axisForwards.z * sin) + (cos * tempX);
+        this->axisForwards.z = (this->axisForwards.z * cos) - (sin * tempX);
+
+        tempX = this->axisUp.x;
+        this->axisUp.x = (this->axisUp.z * sin) + (cos * tempX);
+        this->axisUp.z = (this->axisUp.z * cos) - (sin * tempX);
+
+        tempX = this->axisLeft.x;
+        this->axisLeft.x = (this->axisLeft.z * sin) + (cos * tempX);
+        this->axisLeft.z = (this->axisLeft.z * cos) - (sin * tempX);
```
EllipticEllipsis, line 436: Are these multiplications really all reversed in the different addends/subtrahends?

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -32,50 +37,539 @@ const ActorInit En_Bom_Chu_InitVars = {
...
+                EnBomChu_SpawnRipplesAndSplashes(this, globalCtx, waterY, true);
+            } else {
+                EffectSsBubble_Spawn(globalCtx, &this->actor.world.pos, 0.0f, 3.0f, 15.0f, 0.25f);
+            }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Bom_Chu/func_808F7A84.s")
+            this->actor.bgCheckFlags |= 0x20;
+        } else {
+            this->actor.bgCheckFlags &= ~0x20;
+            this->actor.depthInWater = -32000.0f;
```
EllipticEllipsis, line 536: Is this a `BGCHECK_Y_MIN` or whatever?

EllipticEllipsis (commented):
Good job on this, few small things. Main thing is we need to discourage use of "model" for the actor's own coordinates in world space, as opposed to actual model space: it's just better to think of that separately, given that stuff like the rotations is totally different there.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
tom-overton: By "inlined", do you mean in the same function they're used?

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
EllipticEllipsis: Ugh, I hate it when that happens, so messy 😩

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
EllipticEllipsis: Yeah, if they're this short, static, and only used in one function, I'm pretty sure they're declared as static data in that function.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
tom-overton: Yeah I copied this straight from OoT without really understanding it much. Do you think it's still worthwhile to change the name of this when it will make it different from OoT?

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
tom-overton: Disregard what I said before, you *can* make it match with them in the same order, but that order *has to be* sin/cos first.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
tom-overton: I *could* turn the formatter off here, but (a) it doesn't seem like a super big deal in this case to be honest, and (b) I don't know how I would format this with my human brain either.

EllipticEllipsis: Nah, it doesn't matter, just looks weird.

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -110,8 +104,8 @@ s32 EnBomChu_UpdateFloorPoly(EnBomChu* this, CollisionPoly* floorPoly, GlobalCon
 
     this->actor.floorPoly = floorPoly;
 
-    // This NULL check means if the player releases a Bombchu
-    // out of bounds, the game will not crash, unlike OoT.
+    // This NULL check means if the player releases a Bombchu out of bounds,
```
EllipticEllipsis, line 107: I don't think "out of bounds" is well-defined... you mean something like "with no collision below" or better wording.

EllipticEllipsis (approved):
But would like others' thoughts on the terminology around `EnBomChu_ActorCoordsToWorld`

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
tom-overton: Yeah I was sort of writing that from a speedrunner's perspective, which is imprecise. Let me fix that.

hensldm (approved):
`EnBomChu_ActorCoordsToWorld` seems fine to me

Kenix3 (approved):
Great documentation, very thorough.

PR #680, opened 2022-02-20
ObjectFsn documented

hensldm:
---

Also cleaned up as much of the shopkeepers as I could (was also to remove a few of the fake matches, but not all of them).


hensldm:
Forgot to mention this in the PR header, but because of the bug with Fsn's skeleton, everywhere the FSN_MAX_LIMB is used I had to add 1 to it. Is this how we want to handle it or do we want to do something different like add a special define?

AngheloAlf:
Oh yikes. That's annoying. I guess a define is good workaround.

hensldm:
Just want to note, I've slightly expanded the scope of this PR by documenting the 3 gameplay keep textures used by the shopkeepers.

src/overlays/actors/ovl_En_Fsn/z_en_fsn.c
```diff
@@ -579,8 +602,9 @@ void EnFsn_UpdateCursorAnim(EnFsn* this) {
 void EnFsn_UpdateStickDirectionPromptAnim(EnFsn* this) {
     f32 arrowAnimTween = this->arrowAnimTween;
     f32 stickAnimTween = this->stickAnimTween;
+
+    // POSSIBLY FAKE
     s32 maxColor = 255;
```
AngheloAlf, line 607: I think it would be better to put the comment at the side instead of above
```suggestion
    s32 maxColor = 255; // POSSIBLY FAKE
```

AngheloAlf (approved):
Good work

src/overlays/actors/ovl_En_Fsn/z_en_fsn.c
hensldm: Applied to all four shopkeepers

jpburnett (approved):
Great work!

AngheloAlf (approved):

hensldm:
Just want to note, that I was able to remove the fake looking `trueTmp` from `EnFsn` by changing a variable type from `s32` to `u32`.

Kenix3 (approved):

PR #688, opened 2022-02-22
`z_player_lib`: Part 1

AngheloAlf:
---

I decided to split `z_player_lib` in 3 PRs. Hopefully each PR will have a decent size and should be easier for reviewers to review it.

General overview of the PR:
- More or less 1/3 of the `z_player_lib.c` raw size
- Function prototypes for every `z_player_lib.c` function which is used outside this file
- `PLAYER_STATE` macros. The macros aren't used outside of this file to minimize changes.
- Change "sword" to "meleeWeapon" for player's melee weapons

In case more context of this file is required, here is this file (almost) fully decomped: https://github.com/Decompollaborate/mm/blob/z_player_lib/src/code/z_player_lib.c


assets/xml/objects/object_link_zora.xml
```diff
@@ -56,7 +56,15 @@
...
         <DList Name="object_link_zora_DL_010D08" Offset="0x10D08" />
         <DList Name="object_link_zora_DL_0110A8" Offset="0x110A8" />
+
+        <Array Name="object_link_zora_Vtx_011210" Count="80" Offset="0x011210">
+            <Vtx/>
+        </Array>
+        <Array Name="object_link_zora_U8_011710" Count="80" Offset="0x011710">
+            <Scalar Type="x8"/>
+        </Array>
         <!-- <Blob Name="object_link_zora_Blob_011710" Size="0x50" Offset="0x11710" /> -->
```
EllipticEllipsis, line 66: Something about this blob is wrong.

include/z64.h
```diff
@@ -611,7 +611,7 @@ typedef struct {
     /* 0x214 */ u16 unk_214;
     /* 0x218 */ f32 unk_218;
     /* 0x21C */ s16 unk_21C;
-    /* 0x21E */ s16 unk_21E;
+    /* 0x21E */ s16 unk_21E; // do_action label index for the b button
```
EllipticEllipsis, line 614: Recommend renaming this if it's known.

src/overlays/actors/ovl_En_Wf/z_en_wf.c
```diff
@@ -412,7 +412,7 @@ s32 func_80990948(GlobalContext* globalCtx, EnWf* this, s16 arg2) {
     }
 
     if (func_800BE184(globalCtx, &this->actor, 100.0f, 10000, 12000, this->actor.shape.rot.y) &&
-        ((player->swordAnimation == 0x11) || ((globalCtx->gameplayFrames % 2) != 0))) {
+        ((player->meleeWeaponAnimation == 0x11) || ((globalCtx->gameplayFrames % 2) != 0))) {
```
EllipticEllipsis, line 415: These need to be changed to dec at some point, I suppose. Or actually make an enum once we've documented them, which will be easy enough for once.

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -84,7 +84,7 @@ s32 ObjOshihiki_StrongEnough(ObjOshihiki* this, GlobalContext* globalCtx) {
         return true;
     }
 
-    if (func_80124020() >= 2) {
+    if (Player_GetStrength() >= 2) {
```
EllipticEllipsis, line 87: Was there going to be an enum for this?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    OBJECT_MASK_STONE,  OBJECT_MASK_BREE,    OBJECT_MASK_BAKURETU, OBJECT_MASK_BU_SAN,  OBJECT_MASK_KYOJIN,
+    OBJECT_MASK_BOY,    OBJECT_MASK_GORON,   OBJECT_MASK_ZORA,     OBJECT_MASK_NUTS,
+};
+#endif
+
+// Load mask?
+void func_801229FC(Player* player) {
+    if (player->maskObjectLoading == 1) {
+        // TODO: check if player->maskId is unsigned
+        s16 objectId = D_801BFDA0[(u8)player->maskId - 1];
```
EllipticEllipsis, line 116: How can we check this?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+        player->maskObjectLoading++;
+    } else if (player->maskObjectLoading == 2) {
+        if (osRecvMesg(&player->maskObjectLoadQueue, NULL, OS_MESG_NOBLOCK) == 0) {
+            player->maskObjectLoading = 0;
+
+            if (player->currentMask == PLAYER_MASK_GREAT_FAIRY) {
+                s32 i;
+
+                for (i = 0; i < ARRAY_COUNT(D_801F58B0); i++) {
+                    func_80127B64(D_801F58B0[i], ARRAY_COUNT(D_801F58B0[i]), &player->bodyPartsPos[7]);
```
EllipticEllipsis, line 131: Presumably this is some sort of extra function for the animated stuff on Great Fairy's Mask. Is the plan to not name stuff like this until later player_lib PRs?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    OBJECT_MASK_JSON,   OBJECT_MASK_ROMERNY, OBJECT_MASK_ZACHO,    OBJECT_MASK_POSTHAT, OBJECT_MASK_MEOTO,
+    OBJECT_MASK_BIGELF, OBJECT_MASK_GIBUDO,  OBJECT_MASK_GERO,     OBJECT_MASK_DANCER,  OBJECT_MASK_SKJ,
+    OBJECT_MASK_STONE,  OBJECT_MASK_BREE,    OBJECT_MASK_BAKURETU, OBJECT_MASK_BU_SAN,  OBJECT_MASK_KYOJIN,
+    OBJECT_MASK_BOY,    OBJECT_MASK_GORON,   OBJECT_MASK_ZORA,     OBJECT_MASK_NUTS,
+};
+#endif
+
+// Load mask?
+void func_801229FC(Player* player) {
+    if (player->maskObjectLoading == 1) {
```
EllipticEllipsis, line 114: Looks like this is a sort of state indicator,
Not loading,
Start loading,
Finish loading
?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+}
+
+void func_80122C20(GlobalContext* globalCtx, struct_80122D44_arg1* arg1) {
+    struct_80122D44_arg1_unk_04* temp_v1 = &arg1->unk_04[0];
+    s32 phi_a1;
+    s32 i;
+
+    for (i = 0; i < ARRAY_COUNT(arg1->unk_04); i++, temp_v1++) {
+        // Can't be `temp_v1->unk_01 != 0`
+        if (temp_v1->unk_01) {
```
EllipticEllipsis, line 172: `!!`?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    /* 0x4 */ Gfx* dlist;
+} struct_801BFDD0; // size = 0x08
+
+extern struct_801BFDD0 D_801BFDD0[];
+#if 0
+struct_801BFDD0 D_801BFDD0[] = {
+    { { 180, 200, 255 }, object_link_goron_DL_00BDD8 },
+    { { 155, 0, 0 }, object_link_goron_DL_014690 },
+    { { 255, 0, 0 }, object_link_goron_DL_011AB8 },
+};
```
EllipticEllipsis, line 194: Any idea what these are? Would probably explain the next function. My guess would be punch or spikes or something.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+
+        phi_s2++;
+    }
+
+    CLOSE_DISPS(globalCtx->state.gfxCtx);
+}
+
+extern u8 D_801BFDE8[PLAYER_MASK_MAX - 1];
+#if 0
+u8 D_801BFDE8[PLAYER_MASK_MAX - 1] = {
```
EllipticEllipsis, line 234: Don't need the size on here. Is there something missing?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    ITEM_MASK_CAPTAIN,       // PLAYER_MASK_CAPTAIN
+    ITEM_MASK_STONE,         // PLAYER_MASK_STONE
+    ITEM_MASK_BREMEN,        // PLAYER_MASK_BREMEN
+    ITEM_MASK_BLAST,         // PLAYER_MASK_BLAST
+    ITEM_MASK_SCENTS,        // PLAYER_MASK_SCENTS
+    ITEM_MASK_GIANT,         // PLAYER_MASK_GIANT
+    ITEM_MASK_FIERCE_DEITY,  // PLAYER_MASK_FIERCE_DEITY
+    ITEM_MASK_GORON,         // PLAYER_MASK_GORON
+    ITEM_MASK_ZORA,          // PLAYER_MASK_ZORA
+    ITEM_MASK_DEKU,          // PLAYER_MASK_DEKU
```
EllipticEllipsis, line 258: These comments seem excessive. Could also name this data variable?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+
+u8 Player_CurMaskToItemId(GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+
+    if (player->currentMask != PLAYER_MASK_NONE) {
+        return func_80122ED8(player->currentMask - 1);
+    }
+
+    return ITEM_NONE;
+}
```
EllipticEllipsis, line 274: Seems like the first of these is MaskIdToItemId, and the second is something like GetCurMaskItemId. Also, do we want the prefix for this stuff to be PlayerLib? I think it would be helpful to disambiguate it, even though we know that almost nothing in Player can be accessed from outside it.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+            player->actor.update = func_80160B80;
+            player->actor.draw = func_80160BC0;
+            gSaveContext.equippedMask = 0;
+        }
+    }
+}
+
+extern s16 D_801BFE14[][18];
+
+// OoT's Player_SetBootData?
```
EllipticEllipsis, line 335: I think we can just name this, seems pretty clear that it's the same function. Only problem is OoT's name isn't really right for OoT, let alone for MM. `Player_SetMovementData`, perhaps. Need a way to specify that it's for walking/running.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    REG(39) = bootRegs[8];
+    REG(43) = bootRegs[9];
+    R_RUN_SPEED_LIMIT = bootRegs[10];
+    REG(68) = bootRegs[11];
+    REG(69) = bootRegs[12];
+    IREG(66) = bootRegs[13];
+    IREG(67) = bootRegs[14];
+    IREG(68) = bootRegs[15];
+    IREG(69) = bootRegs[16];
+    MREG(95) = bootRegs[17];
```
EllipticEllipsis, line 379: These are all going to need defines and enums, of course, although we need to work out what they do first. Speedrunners might know.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+
+    if (globalCtx->roomCtx.currRoom.unk3 == 2) {
+        R_RUN_SPEED_LIMIT = 500;
+    }
+
+    if ((player->actor.id == ACTOR_PLAYER) && (player->transformation == PLAYER_FORM_FIERCE_DEITY)) {
+        scale = 0.015f;
+    } else {
+        scale = 0.01f;
+    }
```
EllipticEllipsis, line 389: I'm very curious why they thought the first check was necessary 🤔

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+s32 func_80123448(GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+
+    return (player->stateFlags1 & PLAYER_STATE1_400000) &&
+           (player->transformation != PLAYER_FORM_HUMAN || (!func_80123434(player) && player->unk_730 == 0));
+}
+
+s32 func_801234B0(Player* player) {
+    return player->transformation == PLAYER_FORM_GORON || player->transformation == PLAYER_FORM_DEKU;
+}
```
EllipticEllipsis, line 425: Well, we can probably name this, at least. And I don't really like the logical operations without brackets... we don't all have C's order of operations memorised.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+
+s32 func_801234D4(GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+
+    return (player->stateFlags2 & PLAYER_STATE2_8) || player->actor.speedXZ != 0.0f ||
+           ((player->transformation != PLAYER_FORM_ZORA) && (player->stateFlags1 & PLAYER_STATE1_8000000)) ||
+           ((player->transformation == PLAYER_FORM_ZORA) && (player->stateFlags1 & PLAYER_STATE1_8000000) &&
+            (!(player->actor.bgCheckFlags & 1) || (player->currentBoots < PLAYER_BOOTS_ZORA_UNDERWATER)));
```
EllipticEllipsis, line 433: I mean, obviously this is terrible. But the last condition's the worst of all: it's literally unintelligible from this information.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+        player->actor.home.rot.y = arg2;
+        player->actor.shape.rot.y = arg2;
+        player->unk_B8C = 4;
+        player->invincibilityTimer = 20;
+
+        return true;
+    }
+
+    return false;
+}
```
EllipticEllipsis, line 465: Are these actually used differently, then, that one returns a boolean and the other does not?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+
+extern u8 D_801BFF90[PLAYER_FORM_MAX];
+#if 0
+u8 D_801BFF90[PLAYER_FORM_MAX] = {
+    PLAYER_BOOTS_FIERCE_DEITY, // PLAYER_FORM_FIERCE_DEITY
+    PLAYER_BOOTS_GORON,        // PLAYER_FORM_GORON
+    PLAYER_BOOTS_ZORA_LAND,    // PLAYER_FORM_ZORA
+    PLAYER_BOOTS_DEKU,         // PLAYER_FORM_DEKU
+    PLAYER_BOOTS_HYLIAN,       // PLAYER_FORM_HUMAN
+};
```
EllipticEllipsis, line 532: Again, doesn't really make much sense.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+} TextTriggerEntry; // size = 0x04
+
+extern TextTriggerEntry D_801BFFA0[];
+#if 0
+TextTriggerEntry D_801BFFA0[] = {
+    { 1, 0x26FC },
+    { 2, 0x26FD },
+    { 0, 0 },
+    { 2, 0x26FD },
+};
```
EllipticEllipsis, line 558: What are these textIds? We'd be a long way to understanding this if we had that information.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+                }
+            }
+        }
+    }
+}
+
+void Player_SetModels(Player* player, s32 modelGroup) {
+    u8* aux;
+
+    D_801F59E0 = player->transformation * 2;
```
EllipticEllipsis, line 590: This almost makes me wonder if these are a 3d array... didn't OoT have a PR changing these recently?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+
+    func_80123DC0(player);
+    player->unk_730 = actor;
+    player->unk_A78 = actor;
+    player->stateFlags1 |= PLAYER_STATE1_10000;
+    func_800DFD78(Play_GetCamera(globalCtx, CAM_ID_MAIN), 8, actor);
+    Camera_ChangeMode(Play_GetCamera(globalCtx, CAM_ID_MAIN), 9);
+}
+
+s32 func_80123F14(GlobalContext* globalCtx) {
```
EllipticEllipsis, line 693: Do any of these functions you changed to take globalCtx not match with state? Is that why they're changed?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+}
+
+s32 Player_IsBurningStickInRange(GlobalContext* globalCtx, Vec3f* pos, f32 xzRange, f32 yRange) {
+    Player* this = GET_PLAYER(globalCtx);
+    Vec3f diff;
+    s32 pad;
+
+    if ((this->itemActionParam == PLAYER_AP_STICK) && (this->unk_B28 != 0)) {
+        Math_Vec3f_Diff(&this->meleeWeaponInfo[0].tip, pos, &diff);
+        return ((SQ(diff.x) + SQ(diff.z)) <= SQ(xzRange)) && (0.0f <= diff.y) && (diff.y <= yRange);
```
EllipticEllipsis, line 712: Why does this one get brackets?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    }
+
+    return false;
+}
+
+u8 Player_GetStrength(void) {
+    return D_801BFF98[(void)0, gSaveContext.playerForm];
+}
+
+u8 Player_GetMask(GlobalContext* globalCtx) {
```
EllipticEllipsis, line 722: I suppose this should really be GetCurrentMask.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    Player* player = GET_PLAYER(globalCtx);
+
+    return player->currentMask;
+}
+
+void Player_RemoveMask(GlobalContext* globalCtx) {
```
EllipticEllipsis, line 728: IIRC this one is unused? If so you could comment on it.

include/functions.h
```diff
@@ -1148,7 +1148,7 @@ s32 Camera_GetInputDirYaw(Camera* camera);
 s16 Camera_GetCamDirPitch(Camera* camera);
 s16 Camera_GetCamDirYaw(Camera* camera);
 void func_800DFD04(Camera* camera, s32 arg1, s32 arg2, s32 arg3);
-// void func_800DFD78(void);
+void func_800DFD78(Camera* camera, UNK_TYPE arg1, Actor* actor);
```
engineer124, line 1151:
```suggestion
void func_800DFD78(Camera* camera, s32 paramFlag, void* value);
```
This is the correct prototype for this. In OoT/MM, this function is currently called `Camera_SetParam`, which sets a group of generic data. You can use this to set the target actor like you do in `z_player_lib`, but you can also set Vec3f `eye`/`at` data or `f32` roll/fov data as an example:

https://github.com/engineer124/mm/blob/1e8b04396788f52b6e942cb3e9f1def396c4484a/src/code/z_camera_part3.c#L993-L996

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    return 0;
+}
+
+s32 Player_GetMeleeWeaponHeld(Player* player) {
+    return Player_ActionToMeleeWeapon(player->itemActionParam);
+}
+
+s32 Player_HoldsTwoHandedWeapon(Player* player) {
+    // Relies on two-handed weapons to be contiguous
+    if ((player->itemActionParam >= PLAYER_AP_SWORD_GREAT_FAIRY) && (player->itemActionParam <= PLAYER_AP_STICK)) {
```
EllipticEllipsis, line 786: This is the sort of thing that makes me want an actual macro to refer to this range. If you have to have this comment, it tells you the code is not clear. Much better to make the code clear in the first place.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+        return true;
+    }
+
+    return false;
+}
```
EllipticEllipsis, line 791: Do we prefer doing this, or with an else?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+
+s32 func_80124278(Actor* actor, s32 arg1) {
+    s32 phi_v1 = 0;
+
+    // Fake match?
+    if ((arg1 == 1) || ((phi_v1 = arg1 - 3, (phi_v1 >= 0)) && (phi_v1 < 4))) {
+        return phi_v1;
+    }
+    return -1;
+}
```
EllipticEllipsis, line 831: I feel like I've seen this sort of thing before, and it turned out all the conditions were backwards or something.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+        distXZ = Math_Vec3f_DistXZ(&player->actor.world.pos, &sp30);
+
+        if ((fabsf(player->actor.world.pos.y - sp30.y) < 50.0f) && (distXZ < arg2)) {
+            arg1->unk_01++;
+        }
+
+        return 0;
+    }
+
+    return 1;
```
engineer124, line 65:
```suggestion
        return false;
    }

    return true;
```
This function seems to be a boolean return

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124870.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124190.s")
+#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124CC4.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_801241B4.s")
+extern s32 D_801C0958;
+#if 0
+s32 D_801C0958 = false;
+#endif
```
EllipticEllipsis, line 905: Is there a reason this is here rather than further down? I've noticed a few bits of data in odd places in here.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+extern s16 D_801BFDA0[PLAYER_MASK_MAX - 1];
+#if 0
+s16 D_801BFDA0[PLAYER_MASK_MAX - 1] = {
+    OBJECT_MASK_TRUTH,  OBJECT_MASK_KERFAY,  OBJECT_MASK_YOFUKASI, OBJECT_MASK_RABIT,   OBJECT_MASK_KI_TAN,
+    OBJECT_MASK_JSON,   OBJECT_MASK_ROMERNY, OBJECT_MASK_ZACHO,    OBJECT_MASK_POSTHAT, OBJECT_MASK_MEOTO,
+    OBJECT_MASK_BIGELF, OBJECT_MASK_GIBUDO,  OBJECT_MASK_GERO,     OBJECT_MASK_DANCER,  OBJECT_MASK_SKJ,
+    OBJECT_MASK_STONE,  OBJECT_MASK_BREE,    OBJECT_MASK_BAKURETU, OBJECT_MASK_BU_SAN,  OBJECT_MASK_KYOJIN,
+    OBJECT_MASK_BOY,    OBJECT_MASK_GORON,   OBJECT_MASK_ZORA,     OBJECT_MASK_NUTS,
+};
+#endif
```
engineer124, line 110: What's the motivation for externing this? IIRC, this was the data that was incorrectly split into `z_parameter.c`. Is there a reason that fix isn't included in this PR?
EDIT: Coming back to this, so none of the data is imported yet?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/Player_GetExplosiveHeld.s")
+void func_8012536C(void) {
+    if (D_801C0958) {
+        Matrix_StatePop();
+        D_801C0958 = false;
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124278.s")
+// draws zora shield (?)
```
EllipticEllipsis, line 979: I thought we confirmed this.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+void func_801253A4(GlobalContext* globalCtx, Player* player) {
+    u8* phi_a0;
+    Vtx* sp30;
+    Gfx* dl;
+    f32 sp28; // scale
+    s32 i;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_801242B4.s")
+    sp28 = player->unk_B62 * 0.19607843f;
```
EllipticEllipsis, line 987: I'm sure this is a nice float somehow.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    OPEN_DISPS(globalCtx->state.gfxCtx);
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124420.s")
+    AnimatedMat_DrawXlu(globalCtx, Lib_SegmentedToVirtual(&object_link_zora_Matanimheader_012A80));
+    Matrix_Scale(sp28, sp28, sp28, 1);
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124618.s")
+    // clang-format off
+    sp30 = Lib_SegmentedToVirtual(&object_link_zora_Vtx_011210), phi_a0 = Lib_SegmentedToVirtual(&object_link_zora_U8_011710);
+    // clang-format on
```
EllipticEllipsis, line 997: Is this needed even with a `,` instead of a `;`?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+            } else {
+                temp_v1->unk_01 -= phi_a1;
+            }
+        }
+    }
+}
+
+typedef struct {
+    /* 0x0 */ Color_RGB8 color;
+    /* 0x4 */ Gfx* dlist;
```
engineer124, line 185:
```suggestion
    /* 0x4 */ Gfx* dList;
```

EllipticEllipsis (commented):
Oh boy, there's a lot here already. Great work on this, I have a lot of mostly nitpicky stuff, and what looks like some easy extra documentation.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+void func_80122D44(GlobalContext* globalCtx, struct_80122D44_arg1* arg1) {
+    struct_801BFDD0* temp_s3;
+    struct_80122D44_arg1_unk_04* phi_s2 = arg1->unk_04;
+    s32 phi_s6 = false;
+    s32 i;
+
+    OPEN_DISPS(globalCtx->state.gfxCtx);
+
+    for (i = 0; i != ARRAY_COUNT(arg1->unk_04); i++) {
+        if ((phi_s2->unk_01 != 0) && (phi_s2->unk_01 != 0xFF)) {
```
engineer124, line 206:
```suggestion
        if ((phi_s2->unk_01 != 0) && (phi_s2->unk_01 != 255)) {
```
`phi_s2->unk_01` from the code below is an `envAlpha` so dec. Could maybe call this `phi_s2->envAlpha` or just `phi_s2->alpha`? Could leave it for later docs though

assets/xml/objects/object_link_zora.xml
AngheloAlf: I forgot to delete it :sweat_smile:

src/code/z_player_lib.c
engineer124: `!!` is non-equivalent because that converts all non-zero integers to `1`, unless there's some ido optimization shenenigans at play

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+}
+
+void func_8012301C(Player* player, GlobalContext* globalCtx2) {
+    GlobalContext* globalCtx = globalCtx2;
+    s32 pad;
+
+    player->unk_AE7++;
+
+    if (player->unk_AE7 == 2) {
+        s16 objectId = gLinkFormObjectIndexes[((void)0, gSaveContext.playerForm)];
```
engineer124, line 314:
```suggestion
        s16 objectId = gPlayerFormObjectIndexes[((void)0, gSaveContext.playerForm)];
```
Unless there's good motivation to use `Link` here

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+s32 func_80123434(Player* player) {
+    return player->stateFlags1 & (PLAYER_STATE1_40000000 | PLAYER_STATE1_20000 | PLAYER_STATE1_10000);
+}
+
+// Unused
+s32 func_80123448(GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+
+    return (player->stateFlags1 & PLAYER_STATE1_400000) &&
+           (player->transformation != PLAYER_FORM_HUMAN || (!func_80123434(player) && player->unk_730 == 0));
```
engineer124, line 420:
```suggestion
           (player->transformation != PLAYER_FORM_HUMAN || (!func_80123434(player) && player->unk_730 == NULL));
```
`unk_730` is a pointer

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    Player* player = GET_PLAYER(globalCtx);
+
+    if ((player->stateFlags1 & PLAYER_STATE1_800) && (player->heldActor == actor)) {
+        player->interactRangeActor = NULL;
+        player->heldActor = NULL;
+        player->actor.child = NULL;
+        player->stateFlags1 &= ~PLAYER_STATE1_800;
+        return 1;
+    }
+    return 0;
```
engineer124, line 446:
```suggestion
        return true;
    }
    return false;
```

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    player->rightHandType = gPlayerModelTypes[modelGroup][2];
+    player->sheathType = gPlayerModelTypes[modelGroup][3];
+
+    if (player->sheathType == 14) {
+        if (gSaveContext.equips.buttonItems[CUR_FORM][0] == ITEM_NONE) {
+            player->sheathType = 15;
+        }
+    }
+
+    aux = gPlayerModelTypes[modelGroup];
```
engineer124, line 601: `aux` seems like a weird temp name. Can't you just use something like `playerModelTypes` idk

src/code/z_player_lib.c
engineer124, line 712:
```suggestion
        return SQXZ(diff) <= SQ(xzRange)) && (0.0f <= diff.y) && (diff.y <= yRange);
```

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
+    f32 temp_f14;
+    f32 sp34 = Math_CosS(arg1) * arg0;
+    f32 sp30 = Math_SinS(arg1) * -arg0;
+
+    arg5->x = (Math_SinS(arg3) * sp30) + arg2->x;
+    arg5->y = arg2->y + sp34;
+    arg5->z = (Math_CosS(arg3) * sp30) + arg2->z;
+
+    Math_Vec3f_Diff(arg5, arg4, &sp44);
+    sp40 = sqrtf(SQ(sp44.x) + SQ(sp44.z));
```
engineer124, line 938:
```suggestion
    sp40 = sqrtf(SQXZ(sp44));
```

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1021 @@
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_801242B4.s")
+    sp28 = player->unk_B62 * 0.19607843f;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_801242DC.s")
+    OPEN_DISPS(globalCtx->state.gfxCtx);
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124420.s")
+    AnimatedMat_DrawXlu(globalCtx, Lib_SegmentedToVirtual(&object_link_zora_Matanimheader_012A80));
+    Matrix_Scale(sp28, sp28, sp28, 1);
```
engineer124, line 992:
```suggestion
    Matrix_Scale(sp28, sp28, sp28, MTXMODE_APPLY);
```

engineer124 (commented):
Nice work, some comments to start

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
AngheloAlf: Ohh right, I forgot to make it...
I'll add it

src/code/z_player_lib.c
EllipticEllipsis: Well that's the thing, isn't it? Usually you can have `!= 0` or not, even though that's "locally" nonequivalent.

src/code/z_player_lib.c
AngheloAlf: This is the function.
https://github.com/Decompollaborate/mm/blob/3f7ff526991af1b5e67b0457ca9f8c8e35808253/src/code/z_player_lib.c#L2809-L2818
It just performs some Vec3f copies.

The only other place where `D_801F58B0` is used is in `func_801284A0`, which is currently non equivalent
https://github.com/Decompollaborate/mm/blob/3f7ff526991af1b5e67b0457ca9f8c8e35808253/src/code/z_player_lib.c#L3037-L3069

src/code/z_player_lib.c
AngheloAlf: `!!` matches, but I'm not sure if it is better

EllipticEllipsis: Depends how much we care about our ifs being boolean.

src/code/z_player_lib.c
AngheloAlf: Yeah, it isn't required, but I think it is better to use the `_MAX` enums for cases like this, where the size is not just an arbitrary amount

src/code/z_player_lib.c
AngheloAlf: I can remove the comments if it is preferred.
About the name of the variable, what about `sMasksByItemId` or similar?

src/code/z_player_lib.c
EllipticEllipsis: I only see disadvantages to it really.

src/code/z_player_lib.c
EllipticEllipsis: I'd just say `sMaskItemIds`

src/code/z_player_lib.c
AngheloAlf: I think we would need to discuss if we want to use PlayerLib as a prefix for functions in this file

src/code/z_player_lib.c
AngheloAlf: I haven't named it mainly because the whole "boot data" / "boot enum" stuff doesn't really convince me, specially in MM, where there are no explicit boots like in OoT

src/code/z_player_lib.c
AngheloAlf: Like, why it is checking for both `ACTOR_PLAYER` and `PLAYER_FORM_FIERCE_DEITY`?
I guess it is related to `PLAYER_FORM_FIERCE_DEITY` having value 0. Kafei extends the `Player` and it probably doesn't set the `transformation` member, so it would remain 0.

src/code/z_player_lib.c
AngheloAlf: This function is very dumb imo
It is used twice on player_lib, and 9 times on player_actor.
I was hoping player_actor may give some of explanation to this dumb function, but there's probably isn't one and this is just `Player_IsGoronOrDeku`

src/code/z_player_lib.c
AngheloAlf: No, not really. I just forgot to change that function to use true/false

src/code/z_player_lib.c
AngheloAlf: Those textIds are not present on the text dump I have

src/code/z_player_lib.c
AngheloAlf: I think `Player_GetMask` is fine, but I guess it could be changed.
I would like to know what does other people think about it first

src/code/z_player_lib.c
EllipticEllipsis: Yes, I can't think of any commonality between them.

src/code/z_player_lib.c
AngheloAlf: ... and move the comment to the macro?
I don't think making a macro that would be used only in one place is worth enough. The function itself is already wrapping this dumb behavior in only one place

src/code/z_player_lib.c
AngheloAlf: I don't think we have decided on a style for this specific case.
I personally prefer to avoid the `else` in this specific case

src/code/z_player_lib.c
AngheloAlf: I reused the placement of data in the main player_lib branch

src/code/z_player_lib.c
AngheloAlf: Seems to match as `(10.0f / 51.0f)`

src/code/z_player_lib.c
AngheloAlf: No, I'll change it

src/code/z_player_lib.c
EllipticEllipsis: Hmm, there don't seem to be any `0x26..` ones... are we sure that's what they are?

src/code/z_player_lib.c
AngheloAlf: Yeah, data isn't imported yet to minimize the size of the PR, you can take a look at how much data there is in the linked branch.
I added this as an `#if 0` so people can know what this specific variable contains

src/code/z_player_lib.c
AngheloAlf: Ohh nice catch. That variable is already on master and I forgot to rename it

src/code/z_player_lib.c
isghj5: Neither can swim
(well, someone managed to spawn in pinnacle rock as deku and he does have a swimming animation, but pretty sure that was undefined behavior)

AngheloAlf: Neither being able to swim is interesting. In that case this function could be something like `Player_FormCantSwim` or similar.
But nothing in this file makes it obvious that was the intention for this function. This is why I wanted to wait for player_actor before naming this function

AngheloAlf: Something to keep in consideration, this function is called by `Player_ActionToModelGroup`.
The OoT version of that function looks like this:
```c

s32 Player_IsChildWithHylianShield(Player* this) {
    return gSaveContext.linkAge != LINK_AGE_ADULT && (this->currentShield == PLAYER_SHIELD_HYLIAN);
}

s32 Player_ActionToModelGroup(Player* this, s32 actionParam) {
    s32 modelGroup = sActionModelGroups[actionParam];

    if ((modelGroup == 2) && Player_IsChildWithHylianShield(this)) {
        return 1;
    } else {
        return modelGroup;
    }
}

```

The only similarity I can see here is crouching when you use the shield

isghj5: oh yeah I just wanted to point out one thing I noticed they shared, it might be irrelevant to how the function is used.

Although... naming it `Player_IsGoronOrDeku` now will help us understand what the functions that use it are doing, we can re-name it later if we find out its real purpose. Personally, I prefer naming functions before we have perfect names, as seeing the functions in later functions will help us understand what those are doing (or maybe I'm just too lazy to look up every unnamed function)

EllipticEllipsis: Ah, that actually makes perfect sense, although Goron can't move.

src/code/z_player_lib.c
AngheloAlf: Yes, I am. This is only used on `func_801242DC` (OoT's: `func_8008F2F8`), and that value is passed to `func_801518B0`, which is `Message_StartTextbox`.

src/code/z_player_lib.c
EllipticEllipsis: I'd put a comment above it rather than rename it, I think. It's useful to have the vram handy, and an IDE will show the comment on hover.

src/code/z_player_lib.c
AngheloAlf: I tried changing the type of `maskId` to `u8`, but stuff break everywhere.
I have no idea why there are some random places where this member requires an `u8` cast tho

src/code/z_player_lib.c
AngheloAlf: Considering this function is only used on player_lib and player_actor, I wouldn't be opposed to name it now and rename it when we have a better understanding of it

src/code/z_player_lib.c
EllipticEllipsis: Oh, I see. It's for the environment text. So probably they were going to keep them at some point, but they got dropped. Can explain that... not sure what would happen if it was used, but this probably is not even touched (or presumably you'd get UB).

src/code/z_player_lib.c
EllipticEllipsis: I'm wondering about the whole -1 thing too. Don't the masks start at 0?

src/code/z_player_lib.c
AngheloAlf: Are you referring to this PR? 
https://github.com/zeldaret/oot/pull/1143/files#diff-6352b604d32ba625e89c08478eeaadb8719fe0240427e998ccd7b98c425b2d92

That array changed in player actor, not in player lib

src/code/z_player_lib.c
AngheloAlf: iirc, yes

src/code/z_player_lib.c
EllipticEllipsis: In which case at least add a todo comment with some hypotheses.

src/code/z_player_lib.c
EllipticEllipsis: Probably. But the `* 2` Just Looks Weird.

src/code/z_player_lib.c
AngheloAlf: object_link_goron_DL_00BDD8
Screenshot from 2022-02-22 01-51-00
object_link_goron_DL_014690 (this one uses an AnimatedMaterial)
Screenshot from 2022-02-22 01-51-31
object_link_goron_DL_011AB8
Screenshot from 2022-02-22 01-52-02

EllipticEllipsis: At least the first two are obvious. Perhaps we should get a @tom-overton smiley face on the third...

src/code/z_player_lib.c
AngheloAlf:
```c
    /* 0x00 */ PLAYER_MASK_NONE,
```
Mask 0 is NONE, so everything here skips it and indexes those arrays at 1

src/code/z_player_lib.c
AngheloAlf: Sounds good

src/code/z_player_lib.c
AngheloAlf: It is probably because of all this dlists
https://github.com/Decompollaborate/mm/blob/3f7ff526991af1b5e67b0457ca9f8c8e35808253/src/code/z_player_lib.c#L744-L808

src/code/z_player_lib.c
tom-overton: Your wish is my command; it indeed appears to be the punch:
GLideN64_ZELDA_MAJORA'S_MASK_005

src/overlays/actors/ovl_En_Wf/z_en_wf.c
AngheloAlf: So, you want me to make an enum now? Or should we wait instead?

EllipticEllipsis: I can fire up KZ later, probably won't be hard to find the array in Player.

src/code/z_player_lib.c
AngheloAlf: I'm not convinced by those names.
For example:
```c
    } else if (player->maskObjectLoadState == MASK_LOAD_STATE_FINISH_LOADING) {
        if (osRecvMesg(&player->maskObjectLoadQueue, NULL, OS_MESG_NOBLOCK) == 0) {
            player->maskObjectLoadState = MASK_LOAD_STATE_NOT_LOADING;

            if (player->currentMask == PLAYER_MASK_GREAT_FAIRY) {
                s32 i;

                for (i = 0; i < ARRAY_COUNT(D_801F58B0); i++) {
                    func_80127B64(D_801F58B0[i], ARRAY_COUNT(D_801F58B0[i]), &player->bodyPartsPos[7]);
                }
            }
        }
    }
```
It looks weird

AngheloAlf: Yeah. I think each state mean something like this:
- 0: The mask object is loaded.
- 1: The mask object must be changed (and the DMA request has not been sent yet)
- 2: Waiting for the DMA request to complete.

src/code/z_player_lib.c
AngheloAlf: Nice thanks.
So, what about naming them something like this?:
- `gLinkGoronCrouchingDL`
- `gLinkGoronRollingDL`
- `gLinkGoronPunchDL` / `gLinkGoronPunchEffectDL`

tom-overton: I'm not sure if we prefer "LinkGoron" or "GoronLink", but those all sound pretty sensible to me.

EllipticEllipsis: I think `Curled`, `RollingSpikesAndEffect` and `GoronPunchEffect`: it seems important to me that the latter two are labelled as effects. I assume that when rolling with spikes it just shows the squashed first one inside the second.

src/overlays/actors/ovl_En_Wf/z_en_wf.c
EllipticEllipsis: I fired up KZ, documentation complete: https://docs.google.com/spreadsheets/d/1PPZof5fC0ujw9OupEezdE0XOjCqbGaDUgY24mZ597Xs/edit?usp=sharing
Not saying the names are any good, though.

src/code/z_player_lib.c
AngheloAlf: Currently I have this member named as `alpha`.
I don't think `!!` would be an improvement here

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1045 @@
...
+    OBJECT_MASK_BOY,    OBJECT_MASK_GORON,   OBJECT_MASK_ZORA,     OBJECT_MASK_NUTS,
+};
+#endif
+
+// Load mask object?
+void func_801229FC(Player* player) {
+    /**
+     * Notes:
+     *
+     * player->maskObjectLoadState seem to be able to take 3 possible values
```
EllipticEllipsis, line 122:
```suggestion
     * player->maskObjectLoadState seems to be able to take 3 possible values
```

src/overlays/actors/ovl_En_Maruta/z_en_maruta.c
```diff
@@ -57,7 +57,7 @@ u8 D_80B386C0[] = {
     0xFF, 0x2B, 0xD4, 0x17, 0xE8, 0x55, 0xAA, 0x0F, 0xF0,
 };
 
-s32 D_80B386CC[] = {
+s32 D_80B386CC[PLAYER_MWA_MAX] = {
```
EllipticEllipsis, line 60: Again, I don't really see that adding the explicit size helps here... much better to document what it actually is.

AngheloAlf: My main reason for doing this is if somebody for some reason adds an extra value to this enum, then the array would grow accordingly by adding extra zeroes, so this would not produce an OoB access in case that person forgets to add the extra value here.
Also, I think this way is easier to find the arrays that need to be adjusted when an extra value is added to the enum

EllipticEllipsis: I think actual labels on the elements, or use of the enum where appropriate, are likely a better way to solve this than declaring the dimension, since you can't tell it's wrong at compile time anyway.

hensldm: I'd probably agree with Elliptic here. TBH, I would probably prefer the OOB reads as I feel I'd more likely see that, rather than a silent 0 that while doesn't cause obvious problems, might still give incorrect results and is harder to see.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1047 @@
...
+// Load mask object?
+void func_801229FC(Player* player) {
+    /**
+     * Notes:
+     *
+     * player->maskObjectLoadState seems to be able to take 3 possible values
+     * - 0: The mask object is loaded.
+     * - 1: The mask object must be changed (and the DMA request has not been sent yet)
+     * - 2: Waiting for the DMA request to complete.
+     */
```
hensldm, line 126: I'd prefer these above the function

src/code/z_player_lib.c
hensldm: resolved?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1047 @@
...
+};
+#endif
+
+typedef struct {
+    /* 0x00 */ u8 flag;
+    /* 0x02 */ u16 textId;
+} TextTriggerEntry; // size = 0x04
+
+extern TextTriggerEntry sEnvironmentTextTriggers[];
+// Those textIds are OoT remanents and are not present anymore in this game anymore
```
hensldm, line 572:
```suggestion
// These textIds are OoT remnants and are not present in this game anymore
```

src/code/z_player_lib.c
hensldm: I'd agree. `Player_GetMask` is fine, context is enough.

src/code/z_player_lib.c
hensldm: Agreed with Anghelo. The behavior is already localized enough, the function is basically the macro in this case.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1047 @@
...
+s32 func_801242B4(Player* player) {
+    return (player->stateFlags1 & PLAYER_STATE1_8000000) && player->currentBoots < PLAYER_BOOTS_ZORA_UNDERWATER;
+}
+
+s32 func_801242DC(GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+    TextTriggerEntry* triggerEntry;
+    s32 index;
+
+    if (globalCtx->roomCtx.currRoom.unk2 == 3) { // Room is hot
```
hensldm, line 868: Would it be possible to move the comment to the room struct with unk2?  something like `3: Room is hot`?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1047 @@
...
+            (player->actor.bgCheckFlags & 1)) {
+            index = 1;
+        } else {
+            index = 2;
+        }
+    } else {
+        return 0;
+    }
+
+    // Trigger general textboxes under certain conditions, like "It's so hot in here!". Unused on MM
```
hensldm, line 883:
```suggestion
    // Trigger general textboxes under certain conditions, like "It's so hot in here!". Unused in MM
```

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1047 @@
...
+    Math_Vec3f_Diff(arg5, arg4, &sp44);
+    sp40 = sqrtf(SQXZ(sp44));
+
+    sp3C = (sp40 <= 1.0f) ? arg3 : Math_FAtan2F(sp44.z, sp44.x);
+    sp40 = (Math_CosS(sp3C - arg3) * sp40) + arg8;
+
+    if (ABS_ALT(BINANG_SUB(sp3C, arg3)) > 0x4000) {
+        sp3C = BINANG_ROT180(sp3C);
+    }
+    sp3C = sp3C - arg3;
```
hensldm, line 973:
```suggestion
    sp3C -= arg3;
```

src/code/z_player_lib.c
```diff
@@ -156,17 +1050,102 @@
 
 #pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_801262C8.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_801263FC.s")
+// unused
+s32 func_801263FC(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, Actor* actor) {
```
hensldm, line 1054:
```suggestion
s32 func_801263FC(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, Actor* thisx) {
```
Seems like some kind of limb draw so I think `thisx` makes sense.

src/code/z_player_lib.c
```diff
@@ -156,17 +1050,102 @@
...
+extern u8 D_801C096C[PLAYER_SHIELD_MAX];
+#if 0
+u8 D_801C096C[PLAYER_SHIELD_MAX] = {
+    COLTYPE_METAL,
+    COLTYPE_METAL,
+    COLTYPE_METAL,
+};
+#endif
+
+void func_801265C8(GlobalContext* globalCtx, Player* player, ColliderQuad* arg2, Vec3f arg3[4]) {
```
hensldm, line 1075:
```suggestion
void func_801265C8(GlobalContext* globalCtx, Player* player, ColliderQuad* collider, Vec3f arg3[4]) {
```
Seems reasonable?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1046 @@
...
+    player->currentMask = PLAYER_MASK_NONE;
+}
+
+s32 Player_HasMirrorShieldEquipped(GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+
+    return (player->transformation == PLAYER_FORM_HUMAN) && (player->currentShield == PLAYER_SHIELD_MIRROR_SHIELD);
+}
+
+s32 Player_HasMirrorShieldSetToDraw(GlobalContext* globalCtx) {
```
engineer124, line 763: I think `Player_IsHoldingMirrorShield` would be a better name, see discussion here:
https://discord.com/channels/688807550715560050/688851317593997489/974840246657232967

Same with:
`Player_HoldsHookshot` -> `Player_IsHoldingHookshot`
`Player_HoldsTwoHandedWeapon` -> `Player_IsHoldingTwoHandedWeapon`

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1046 @@
...
+
+s32 Player_GetExplosiveHeld(Player* player) {
+    return Player_ActionToExplosive(player, player->itemActionParam);
+}
+
+s32 func_80124278(Actor* actor, s32 arg1) {
+    s32 phi_v1 = 0;
+
+    // Fake match?
+    if ((arg1 == 1) || ((phi_v1 = arg1 - 3, (phi_v1 >= 0)) && (phi_v1 < 4))) {
```
engineer124, line 851:
```suggestion
    //! FAKE:
    if ((arg1 == 1) || ((phi_v1 = arg1 - 3, (phi_v1 >= 0)) && (phi_v1 < 4))) {
```
We're starting to use this label to quickly find fake matches (or likely fake matches, you never know with devs)

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1046 @@
...
+            index = 1;
+        } else {
+            index = 2;
+        }
+    } else {
+        return 0;
+    }
+
+    // Trigger general textboxes under certain conditions, like "It's so hot in here!". Unused in MM
+    triggerEntry = &sEnvironmentTextTriggers[index];
```
engineer124, line 883:
```suggestion
    triggerEntry = &sEnvironmentTextTriggers[envIndex];
```
Maybe to give it a more interesting name?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1046 @@
...
+
+    if (ABS_ALT(BINANG_SUB(sp3C, arg3)) > 0x4000) {
+        sp3C = BINANG_ROT180(sp3C);
+    }
+    sp3C -= arg3;
+
+    temp_v0 = Math_FAtan2F(sp44.y, sp40);
+    temp_v0 = CLAMP(temp_v0, (s16)-arg9, arg9);
+    // fake match?
+    if (sp3C) {}
```
engineer124, line 977:
```suggestion
    //! FAKE:
    if (sp3C) {}
```

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1046 @@
...
+        Matrix_StatePop();
+        D_801C0958 = false;
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80124278.s")
+void Player_DrawZoraShield(GlobalContext* globalCtx, Player* player) {
+    u8* phi_a0;
+    Vtx* vtx;
+    Gfx* dl;
```
engineer124, line 1008: I'd prefer `Gfx* gfx` or `Gfx* dList`, `dl` seems too abbreviated on its own

src/overlays/actors/ovl_En_Maruta/z_en_maruta.c
```diff
@@ -58,7 +58,40 @@ u8 D_80B386C0[] = {
 };
 
 s32 D_80B386CC[] = {
-    5, 5, 3, 3, 7, 7, 7, 7, 3, 3, 3, 3, 7, 7, 3, 3, 0, 0, 0, 0, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 7, 7, 7, 7,
+    /* PLAYER_MWA_FORWARD_SLASH_1H */ 5,
+    /* PLAYER_MWA_FORWARD_SLASH_2H */ 5,
+    /* PLAYER_MWA_FORWARD_COMBO_1H */ 3,
+    /* PLAYER_MWA_FORWARD_COMBO_2H */ 3,
+    /* PLAYER_MWA_RIGHT_SLASH_1H */ 7,
```
engineer124, line 65:
```suggestion
    5, // PLAYER_MWA_FORWARD_SLASH_1H
    5, // PLAYER_MWA_FORWARD_SLASH_2H
    3, // PLAYER_MWA_FORWARD_COMBO_1H 
    3, // PLAYER_MWA_FORWARD_COMBO_2H
    7, // PLAYER_MWA_RIGHT_SLASH_1H
```
These should be stylized like this:

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1043 @@
...
+void Player_SetModels(Player* player, s32 modelGroup) {
+    u8* playerModelTypes;
+
+    D_801F59E0 = player->transformation * 2;
+    player->leftHandType = gPlayerModelTypes[modelGroup][1];
+    player->rightHandType = gPlayerModelTypes[modelGroup][2];
+    player->sheathType = gPlayerModelTypes[modelGroup][3];
+
+    if (player->sheathType == 14) {
+        if (gSaveContext.save.equips.buttonItems[CUR_FORM][EQUIP_SLOT_B] == ITEM_NONE) {
```
hensldm, line 613:
```suggestion
if (CUR_FORM_EQUIP(EQUIP_SLOT_B) == ITEM_NONE) {
```

hensldm (commented):
Macro suggestion

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1043 @@
...
+}
+
+s32 Player_ActionToMeleeWeapon(s32 actionParam) {
+    s32 weapon = actionParam - PLAYER_AP_UNK_2;
+
+    if ((weapon > (PLAYER_AP_UNK_2 - PLAYER_AP_UNK_2)) && (weapon <= (PLAYER_AP_UNK_8 - PLAYER_AP_UNK_2))) {
+        return weapon;
+    }
+    return 0;
+}
```
EllipticEllipsis, line 799: Did a quick rummage in the item to action param function, UNK_2 is the fishing rod, as in OoT. So Kokiri Sword - 1 is probably a better thing to use. UNK_8 is zora fins, so I suppose this counts it as a sword?

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1043 @@
...
+
+s32 Player_GetMeleeWeaponHeld(Player* player) {
+    return Player_ActionToMeleeWeapon(player->itemActionParam);
+}
+
+s32 Player_IsHoldingTwoHandedWeapon(Player* player) {
+    // Relies on two-handed weapons to be contiguous
```
EllipticEllipsis, line 806:
```suggestion
    // Relies on the actionParams for two-handed weapons being contiguous.
```

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1043 @@
...
+    }
+
+    return -1;
+}
+
+s32 Player_GetExplosiveHeld(Player* player) {
+    return Player_ActionToExplosive(player, player->itemActionParam);
+}
+
+s32 func_80124278(Actor* actor, s32 arg1) {
```
EllipticEllipsis, line 844: Think this one is for converting actionparam to a sword too, it looks like `func_8008F2BC` in OoT.

src/code/z_player_lib.c
```diff
@@ -176,11 +1151,37 @@
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_80127488.s")
+    CLOSE_DISPS(play->state.gfxCtx);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_player_lib/func_8012754C.s")
+void func_8012754C(PlayState* play, Player* player) {
+    gSegments[0xA] = PHYSICAL_TO_VIRTUAL(player->maskObjectSegment);
+    AnimatedMat_DrawOpa(play, Lib_SegmentedToVirtual(&object_mask_meoto_Matanimheader_001CD8));
+}
```
EllipticEllipsis, line 1184: `DrawCouplesMask`, I suppose, probably should be a bit more specific.

EllipticEllipsis (commented):
Few more things.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1043 @@
...
+    ITEM_MASK_STONE,
+    ITEM_MASK_BREMEN,
+    ITEM_MASK_BLAST,
+    ITEM_MASK_SCENTS,
+    ITEM_MASK_GIANT,
+    ITEM_MASK_FIERCE_DEITY,
+    ITEM_MASK_GORON,
+    ITEM_MASK_ZORA,
+    ITEM_MASK_DEKU,
+};
```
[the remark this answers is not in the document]
AngheloAlf: I'm not sure what you mean. Something like this https://github.com/zeldaret/mm/pull/688#discussion_r811527207 ?

src/code/z_player_lib.c
AngheloAlf: I guess zora fins being a melee weapon (not sword) kinda makes sense.
The fishing rod does not count as a two handed melee weapon?

src/code/z_player_lib.c
AngheloAlf: The item enum is `ITEM_MASK_COUPLE`, so shouldn't it be `Player_DrawMaskCouple`?

EllipticEllipsis: It's not going to come up in a search anyway, and I intend to move the item names over to be more like the actual item names soon anyway.

src/code/z_player_lib.c
EllipticEllipsis: Well, the point is that it returns 0 for everything else, right? I don't think it's coercing every other value into a fishing rod.

src/code/z_player_lib.c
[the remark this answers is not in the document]
EllipticEllipsis: Hmm, obviously I'm not sure if I want them... either way is fine, just would be more consistent with what is done elsewhere.

hensldm (approved):

src/code/z_player_lib.c
```diff
@@ -773,11 +774,11 @@ s32 func_801240DC(Player* player) {
...
-    s32 temp_v0 = actionParam - PLAYER_AP_UNK_2;
+    s32 temp_v0 = actionParam - PLAYER_AP_FISHING_POLE;
 
     if (player->transformation != PLAYER_FORM_GORON) {
-        if (((actionParam - PLAYER_AP_UNK_2) > (PLAYER_AP_UNK_2 - PLAYER_AP_UNK_2)) &&
-            ((actionParam - PLAYER_AP_UNK_2) < (PLAYER_AP_SWORD_GREAT_FAIRY - PLAYER_AP_UNK_2))) {
+        if (((actionParam - PLAYER_AP_FISHING_POLE) > (PLAYER_AP_FISHING_POLE - PLAYER_AP_FISHING_POLE)) &&
+            ((actionParam - PLAYER_AP_FISHING_POLE) < (PLAYER_AP_SWORD_GREAT_FAIRY - PLAYER_AP_FISHING_POLE))) {
             return temp_v0;
         }
```
EllipticEllipsis, line 783: I still think it would be better to use `KOKIRI_SWORD - 1` for these, but it doesn't really matter for now.

src/code/z_player_lib.c
```diff
@@ -1,153 +1,1047 @@
...
+}
+
+void func_80122C20(PlayState* play, struct_80122D44_arg1* arg1) {
+    struct_80122D44_arg1_unk_04* temp_v1 = &arg1->unk_04[0];
+    s32 phi_a1;
+    s32 i;
+
+    for (i = 0; i < ARRAY_COUNT(arg1->unk_04); i++, temp_v1++) {
+        // Can't be `temp_v1->alpha != 0`
+        if (temp_v1->alpha) {
```
engineer124, line 184:
```suggestion
        if (temp_v1->alpha) { // != 0
```
I prefer this style of comment when it's suppose to be like this, indicating that it's only a matching issue. But up to you

EllipticEllipsis: I think the first is clearer, tbh.

AngheloAlf: I don't like it either, feels a bit more confusing

EllipticEllipsis (approved):
We can decide on the unused MWAs later.

Kenix3 (approved):

PR #692, opened 2022-02-22
Git tips and tricks

jpburnett:
---

Here are just a list of some of the git tips and tricks that I think are relevant for decomp. They could be helpful for those either new to git or unfamiliar with how to do some things. It is by far not a comprehensive list of all commands, so if there are any that others find useful I can add them to this.

Things that are missing is what to do if/when certain errors are encountered. I have not run in to any personally, but can see how it would be useful since it could possibly happen to others.


jpburnett:
Going to iron this out a little more.

PR #704, opened 2022-03-05
Obj_Nozoki

Kelebek1:
---


src/overlays/actors/ovl_Obj_Nozoki/z_obj_nozoki.c
```diff
@@ -37,54 +45,427 @@ const ActorInit Obj_Nozoki_InitVars = {
...
+        }
+    }
+}
+
+s32 func_80BA26A8(ObjNozoki* this) {
+    if (this->unk_15F < 0) {
+        return true;
+    }
+
+    if (ActorCutscene_GetCanPlayNext(this->unk_15F) != 0) {
```
hensldm, line 137:
```suggestion
    if (ActorCutscene_GetCanPlayNext(this->unk_15F)) {
```

hensldm (approved):

src/overlays/actors/ovl_Obj_Nozoki/z_obj_nozoki.c
```diff
@@ -37,54 +44,427 @@ const ActorInit Obj_Nozoki_InitVars = {
...
+
+void ObjNozoki_Update(Actor* thisx, GlobalContext* globalCtx) {
+    ObjNozoki* this = THIS;
+
+    this->actionFunc(this, globalCtx);
+}
+
+Gfx* D_80BA34FC[] = {
+    object_secom_obj_DL_000080,
+    (Gfx*)0x0A0001A0,
```
AngheloAlf, line 457:
```suggestion
    (Gfx*)0x0A0001A0, //! @bug This dlist should point to a dlist on OBJECT_GI_MSSA, but the segment and the offset is wrong. This doesn't have side-effects because of the extra check on the Draw function
```

You'll probably need to run the formatter after this

src/overlays/actors/ovl_Obj_Nozoki/z_obj_nozoki.c
```diff
@@ -37,54 +44,427 @@ const ActorInit Obj_Nozoki_InitVars = {
...
+    object_secom_obj_DL_001230,
+    object_secom_obj_DL_001300,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_Obj_Nozoki/ObjNozoki_Update.s")
+void func_80BA3434(Actor* thisx, GlobalContext* globalCtx) {
```
AngheloAlf, line 462: This seems to be the main and only draw function, because it is the only one set to `actor.draw`, so I think it should be named accordingly
```suggestion
void ObjNozoki_Draw(Actor* thisx, GlobalContext* globalCtx) {
```

EllipticEllipsis (approved):

Kenix3 (approved):

PR #716, opened 2022-03-11
En_Syateki_Dekunuts (Shooting Gallery Deku Scrub) OK

tom-overton:
---

Pretty simple actor. Like the other Syateki actors, I'm leaving this undocumented until all of them are decompiled. Petrie also found a better way to match one of the loops in Init, so I went and changed them for Syateki_Wf and Syateki_Crow too.


AngheloAlf (approved):
Looks good
I would prefer if the limb enum were used, but since you didn't used it anywhere I think it is fine

tom-overton:
> Looks good I would prefer if the limb enum were used, but since you didn't used it anywhere I think it is fine

Sure, that makes sense. Back when I worked on the other two Syateki actors, the limb enum stuff wasn't checked in, so I would've had to add it myself. But now that it *is* in, it makes sense to use it. I'll add it to all three Syateki actors I've done.

src/overlays/actors/ovl_En_Syateki_Crow/z_en_syateki_crow.c
```diff
@@ -292,9 +290,9 @@ s32 EnSyatekiCrow_OverrideLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx*
                                    Actor* thisx) {
     EnSyatekiCrow* this = THIS;
 
-    if (limbIndex == 7) {
+    if (limbIndex == OBJECT_CROW_LIMB_07) {
         rot->y += (s16)(3072.0f * sin_rad(this->skelAnime.curFrame * 0.7853982f));
-    } else if (limbIndex == 8) {
+    } else if (limbIndex == OBJECT_CROW_LIMB_08) {
         rot->y += (s16)(5120.0f * sin_rad((this->skelAnime.curFrame + 2.5f) * 0.7853982f));
```
isghj5, line 296: These two floats `0.7853982`
Is that M_PI/4 ?

isghj5 (approved):

src/overlays/actors/ovl_En_Syateki_Crow/z_en_syateki_crow.c
tom-overton: Yep

jpburnett (approved):
nice 👍

Kenix3 (approved):

PR #718, opened 2022-03-13
Script for checking relocs correctly updated on spec

AngheloAlf:
---

I made a small script which checks for relocs to be correctly used on the spec depending if the overlay is matched, has NON_MATCHINGs and/or NON_EQUIVALENTs

The script performs the following checks:
- The extracted reloc isn't being used on the spec if the overlay has no other GLOBAL_ASM left (it has been matched).
- Warn if an overlay is using both NON_MATCHING and NON_EQUIVALENT.
- Warn if an overlay has NON_MATCHING ifdefs but some functions are still unattempted.
- Both relocs (generated and extracted) are used on the spec on NON_MATCHING overlays.
- The generated reloc isn't being used on NON_EQUIVALENT overlays.

I also added this script as another step to the Jenkinsfile, so Jenkins can check for this automatically.
Because of this (and as an example of this script), Jenkins will be mad at this PR, because currently `ovl_En_Door` isn't using the correct reloc on the master branch.

This script isn't perfect tho, there are probably some edge cases I haven't considered. 
Any feedback is welcome

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,6 +14,12 @@ pipeline {
                 sh 'bash -c "tools/check_format.sh 2>&1 >(tee tools/check_format.txt)"'
             }
         }
+        stage('Check relocs') {
+            steps {
+                echo 'Checking relocs on spec...'
+                sh 'bash -c "tools/reloc_spec_check.sh"'
+            }
+        }
         stage('Copy ROM') {
             steps {
                 echo 'Setting up ROM...'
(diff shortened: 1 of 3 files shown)


hensldm (approved):

EllipticEllipsis:
I wonder if it should have colourised output to make it stick out more in the Jenkins output, since it's quite a long way up. Same for the formatting checker, come to that.

AngheloAlf:
Added colors as requested
image
https://jenkins.deco.mp/job/MM/job/PR-718/2/console

hensldm:
Since there is a Jenkins jobs showing what happens on failure, I think it would be appropriate to fix `EnDoor's` reloc so that this PR now passes.

EllipticEllipsis (approved):

Kenix3 (approved):

PR #748, opened 2022-03-24
`EffectEnIceBlock` OK, `object_ice_block` documented (and minor `Obj_Iceblock` documented)

isghj5:
---

https://user-images.githubusercontent.com/42048411/159821871-9dc75964-2f81-4ae9-9a88-f0eebfda4830.mov

I got ahead of myself: this is my first effect so I need verification that I am decomping effects correctly but I got distracted and started documenting actor:`Obj_Iceblock` before I realized how large and complicated of an actor it is, so this includes some basic documentation from that actor, but I don't want to document that actor any further right now since the Effect should be the focus, and it could take me awhile to tackle Obj_Iceblock.


AngheloAlf (approved):

src/overlays/effects/ovl_Effect_En_Ice_Block/z_eff_en_ice_block.c
```diff
@@ -1,31 +1,85 @@
...
+        Matrix_InsertTranslation(this->pos.x, this->pos.y, this->pos.z, MTXMODE_NEW);
+        Matrix_Scale(scale, scale, scale, MTXMODE_APPLY);
+        Matrix_InsertMatrix(&globalCtx->billboardMtxF, MTXMODE_APPLY);
+        Matrix_InsertZRotation_s(this->rRot, MTXMODE_APPLY);
+
+        gSPMatrix(POLY_XLU_DISP++, Matrix_NewMtx(gfxCtx), G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+
+        gSPDisplayList(POLY_XLU_DISP++, &gIceBlockShardEffectDL);
+
+        CLOSE_DISPS(globalCtx->state.gfxCtx);
```
hensldm, line 78:
```suggestion
        CLOSE_DISPS(gfxCtx);
```
Was looking at the decompme scratch, and this still matches and seems better imo.

hensldm (approved):
Nice work. Just one minor comment about the draw function in the effect.

src/overlays/effects/ovl_Effect_En_Ice_Block/z_eff_en_ice_block.c
AngheloAlf: I think I prefer if we used the same parameter as the one used on the  `OPEN_DISP

isghj5: Oh cool it does match, yeah I just assumed it didn't, will change.

hensldm (commented):
Just noticed this while decomping one of the other effects, but you should be able to remove the `ovl_Effect_En_Ice_Block` section from `undefined_syms.txt`

Kenix3 (approved):

PR #763, opened 2022-03-26
`code_8013EC10` and `code_80182CE0` (`z_rumble` and `sys_rumble`)

AngheloAlf:
---

This small file seems to contain the functions to rumble the controller, specifically `func_8013EC44` and `func_8013ECE0`. Because of this I decided to rename this file to `z_rumble.c`

I didn't name most functions because I don't understand what they do, or what is the difference between the two functions which trigger the controller rumble.

I introduced the small header `z64rumble.h` and moved the corresponding functions prototypes to that file. Since not every file uses the rumble functions, I `#include`d this header only on the files where it is used instead of adding it on a global header like `z64.h`

I also renamed a few functions from padmgr. I took the name of that function from Tharo's branch.

I didn't add any of the functions I renamed to actorfixer since they are used only on decomped functions or only once in undecomped functions.

edit: I removed the `static`s on Boss02 because I had bss problems with that file.


src/code/z_rumble.c
```diff
@@ -0,0 +1,86 @@
...
+    s32 i;
+
+    if (SQ(1000.0f) < xyzDistToPlayerSq) {
+        phi_v0 = 1000;
+    } else {
+        phi_v0 = sqrtf(xyzDistToPlayerSq);
+    }
+
+    if ((phi_v0 < 1000) && (arg1 != 0) && (arg3 != 0)) {
+        temp_a0 = arg1 - ((s32)(phi_v0 * 0xFF) / 1000);
```
hensldm, line 47:
```suggestion
        temp_a0 = arg1 - (phi_v0 * 255) / 1000;
```
Cast necessary? Also the above functions has dec for 255.

src/code/z_rumble.c
```diff
@@ -0,0 +1,86 @@
...
+void func_8013EE24(void) {
+    D_801F69D0.unk_104 = 2;
+}
+
+void func_8013EE38(void) {
+    D_801F69D0.unk_104 = 0;
+}
+
+void func_8013EE48(s32 arg0) {
+    D_801F69D0.unk_105 = arg0 != false;
```
hensldm, line 85:
```suggestion
    D_801F69D0.unk_105 = !!arg0;
```
In this instance I think I like this syntax better, but completely up to you.

hensldm (approved):
LGTM. Just a couple thoughts.

EllipticEllipsis (approved):
I'm not completely sure about making a separate header for this considering how widely it's used, but we can worry about that later.

AngheloAlf:
I updated the PR with the new file and I renamed some functions. 
I suggest to review it again.

AngheloAlf:
> I'm not completely sure about making a separate header for this considering how widely it's used, but we can worry about that later.

Instead of `#include`ing the separate header on each specific C file, I could `#include` it on `z64.h` if you think that's better. I tried to avoid that for bss purposes tho

src/code/z_rumble.c
```diff
@@ -0,0 +1,85 @@
...
+        if (temp > 0) {
+            sRumbleMgr.unk_10A = temp;
+            sRumbleMgr.unk_10B = arg2;
+            sRumbleMgr.unk_10C = arg3;
+        }
+    }
+}
+
+// Normal rumble?
+void func_8013ECE0(f32 xyzDistToPlayerSq, u8 arg1, u8 arg2, u8 arg3) {
```
hensldm, line 34:
```suggestion
void func_8013ECE0(f32 distSq, u8 arg1, u8 arg2, u8 arg3) {
```
probably better since I don't think the function is specific to player distance, also would mirror `func_8013EC44` better.
Note this would also need to be updated in the `.h` file.

hensldm (commented):
I like the new docs. Noticed something else I think I missed in my previous passes.

Kenix3 (approved):

hensldm (approved):
LGTM. I like the new docs

src/overlays/actors/ovl_Bg_Iknin_Susceil/z_bg_iknin_susceil.c
```diff
@@ -75,11 +76,11 @@ void func_80C0A86C(BgIkninSusceil* this, GlobalContext* globalCtx, s16 verticalM
     Quake_SetQuakeValues(quake, verticalMag, 0, 0, 0);
     Quake_SetCountdown(quake, countdown);
     if (arg4 == 1) {
-        func_8013ECE0(10000.0f, 255, 20, 150);
+        Rumble_Request(10000.0f, 255, 20, 150);
```
engineer124, line 79:
```suggestion
        Rumble_Request(SQ(100.0f), 255, 20, 150);
```

src/overlays/actors/ovl_Bg_Iknin_Susceil/z_bg_iknin_susceil.c
```diff
@@ -75,11 +76,11 @@ void func_80C0A86C(BgIkninSusceil* this, GlobalContext* globalCtx, s16 verticalM
...
     } else if (arg4 == 2) {
-        func_8013ECE0(10000.0f, 180, 20, 100);
+        Rumble_Request(10000.0f, 180, 20, 100);
```
engineer124, line 81:
```suggestion
        Rumble_Request(SQ(100.0f), 180, 20, 100);
```

src/overlays/actors/ovl_Bg_Iknin_Susceil/z_bg_iknin_susceil.c
```diff
@@ -75,11 +76,11 @@ void func_80C0A86C(BgIkninSusceil* this, GlobalContext* globalCtx, s16 verticalM
...
     } else if (arg4 == 3) {
-        func_8013ECE0(10000.0f, 120, 20, 10);
+        Rumble_Request(10000.0f, 120, 20, 10);
```
engineer124, line 83:
```suggestion
        Rumble_Request(SQ(100.0f), 120, 20, 10);
```

src/code/sys_rumble.c
```diff
@@ -0,0 +1,156 @@
...
+
+        if (rumbleMgr->unk_10A != 0) {
+            temp = rumbleMgr->unk_10A;
+        } else if (index == -1) {
+            temp = 0;
+        } else {
+            temp = rumbleMgr->unk_04[index];
+        }
+
+        // Keep track for how long this have been rumbling (almost) non stop
```
Dragorn421, line 114:
```suggestion
        // Keep track of how long rumbling has been (almost) nonstop enabled
```

src/code/sys_rumble.c
```diff
@@ -0,0 +1,156 @@
+#include "global.h"
```
EllipticEllipsis, line 1: Can make some file comments too: this is intermediate between padmgr and the external file, whereas the external file is how you're actually supposed to interact with this? I suppose this is sort of the scheduling/processing file, the other is the requesting file. There's also the question of where to put an explanatory block about how the system works, since Update is really doing all the work here.

EllipticEllipsis (commented):
Sorry for the very late review, I finally sat down to understand this properly and have some documentation suggestions.

src/code/z_rumble.c
```diff
@@ -0,0 +1,95 @@
...
+ */
+void Rumble_StateWipeRequests(void) {
+    gRumbleMgr.state = RUMBLEMANAGER_STATE_WIPE;
+}
+
+/**
+ * Request processing is paused if updateEnabled is set to false
+ */
+void Rumble_SetUpdateEnabled(s32 updateEnabled) {
+    gRumbleMgr.updateEnabled = !!updateEnabled;
```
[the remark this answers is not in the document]
AngheloAlf: https://github.com/zeldaret/mm/pull/763#discussion_r835784195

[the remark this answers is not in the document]
EllipticEllipsis: I still don't think this is a style we should have in the codebase, imo it's way less clear than explicit `(x == 0)`.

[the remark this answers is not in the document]
hensldm: While I still think `!!` is probably the best syntax, I don't have too strong an opinion. Though I don't really like `updateEnabled != false` and would probably prefer `updateEnabled != 0`.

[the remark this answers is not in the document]
EllipticEllipsis: Semantically it's more correct with `false`: if it was a true bool in a type-safe language you'd have to write that if you wanted to write it that way.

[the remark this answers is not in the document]
hensldm: Though that is assuming `updateEnabled` is in fact a bool (which I'm not 100% sure it truly was). It's possible it had multiple values/states, and only state 0 would have disabled update, but of course this is all speculation and I think I just don't like how it ended up like this :) .

src/code/sys_rumble.c
```diff
@@ -0,0 +1,156 @@
...
+#include "z64rumble.h"
+
+// sRumbleWasEnabledOnLastTick? Probably name it after updateEnabled
+u8 D_801D1E70 = true;
```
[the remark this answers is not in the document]
EllipticEllipsis: Basically it makes the block that checks it run once after the state flips to off, and that turns it to false. Don't know if that is what "lock" means.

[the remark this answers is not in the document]
hensldm: Hmm, at least to me sounds similar to a locking mechanism, but not quite so shouldn't name it after that.

src/code/sys_rumble.c
```diff
@@ -0,0 +1,156 @@
...
+
+    // Turn rumbling off for all controllers
+    for (i = 0; i < ARRAY_COUNT(rumbleMgr->rumbleEnabled); i++) {
+        rumbleMgr->rumbleEnabled[i] = false;
+    }
+
+    if (!rumbleMgr->updateEnabled) {
+        // Rumbling update is disabled
+
+        if (D_801D1E70) {
```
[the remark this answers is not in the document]
hensldm: I don't think `will` is necessary. `sDisableRumble` seems fine to me. Could go even more specific with `sDisablePadmgrRumble`?

include/z64rumble.h
```diff
@@ -0,0 +1,46 @@
...
+    /* 0x004 */ u8 requestIntensities[RUMBLE_REQUEST_BUFFER_SIZE]; // seems to have two purposes: act as a priority value over the other rumble requests and to contribute to requestAccumulators on each tick
+    /* 0x044 */ u8 requestDecayTimers[RUMBLE_REQUEST_BUFFER_SIZE]; // amount of ticks this request will run before requestIntensities starts decreasing
+    /* 0x084 */ u8 requestDecaySteps[RUMBLE_REQUEST_BUFFER_SIZE]; // used to decrease requestIntensities by this amount each tick after requestDecayTimers runs out
+    /* 0x0C4 */ u8 requestAccumulators[RUMBLE_REQUEST_BUFFER_SIZE]; // the value contained here plus requestIntensities must meet a certain threshold to start rumbling this tick
+    /* 0x104 */ u8 state;
+    /* 0x105 */ u8 updateEnabled; // boolean value which controls if the Update function should be run. Unlike RUMBLEMANAGER_STATE_WIPE, disabling this does not wipe the state of the requests
+    /* 0x106 */ u16 rumblingDuration; // amount of "ticks" updates the Rumble Pak has been rumbling without pause. It is restarted after 5 ticks of not rumbling
+    /* 0x108 */ u16 downTime; // small counter/timer used to restart rumblingDuration?
```
EllipticEllipsis, line 24:
```suggestion
    /* 0x108 */ u16 downTime; // small counter/timer used to restart rumblingDuration
```

Confirmed now, right?

src/code/z_rumble.c
```diff
@@ -45,11 +54,11 @@ void Rumble_Request(f32 distSq, u8 arg1, u8 ticksTimer, u8 decreaseStep) {
...
         for (i = 0; i < RUMBLE_REQUEST_BUFFER_SIZE; i++) {
-            if (gRumbleMgr.unk_04[i] == 0) {
+            if (gRumbleMgr.requestIntensities[i] == 0) {
                 if (temp > 0) {
-                    gRumbleMgr.unk_04[i] = temp;
-                    gRumbleMgr.timer[i] = ticksTimer;
-                    gRumbleMgr.decreaseStep[i] = decreaseStep;
+                    gRumbleMgr.requestIntensities[i] = temp;
+                    gRumbleMgr.requestDecayTimers[i] = ticksTimer;
+                    gRumbleMgr.requestDecaySteps[i] = decreaseStep;
```
EllipticEllipsis, line 61: I think these (and the names in the cutscene files) should be closer to the RumbleManager names, I'd already forgotten which was which when I was looking at them.

EllipticEllipsis (commented):
Last few things.

EllipticEllipsis (approved):

src/code/sys_rumble.c
```diff
@@ -0,0 +1,160 @@
+/*
+ * File: sys_rumble.c
+ * Description: Internal scheduller system for rumble requests
```
engineer124, line 3:
```suggestion
 * Description: Internal scheduler system for rumble requests
```

src/code/z_rumble.c
```diff
@@ -0,0 +1,104 @@
+/*
+ * File: z_rumble.c
+ * Description: Rumble request system
+ *
+ * Provides a simple interface to allow schedulling up to RUMBLE_REQUEST_BUFFER_SIZE rumble requests to the RumblePak.
+ * There's an aditional Override type of rumble request for requests which should take priorities over any other
+ * schedulled request.
```
engineer124, line 7:
```suggestion
 * Provides a simple interface to allow scheduling up to RUMBLE_REQUEST_BUFFER_SIZE rumble requests to the RumblePak.
 * There's an additional Override type of rumble request for requests which should take priorities over any other
 * scheduled request.
```

engineer124 (commented):
Just some quick typos

src/code/z_rumble.c
```diff
@@ -0,0 +1,104 @@
...
+
+RumbleManager gRumbleMgr;
+
+void Rumble_Update(void* arg0) {
+    RumbleManager_Update(&gRumbleMgr);
+    PadMgr_RumbleSet(gRumbleMgr.rumbleEnabled);
+}
+
+// Used by some bosses (and fishing)
+void Rumble_Override(f32 distSq, u8 intensity, u8 decayTimer, u8 decayStep) {
```
hensldm, line 21: Looking at how its used, I think `intensity` would be better off named `maxIntensity`, while `temp` would be the actual `intensity`.

src/code/z_rumble.c
```diff
@@ -0,0 +1,104 @@
...
+
+        if (temp > 0) {
+            gRumbleMgr.overrideIntensity = temp;
+            gRumbleMgr.overrideDecayTimer = decayTimer;
+            gRumbleMgr.overrideDecayStep = decayStep;
+        }
+    }
+}
+
+void Rumble_Request(f32 distSq, u8 intensity, u8 decayTimer, u8 decayStep) {
```
hensldm, line 42: Same as above.

src/code/z_rumble.c
EllipticEllipsis: Or perhaps `sourceIntensity`. The idea is that the source has an inherent intensity attached to it, which is damped linearly with distance to give the received intensity.

AngheloAlf: I like `sourceIntensity` for the parameter

EllipticEllipsis (approved):

PR #765, opened 2022-03-26
Update Readme

hensldm:
---

This was brought up in discord awhile ago (forget when or where), but cloning the repo should be the first step, since our setup requires using `requirments.txt` in the repo for python dependencies. So this PR changes steps 1 and 2, as well as fix the linking bug in the MacOS instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,12 +50,26 @@ We recommend using Debian or Ubuntu 20.04 Linux distributions.
 
 ### MacOS
 
-Preparation is covered in a [separate document](docsBUILDING_MACOS.md).
+Preparation is covered in a [separate document](docs/BUILDING_MACOS.md).
 
 
 ### Linux (Native or under WSL / VM)
 
-#### 1. Install build dependencies
+#### 1. Fork the repository
+
+Create your own fork of the repository at `https://github.com/zeldaret/mm`. Then clone your fork where you wish to have the project, with the command:
+
+```bash
+git clone https://github.com/<YOUR_USERNAME>/mm.git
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,12 +50,26 @@ We recommend using Debian or Ubuntu 20.04 Linux distributions.
...
 
 ### Linux (Native or under WSL / VM)
 
-#### 1. Install build dependencies
+#### 1. Clone the repository
+
+Create your own fork of the repository at `https://github.com/zeldaret/mm`. Then clone your fork where you wish to have the project, with the command:
+
+```bash
+git clone https://github.com/<YOUR_USERNAME>/mm.git
```
AngheloAlf, line 63: On the build dependencies we are listing git, so probably the user may not have git installed at this point. So I think the cloning step should go after the dependencies one

EllipticEllipsis: Yes, and there's no point cloning the repo if the dependencies don't work.

hensldm: Hmm, didn't see that. We would still have the problem of `python3 -m pip install -r requirements.txt` not working until the repo has been cloned, since the `requirements.txt` is part of the repo. Going to move the conversation over to discord for ease.

hensldm: Okay updated based on Discord convo https://discord.com/channels/688807550715560050/740615251011174401/957434679248707614

AngheloAlf (approved):

jpburnett (approved):

Kenix3 (approved):

PR #767, opened 2022-03-27
ABI FPR names

EllipticEllipsis:
---

Use the o32 ABI names for floating-point registers in the assembly. Finally the float registers usage makes some sense!

image

This involved changing quite a lot of tools:
- Add the aliases to `macro.inc`.
- Add support to `diff.py`.
- Add support to `mips_to_c`.
- Change the disassembler to enable the choice of abi or numeric names for float registers. This required tweaking the imports, since `*` imports apparently don't let you edit variables in other files....
- Change the makefile to support this flag.
- Add aliases to `asm-processor`'s `prelude.inc`.
- Also did some other cleanup of the makefile to use the newest asm-processor.

Everything here should be automatic: the end user should not have to do anything apart from updating mips_to_c and any other forks they've made of the external scripts. If you really want to still use the numeric names, you can edit the disassembler line in the makefile to remove the new flag (since I don't expect this to be a popular option, I didn't add a makefile variable).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,8 +73,9 @@ AS         := $(MIPS_BINUTILS_PREFIX)as
 LD         := $(MIPS_BINUTILS_PREFIX)ld
 OBJCOPY    := $(MIPS_BINUTILS_PREFIX)objcopy
 OBJDUMP    := $(MIPS_BINUTILS_PREFIX)objdump
-# ASM_PROC   := python3 tools/asm-processor/build.py --input-enc=utf-8 --output-enc=euc-jp
 ASM_PROC   := python3 tools/asm-processor/build.py
+
+ASM_PROC_FLAGS := --input-enc=utf-8 --output-enc=euc-jp
 
 IINC       := -Iinclude -Isrc -Iassets -Ibuild -I.
 
@@ -183,6 +184,7 @@ build/src/libultra/flash/%.o: MIPS_VERSION := -mips1
 build/src/code/audio/%.o: OPTFLAGS := -O2
 
 build/assets/%.o: OPTFLAGS := -O1
+build/assets/%.o: ASM_PROC_FLAGS := 
 
 # file flags
 build/src/boot_O2_g3/fault.o: CFLAGS += -trapuv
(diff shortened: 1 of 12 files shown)


include/macro.inc
```diff
@@ -37,3 +37,39 @@
...
+.set $fs1,          $f22
+.set $fs1f,         $f23
+.set $fs2,          $f24
+.set $fs2f,         $f25
+.set $fs3,          $f26
+.set $fs3f,         $f27
+.set $fs4,          $f28
+.set $fs4f,         $f29
+.set $fs5,          $f30
+.set $fs5f,         $f31
```
AngheloAlf, line 75: Isn't this the FpcCsr register? Shouldn't it be named like that instead?

tools/asm-processor/prelude.inc
```diff
@@ -1,5 +1,43 @@
...
+.set $fs1,          $f22
+.set $fs1f,         $f23
+.set $fs2,          $f24
+.set $fs2f,         $f25
+.set $fs3,          $f26
+.set $fs3f,         $f27
+.set $fs4,          $f28
+.set $fs4f,         $f29
+.set $fs5,          $f30
+.set $fs5f,         $f31
```
AngheloAlf, line 43: same here

tools/disasm/disasm.py
```diff
@@ -2,9 +2,16 @@
 
 import argparse, ast, math, os, re, struct
 import bisect
-from mips_isa import *
-from multiprocessing import *
+import multiprocessing
 from pathlib import Path
+# global mips_isa.mips_fpr_names
```
AngheloAlf, line 7: what is this comment?

tools/disasm/disasm.py
```diff
@@ -26,10 +33,14 @@
     required=False,
     help="Optional list of files to diassemble separated by a space. This is a whitelist, all files will be skipped besides the ones listed here if used.",
 )
+parser.add_argument("-Mreg-names", choices=fpr_name_options.keys(), help="How to name registers in the output")
```
AngheloAlf, line 36: This is a long argument name, so it should have 2 `-`, doesn't it?
```suggestion
parser.add_argument("--Mreg-names", choices=fpr_name_options.keys(), help="How to name registers in the output")
```

EllipticEllipsis: It mirrors the flag from objdump, that's why it's not POSIX.

Makefile
```diff
@@ -263,7 +268,7 @@ setup:
 ## Assembly generation
 disasm:
 	$(RM) -rf asm data
-	python3 tools/disasm/disasm.py -j $(N_THREADS) $(DISASM_FLAGS)
+	python3 tools/disasm/disasm.py -j $(N_THREADS) $(DISASM_FLAGS) -Mreg-names=o32
```
LtPeriwinkle, line 271: Can the `-Mreg-names=` be added to `DISASM_FLAGS`?

tools/disasm/disasm.py
AngheloAlf: That objdump flag is POSIX complaint. The flag itself is `-M`. From its man page:
```
       -M options
       --disassembler-options=options
           Pass target specific information to the disassembler.  Only supported on some targets.  If it is
           necessary to specify more than one disassembler option then multiple -M options can be used or can be
           placed together into a comma separated list.

```

tools/disasm/disasm.py
```diff
@@ -2,9 +2,16 @@
...
-from multiprocessing import *
+import multiprocessing
 from pathlib import Path
+# global mips_isa.mips_fpr_names
+import mips_isa
+
+# Consider implementing gpr naming too, but already uses abi names by default
+fpr_name_options = {
+    "numeric": mips_isa.numeric_fpr_names,
+    "o32": mips_isa.o32_fpr_names,
```
LtPeriwinkle, line 13: Is it better to leave this as `o32` for clarity or change it to just `32` for consistency with objdump flags?

tools/disasm/disasm.py
hensldm: Speaking of argument names, the `files` argument above should probably also be adjusted (though might be out of scope of this PR). In particular I would suggest the `-f/--full` flag be changed to `-a/-all` and the `-files/--files` flag be `-f/--files`.

include/macro.inc
Thar0: The FpcCsr register actually belongs to a separate set of registers, the floating-point "control" registers. There are two such registers, $1 the floating-point unit version register, and $31 which is the well-known FpcCsr. The disassembler doesn't implement these properly since $f31/$fs5f is never used and neither is $1, the version register, so I cut corners to avoid adding another operand type. I'm not sure if this is now an issue with this additional machinery being added to the float registers.

Makefile
```diff
@@ -99,6 +102,9 @@ MIPS_VERSION := -mips2
 # we support Microsoft extensions such as anonymous structs, which the compiler does support but warns for their usage. Surpress the warnings with -woff.
 CFLAGS += -G 0 -non_shared -Xfullwarn -Xcpluscomm $(IINC) -nostdinc -Wab,-r4300_mul -woff 624,649,838,712
 
+# Use relocations and abi fpr names in the dump
+OBJDUMP_FLAGS := -d -r -Mreg-names=32
```
Thar0, line 106:
```suggestion
OBJDUMP_FLAGS := -drz -Mreg-names=32
```
Might as well disassemble zeros as well for full clarity?

include/macro.inc
EllipticEllipsis: It's still not implemented in the disassembler, so it doesn't really matter, I don't think... $f31 is not the same as $31 anyway, right?

Makefile
EllipticEllipsis: Good idea. Interestingly, I tried combining the flags like this and it threw an error.

include/macro.inc
EllipticEllipsis: If I remember correctly, the point is that different float instructions treat `$31` in different ways, and it so happens that it's never used by instructions that would use register 31 as `$f31`/`$fs5f`.

tools/disasm/disasm.py
EllipticEllipsis: Oh goodness, yes the files one should change.

AngheloAlf (approved):

hensldm (approved):
Checked it out and it works great. Definitely enjoy the clearer float reg names.

Kenix3 (approved):

hensldm (changes requested):
Just noting that the permuter should be adjusted before this gets merged.

PR #769, opened 2022-03-27
Boot_800862E0 OK

hensldm:
---

This PR supersedes #382. Thank you for the groundwork done.

A few notes about this file: 
* Everything is prefixed with SystemArena, though not 100% sure that is right, especially since most of the functions are unused.
     - If SystemArena is fine, perhaps system_arena.c would be an appropriate name for this file
* In particular the blockFunctions are unused, so have no Idea what they would actually do so left them with generic names.
* This looks to be equivalent to OOT's code_800FC620.c


EllipticEllipsis (approved):

AngheloAlf (approved):

Kenix3 (commented):
Why not rename the file to something like "systemarena.c?"

hensldm:
> Why not rename the file to something like "systemarena.c?"

The only reason was I was hesitant since I wasn't sure if SystemArena was the right prefix/system name. However, it seems nobody objects so I will go ahead and rename the file.

AngheloAlf:
Technically all the SystemArena functions live on the `system_malloc.c` file, and the functions here are kinda wrappers for those, so I'm not sure about naming this file systemarena.c

Kenix3 (approved):

PR #771, opened 2022-03-29
En_Dragon (Deep Python) OK and documented. object_utubo documented

tom-overton:
---

Definitely one of the weirder-implemented actors I've worked on. I found myself numerous times thinking "that's not how I would've done this". I don't think my documentation is fantastic, but at least for me it's good enough for how annoying this actor was to understand.


src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
```diff
@@ -37,130 +72,777 @@ const ActorInit En_Dragon_InitVars = {
...
+
+    Collider_DestroyJntSph(globalCtx, &this->collider);
+}
+
+AnimationHeader* animations[] = {
+    &gDeepPythonSmallSideSwayAnim,
+    &gDeepPythonLargeSideSwayAnim,
+    &gDeepPythonVerticalSwayAnim,
+    &gDeepPythonSmallSideSwayAnim,
+};
```
AngheloAlf, line 273: You should either use a more unique name for named variables to avoid possibly future problems with repeated names or use the `static` keyword.
If you want to keep this name, you should use something like `sAnimations`.
This comment applies for the rest of the data of this actor

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
```diff
@@ -37,130 +72,777 @@ const ActorInit En_Dragon_InitVars = {
...
+
+void EnDragon_Attack(EnDragon* this, GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+    f32 currentFrame = this->skelAnime.curFrame;
+    Vec3f sp4C; // used as both the extended position and the camera eye
+    Vec3f at;
+
+    SkelAnime_Update(&this->skelAnime);
+    Math_SmoothStepToS(&this->actor.shape.rot.z, 0, 0xA, 0x1388, 0);
+    if (!(globalCtx->gameplayFrames & 0xF)) {
```
AngheloAlf, line 578: I assume this doesn't match, but I'm going to ask anyways
```suggestion
    if ((globalCtx->gameplayFrames % 16) == 0) {
```

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
```diff
@@ -37,130 +72,777 @@ const ActorInit En_Dragon_InitVars = {
...
+            this->actor.flags &= ~ACTOR_FLAG_1;
+            this->actor.flags |= ACTOR_FLAG_100000;
+            this->action = DEEP_PYTHON_ACTION_SETUP_DEAD;
+            this->actionFunc = EnDragon_SetupDead;
+        }
+    }
+
+    if ((this->action == DEEP_PYTHON_ACTION_EXTEND) && (this->grabWaitTimer == 0) &&
+        (player->invincibilityTimer == 0) && (this->collider.elements[0].info.ocElemFlags & OCELEM_HIT) &&
+        (!(func_800B64FC(globalCtx, 1000.0f, &this->actor.world.pos, &sp30) >= 0.0f) || (sp30 != 1))) {
```
AngheloAlf, line 762: This line has to be written like this? a `<` wouldn't work?
```suggestion
        ((func_800B64FC(globalCtx, 1000.0f, &this->actor.world.pos, &sp30) < 0.0f) || (sp30 != 1))) {
```

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
```diff
@@ -37,130 +72,777 @@ const ActorInit En_Dragon_InitVars = {
...
-        { ELEMTYPE_UNK1, { 0xF7CFFFFF, 0x00, 0x00 }, { 0xF7CFFFFF, 0x00, 0x00 }, TOUCH_NONE | TOUCH_SFX_NORMAL, BUMP_ON, OCELEM_ON, },
+        {
+            ELEMTYPE_UNK1,
+            { 0xF7CFFFFF, 0x00, 0x00 },
+            { 0xF7CFFFFF, 0x00, 0x00 },
+            TOUCH_NONE | TOUCH_SFX_NORMAL,
+            BUMP_ON,
+            OCELEM_ON,
+        },
         { 13, { { 0, 0, 0 }, 0 }, 1 },
```
AngheloAlf, line 125: The first value here is the limb to which this sphere element is attached, so you can use the limb enum here
```suggestion
        { DEEP_PYTHON_LIMB_HEAD, { { 0, 0, 0 }, 0 }, 1 },
```

AngheloAlf (commented):
Good work on this actor!
I left some thoughts

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
tom-overton: I am absolutely shocked, but it *does* match.

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
```diff
@@ -37,130 +72,777 @@ const ActorInit En_Dragon_InitVars = {
...
+void EnDragon_UpdateDamage(EnDragon* this, GlobalContext* globalCtx) {
+    Player* player = GET_PLAYER(globalCtx);
+    u32 sp30;
+
+    if ((this->action == DEEP_PYTHON_ACTION_EXTEND) && ((this->collider.elements[2].info.bumperFlags & BUMP_HIT) ||
+                                                        (this->collider.elements[3].info.bumperFlags & BUMP_HIT) ||
+                                                        (this->collider.elements[4].info.bumperFlags & BUMP_HIT) ||
+                                                        (this->collider.elements[5].info.bumperFlags & BUMP_HIT) ||
+                                                        (this->collider.elements[6].info.bumperFlags & BUMP_HIT) ||
+                                                        (this->collider.elements[7].info.bumperFlags & BUMP_HIT))) {
```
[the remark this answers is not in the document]
tom-overton: It doesn't seem like this actually matches? Which is silly because it's *logically* the same thing. So I guess we're stuck with this ugly-looking thing.

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
tom-overton: Correct, it has to be written like this, as silly as it looks:
image

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
```diff
@@ -37,130 +72,777 @@ const ActorInit En_Dragon_InitVars = {
...
+static ColliderJntSphInit sJntSphInit = {
+    {
+        COLTYPE_HIT6,
+        AT_NONE,
+        AC_ON | AC_TYPE_PLAYER,
+        OC1_ON | OC1_TYPE_PLAYER,
+        OC2_TYPE_1,
+        COLSHAPE_JNTSPH,
+    },
+    8,
```
hensldm, line 215:
```suggestion
    ARRAY_COUNT(sJntSphElementsInit),
```

hensldm (approved):
Nice. Just one small suggestions.

AngheloAlf (approved):

src/overlays/actors/ovl_En_Dragon/z_en_dragon.c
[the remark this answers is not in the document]
tom-overton: Actually, I was just being silly, and this *does* match. I think it definitely looks better, so I'll go with that instead.

Kenix3 (approved):
Very good documentation on this actor. Great work.

PR #775, opened 2022-03-31
Add OoT's `namefixer.py`

EllipticEllipsis:
---

This should be able to effect whole-word replacements. This needs feedback on several things:
- What should be in the list? I added the gSaveContext ones partly as an example, partly because I know they're recent, but since the list is generally not sorted in order of addition, I'm not sure what else should be addressed. I also left some other examples from OoT in to make it clear how to use it.
- Which files should it cover? I added headers and functions.txt/variables.txt, but perhaps these are not needed, since we'd expect the person who did the rename in the first place to cover it?
- It also will not rename files in the nonmatchings folder to match the name of the function, so you still have to run `rename_global_asm.py` if you rename a function included via `GLOBAL_ASM`. This should probably stay like this, but we could merge them if there's interest.


AngheloAlf:
So this PR adds namefixer without removing actorfixer?
Any reason to not move all the logic from namefixer and move it to actorfixer? This way we could keep the script and the current dictionary we have and it would prevent lossing new renames people have added to actorfixer on their corresponding PRs/branches

AngheloAlf:
Yes, we should do some cleanup on actorfixer, but maintaining two scripts sounds tedious and a bad idea imo.
I recommend moving the whole logic from namefixer to actorfixer, only keeping the current dictionary

EllipticEllipsis:
Yes, this intended to be a replacement.

EllipticEllipsis:
And this isn't just for actors anyway, hence changing the name.

hensldm:
I like the idea of starting over with a new dictionary, the current one is getting long and messy. I'd say this is the start of deprecating `actorfixer.py` (maybe even add a warning that this is now deprecated and not maintained any more) which I'd prefer to just be deleted, but should probably stick around for a little while longer just in case.

To address bullet point 2, I don't have too strong an opinion, but I think running `namerfixer` on the whole repo should be limited to just `c` files, since as you said the others should have already been addressed by the original PR that changed the name in the first place.

tom-overton:
I really don't think we should start over with a new dictionary for a long time, potentially even ever, because people with in-progress branches will need to use it to bring their branches up-to-date. I just had to use it for that purpose today to mess with one of Boss07's non-matchings locally. If we clear the dictionary out, bringing these in-progress branches up-to-date will be very annoying (though not impossible, I'll admit).

If we deprecate actorfixer but still keep it around, then at least there is *some* way to get your code fixed, even if it takes two scripts instead of one.

TL;DR: the dictionary in actorfixer should probably live in *some* way in the repo for perpetuity, or at least until we know that everything that's in-progress won't need it.

hensldm:
Thinking about it some more, while I'd prefer to start with a new dictionary, it probably is just better to move the old one in its entirety (with a little cleanup here and there, like removing commented out ones, and with whole word replace removing the `(` from the ones that have it) to the new `namefixer.py` and just delete `actorfixer.py`.

EllipticEllipsis:
I think I've addressed everything discussed here, although there's more to talk about in Discord, I think?

hensldm (approved):
Seems fine to me. What else would need to be discussed?

AngheloAlf (approved):

Kenix3 (approved):

PR #780, opened 2022-04-03
En_Bbfall (Red Bubble) OK and documented

tom-overton:
---

Some really nasty loops in this one.


assets/xml/objects/object_bb.xml
```diff
@@ -3,7 +3,7 @@
     <File Name="object_bb" Segment="6">
         <!-- Bubble Animations -->
         <Animation Name="gBubbleAttackAnim" Offset="0x184" /> <!-- Original name is "bb_atack" -->
-        <Animation Name="object_bb_Anim_0002AC" Offset="0x2AC" /> <!-- Original name might be "bb_damage". Maybe unused? -->
+        <Animation Name="gBubbleUnusedPainAnim" Offset="0x2AC" /> <!-- Original name might be "bb_damage". Unused. -->
```
AngheloAlf, line 6: I think we should try to avoid naming stuff "unused" if we can, the comment should be enough.
Either `gBubblePainAnim` or `gBubbleDamagedAnim` should be fine

src/overlays/actors/ovl_En_Bbfall/z_en_bbfall.c
```diff
@@ -37,140 +47,717 @@ const ActorInit En_Bbfall_InitVars = {
...
+                EnBbfall_SetupDown(this);
+            }
+        }
+    }
+}
+
+void EnBbfall_Update(Actor* thisx, GlobalContext* globalCtx) {
+    EnBbfall* this = THIS;
+    Sphere16* sphere;
+    Vec3f sp5C;
```
hensldm, line 591:
```suggestion
    Vec3f diff;
```
Looks to be the diff/distance between the the previous 2 flame positions.

src/overlays/actors/ovl_En_Bbfall/z_en_bbfall.c
```diff
@@ -37,140 +47,717 @@ const ActorInit En_Bbfall_InitVars = {
...
+    }
+}
+
+void EnBbfall_UpdateDamage(EnBbfall* this, GlobalContext* globalCtx) {
+    if (this->collider.base.acFlags & AC_HIT) {
+        this->collider.base.acFlags &= ~AC_HIT;
+        this->collider.base.atFlags &= ~(AT_HIT | AT_BOUNCED);
+        this->collider.base.atFlags &= ~AT_ON;
+        if ((this->drawDmgEffType != ACTOR_DRAW_DMGEFF_FROZEN_NO_SFX) ||
+            (!(this->collider.elements[0].info.acHitInfo->toucher.dmgFlags & 0xDB0B3))) {
```
hensldm, line 539:
```suggestion
            !(this->collider.elements[0].info.acHitInfo->toucher.dmgFlags & 0xDB0B3)) {
```
Personally I don't think it is necessary to add parenthesis when the operation is just `!`, but up to you.

hensldm (approved):
Good work as always. Just a couple suggestions.

src/overlays/actors/ovl_En_Bbfall/z_en_bbfall.c
tom-overton: I agree. I copied this styling from En_Bb so let me change it there too.

src/overlays/actors/ovl_En_Bbfall/z_en_bbfall.c
```diff
@@ -37,140 +47,717 @@ const ActorInit En_Bbfall_InitVars = {
...
+                              0.15f);
+        this->actor.flags |= ACTOR_FLAG_200;
+    }
+}
+
+s32 EnBbfall_IsTouchingLava(EnBbfall* this, GlobalContext* globalCtx) {
+    if (!SurfaceType_IsWallDamage(&globalCtx->colCtx, this->actor.floorPoly, this->actor.floorBgId)) {
+        u32 floorType = func_800C99D4(&globalCtx->colCtx, this->actor.floorPoly, this->actor.floorBgId);
+
+        if ((floorType == 2) || (floorType == 3) || (floorType == 9)) {
```
jpburnett, line 214: Just random thoughts here is this checking for different lava floor types? Could this function be `EnBbfall_IsTouchingFloor`?

jpburnett (approved):
Excellent job with this, I just had one thought/question on the name, but nothing preventing my +1

src/overlays/actors/ovl_En_Bbfall/z_en_bbfall.c
tom-overton: It's definitely checking for lava floor types. The goal of this function is to return true if the floor is the type that the Red Bubble can "sink into". If I make the function return true no matter what, then this happens:

https://user-images.githubusercontent.com/12245827/162553551-46f07a00-46fc-48d0-8bea-d8506a68dc3e.mp4

I'll add a comment above the function explaining this.

AngheloAlf (approved):

Kenix3 (approved):

PR #789, opened 2022-04-12
`EnFamos` OK with `object_famos`

isghj5:
---


assets/xml/objects/object_famos.xml
```diff
@@ -1,28 +1,30 @@
...
+        <Limb Name="gFamosBodyLimb" Type="Standard" EnumName="FAMOS_LIMB_BODY" Offset="0x3CE8" />
+        <Limb Name="gFamosEmblemLimb" Type="Standard" EnumName="FAMOS_LIMB_EMBLEM" Offset="0x3CF4" />
+        <Limb Name="gFamosSwordLimb" Type="Standard" EnumName="FAMOS_LIMB_SWORD" Offset="0x3D00" />
+        <Limb Name="gFamosShieldLimb" Type="Standard" EnumName="FAMOS_LIMB_SHIELD" Offset="0x3D0C" />
+        <Limb Name="gFamosHeadLimb" Type="Standard" EnumName="FAMOS_LIMB_HEAD" Offset="0x3D18" />
+        <Skeleton Name="gFamosSkeleton" Type="Normal" LimbType="Standard" LimbNone="FAMOS_LIMB_NONE" LimbMax="FAMOS_LIMB_MAX" EnumName="ObjectFamosLimbs" Offset="0x3D38" />
+
+        <Animation Name="gFamosLowerSNSAnim" Offset="0x3DC8" />
+        <TextureAnimation Name="gFamosNormalGlowingEmblemAnimTex" Offset="0x3E30" />
+        <TextureAnimation Name="gFamosFlippedGlowingEmblemAnimTex" Offset="0x3E38" />
```
EllipticEllipsis, line 28:
```suggestion
        <TextureAnimation Name="gFamosNormalGlowingEmblemTexAnim" Offset="0x3E30" />
        <TextureAnimation Name="gFamosFlippedGlowingEmblemTexAnim" Offset="0x3E38" />
```
(https://github.com/zeldaret/ZAPD/blob/master/docs/zapd_extraction_xml_reference.md#textureanimation)
Possibly could get away with not mentioning the emblem. Could also name it to reflect that they're colour-changing ones, but doesn't matter that much.

assets/xml/objects/object_famos.xml
```diff
@@ -1,28 +1,30 @@
...
+        <Animation Name="gFamosShakeAnim" Offset="0xF8" />
+        <DList Name="gFamosHelmetDL" Offset="0x12B0" />
+        <DList Name="gFamosMainBodyDL" Offset="0x1528" />
+        <DList Name="gFamosShieldDL" Offset="0x1C88" />
+        <DList Name="gFamosSwordDL" Offset="0x1DE8" />
+        <DList Name="gFamosLightSwitchDL" Offset="0x1FA8" />
         <Texture Name="object_famos_TLUT_002068" OutName="tlut_002068" Format="rgba16" Width="4" Height="4" Offset="0x2068" />
         <Texture Name="object_famos_TLUT_002088" OutName="tlut_002088" Format="rgba16" Width="4" Height="4" Offset="0x2088" />
         <Texture Name="object_famos_TLUT_0020A8" OutName="tlut_0020A8" Format="rgba16" Width="4" Height="4" Offset="0x20A8" />
         <Texture Name="object_famos_TLUT_0020C8" OutName="tlut_0020C8" Format="rgba16" Width="4" Height="4" Offset="0x20C8" />
```
EllipticEllipsis, line 12: Is there any pattern to how these are used? If not can just name them `gFamos0TLUT` etc.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
-        { ELEMTYPE_UNK0, { 0x00000000, 0x00, 0x00 }, { 0x00002000, 0x00, 0x00 }, TOUCH_NONE | TOUCH_SFX_NORMAL, BUMP_ON, OCELEM_NONE, },
+        {
+            ELEMTYPE_UNK0,
+            { 0x00000000, 0x00, 0x00 },
+            { 0x00002000, 0x00, 0x00 },
+            TOUCH_NONE | TOUCH_SFX_NORMAL,
+            BUMP_ON,
+            OCELEM_NONE,
+        },
         { 2, { { 2500, 0, 0 }, 20 }, 100 },
```
EllipticEllipsis, line 119: The 2 is a limbIndex, this system is designed to attach the spheres to the limbs.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+        COLSHAPE_JNTSPH,
+    },
+    2,
+    sJntSphElementsInit,
 };
 
-// static InitChainEntry sInitChain[] = {
-static InitChainEntry D_808AE6B8[] = {
+#define FAMOS_ANIMATED_MATERIAL_NORMAL 0  // normal is greenish
+#define FAMOS_ANIMATED_MATERIAL_FLIPPED 1 // flipped is orange/yellowish
```
EllipticEllipsis, line 148: Is there any reason not to make this an enum? Also, throughout you can abbreviate `material` to `mat`. since it is elsewhere.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+    } else {
+        EnFamos_SetupStillIdle(this);
+    }
+}
+
+void EnFamos_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    EnFamos* this = THIS;
+
+    Collider_DestroyCylinder(globalCtx, &this->collider1);
+    Collider_DestroyCylinder(globalCtx, &this->collider2);
```
EllipticEllipsis, line 211: Is there a reasonable way to distinguish these colliders?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+            } else {
+                this->animatedMaterialIndex = FAMOS_ANIMATED_MATERIAL_FLIPPED;
+                this->flippedTimer = 100;
+                Actor_PlaySfxAtPos(&this->actor, NA_SE_EN_FAMOS_REVERSE1);
+                Actor_PlaySfxAtPos(&this->actor, NA_SE_EN_EYEGOLE_DAMAGE);
+            }
+            this->stableRotation = false;
+        }
+
+    } else if (this->flippedTimer > 0 && --this->flippedTimer == 0) { // check flipped timer
```
EllipticEllipsis, line 320:
```suggestion
    } else if ((this->flippedTimer) > 0 && (--this->flippedTimer == 0)) { // check flipped timer
```
I suppose this isn't DECR, but would it match as separate ifs? We try to avoid assignments in conditionals where possible.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+        if (this->animatedMaterialIndex != FAMOS_ANIMATED_MATERIAL_NORMAL) {
+            Actor_PlaySfxAtPos(&this->actor, NA_SE_EN_FAMOS_REVERSE2);
+        }
+        this->animatedMaterialIndex = FAMOS_ANIMATED_MATERIAL_NORMAL;
+        this->stableRotation = false;
+    }
+
+    if (this->stableRotation == false) {
+        if (this->animatedMaterialIndex != FAMOS_ANIMATED_MATERIAL_NORMAL) {
+            famosRotTest = Math_ScaledStepToS(&this->flipRot, -0x8000, 0x1000);
```
EllipticEllipsis, line 330: Something like `hasFinishedFlipping` might be more appropriate here?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+ * Summary: Famos has finished rotating toward home; Take off flying in straight line.
+ */
+void EnFamos_SetupReturnHome(EnFamos* this) {
+    this->actor.world.rot.y = this->actor.shape.rot.y;
+    this->actor.world.rot.x = -Actor_PitchToPoint(&this->actor, &this->targetDest);
+    this->actionFunc = EnFamos_ReturnHome;
+}
+
+void EnFamos_ReturnHome(EnFamos* this, GlobalContext* globalCtx) {
+    f32 distance = Actor_XZDistanceToPoint(&this->actor, &this->targetDest);
```
EllipticEllipsis, line 417: Maybe make this more specific, `distToTargetPos`?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+
+    this->actor.shape.rot.y = Actor_YawToPoint(&this->actor, &this->targetDest);
+    this->actor.world.rot.y = this->actor.shape.rot.y;
```
EllipticEllipsis, line 420: Chained assignment?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+        func_800B9010(&this->actor, NA_SE_EN_FAMOS_FLOAT_REVERSE - SFX_FLAG);
+    } else {
+        func_800B9010(&this->actor, NA_SE_EN_FAMOS_FLOAT - SFX_FLAG);
+    }
+
+    if (--this->stateTimer == 0) {
+        this->actor.world.pos.y = this->baseHeight;
+        EnFamos_SetupChase(this);
+    } else {
+        this->actor.world.pos.y = (Math_SinS(this->stateTimer * 0x1000) * 30.0f) + this->baseHeight;
```
EllipticEllipsis, line 467:
```suggestion
        this->actor.world.pos.y = Math_SinS(this->stateTimer * 0x1000) * 30.0f + this->baseHeight;
```
if you want.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+    }
+}
+
+void EnFamos_UpdateBobbingHeight(EnFamos* this) {
+    if (this->hoverClk == 0) {
+        this->hoverClk = 30;
+    }
+
+    this->hoverClk--;
+    this->actor.world.pos.y = (Math_SinS(this->hoverClk * 2184) * 10.0f) + this->baseHeight;
```
EllipticEllipsis, line 289:
```suggestion
    this->actor.world.pos.y = (Math_SinS(this->hoverClk * 0x888) * 10.0f) + this->baseHeight;
```
although it's probably something sensible in degrees anyway.

src/overlays/actors/ovl_En_Famos/z_en_famos.h
```diff
@@ -2,18 +2,60 @@
...
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ Vec3s jointTable[FAMOS_LIMB_MAX];
+    /* 0x1AC */ Vec3s morphTable[FAMOS_LIMB_MAX];
+    /* 0x1D0 */ EnFamosActionFunc actionFunc;
+    /* 0x1D4 */ u8 animatedMaterialIndex;
+    /* 0x1D5 */ u8 stableRotation;
+    /* 0x1D6 */ u8 pathNodeCount;
+    /* 0x1D7 */ u8 currentPathNode;
+    /* 0x1D8 */ u8 isCalm; // is NOT aware of player
+    /* 0x1DA */ s16 hoverClk; // 30->0, repeat, for height adjustment when hovering
```
EllipticEllipsis, line 40: Not clear what this is supposed to mean.

src/overlays/actors/ovl_En_Famos/z_en_famos.h
```diff
@@ -2,18 +2,60 @@
...
-    /* 0x0144 */ char unk_144[0x8C];
-    /* 0x01D0 */ EnFamosActionFunc actionFunc;
-    /* 0x01D4 */ char unk_1D4[0x440];
+    /* 0x000 */ Actor actor;
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ Vec3s jointTable[FAMOS_LIMB_MAX];
+    /* 0x1AC */ Vec3s morphTable[FAMOS_LIMB_MAX];
+    /* 0x1D0 */ EnFamosActionFunc actionFunc;
+    /* 0x1D4 */ u8 animatedMaterialIndex;
+    /* 0x1D5 */ u8 stableRotation;
```
EllipticEllipsis, line 36: What does this do?

EllipticEllipsis (commented):
Good job on this. I have quite a lot of comments, but most are minor.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: Since when? Why?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+    this->actor.world.rot.x = -Actor_PitchToPoint(&this->actor, &abovePlayerPos);
+    Math_StepToF(&this->actor.speedXZ, 6.0f, 0.5f);
+
+    surfaceType = func_800C9B18(&globalCtx->colCtx, this->actor.floorPoly, this->actor.floorBgId);
+    if (this->actor.xzDistToPlayer < 30.0f && this->actor.floorHeight > BGCHECK_Y_MIN && // close enough
+        (surfaceType != 0xC && surfaceType != 0xD)) { // two surfaces he wont attack you over? wonder what they are
+        EnFamos_SetupAttackAim(this);
+
+    } else if (Player_GetMask(globalCtx) == PLAYER_MASK_STONE ||
+               this->agroDistance < Actor_XZDistanceToPoint(&GET_PLAYER(globalCtx)->actor, &this->calmPos) ||
```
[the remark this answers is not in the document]
isghj5:
```c
 [this->agroDistance < Actor_XZDistanceToPoint(&GET_PLAYER(globalCtx)->actor, &this->calmPos)] ||
```
wat

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+        EnBom* blast = (EnBom*)Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_BOM, this->actor.world.pos.x,
+                                           this->actor.world.pos.y + 40.0f, this->actor.world.pos.z, 0, 0, 0, ENBOM_0);
+        if (blast != NULL) {
+            blast->timer = 0; // instant explosion
+        }
+        this->stateTimer--;
+
+    } else if (this->stateTimer == 0) {
+        Item_DropCollectibleRandom(globalCtx, &this->actor, &this->actor.world.pos,
+                                   (0xD << 4)); // random item from droptable 0xD
```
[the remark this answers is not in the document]
isghj5: Disagree. `0xD0` being passed as a parameter is unreadable magic number.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+    s32 i;
+
+    if (this->debrisTimer > 0) {
+        Gfx* dispOpa;
+        EnFamosRock* rock;
+
+        OPEN_DISPS(globalCtx->state.gfxCtx);
+        dispOpa = POLY_OPA_DISP;
+
+        gSPDisplayList(&dispOpa[0], &sSetupDL[0x96]);
```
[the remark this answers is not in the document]
isghj5: Oh this is actually worse... sSetupDL uses hex values to identify them in the list but it only goes up to 0x48... this is reeading from outside of sSetupDL, hmm

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+        rock = &this->rocks[0];
+        for (i = 0; i < ARRAY_COUNT(this->rocks); i++, rock++) {
+
+            Matrix_SetStateRotationAndTranslation(rock->pos.x, rock->pos.y, rock->pos.z, &rock->rotation);
+            Matrix_Scale(rock->scale, rock->scale, rock->scale, MTXMODE_APPLY);
+
+            gSPMatrix(&dispOpa[3 + (i * 2)], Matrix_NewMtx(globalCtx->state.gfxCtx),
+                      G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+
+            gSPDisplayList(&dispOpa[4 + (i * 2)], &gameplay_keep_DL_06AB30); // greenish brown rock DL
```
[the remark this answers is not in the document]
isghj5: ???

It's an array index, of course it uses brackets...?

src/overlays/actors/ovl_En_Famos/z_en_famos.h
isghj5: bool, true if Famos is NOT busy rotating, and is either perfectly upright or upside down

I should rename to `isNotRotating` probably

src/overlays/actors/ovl_En_Famos/z_en_famos.h
isghj5: Timer (count down), starts at 30 counts down once per frame to zero then resets to 30, used by the trig functions to identify what height adjustment Famos should have as it bobs up and down on a sin wave.

src/overlays/actors/ovl_En_Famos/z_en_famos.h
```diff
@@ -2,18 +2,60 @@
...
+    /* 0x1D8 */ u8 isCalm; // is NOT aware of player
+    /* 0x1DA */ s16 hoverClk; // 30->0, repeat, for height adjustment when hovering
+    /* 0x1DC */ s16 stateTimer; // reused for different actionFunc
+    /* 0x1DE */ s16 debrisTimer; // also counts frames until despawning bom after death
+    /* 0x1E0 */ s16 flippedTimer; // frames until it flips right side up
+    /* 0x1E2 */ s16 cratorDespawnTimer;
+    /* 0x1E4 */ s16 targetYaw;
+    /* 0x1E6 */ s16 flipRot; // is used to detect if the actor has been flipped upside down by light arrows
+    /* 0x1E8 */ Vec3s* pathPoints;
+    /* 0x1EC */ f32 baseHeight; // (because hover can hit the floor, we need to keep track so famos cannot by adjusted/moved by terain)
```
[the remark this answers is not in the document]
isghj5: Wait by "brackets" do you mean Parenthesis?

src/overlays/actors/ovl_En_Famos/z_en_famos.h
```diff
@@ -2,18 +2,60 @@
...
+    /* 0x1F0 */ f32 agroDistance;
+    /* 0x1F4 */ Vec3f targetDest;
+    /* 0x200 */ Vec3f calmPos;
+    // havent figured out what the difference between the colliders are, only one cylinder is visible in KZ
+    /* 0x20C */ ColliderCylinder collider1;
+    /* 0x258 */ ColliderCylinder collider2;
+    /* 0x2A4 */ ColliderJntSph   emblemCollider;
+    /* 0x2C4 */ ColliderJntSphElement emblemColliderElements;
+    /* 0x340 */ u8 pad340[0x40];
```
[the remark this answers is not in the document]
isghj5: What init data?

assets/xml/objects/object_famos.xml
```diff
@@ -1,28 +1,30 @@
...
+        <Texture Name="gFamosLightSwitchTex" OutName="famos_light_switch" Format="rgba16" Width="32" Height="32" Offset="0x34E8" />
+
+        <Limb Name="gFamosBodyLimb" Type="Standard" EnumName="FAMOS_LIMB_BODY" Offset="0x3CE8" />
+        <Limb Name="gFamosEmblemLimb" Type="Standard" EnumName="FAMOS_LIMB_EMBLEM" Offset="0x3CF4" />
+        <Limb Name="gFamosSwordLimb" Type="Standard" EnumName="FAMOS_LIMB_SWORD" Offset="0x3D00" />
+        <Limb Name="gFamosShieldLimb" Type="Standard" EnumName="FAMOS_LIMB_SHIELD" Offset="0x3D0C" />
+        <Limb Name="gFamosHeadLimb" Type="Standard" EnumName="FAMOS_LIMB_HEAD" Offset="0x3D18" />
+        <Skeleton Name="gFamosSkeleton" Type="Normal" LimbType="Standard" LimbNone="FAMOS_LIMB_NONE" LimbMax="FAMOS_LIMB_MAX" EnumName="ObjectFamosLimbs" Offset="0x3D38" />
+
+        <Animation Name="gFamosLowerSNSAnim" Offset="0x3DC8" />
```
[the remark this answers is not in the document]
isghj5: Common gaming acronym used for "Sword and Shield"

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: I haven't figured them out yet, my guess that one was an attack damage collider and the other was hard armor/rock collider seems to have been wrong they both seem to do both.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: chain asignment

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
hensldm: `sSetupDL` is indexed by `6 * n` where `n` is what goes up to `0x48`. So in this case it would be `sSetupDL[6 * 0x19]`.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
hensldm: It would probably be better to document `Item_DropCollectibleRandom`, as this will probably have a bit packing macro eventually, and having the documentation split through different actors will just be messy. Perhaps adding a comment like "`0-16 << 4` designates the drop table" above `Item_DropCollectibleRandom`.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
EllipticEllipsis: Since always, there's only a few in the entire codebase, most of which IIRC are required to match. It makes it clearer if you don't have to figure out if it's using the old or new (and it's bad style to mutate things in conditionals anyway).

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
EllipticEllipsis: ()

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
EllipticEllipsis: Exactly, this is something that needs centralised documentation, not isolated comments scattered throughout the codebase.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
EllipticEllipsis: Around the `i * 2`.

src/overlays/actors/ovl_En_Famos/z_en_famos.h
EllipticEllipsis: I don't think you need to explain what values it takes, someone can read the code to find that out. And what does "Clk" mean?

src/overlays/actors/ovl_En_Famos/z_en_famos.h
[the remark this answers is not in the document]
EllipticEllipsis: The `ColliderJntSphElementInit` in the C file.

src/overlays/actors/ovl_En_Famos/z_en_famos.h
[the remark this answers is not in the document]
EllipticEllipsis: Oh, is this an American thing? Most of the world mean () when they say brackets, [] are square brackets.

assets/xml/objects/object_famos.xml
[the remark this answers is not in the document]
EllipticEllipsis: Apparently this is Monster Hunter dialect, not general terminology. I'd much rather use something more recognisable, I had to Google quite specifically to find confirm that.

src/overlays/actors/ovl_En_Famos/z_en_famos.h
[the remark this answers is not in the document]
isghj5: No no that's too weird. When I think of Brackets I think of metal you use to hold things together, its usually square or angled, it's never round. "Round" is the corners cut out by the cheap companies.

src/overlays/actors/ovl_En_Famos/z_en_famos.h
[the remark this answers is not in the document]
hensldm: In particular `emblemColliderElements` should be an array of size `2` which looks to include the entire pad section.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Famos/func_808AE3FC.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Famos/EnFamos_Draw.s")
+static s32 animatedMaterialsVirtualized = false;
+
+void EnFamos_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnFamos* this = THIS;
+    Path* path;
+    // prob fake match
+    s32 sTrue = 1; // required to be separate to match
```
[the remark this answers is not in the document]
hensldm: Could you make a scratch of this? I was able to remove a similar fake match by changing an `s32` struct member to `u32`, and would like to try a few things here.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,801 @@ const ActorInit En_Famos_InitVars = {
...
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Famos/EnFamos_Draw.s")
+static s32 animatedMaterialsDesgmented = false;
+
+void EnFamos_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnFamos* this = THIS;
+    Path* path;
+    // prob fake match
+    s32 sTrue = 1; // required to be separate to match
+    int i;
```
hensldm, line 165:
```suggestion
    s32 i;
```

src/overlays/actors/ovl_En_Famos/z_en_famos.h
```diff
@@ -2,18 +2,60 @@
...
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ Vec3s jointTable[FAMOS_LIMB_MAX];
+    /* 0x1AC */ Vec3s morphTable[FAMOS_LIMB_MAX];
+    /* 0x1D0 */ EnFamosActionFunc actionFunc;
+    /* 0x1D4 */ u8 animatedMaterialIndex;
+    /* 0x1D5 */ u8 hasFinishedRotating; // stable up or down
+    /* 0x1D6 */ u8 pathNodeCount;
+    /* 0x1D7 */ u8 currentPathNode;
+    /* 0x1D8 */ u8 isCalm; // is NOT aware of player
+    /* 0x1DA */ s16 hoverClk; // timer, 30->0 then repeat, for sin height adjustment when hovering
```
hensldm, line 40: What about just `hoverTimer`? Seems clearer imo.

hensldm (commented):
A few thoughts

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,801 @@ const ActorInit En_Famos_InitVars = {
...
+
+    ActorShape_Init(&this->actor.shape, 0.0f, ActorShadow_DrawSquare, 30.0f);
+    SkelAnime_Init(globalCtx, &this->skelAnime, &gFamosSkeleton, &gFamosLowerSNSAnim, this->jointTable,
+                   this->morphTable, FAMOS_LIMB_MAX);
+    Collider_InitAndSetCylinder(globalCtx, &this->collider1, &this->actor, &sCylinderInit1);
+    Collider_InitAndSetCylinder(globalCtx, &this->collider2, &this->actor, &sCylinderInit2);
+    Collider_InitAndSetJntSph(globalCtx, &this->emblemCollider, &this->actor, &sJntSphInit,
+                              &this->emblemColliderElements);
+
+    if (!animatedMaterialsDesgmented) { // init animated materials
```
hensldm, line 186:
```suggestion
    if (!sAnimatedMaterialsDesegmented) { // init animated materials
```
Should have an `s` prefix still for `static`, as well as small typo in `desegmented`

src/overlays/actors/ovl_En_Famos/z_en_famos.h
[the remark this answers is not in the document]
MrPolymorph: Just going to chime in here and say, when programmers say brackets its pretty much '()'

src/overlays/actors/ovl_En_Famos/z_en_famos.h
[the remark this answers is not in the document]
isghj5: Oh emblemColliderElements is an array, for some reason I was fixated on finding something that would use offset 340

src/overlays/actors/ovl_En_Famos/z_en_famos.h
isghj5: Sorry Clk is common abbreviation of "Clock" in hardware, wrong project.

assets/xml/objects/object_famos.xml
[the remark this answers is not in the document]
isghj5: No no, it was used in World of Warcraft in the mid aughts at least, and I thought it was used in Everquest before that, but sure we could rename it.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +66,792 @@ const ActorInit En_Famos_InitVars = {
...
+}
+
+void EnFamos_Update(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    EnFamos* this = THIS;
+    f32 oldHeight;
+    s32 oldHoverClk; // save old value to test if changed
+
+    if (this->debrisTimer <= 0 ||
+        (this->debrisTimer--, EnFamos_UpdateDebrisPosRot(this), (this->actionFunc != EnFamos_DeathFade))) {
```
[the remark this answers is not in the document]
isghj5: Don't know how to format this one to remove it.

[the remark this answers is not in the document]
EllipticEllipsis: Could this be an early return? That might avoid the comma expression.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,809 @@ const ActorInit En_Famos_InitVars = {
...
 
-// static InitChainEntry sInitChain[] = {
-static InitChainEntry D_808AE6B8[] = {
+typedef enum {
+    /* 0 */ FAMOS_ANIMATED_MAT_NORMAL, // normal is greenish
+    /* 1 */ FAMOS_ANIMATED_MAT_FLIPPED // flipped is orange/yellowish
+} FamosAnimatedMatArrayIndexes;
+
+static AnimatedMaterial* sEmblemAnimatedMats[] = { gFamosNormalGlowingEmblemTexAnim,
+                                                   gFamosFlippedGlowingEmblemTexAnim };
```
hensldm, line 151:
```suggestion
static AnimatedMaterial* sEmblemAnimatedMats[] = { gFamosNormalGlowingEmblemTexAnim,
                                                   gFamosFlippedGlowingEmblemTexAnim, };
```
Probably formats a little better

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,809 @@ const ActorInit En_Famos_InitVars = {
...
+    Collider_InitAndSetCylinder(globalCtx, &this->collider1, &this->actor, &sCylinderInit1);
+    Collider_InitAndSetCylinder(globalCtx, &this->collider2, &this->actor, &sCylinderInit2);
+    Collider_InitAndSetJntSph(globalCtx, &this->emblemCollider, &this->actor, &sJntSphInit,
+                              this->emblemColliderElements);
+
+    if (!sAnimatedMaterialsDesgmented) { // init animated materials
+        for (i = 0; i < ARRAY_COUNT(sEmblemAnimatedMats); i++) {
+            sEmblemAnimatedMats[i] = Lib_SegmentedToVirtual(sEmblemAnimatedMats[i]);
+        }
+        sAnimatedMaterialsDesgmented = 1;
```
hensldm, line 188:
```suggestion
        sAnimatedMaterialsDesgmented = true;
```

hensldm (approved):
Just a couple more minor things.

assets/xml/objects/object_famos.xml
```diff
@@ -1,28 +1,30 @@
 ﻿<Root>
     <File Name="object_famos" Segment="6">
```
tom-overton, line 2: Can you put a header above the `File` node explaining what this object is?

assets/xml/objects/object_famos.xml
```diff
@@ -1,28 +1,30 @@
...
+        <Animation Name="gFamosShakeAnim" Offset="0xF8" />
+        <DList Name="gFamosHelmetDL" Offset="0x12B0" />
+        <DList Name="gFamosMainBodyDL" Offset="0x1528" />
+        <DList Name="gFamosShieldDL" Offset="0x1C88" />
+        <DList Name="gFamosSwordDL" Offset="0x1DE8" />
+        <DList Name="gFamosLightSwitchDL" Offset="0x1FA8" />
+        <Texture Name="gFamosTLUT2068" OutName="tlut_002068" Format="rgba16" Width="4" Height="4" Offset="0x2068" />
+        <Texture Name="gFamosTLUT2088" OutName="tlut_002088" Format="rgba16" Width="4" Height="4" Offset="0x2088" />
+        <Texture Name="gFamosTLUT20A8" OutName="tlut_0020A8" Format="rgba16" Width="4" Height="4" Offset="0x20A8" />
+        <Texture Name="gFamosTLUT20C8" OutName="tlut_0020C8" Format="rgba16" Width="4" Height="4" Offset="0x20C8" />
```
tom-overton, line 12: These TLUTs can all be named. The first is the TLUT for `gFamosMainBodyTex`, so it could be named something like `gFamosMainBodyTLUT` or something of that sort. The second is for `gFamosBottomTex`, the third is for `gFamosShieldTex`, and the last is for `gFamosSwordHiltTex`. If you're wondering how I figured that out, you can just look for which TLUT is used with `gsDPLoadTLUT_pal16` before the texture is used in a displaylist.

assets/xml/objects/object_famos.xml
```diff
@@ -1,28 +1,30 @@
...
+        <Texture Name="gFamosShieldTex" OutName="famos_shield" Format="ci4" Width="64" Height="64" Offset="0x2AE8" />
+        <Texture Name="gFamosSwordHiltTex" OutName="famos_sword_hilt" Format="ci4" Width="32" Height="32" Offset="0x32E8" />
+        <Texture Name="gFamosLightSwitchTex" OutName="famos_light_switch" Format="rgba16" Width="32" Height="32" Offset="0x34E8" />
+
+        <Limb Name="gFamosBodyLimb" Type="Standard" EnumName="FAMOS_LIMB_BODY" Offset="0x3CE8" />
+        <Limb Name="gFamosEmblemLimb" Type="Standard" EnumName="FAMOS_LIMB_EMBLEM" Offset="0x3CF4" />
+        <Limb Name="gFamosSwordLimb" Type="Standard" EnumName="FAMOS_LIMB_SWORD" Offset="0x3D00" />
+        <Limb Name="gFamosShieldLimb" Type="Standard" EnumName="FAMOS_LIMB_SHIELD" Offset="0x3D0C" />
+        <Limb Name="gFamosHeadLimb" Type="Standard" EnumName="FAMOS_LIMB_HEAD" Offset="0x3D18" />
+        <Skeleton Name="gFamosSkeleton" Type="Normal" LimbType="Standard" LimbNone="FAMOS_LIMB_NONE" LimbMax="FAMOS_LIMB_MAX" EnumName="ObjectFamosLimbs" Offset="0x3D38" />
```
tom-overton, line 24: Since you use `FAMOS_LIMB_[whatever]` for the enum values, I'd drop the "Object" from the `EnumName` like this:
```suggestion
        <Skeleton Name="gFamosSkeleton" Type="Normal" LimbType="Standard" LimbNone="FAMOS_LIMB_NONE" LimbMax="FAMOS_LIMB_MAX" EnumName="FamosLimbs" Offset="0x3D38" />
```

Since we know this is a Death Armos, I think it'd be better to use `DEATH_ARMOS_LIMB_[whatever]` and then call it `DeathArmosLimbs`, but the consistency is the important point. If you *do* change the name like that, you probably want to swap all the `gFamos[whatever]` stuff in here to `gDeathArmos[whatever]`

isghj5: Until we decide when/if we change actor names from ripped names to official names, this just seems to make all variable names longer IMO.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5:
```
-static AnimatedMaterial* sEmblemAnimatedMats[] = { gFamosNormalGlowingEmblemTexAnim,
-                                                   gFamosFlippedGlowingEmblemTexAnim, };
+static AnimatedMaterial* sEmblemAnimatedMats[] = {
+    gFamosNormalGlowingEmblemTexAnim,
+    gFamosFlippedGlowingEmblemTexAnim,
+};
```
Not really convinced this is better, but don't care enough to not leave as it is now.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,811 @@ const ActorInit En_Famos_InitVars = {
...
+void EnFamos_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    EnFamos* this = THIS;
+
+    Collider_DestroyCylinder(globalCtx, &this->collider1);
+    Collider_DestroyCylinder(globalCtx, &this->collider2);
+    Collider_DestroyJntSph(globalCtx, &this->emblemCollider);
+}
+
+/**
+ * Summary: Sets 20 rocks to draw from an explosion on the ground (slam attack).
```
EllipticEllipsis, line 217: You don't have to write "Summary" every time: if it's not detailed documentation you can just use `@brief` or make the comment.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,811 @@ const ActorInit En_Famos_InitVars = {
...
+        }
+    } else if (distanceToHome < 40.0f) {
+        Math_StepToF(&this->actor.speedXZ, 0.5f, 0.3f);
+    } else {
+        Math_StepToF(&this->actor.speedXZ, 3.0f, 0.3f);
+    }
+}
+
+/**
+ * Summary: Famos has spotted the player; Delayed by surprise.
```
EllipticEllipsis, line 449: Either this is two sentences with a `.` (although the second one is a fragment), or it's one with a `;` and no capital D.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,811 @@ const ActorInit En_Famos_InitVars = {
...
+    this->actor.world.rot.x = -Actor_PitchToPoint(&this->actor, &abovePlayerPos);
+    Math_StepToF(&this->actor.speedXZ, 6.0f, 0.5f);
+
+    surfaceType = func_800C9B18(&globalCtx->colCtx, this->actor.floorPoly, this->actor.floorBgId);
+    if (this->actor.xzDistToPlayer < 30.0f && this->actor.floorHeight > BGCHECK_Y_MIN && // close enough
+        (surfaceType != 0xC && surfaceType != 0xD)) {
+        EnFamos_SetupAttackAim(this);
+
+    } else if (Player_GetMask(globalCtx) == PLAYER_MASK_STONE ||
+               (this->aggroDistance < Actor_XZDistanceToPoint(&GET_PLAYER(globalCtx)->actor, &this->calmPos)) ||
```
EllipticEllipsis, line 509: ()

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,811 @@ const ActorInit En_Famos_InitVars = {
...
+
+    Math_StepToF(&this->actor.speedXZ, 20.0f, 2.0f);
+    this->stateTimer--;
+    if (this->stateTimer == 0) {
+        this->emblemCollider.base.acFlags &= ~AC_ON;
+    }
+
+    surfaceType = func_800C9B18(&globalCtx->colCtx, this->actor.floorPoly, this->actor.floorBgId);
+    hitFloor = this->actor.bgCheckFlags & 1;
+    if (hitFloor || this->actor.floorHeight == BGCHECK_Y_MIN || (surfaceType == 0xC || surfaceType == 0xD)) {
```
EllipticEllipsis, line 548:
```suggestion
    if (hitFloor || (this->actor.floorHeight == BGCHECK_Y_MIN) || (surfaceType == 0xC) || (surfaceType == 0xD)) {
```
Unless you think it's semantically valuable to have () round both; it should be equivalent?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,811 @@ const ActorInit En_Famos_InitVars = {
...
+    Math_StepToF(&this->actor.speedXZ, 5.0f, 0.3f);
+    if (this->actor.speedXZ > 1.0f) {
+        if (ABS_ALT(this->flipRot) > 0x4000) {
+            func_800B9010(&this->actor, NA_SE_EN_FAMOS_FLOAT_REVERSE - SFX_FLAG);
+        } else {
+            func_800B9010(&this->actor, NA_SE_EN_FAMOS_FLOAT - SFX_FLAG);
+        }
+    }
+
+    if (this->baseHeight < this->actor.world.pos.y || this->actor.bgCheckFlags & 0x10) { // touching ceiling
```
EllipticEllipsis, line 620: ()

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,811 @@ const ActorInit En_Famos_InitVars = {
...
+}
+
+/**
+ * Summary: Famos was upside down and hit the ground headfirst. This kills the Famos.
+ */
+void EnFamos_SetupDeathSlam(EnFamos* this) {
+    this->emblemCollider.base.acFlags &= ~AC_ON;
+    this->stateTimer = 20;
+    this->actor.speedXZ = 0.0f;
+    Actor_SetColorFilter(&this->actor, 0x4000, 0xFF, false, 20);
```
EllipticEllipsis, line 655:
```suggestion
    Actor_SetColorFilter(&this->actor, 0x4000, 255, 0, 20);
```
The fourth argument is not boolean, for example
```
src/overlays/actors/ovl_En_Firefly/z_en_firefly.c:        Actor_SetColorFilter(&this->actor, 0x4000, 255, 0x2000, 40);
src/overlays/actors/ovl_En_Firefly/z_en_firefly.c:        Actor_SetColorFilter(&this->actor, 0x4000, 255, 0, 40);
src/overlays/actors/ovl_En_Firefly/z_en_firefly.c:        Actor_SetColorFilter(&this->actor, 0, 255, 0x2000, this->timer);
src/overlays/actors/ovl_En_Firefly/z_en_firefly.c:        Actor_SetColorFilter(&this->actor, 0, 255, 0, this->timer);
```
(applies elsewhere)

EllipticEllipsis (commented):
There's still a few unaddressed things here. I spotted a few more things as well.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,811 @@ const ActorInit En_Famos_InitVars = {
...
+        rock->pos.y = randPlusMinusPoint5Scaled(60.0f) + (this->actor.world.pos.y + 40.0f);
+        rock->pos.z = Math_CosS(randVelDirection) * 20.0f + this->actor.world.pos.z;
+        rock->scale = Rand_ZeroFloat(0.002f) + (2.5f * 0.001f);
+    }
+}
+
+s32 EnFamos_IsPlayerSeen(EnFamos* this, GlobalContext* globalCtx) {
+    if (Player_GetMask(globalCtx) != PLAYER_MASK_STONE &&
+        Actor_XZDistanceToPoint(&GET_PLAYER(globalCtx)->actor, &this->calmPos) < this->aggroDistance &&
+        (Actor_IsFacingPlayer(&this->actor, 0x5000))) {
```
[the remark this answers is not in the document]
isghj5: Your choices on which expressions gets parenthesis is confusing... the second one is long and complicated, parenthesis helps visually isolate, but the top one is simple, it doesn't seem necessary to me.

I will admit the last one does not need any.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: You... want me to remove it from this one? It already has one
```
                this->aggroDistance < Actor_XZDistanceToPoint(&GET_PLAYER(globalCtx)->actor, &this->calmPos) ||
```
adding one more seems silly

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: I do like to visually isolate when one or the other condition are relevant to each other (in this case which surface type would appear to be its own meta condition _if surface type is valid exception_), even if it matches without, but I could remove them all or add to each here, either or seems fine.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: Weird, not sure how I got the impression that was a bool

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
EllipticEllipsis: It's around the arithmetic comparisons, to distinguish them from the subsequent logical operations.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
isghj5: I agree there should be macros for this, but it is beyond the scope of this PR to add one. Until the macros are added I see no reason to remove documentation explaining what the code is doing. At the time the macro is added, it is highly likely the macros will be applied to existing code, at which time this documentation can be easily spotted as they are on the same line.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,810 @@ const ActorInit En_Famos_InitVars = {
...
+    if (!sAnimatedMaterialsDesgmented) { // init animated materials
+        for (i = 0; i < ARRAY_COUNT(sEmblemAnimatedMats); i++) {
+            sEmblemAnimatedMats[i] = Lib_SegmentedToVirtual(sEmblemAnimatedMats[i]);
+        }
+        sAnimatedMaterialsDesgmented = true;
+    }
+
+    this->actor.colChkInfo.mass = 250;
+    this->baseHeight = this->actor.world.pos.y;
+    // params: [this->actor.shape.rot.x] is used to set agro distance
```
EllipticEllipsis, line 195:
```suggestion
    // params: [this->actor.shape.rot.x] is used to set aggro distance
```
Consider making a define for this instead.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,810 @@ const ActorInit En_Famos_InitVars = {
...
+    this->actor.world.rot.y = this->actor.shape.rot.y;
+    abovePlayerPos.x = player->actor.world.pos.x;
+    abovePlayerPos.y = player->actor.world.pos.y + 100.0f;
+    abovePlayerPos.z = player->actor.world.pos.z;
+    this->actor.world.rot.x = -Actor_PitchToPoint(&this->actor, &abovePlayerPos);
+    Math_StepToF(&this->actor.speedXZ, 6.0f, 0.5f);
+
+    surfaceType = func_800C9B18(&globalCtx->colCtx, this->actor.floorPoly, this->actor.floorBgId);
+    if (this->actor.xzDistToPlayer < 30.0f && this->actor.floorHeight > BGCHECK_Y_MIN && // close enough
+        (surfaceType != 0xC && surfaceType != 0xD)) {
```
EllipticEllipsis, line 505:
```suggestion
    if ((this->actor.xzDistToPlayer < 30.0f) && (this->actor.floorHeight > BGCHECK_Y_MIN) && // close enough
        (surfaceType != 0xC && surfaceType != 0xD)) {
```
This can't be 2 ifs, I suppose? Comment not needed IMO.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,810 @@ const ActorInit En_Famos_InitVars = {
...
+        EnFamos_SetupAttackAim(this);
+
+    } else if ((Player_GetMask(globalCtx) == PLAYER_MASK_STONE) ||
+               (this->aggroDistance < Actor_XZDistanceToPoint(&GET_PLAYER(globalCtx)->actor, &this->calmPos)) ||
```
EllipticEllipsis, line 509: Possibly this should be swapped so the `this->aggroDistance` is second, if it matches.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,810 @@ const ActorInit En_Famos_InitVars = {
...
+    Math_StepToF(&this->actor.speedXZ, 5.0f, 0.3f);
+    if (this->actor.speedXZ > 1.0f) {
+        if (ABS_ALT(this->flipRot) > 0x4000) {
+            func_800B9010(&this->actor, NA_SE_EN_FAMOS_FLOAT_REVERSE - SFX_FLAG);
+        } else {
+            func_800B9010(&this->actor, NA_SE_EN_FAMOS_FLOAT - SFX_FLAG);
+        }
+    }
+
+    if (this->baseHeight < this->actor.world.pos.y || (this->actor.bgCheckFlags & 0x10)) { // touching ceiling
```
EllipticEllipsis, line 620:
```suggestion
    if ((this->baseHeight < this->actor.world.pos.y) || (this->actor.bgCheckFlags & 0x10)) { // touching ceiling
```

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
EllipticEllipsis: Assuming it's not done via a script.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
EllipticEllipsis: Can it be?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
```diff
@@ -28,133 +62,810 @@ const ActorInit En_Famos_InitVars = {
...
+s32 EnFamos_OverrideLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot,
+                             Actor* thisx) {
+    EnFamos* this = THIS;
+
+    if (limbIndex == FAMOS_LIMB_BODY) {
+        Matrix_InsertTranslation(0.0f, 4000.0f, 0.0f, MTXMODE_APPLY);
+        Matrix_InsertZRotation_s(this->flipRot, MTXMODE_APPLY);
+        Matrix_InsertTranslation(0.0f, -4000.0f, 0.0f, MTXMODE_APPLY);
+
+    } else if (this->flippedTimer < 0 && // if set to -1, famos is dying
```
EllipticEllipsis, line 806:
```suggestion
    } else if ((this->flippedTimer < 0) && // if set to -1, famos is dying
```
This would probably also look better as 2 ifs.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: order

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: Why do you want to keep separating if conditions?

src/overlays/actors/ovl_En_Famos/z_en_famos.c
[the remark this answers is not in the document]
isghj5: Screenshot_2022-05-14_14-04-10
I dont see how, it introduces another branch

src/overlays/actors/ovl_En_Famos/z_en_famos.c
isghj5: Screenshot_2022-05-14_14-11-43
Doesn't look like it, because of the following elseif: If condition 1 == TRUE If condition 2 == false leads to falling out of of the if/elseif instead of being caught by the elseif.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
EllipticEllipsis: I will believe you if you just say it doesn't match, don't have to show a screenshot!

src/overlays/actors/ovl_En_Famos/z_en_famos.c
EllipticEllipsis: It makes them easier to read than if they're one huge run-on line that gets split up. And it makes sense, if the things that are being checked are sufficiently different.

src/overlays/actors/ovl_En_Famos/z_en_famos.c
EllipticEllipsis: Yeah, that makes sense.

EllipticEllipsis (approved):

Kenix3 (approved):

PR #792, opened 2022-04-17
Load OK

hensldm:
---

Matches all the functions in both loadfragment.c and loadfragment2.c. 
I used oot's us 1.0 and mm's jp 1.1 to determine the debugging if statements (both seem to have not stubbed out `osSyncPrintf`).
Comments come from OOT.


include/functions.h
```diff
@@ -104,11 +104,11 @@ void FaultDrawer_SetInputCallback(FaultDrawerCallback callback);
...
-void Load_Relocate(u32 allocatedVRamAddr, OverlayRelocationSection* overlayInfo, u32 vRamStart);
-s32 Load_LoadOverlay(u32 vRomStart, u32 vRomEnd, u32 vRamStart, u32 allocatedVRamAddr, u32 allocatedBytes);
+void Load_Relocate(void* allocatedVRamAddr, OverlayRelocationSection* overlayInfo, u32 vRamStart);
+size_t Load_LoadOverlay(u32 vRomStart, u32 vRomEnd, u32 vRamStart, void* allocatedVRamAddr, size_t allocatedBytes);
 void* Load_AllocateAndLoad(u32 vRomStart, u32 vRomEnd, u32 vRamStart);
-void Load2_Relocate(u32 allocatedVRamAddr, OverlayRelocationSection* overlayInfo, u32 vRamStart);
-s32 Load2_LoadOverlay(u32 vRomStart, u32 vRomEnd, u32 vRamStart, u32 vRamEnd, u32 allocatedVRamAddr);
+void Load2_Relocate(void* allocatedVRamAddr, OverlayRelocationSection* overlayInfo, u32 vRamStart);
+size_t Load2_LoadOverlay(u32 vRomStart, u32 vRomEnd, u32 vRamStart, u32 vRamEnd, void* allocatedVRamAddr);
 void* Overlay_AllocateAndLoad(u32 vRomStart, u32 vRomEnd, u32 vRamStart, u32 vRamEnd);
```
EllipticEllipsis, line 112: uintptr_t's insteda of `u32`s, I think, at least for VRAM. I can't remember what we decided for VROM.

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,171 @@
...
+    if (gLoadLogSeverity >= 3) {}
+
+    DmaMgr_SendRequest0(allocatedVRamAddr, vRomStart, size);
+
+    if (gLoadLogSeverity >= 3) {}
+
+    ovlOffset = (uintptr_t)end - 4;
+    ovl = (OverlayRelocationSection*)((uintptr_t)end - ((s32*)end)[-1]);
+
+    if (1) {}
```
hensldm, line 141: Just noting here. This actually does not match as `if(gLoadLogSeverity >= 3){}` so feels like a fake match. Based on jp 1.1 I think there should be a debug statement here, but its hard to tell with just the raw m2c output.

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,171 @@
...
+    for (i = 0; i < ovl->nRelocations; i++) {
+        reloc = ovl->relocations[i];
+        relocDataP = (u32*)(sections[reloc >> 0x1E] + (reloc & 0xFFFFFF));
+
+        switch (reloc & 0x3F000000) {
+            case 0x2000000:
+                /* R_MIPS_32
+                 * Handles 32-bit address relocation.  Used in things such as
+                 * jump tables.
+                 */
```
EllipticEllipsis, line 35: Formatting looks a bit funny here, can the second two lines be combined? Also in-function comments are usually `//`

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,171 @@
...
+                if ((*relocDataP & 0xF000000) == 0) {
```
EllipticEllipsis, line 36: Hmm, not clear if it's better with a `!` or not. I suppose this is to rule out segmented stuff? Actually I'm not sure. Either way it probably needs a comment.

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,171 @@
...
+                 */
+                luiRefs[(*relocDataP >> 0x10) & 0x1F] = relocDataP;
+                luiVals[(*relocDataP >> 0x10) & 0x1F] = *relocDataP;
+                break;
+            case 0x6000000:
+                /* R_MIPS_LO16
+                 * Updates the LUI instruction to reflect the relocated address.
+                 * The full address is calculated from the LUI and lo parts, and then updated.
+                 * if the lo part is negative, add 1 to the lui.
+                 */
```
EllipticEllipsis, line 63: I mean, it also has to update the other half of the address usually 🤔 That's what the main part is doing, I suppose.

src/boot_O2/loadfragment2.c
```diff
@@ -1,120 +1,120 @@
...
-#ifdef NON_MATCHING
-// This needs lots of work. Mostly regalloc and getting the address of D_80096C30 placed in s5 at the beginning of the
-// function
-void Load2_Relocate(u32 allocatedVRamAddr, OverlayRelocationSection* overlayInfo, u32 vRamStart) {
-    s32 sectionLocations[4];
-    u32* regReferences[32];
-    u32 regValues[32];
+s32 gLoad2LogSeverity = 2;
+
+void Load2_Relocate(void* allocatedVRamAddr, OverlayRelocationSection* ovl, u32 vRamStart) {
```
EllipticEllipsis, line 6: I think it's worth pointing out the difference between the two functions/files, it's confusing enough having two slightly different versions of the same file, only one of which is used, I think?

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,171 @@
...
+                luiInstRef = luiRefs[(*relocDataP >> 0x15) & 0x1F];
+                regValP = &luiVals[((*relocDataP) >> 0x15) & 0x1F];
+                if ((((*luiInstRef << 0x10) + (s16)*relocDataP) & 0x0F000000) == 0) {
+                    relocatedAddress = (*regValP << 0x10) + (s16)*relocDataP - vRamStart + allocu32;
+                    isLoNeg = (relocatedAddress & 0x8000) ? 1 : 0;
+                    *luiInstRef = (*luiInstRef & 0xFFFF0000) | (((relocatedAddress >> 0x10) & 0xFFFF) + isLoNeg);
+                    *relocDataP = (*relocDataP & 0xFFFF0000) | (relocatedAddress & 0xFFFF);
+                } else {
+                    if (gLoadLogSeverity >= 3) {}
```
EllipticEllipsis, line 72: Does it make sense for this to be an else if, or is there stuff in other versions that rules that out?

EllipticEllipsis (commented):
Good job on this, a few comments based on having looked at this stuff for too long.

src/boot_O2/loadfragment2.c
hensldm: Yeah, only Load2 is actually used. To be honest I don't really understand the differences between the two files, other than that Load2 is used and Load just exists.

src/boot_O2/loadfragment.c
hensldm: Yeah kind of mentioned in `R_MIPS_HI16`, though I updated the comments to hopefully be more clear.

src/boot_O2/loadfragment.c
hensldm: I can't seem to find the answer, but what all is `R_MIPS_32` used for? I don't think its just jump tables.

src/boot_O2/loadfragment.c
hensldm: Possibly segmented stuff, but this would also reject address 0x81000000 to 0x8FFFFFFF. Not quite sure what the check is for. Note, the same condition is also used in the `R_MIPS_LO16` case.

src/boot_O2/loadfragment.c
EllipticEllipsis: It's any time you get a full word that's an address, so jump tables, tables of functions, etc.

src/boot_O2/loadfragment.c
EllipticEllipsis: Yeah, I suppose that might include stuff like hardware registers, although this is beyond the limits of my knowledge. You know who's going to know, though.

hensldm: Added a small comment based on discord convo

src/boot_O2/loadfragment2.c
hensldm: So looking at it closer, I noticed that the `Load` functions (the unused one in `loadfragment.c`) don't have `vRamEnd` arguments. I think its purpose was to dynamically determine `vRamEnd` and why it calculates `ovl->bssSize + size` and reallocates if `size` is to small.

src/boot_O2/loadfragment.c
```diff
@@ -29,54 +29,59 @@ void Load_Relocate(void* allocatedVRamAddr, OverlayRelocationSection* ovl, u32 v
...
-                /* R_MIPS_LO16
-                 * Updates the LUI instruction to reflect the relocated address.
-                 * The full address is calculated from the LUI and lo parts, and then updated.
-                 * if the lo part is negative, add 1 to the lui.
-                 */
+                // R_MIPS_LO16
+                // Handles relocation for a lui instruction, part 2.
+                // Updates the LUI instruction to reflect the relocated address.
+                // The full address is calculated from the LUI and lo parts, and then updated.
+                // If the lo part is negative, add 1 to the lui.
```
EllipticEllipsis, line 65: It's probably truer to say that together these handle the halves of a hi/lo pair. The hi part is always a `lui`, the bottom part can be pretty much any other immediate instruction.

hensldm: Perhaps something like
```c
case 0x5000000:
                // R_MIPS_HI16
                // Handles relocation for a hi/lo pair, part 1.
                // Store the reference to the instruction, and update it in the R_MIPS_LO16 section.
...
case 0x6000000:
                // R_MIPS_LO16
                // Handles relocation for a hi/lo pair, part 2.
                // Updates the LUI instruction (hi) to reflect the relocated address.
                // The full address is calculated from the LUI and lo parts, and then updated.
                // If the lo part is negative, add 1 to the LUI.

Comment (EllipticEllipsis) [reply src/boot_O2/loadfragment.c:65]: I think that's clearer, yes. Would also like @Thar0 's approval, though.

File: src/boot_O2/loadfragment.c [hunk 9]
```diff
@@ -1,8 +1,187 @@
...
+                }
+                break;
+
+            case 0x4000000:
+                // R_MIPS_26
+                // Handles 26-bit address relocation, used for jumps and jals.
+
+                *relocDataP =
+                    (*relocDataP & 0xFC000000) |
+                    ((((((*relocDataP & 0x3FFFFFF) << 2) | 0x80000000) - vRamStart + allocu32) & 0xFFFFFFF) >> 2);
```
Thar0, line 59: These various shifts are extracting the target field from a J-type MIPS instruction (`j` and `jal`)
Not sure if we want to make that more clear or just leave it as-is.
If we had the `R4300.h` header I'd suggest changing 0x80000000 to the `K0BASE` define, but it's up to you whether you want to add that header now.

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,187 @@
...
+                break;
+
+            case 0x5000000:
+                // R_MIPS_HI16
+                // Handles relocation for a hi/lo pair, part 1.
+                // Store the reference to the LUI instruction, and update it in the R_MIPS_LO16 section.
+
+                luiRefs[(*relocDataP >> 0x10) & 0x1F] = relocDataP;
+                luiVals[(*relocDataP >> 0x10) & 0x1F] = *relocDataP;
```
Thar0, line 68: `>> 0x10 & 0x1F` extracts the `rt` field, again not sure if that should be made more clear

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,187 @@
...
+
+            case 0x6000000:
+                // R_MIPS_LO16
+                // Handles relocation for a hi/lo pair, part 2.
+                // Updates the LUI instruction (hi) to reflect the relocated address.
+                // The full address is calculated from the LUI and lo parts, and then updated.
+                // If the lo part is negative, add 1 to the LUI.
+
+                luiInstRef = luiRefs[(*relocDataP >> 0x15) & 0x1F];
+                regValP = &luiVals[(*relocDataP >> 0x15) & 0x1F];
```
Thar0, line 79: As above but it's the `rs` field here

Thar0 (commented):
Some thoughts on documentation, most comments apply to both files

src/boot_O2/loadfragment.c
hensldm: Would it be fine to just take the header from OOT? Also would using `PHYS_TO_K0` be fine, or should I use the K0Base define directly?

include/functions.h
```diff
@@ -264,7 +264,7 @@ void __osPackReadData(void);
 uintptr_t osVirtualToPhysical(void* vaddr);
 u32 __osGetSR(void);
 void __osSetSR(u32 value);
-void osWritebackDCache(void* vaddr, s32 nbytes);
+void osWritebackDCache(void* vaddr, size_t nbytes);
```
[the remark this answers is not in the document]
hensldm: Just a note, this does affect matching in `Load2_LoadOverlay` in that I had to change size to `s32` instead of `size_t`. `Load_LoadOverlay` seems fine to leave as `size_t`.

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,187 @@
...
+    u32* luiRefs[32];
+    u32 luiVals[32];
+    u32 isLoNeg;
 
-#pragma GLOBAL_ASM("asm/non_matchings/boot/loadfragment/Load_AllocateAndLoad.s")
+    if (gLoadLogSeverity >= 3) {}
+
+    sections[0] = 0;
+    sections[1] = allocu32;
+    sections[2] = allocu32 + ovl->textSize;
```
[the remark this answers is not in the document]
hensldm: Unfortunately it causes regalloc issues.

src/boot_O2/loadfragment.c
```diff
@@ -1,8 +1,187 @@
...
+    sections[3] = sections[2] + ovl->dataSize;
+
+    for (i = 0; i < ovl->nRelocations; i++) {
+        reloc = ovl->relocations[i];
+        relocDataP = (u32*)(sections[reloc >> 0x1E] + (reloc & 0xFFFFFF));
+
+        switch (reloc & 0x3F000000) {
+            case 0x2000000:
```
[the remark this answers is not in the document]
hensldm: Decided to go with the value that would match the elf file, but just shift it in the case.

src/boot_O2/loadfragment.c
hensldm: Yeah, not sure either if that needs to be clearer. Thoughts from others?

EllipticEllipsis: I think a comment about it being the destination register would be reasonable. Either that or explain actually dissecting the instruction into `opcode | rt | imm` and so on for each.

hensldm: Does this seem fine? Explaining the process a little more in detail in the header comments of each case
```
            case R_MIPS_32 << 24:
                // Handles 32-bit address relocation, used for things such as jump tables and pointers in data.
                // Just relocate the full address

            [...]

            case R_MIPS_26 << 24:
                // Handles 26-bit address relocation, used for jumps and jals.
                // Extract the address from the target field of the J-type MIPS instruction.
                // Relocate the address and update the instruction.

               [...]

            case R_MIPS_HI16 << 24:
                // Handles relocation for a hi/lo pair, part 1.
                // Store the reference to the LUI instruction (hi) using the `rt` register of the instruction.
                // This will be updated later in the `R_MIPS_LO16` section.

               [...]

            case R_MIPS_LO16 << 24:
                // Handles relocation for a hi/lo pair, part 2.
                // Grab the stored LUI (hi) from the `R_MIPS_HI16` section using the `rs` register of the instruction.
                // The full address is calculated, relocated, and then used to update both the LUI and lo parts.
                // If the lo part is negative, add 1 to the LUI.

               [...]
```

hensldm: I'm also debating if I should add a
`#define RELOCATE_ADDRESS(address, vRamStart, allocu32) ((address) - (vramStart) + (allocu32))`
macro just to make it clear what is actually getting relocated, but not sure.

src/boot_O2/loadfragment.c
Thar0: Sorry I missed this, I think it looks good although strictly the LO instruction must be one with a signed immediate like `addiu`, due to the `If the lo part is negative, add 1 to the LUI.`. This would break if for example `ori` was relocated.

src/boot_O2/loadfragment.c
Thar0: `PHYS_TO_K0` like you've done looks good. It's better to use that where possible, I failed to notice it was applicable here 😅

src/boot_O2/loadfragment.c
AngheloAlf: I like the idea of this macro

hensldm: Added. Let me know what you think.

AngheloAlf (approved):

EllipticEllipsis (approved):
Looks good now, the other stuff might get documented further in future but for now this is fine IMO. Couple of remaining things, mainly to avoid putting *even more* in z64.h.

include/z64.h
```diff
@@ -287,6 +287,19 @@ typedef enum IRQ_TYPE {
...
+
+#define RELOC_SECTION(reloc) ((reloc) >> 0x1E)
+#define RELOC_OFFSET(reloc) ((reloc) & 0xFFFFFF)
+#define RELOC_TYPE(reloc) ((reloc) & 0x3F000000)
+
+typedef enum OverlayRelocationType {
+	R_MIPS_32 = 2,
+	R_MIPS_26 = 4,
+	R_MIPS_HI16 = 5,
+	R_MIPS_LO16 = 6
```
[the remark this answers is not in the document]
hensldm: Personally I feel this stuff makes more sense to be in the same header as `OverlayRelocationSection`, which we know gets used in multiple files in the GC roms. I could split all of them out into a `z64load.h` or something like that.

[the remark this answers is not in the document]
EllipticEllipsis: I think you might as well, it's a classic case where only a few files should ever need to access it, so that sounds appropriate.

hensldm:
That is the last little bit of changes to keep things consistent with Oot. Not sure if you want to review again @AngheloAlf, @EllipticEllipsis, @Thar0.

src/boot_O2/loadfragment.c
```diff
@@ -38,29 +38,30 @@ void Load_Relocate(void* allocatedVRamAddr, OverlayRelocationSection* ovl, uintp
         reloc = ovl->relocations[i];
         relocDataP = (u32*)(sections[RELOC_SECTION(reloc)] + RELOC_OFFSET(reloc));
 
-        switch (RELOC_TYPE(reloc)) {
-            case R_MIPS_32 << 24:
+        switch (RELOC_TYPE_MASK(reloc)) {
+            case R_MIPS_32 << RELOC_TYPE_SHIFT:
```
EllipticEllipsis, line 42: I don't think making defines for shifts is generally worth it over actual shifting macros most of the time, although I sort of see the point here.

Kenix3 (approved):

PR #794, opened 2022-04-18
En_Raf (Carnivorous Lily Pad) OK and documented, object_raf documented

tom-overton:
---

I don't think the names in this actor are really the best, but they're about as good as I can get by myself; please feel free to comment on the bad ones and suggest improvements. I also didn't document the damage table because I believe it's irrelevant? If you feed it an explosive, it will blow up, and otherwise the damage table isn't even looked at.


hensldm (approved):
Seems good to me. Can't think of any improvements at the moment.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Raf/func_80A175E4.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Raf/func_80A17848.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Raf/func_80A178A0.s")
+/**
+ * Sets the `index`th pixel of the trap teeth texture to 0 (transparent black)
+ * according to the `clearPixelTable`
+ */
+void EnRaf_ClearPixelsTeeth(u16* texture, u8* clearPixelTable, s32 index) {
```
EllipticEllipsis, line 185: This is one pixel at a time since it's an rgba16, so I think to singular instead.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+void EnRaf_ClearPixelsPetal(u16* texture, u8* clearPixelTable, s32 index) {
+    if (clearPixelTable[index] != 0) {
+        texture[index] = 0;
+    }
+}
+
+void EnRaf_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnRaf* this = THIS;
+    Vec3f limbScale = { 1.0f, 1.0f, 1.0f };
+    s32 pad;
```
EllipticEllipsis, line 204: Does the pad have to go exactly here?

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+    &gCarnivorousLilyPadSpitAnim, &gCarnivorousLilyPadConvulseAnim, &gCarnivorousLilyPadDeathAnim,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Raf/func_80A179C8.s")
+static u8 sAnimationModes[] = {
+    ANIMMODE_ONCE, ANIMMODE_ONCE, ANIMMODE_LOOP, ANIMMODE_ONCE, ANIMMODE_LOOP, ANIMMODE_ONCE,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Raf/func_80A17C6C.s")
+void EnRaf_ChangeAnimation(EnRaf* this, s32 index) {
```
EllipticEllipsis, line 272: Might as well go in the function, I suppose.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+            return;
+        }
+
+        if ((globalCtx->gameplayFrames % 2) == 0) {
+            return;
+        }
+
+        explosive = globalCtx->actorCtx.actorLists[ACTORCAT_EXPLOSIVES].first;
+        while (explosive != NULL) {
+            if ((EnRaf*)explosive == this) {
```
EllipticEllipsis, line 337: Well that's weird.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+
+void EnRaf_SetupConvulse(EnRaf* this) {
+    EnRaf_ChangeAnimation(this, EN_RAF_ANIMATION_CONVULSE);
+    this->chewCount = 0;
+    this->action = EN_RAF_ACTION_CONVULSE;
+    this->actionFunc = EnRaf_Convulse;
+}
+
+/**
+ * Plays the convulsing animation and sets the lily pad's switch flag to prevent them from
```
EllipticEllipsis, line 568:
```suggestion
 * Plays the convulsing animation and sets the lily pad's switch flag to prevent it from
```

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+
+    if (this->endFrame <= curFrame) {
+        this->dissolveTimer++;
+        if (this->dissolveTimer < (BREG(3) + 105)) {
+            for (i = 0; i < (BREG(4) + 5); i++) {
+                EnRaf_ClearPixelsPetal(Lib_SegmentedToVirtual(&gCarnivorousLilyPadTrapPetalTex),
+                                       sPetalClearPixelTableFirstPass, this->petalClearPixelFirstPassIndex);
+                EnRaf_ClearPixelsTeeth(Lib_SegmentedToVirtual(&gCarnivorousLilyPadTrapTeethTex),
+                                       sTeethClearPixelTableFirstPass, this->teethClearPixelFirstPassIndex);
+                if (this->petalClearPixelFirstPassIndex < 0x200) {
```
EllipticEllipsis, line 611: I suppose this is related to the texture dimensions?

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+    }
+
+    this->actionFunc = EnRaf_DeadIdle;
+}
+
+/**
+ * Simply sits around doing nothing. If the revive timer is non-zero, then this function
+ * will decrement the revive timer and revive the trap petals once it reaches 0.
+ */
+void EnRaf_DeadIdle(EnRaf* this, GlobalContext* globalCtx) {
```
EllipticEllipsis, line 663: I suppose there is a question of whether it can be dead if it gets better... how about "dormant"?

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+                (limbIndex == CARNIVOROUS_LILY_PAD_LIMB_TRAP_3_MIDDLE_SEGMENT) ||
+                (limbIndex == CARNIVOROUS_LILY_PAD_LIMB_TRAP_2_MIDDLE_SEGMENT)) {
+                Math_Vec3f_Copy(&this->targetLimbScale[limbIndex], &sMiddleSegmentTargetScaleDuringGrab[2]);
+            } else if ((limbIndex == CARNIVOROUS_LILY_PAD_LIMB_TRAP_1_UPPER_SEGMENT) ||
+                       (limbIndex == CARNIVOROUS_LILY_PAD_LIMB_TRAP_3_UPPER_SEGMENT) ||
+                       (limbIndex == CARNIVOROUS_LILY_PAD_LIMB_TRAP_2_UPPER_SEGMENT)) {
+                Math_Vec3f_Copy(&this->targetLimbScale[limbIndex], &sUpperSegmentTargetScaleDuringGrab[2]);
+            }
+
+            // These matrix operations make the trap petals look a bit more "wobbly" as it chews.
```
EllipticEllipsis, line 831: I know this was discussed in Discord, I think it essentially stretches them a bit in a particular direction, right? it undoes all the limb rotations, scales/stretches, then puts them all back to get it in the same place.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -80,65 +178,787 @@ static DamageTable D_80A1939C = {
...
+    func_8012C2DC(globalCtx->state.gfxCtx);
+    SkelAnime_DrawTransformFlexOpa(globalCtx, this->skelAnime.skeleton, this->skelAnime.jointTable,
+                                   this->skelAnime.dListCount, NULL, NULL, EnRaf_TransformLimbDraw, &this->dyna.actor);
+
+    if (this->action == EN_RAF_ACTION_EXPLODE) {
+        EnRaf_DrawParticles(this, globalCtx);
+    }
+}
+
+void EnRaf_InitializeParticle(EnRaf* this, Vec3f* position, Vec3f* velocity, Vec3f* acceleration, f32 scale,
```
EllipticEllipsis, line 883: I think OoT has moved away from "particle" to just calling these "special effects" or something, would recommend investigating the current terminology to keep us in sync. (I don't have a strong opinion on which is better, so happy to go the OoT way.)

src/overlays/actors/ovl_En_Raf/z_en_raf.h
```diff
@@ -2,16 +2,64 @@
...
 #include "global.h"
+#include "objects/object_raf/object_raf.h"
+
+#define EN_RAF_GET_TYPE(thisx) (((thisx)->params >> 12) & 0xF)
+#define EN_RAF_GET_REVIVE_TIMER(thisx) (((thisx)->params >> 7) & 0x1F)
+#define EN_RAF_GET_SWITCH_FLAG(thisx) ((thisx)->params & 0x7F)
+
+typedef enum {
+    /* 0 */ EN_RAF_TYPE_NORMAL,
+    /* 1 */ EN_RAF_TYPE_ALEADY_DEAD,          // Spawns without trap, so it can't eat bombs/player
```
EllipticEllipsis, line 13: Nice, keeping up the Hokuto no Ken references...

src/overlays/actors/ovl_En_Raf/z_en_raf.h
```diff
@@ -2,16 +2,64 @@
...
+    /* 2 */ EN_RAF_TYPE_NO_WATER_INTERACTIONS // Won't produce ripples, and particles won't produce splashes
```
EllipticEllipsis, line 14: I suppose this is never used, since afaik there's only a few of these in the entire game?

src/overlays/actors/ovl_En_Raf/z_en_raf.h
```diff
@@ -2,16 +2,64 @@
...
+    /* 0x1E8 */ Vec3s morphTable[CARNIVOROUS_LILY_PAD_LIMB_MAX];
+    /* 0x230 */ EnRafActionFunc actionFunc;
+    /* 0x234 */ Vec3f limbScale[CARNIVOROUS_LILY_PAD_LIMB_MAX];
+    /* 0x2C4 */ Vec3f targetLimbScale[CARNIVOROUS_LILY_PAD_LIMB_MAX];
+    /* 0x354 */ Vec3s chewLimbRot[CARNIVOROUS_LILY_PAD_LIMB_MAX];
+    /* 0x39C */ s16 grabTarget;
+    /* 0x39E */ u8 isCurrentlyInRidingMovingState;
+    /* 0x3A0 */ f32 endFrame;
+    /* 0x3A4 */ f32 chewScale;
+    /* 0x3A8 */ f32 heightDiffFromBob;
```
EllipticEllipsis, line 44: I think just call this "bobOffset` or something... otherwise given how it's next to `heightDiffFromPlayer`, the inevitable question is "who's Bob?".

src/overlays/actors/ovl_En_Raf/z_en_raf.h
```diff
@@ -2,16 +2,64 @@
...
+    /* 0x3AC */ f32 heightDiffFromPlayer;
+    /* 0x3B0 */ f32 bobPhase;
+    /* 0x3B4 */ s16 timer;
+    /* 0x3B6 */ s16 dissolveTimer;
+    /* 0x3B8 */ s16 rippleTimer;
+    /* 0x3BA */ s16 reviveTimer;
+    /* 0x3BC */ s16 playerYRotWhenGrabbed;
```
EllipticEllipsis, line 51: Maybe `RotY` over `YRot`?

src/overlays/actors/ovl_En_Raf/z_en_raf.h
```diff
@@ -2,16 +2,64 @@
...
+    /* 0x3BE */ s16 type;
```
EllipticEllipsis, line 52: Better make this `mainType`, I think, or it disappears into all the other types.

src/overlays/actors/ovl_En_Raf/z_en_raf.h
```diff
@@ -2,16 +2,64 @@
...
+    /* 0x3C0 */ s16 switchFlag;
+    /* 0x3C2 */ s16 petalScaleType;
+    /* 0x3C4 */ s16 chewCount;
+    /* 0x3C6 */ s16 action;
+    /* 0x3C8 */ s16 petalClearPixelFirstPassIndex;
+    /* 0x3CA */ s16 teethClearPixelFirstPassIndex;
+    /* 0x3CC */ s16 petalClearPixelSecondPassIndex;
```
EllipticEllipsis, line 59: Do these come from OoT? I remember several OoT actors (Gohma, Phantom Ganon) do similar things.

EllipticEllipsis (commented):
Good job on this, I think most of the names are fine, but I have a few thoughts for where they might be improved.

src/overlays/actors/ovl_En_Raf/z_en_raf.h
tom-overton: I'll be 100% honest, it actually came from Bloodborne: https://www.youtube.com/watch?v=4F54phkQgiU

But that in itself might be a reference to something else. Either way, this is probably getting replaced if I switch from `DeadIdle` to `Dormant`, but it was fun while it lasted.

assets/xml/objects/object_raf.xml
```diff
@@ -1,42 +1,58 @@
...
+        <!-- Carnivorous Lily Pad Textures -->
+        <Texture Name="gCarnivorousLilyPadRootsTex" OutName="carnivorous_lily_pad_roots" Format="rgba16" Width="16" Height="32" Offset="0x24F8" />
+        <Texture Name="gCarnivorousLilyPadFlowerPetalTex" OutName="carnivorous_lily_pad_flower_petal" Format="rgba16" Width="16" Height="32" Offset="0x28F8" />
+        <Texture Name="gCarnivorousLilyPadFlowerCenterTex" OutName="carnivorous_lily_pad_flower_center" Format="rgba16" Width="16" Height="16" Offset="0x2CF8" />
+        <Texture Name="gCarnivorousLilyPadTrapPetalTex" OutName="carnivorous_lily_pad_trap_petal" Format="rgba16" Width="16" Height="32" Offset="0x2EF8" />
+        <Texture Name="gCarnivorousLilyPadTrapTeethTex" OutName="carnivorous_lily_pad_trap_teeth" Format="rgba16" Width="8" Height="8" Offset="0x32F8" />
+        <Limb Name="gCarnivorousLilyPadFlowerLimb" Type="Standard" EnumName="CARNIVOROUS_LILY_PAD_LIMB_FLOWER" Offset="0x3378" />
+
+        <!-- Carnivorous Lily Pad Limbs -->
+        <Limb Name="gCarnivorousLilyPadTrap1LowerSegmentLimb" Type="Standard" EnumName="CARNIVOROUS_LILY_PAD_LIMB_TRAP_1_LOWER_SEGMENT" Offset="0x3384" />
```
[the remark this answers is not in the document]
tom-overton: Whoops, no it is not. Thanks for catching it.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Yes, it has to be right there to get the stack to line up. I'm guessing the `GlobalContext *` temp was just there for some reason instead of at the top.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Yep, it's 16 * 32. Should I express it as such, like this? I would do this with all texture dimension stuff.
```c
if (this->petalClearPixelFirstPassIndex < (16 * 32))
```

src/overlays/actors/ovl_En_Raf/z_en_raf.h
tom-overton: Yeah, there are only 7 of these things, and every single one is of type `EN_RAF_TYPE_NORMAL`.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Yeah, I have no clue why they did this either. It's definitely some sort of sanity check, but it seems excessive.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Yeah, that's what it does. I'll make the comment a little more clear as to what exactly is happening here.

src/overlays/actors/ovl_En_Raf/z_en_raf.h
tom-overton: Gohma uses `decayingProgress`, whereas Phantom Ganon uses `work[GND_BODY_DECAY_INDEX]`. However, Gohma has `sClearPixelTableFirstPass` and `sClearPixelTableSecondPass`, which is where I got the names for these things.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Hmmm, well, most of the bosses call them "effects" because they use `globalCtx->specialEffects`. I can find one instance of an actor in OoT using "effect" *without* using `globalCtx->specialEffects`, which is En_G_Switch. Meanwhile, there are a few more instances of "particle" being used for this purpose instead, including for actors whose code for handling these things looks nearly-identical to Raf's. Unfortunately, there isn't a ton of consistency in OoT either. My gut says to leave this for now and we can decide on the particle vs effect nomenclature later, but what do you think?

src/overlays/actors/ovl_En_Raf/z_en_raf.c
EllipticEllipsis: Yes, once zeldaret/ZAPD#233 is in we'll have macros for them anyway though.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
EllipticEllipsis: Yeah, just unusual, that's all. Presumably the inline data forces the stack to be exactly this.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Ah, well, let me just do them manually for now then. If I forget to come back to this once the ZAPD change is merged, feel free to remind me.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
EllipticEllipsis: More than that, it just doesn't make any sense... I'd be tempted to `@bug` it except that it doesn't do anything.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
EllipticEllipsis: Yeah, let's not worry about it for now then... I did think there'd been a push for consensus on effects. 🤔

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Yeah, that's probably it. I remember having a devil of a time getting the stack in this function to work because of this pad.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: I did a quick discord search for "particle" and I didn't find anything recently. So if there *was* a consensus, maybe it was a while ago?

src/overlays/actors/ovl_En_Raf/z_en_raf.h
EllipticEllipsis: https://knowyourmeme.com/memes/you-are-already-dead-omae-wa-mou-shindeiru

src/overlays/actors/ovl_En_Raf/z_en_raf.c
tom-overton: Do you want me to slap a comment above this saying it's a pointless check?

EllipticEllipsis: Yes, that's probably worth doing, if it was totally obvious it wouldn't still be there... I hope.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -182,8 +182,8 @@ static DamageTable sDamageTable = {
  * Sets the `index`th pixel of the trap teeth texture to 0 (transparent black)
  * according to the `clearPixelTable`
  */
-void EnRaf_ClearPixelsTeeth(u16* texture, u8* clearPixelTable, s32 index) {
-    if ((index < 0x40) && (clearPixelTable[index] != 0)) {
+void EnRaf_ClearPixelTeeth(u16* texture, u8* clearPixelTable, s32 index) {
```
EllipticEllipsis, line 185: I suppose the other thing to note here is that this function will *only* work with a texture that uses 16-bit pixels, hence why `texture` is a `u16*` here.

src/overlays/actors/ovl_En_Raf/z_en_raf.c
```diff
@@ -334,6 +332,7 @@ void EnRaf_Idle(EnRaf* this, GlobalContext* globalCtx) {
 
         explosive = globalCtx->actorCtx.actorLists[ACTORCAT_EXPLOSIVES].first;
         while (explosive != NULL) {
+            // This check is pointless, since EnRaf never spawns as an explosive.
```
EllipticEllipsis, line 335: I think I'd just say
```suggestion
            // This check is pointless, since EnRaf is never in the explosive category.
```

EllipticEllipsis (approved):

Kenix3 (approved):
We have generally been preferring to use the term "effects" over "particles," but no reason to hold up this PR in my opinion. We have a lot of other places in MM that are using the term "particle right now."

Other than that, the names aren't bad.

PR #795, opened 2022-04-18
Document object_boss02 (Twinmold assets)

tom-overton:
---

Relatively straightforward object, though the dust thing took me some time and testing to figure out. I couldn't grab animation names from MM3D because the way Twinmold's rig is set up in that game is so different; his head and body are just a single skeleton instead of being split up like in MM64.

### DIFF
--- a/assets/xml/objects/object_boss02.xml
+++ b/assets/xml/objects/object_boss02.xml
@@ -1,99 +1,125 @@
 ﻿<Root>
+    <!-- Assets for the Twinmold boss fight (Twinmold itself, the props in its arena, its title card, etc.)-->
     <File Name="object_boss02" Segment="6">
-        <Texture Name="object_boss02_Tex_000000" OutName="tex_000000" Format="i4" Width="32" Height="32" Offset="0x0" />
-        <DList Name="object_boss02_DL_000230" Offset="0x230" />
-        <DList Name="object_boss02_DL_0002E0" Offset="0x2E0" />
-        <Texture Name="object_boss02_Tex_000300" OutName="tex_000300" Format="rgba16" Width="4" Height="4" Offset="0x300" />
-        <Texture Name="object_boss02_TLUT_000320" OutName="tlut_000320" Format="rgba16" Width="4" Height="4" Offset="0x320" />
-        <Texture Name="object_boss02_TLUT_000340" OutName="tlut_000340" Format="rgba16" Width="4" Height="4" Offset="0x340" />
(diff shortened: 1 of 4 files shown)


EllipticEllipsis (approved):

assets/xml/objects/object_boss02.xml
```diff
@@ -1,99 +1,125 @@
...
+        <Texture Name="gTwinmoldVeinsTex" OutName="twinmold_veins" Format="ci4" Width="64" Height="64" Offset="0x1BA0" />
+        <Texture Name="gTwinmoldTeethTex" OutName="twinmold_teeth" Format="rgba16" Width="16" Height="32" Offset="0x23A0" />
+        <Texture Name="gTwinmoldUndersideTex" OutName="twinmold_underside" Format="ci4" Width="64" Height="64" Offset="0x27A0" />
+        <Texture Name="gTwinmoldSpikeAndMandibleTex" OutName="twinmold_spike_and_mandible" Format="ci4" Width="64" Height="64" Offset="0x2FA0" />
+        <Texture Name="gTwinmoldAntennaTex" OutName="twinmold_antenna" Format="rgba16" Width="16" Height="16" Offset="0x37A0" />
+        <Texture Name="gTwinmoldTailGemTex" OutName="twinmold_tail_gem" Format="rgba16" Width="32" Height="32" Offset="0x39A0" />
+        <Texture Name="gTwinmoldBlueSkinTex" OutName="twinmold_blue_skin" Format="ci4" Width="64" Height="64" Offset="0x41A0" TlutOffset="0x300"/>
+
+        <!-- Unused, completely motionless Twinmold Head animation -->
+        <Animation Name="gTwinmoldHeadStationaryAnim" Offset="0x4A1C" />
```
hensldm, line 27: Hmm, is this not used when a Twinmold dies and the head falls to the ground?

hensldm (approved):
LGTM

assets/xml/objects/object_boss02.xml
tom-overton: That may have been its intention, but that's not what it's used for in the final game (it's not used for anything, mind you). I think what happens is they just stop calling `SkelAnime_Update` once it dies.

tom-overton: Just looked at it in-game and I'm almost certain I'm right. The head's animation "freezes" in whatever state it's in when it dies, rather than going back to the "default" state shown in this unused animation.

Kenix3 (approved):

PR #800, opened 2022-04-19
EffectSsLightning OK

hensldm:
---

More or less just copy paste from OOT with just a few differences in how the `gfx` and `tex` are initialized.


src/overlays/effects/ovl_Effect_Ss_Lightning/z_eff_ss_lightning.c
```diff
@@ -1,29 +1,157 @@
 /*
  * File: z_eff_ss_lightning.c
  * Overlay: ovl_Effect_Ss_Lightning
- * Description:
+ * Description: Lightning
```
EllipticEllipsis, line 4: As in electrical arcs, presumably, rather than weather? I think "lightning" is usually reserved for cloud-based arcs.
> Lightning is a naturally occurring [electrostatic discharge](https://en.wikipedia.org/wiki/Electrostatic_discharge) during which two [electrically charged](https://en.wikipedia.org/wiki/Electric_charge) regions, both in the [atmosphere](https://en.wikipedia.org/wiki/Atmosphere) or with one on the ground, temporarily neutralize themselves,

(https://en.wikipedia.org/wiki/Lightning)

(Yes I know it's in the name, but better to be more precise if possible, I think.)

src/overlays/effects/ovl_Effect_Ss_Lightning/z_eff_ss_lightning.c
```diff
@@ -1,29 +1,157 @@
...
+        texIndex = ARRAY_COUNT(sLightningTextures) - 1;
+    }
+
+    SkinMatrix_SetTranslate(&mfTrans, this->pos.x, this->pos.y, this->pos.z);
+    xzScale = yScale * 0.6f;
+    SkinMatrix_SetScale(&mfScale, xzScale, yScale, xzScale);
+    SkinMatrix_SetRotateRPY(&mfRot, this->vec.x, this->vec.y, this->rYaw);
+    SkinMatrix_MtxFMtxFMult(&mfTrans, &globalCtx->billboardMtxF, &mfTransBillboard);
+    SkinMatrix_MtxFMtxFMult(&mfTransBillboard, &mfRot, &mfTransBillboardRot);
+    SkinMatrix_MtxFMtxFMult(&mfTransBillboardRot, &mfScale, &mfResult);
```
EllipticEllipsis, line 114: Wow, what a weird way to do this.

EllipticEllipsis (commented):
Good job on this, just one thing.

src/overlays/effects/ovl_Effect_Ss_Lightning/z_eff_ss_lightning.c
hensldm: Fair point. Probably better to just go with the general `Electric Discharge` since it does have nothing to do with weather. (It also doesn't look to be used in MM, just OOT leftover for specific enemies such as the jellyfish).

EllipticEllipsis (approved):

src/overlays/effects/ovl_Effect_Ss_Lightning/z_eff_ss_lightning.c
```diff
@@ -1,29 +1,157 @@
...
+void EffectSsLightning_Update(GlobalContext* globalCtx, u32 index, EffectSs* this) {
+    s32 pad;
+    Vec3f pos;
+    s32 yaw;
+    f32 scale;
+    s32 sign;
+
+    if ((this->rNumBolts != 0) && ((this->life + 1) == this->rLifespan)) {
+        sign = (Rand_ZeroOne() < 0.5f) ? -1 : 1;
+        yaw = this->rYaw + sign * ((s32)((Rand_ZeroOne() * 3640.0f)) + 0xE38);
```
Rozelette, line 142:
```suggestion
        yaw = this->rYaw + sign * ((s32)((Rand_ZeroOne() * 3640.0f)) + 3640);
```

EllipticEllipsis: This is an angle, so for now it goes in hex. We'd write both in hex if we had support for hex floats, but unfortunately that's C99. It's really 20 degrees, it seems (0x10000 * 20 / 360).

Rozelette: Ah that's right, it is an angle. Although it's a bit unfortunate, since it hides the fact that both values are the same.

I wonder if that degree function would work as a macro (outside the scope of this pr ofc)

jpburnett (approved):

Kenix3 (approved):

PR #803, opened 2022-04-20
Document object_dekunuts (and minor cleanup of object_dnk)

tom-overton:
---

Pretty simple object that's mostly an OoT port. While I was putting the limb enums in for EnDnk, I noticed that the enum for object_dnk isn't in the form we like; we prefer *not* to reference the object in the enum name, so I went ahead and did some minor cleanup on that.


hensldm (approved):

assets/xml/objects/object_dekunuts.xml
```diff
@@ -1,42 +1,55 @@
...
+        <Animation Name="gDekuScrubSpitAnim" Offset="0x168" /> <!-- Original name is "dn_atack" -->
+
+        <!-- Deku Scrub Limb DisplayLists -->
+        <DList Name="gDekuScrubRightFootDL" Offset="0xA00" />
+        <DList Name="gDekuScrubRightLegDL" Offset="0xAA8" />
+        <DList Name="gDekuScrubLeftFootDL" Offset="0xB40" />
+        <DList Name="gDekuScrubLeftLegDL" Offset="0xBE8" />
+        <DList Name="gDekuScrubSnoutDL" Offset="0xC80" />
+        <DList Name="gDekuScrubCrestDL" Offset="0xD90" />
+        <DList Name="gDekuScrubHairDL" Offset="0xE30" />
```
EllipticEllipsis, line 13: I don't think "hair" is a very good description for this, even if it does seem to be coming out of its head. Deku Link seems to have actual hair, which is pretty weird, but every other Deku it's more a sort of leafy shawl or cloak or hood?

tom-overton: One of the reasons I named it "hair" is because of this:
NKZE01_2022-04-21_15-37-59
To make the Deku Scrubs look like this, they just rotate this limb upwards. Would it make sense for a "cloak" or "shawl" or "hood" to stick straight up like this? An argument *against* hair is that they just wanted some visual variety, so they did the rotation without caring if it made any logical sense. I'm totally sympathetic to this line of thinking, but I'm curious if seeing this screenshot changes your opinion on my naming. If it doesn't, that's totally understandable, and I'll change it.

EllipticEllipsis: Ah, I hadn't realised it was the same thing. I think basically I don't want to call it that when it looks like leaves.... "Headdress"?

tom-overton: Sure, that works too. Let's go with that for now.

EllipticEllipsis (approved):

Kenix3 (approved):

PR #818, opened 2022-05-03
En_Snowman (Eeno and snowballs) OK and documented. object_snowman documented

tom-overton:
---

This was a tough one to document, so I hope it's understandable to someone who isn't me. I also documented the [Mega Eeno](https://www.youtube.com/watch?v=EI__RYjSnrY) bug, which has a very silly cause. I set up the `work` variable the way I did precisely to show how this variable reuse causes the Mega Eeno bug.

In EnSnowman_Init, there's an `if (1)` that I can't seem to get rid of. If you want to try and get rid of it, I made a scratch to check out: https://decomp.me/scratch/5DzO7


src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -42,155 +73,1123 @@ const ActorInit En_Snowman_InitVars = {
...
+
+    this->actor.scale.y = this->actor.scale.x;
+    this->actor.scale.z = this->actor.scale.x;
+}
+
+void EnSnowman_UpdateDamage(EnSnowman* this, GlobalContext* globalCtx) {
+    if (this->collider.base.acFlags & AC_HIT) {
+        this->collider.base.acFlags &= ~AC_HIT;
+        Actor_SetDropFlag(&this->actor, &this->collider.info);
+        if ((this->actor.colChkInfo.damageEffect != EN_SNOWMAN_DMGEFF_HOOKSHOT) ||
```
EllipticEllipsis, line 968: I think we can actually name all the damage effects after the thing it does rather than the thing that does it: hookshot does nothing, Zora is Electric Stun, Fire Arrows is Melt?

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -42,155 +73,1123 @@ const ActorInit En_Snowman_InitVars = {
...
+        DECR(this->combineTimer);
+
+        EnSnowman_UpdateDamage(this, globalCtx);
+        this->actionFunc(this, globalCtx);
+
+        if (this->actionFunc != EnSnowman_MarkForDeath) {
+            Actor_MoveWithGravity(&this->actor);
+            if ((EN_SNOWMAN_GET_TYPE(&this->actor) == EN_SNOWMAN_TYPE_LARGE) &&
+                (this->actionFunc == EnSnowman_ReadySnowball)) {
+                wallCheckRadius = (this->skelAnime.curFrame * 0.016666668f) + 1.0f;
```
EllipticEllipsis, line 1031: `1.0f / 60.0f`

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -42,155 +73,1123 @@ const ActorInit En_Snowman_InitVars = {
...
+    s16 scale;
+    s32 i;
+
+    if (this->timer > 0) {
+        this->timer--;
+    } else {
+        this->collider.base.ocFlags1 |= OC1_ON;
+    }
+
+    if ((this->actor.bgCheckFlags & 8) || (this->actor.bgCheckFlags & 1) || (this->actor.bgCheckFlags & 0x10) ||
```
EllipticEllipsis, line 1082: Hmm, interesting. I wonder if these flags are always checked like this.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -42,155 +73,1123 @@ const ActorInit En_Snowman_InitVars = {
...
+ */
+static Vec3f sBodyBottomBodyPartOffsets[] = {
+    { 2000.0f, 3000.0f, 0.0f }, { 2000.0f, -2000.0f, 0.0f }, { 3000.0f, 0.0f, 0.0f },
+    { 1000.0f, 0.0f, 3000.0f }, { 1000.0f, 0.0f, -3000.0f },
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Snowman/func_80B189C4.s")
+void EnSnowman_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx) {
+    EnSnowman* this = THIS;
+    s32 pad;
```
EllipticEllipsis, line 1138: Necessary, can it be above THIS?

src/overlays/actors/ovl_En_Snowman/z_en_snowman.h
```diff
@@ -2,16 +2,56 @@
...
+                    s16 snowballsToThrowBeforeIdling;
+                };
+    /* 0x28E */ s16 snowPileTargetRotY;
+    /* 0x290 */ s16 combineTimer;
+    /* 0x294 */ f32 eenoScale;
+    /* 0x298 */ union {
+                    f32 frameToStartHoldingSnowball;
+                    f32 frameToThrowSnowball;
+                    f32 targetScaleDuringCombine;
+                } work;
```
EllipticEllipsis, line 47: Usually fwork for floats... indeed, I'm not sure we usually use work for unions. Also, I think this needs to telegraph better that it holds 3 distinct things. Yes it's the source of a glitch, but we need to signpost it so that people don't make the same mistake if modding!

src/overlays/actors/ovl_En_Snowman/z_en_snowman.h
```diff
@@ -2,16 +2,56 @@
...
+    /* 0x280 */ EnSnowmanActionFunc actionFunc;
+    /* 0x284 */ EnSnowmanActionFunc oldActionFunc;
+    /* 0x288 */ u8 isHoldingSnowball;
+    /* 0x289 */ u8 combineState;
+    /* 0x28A */ u8 turningOnSteepSlope;
+    /* 0x28B */ u8 drawDmgEffType;
+    /* 0x28C */ union {
+                    s16 timer;
+                    s16 snowballsToThrowBeforeIdling;
+                };
```
EllipticEllipsis, line 39: I don't think we should use anonymous unions to hold genuinely distinct things.

EllipticEllipsis (commented):
Very good job on this, I found nicer numbers for the floats, and a few other things.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -42,155 +73,1123 @@ const ActorInit En_Snowman_InitVars = {
...
+            Actor_MarkForDeath(this->actor.child);
+            Actor_MarkForDeath(&this->actor);
+        } else {
+            Actor_MarkForDeath(&this->actor);
+        }
+    }
+
+    this->actor.scale.y = this->timer * 0.0002f * this->eenoScale;
+    this->actor.scale.x = (this->eenoScale * 0.0139999995f) - (0.4f * this->actor.scale.y);
+    this->actor.scale.z = (this->eenoScale * 0.0139999995f) - (0.4f * this->actor.scale.y);
```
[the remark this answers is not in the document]
tom-overton: Actually neither work.
`1.4f / 100.0f`
image

`1.4f * 0.01f`
image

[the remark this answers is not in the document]
tom-overton: Never mind, just have to wrap it in parentheses.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
tom-overton: Yeah, it can go above THIS.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -42,155 +73,1123 @@ const ActorInit En_Snowman_InitVars = {
...
+            this->actor.flags |= ACTOR_FLAG_400;
+            this->collider.base.ocFlags1 |= OC1_ON;
+            this->combineState = EN_SNOWMAN_COMBINE_STATE_BEING_ABSORBED_OR_DONE;
+            this->eenoScale = 2.0f;
+            EnSnowman_SetupMoveSnowPile(this);
+        }
+    }
+
+    this->actor.scale.y = this->actor.scale.x;
+    this->actor.scale.z = this->actor.scale.x;
```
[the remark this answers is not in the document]
tom-overton: Can't seem to get this to work chained. It messes with the stack when I do `this->actor.scale.z = this->actor.scale.y = this->actor.scale.x;`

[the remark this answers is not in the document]
EllipticEllipsis: Surprising, but never mind.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -42,155 +73,1123 @@ const ActorInit En_Snowman_InitVars = {
...
+        Math_Vec3f_Copy(&this->combinePos, combinePos);
+        this->combineState = EN_SNOWMAN_COMBINE_STATE_ACTIVE;
+
+        if (this->actionFunc != EnSnowman_Submerge) {
+            this->combineTimer = 400;
+
+            //! @bug: Skipping this based the current actionFunc results in whatever is currently in work
+            // being treated as the target scale when this function is called from EnSnowman_Submerge.
+            // Since the only other things that get put in work are animation frame numbers, this results
+            // in an enormous target scale, potentially creating the so-called "Mega Eeno" glitch.
```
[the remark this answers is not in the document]
tom-overton: No, because of how EnSnowman_Combine works. If two Eenos have the same scale, then the one that is currently executing EnSnowman_Combine absorbs the other one. The bug gives the Eeno a massive target scale, but it has to call EnSnowman_Combine to start changing its actual scale, and it grows rather slowly. So even if an Eeno is setup to become the Mega Eeno, it could be absorbed before it even gets a chance to run EnSnowman_Combine.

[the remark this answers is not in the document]
EllipticEllipsis: Hmm, that needs to be explained a bit more explicitly somewhere, I think. Also, when you do a `@bug` comment, you have to do `//!` on every line or Doxygen ignores it, it turns out.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
tom-overton: Hookshot is weird in that Link hooks onto and pulls himself towards big Eenos, but it damages small Eenos. I'm not sure what to even call that.

EllipticEllipsis: That's handled by the collider rather than the damage effect system, right? But in that case it probably should stay as hookshot.

tom-overton: Yeah, it's handled by the collider. I'll leave it as HOOKSHOT and change the rest to reflect what they do.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.h
tom-overton: When you say "telegraph better", do you mean changing the name of `fwork` to something else? Or do you mean some sort of comment? I'm not sure what I would name this that would better signal it holds distinct things.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.h
tom-overton: I named it `work` for now, but my comment on `fwork` above still applies; I'm not sure there's a better name for something that holds two different things.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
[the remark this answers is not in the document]
tom-overton: OK, added a sentence to this comment with that info. Let me know if it can be improved.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
tom-overton: Yeah, it seems macro-y, but this is also the only instance in the entire repo of them checking the flags in this order.

EllipticEllipsis: Yeah, this seems more an anomaly than anything, they check several at once elsewhere.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.h
EllipticEllipsis: Yeah, but my point is that you don't want it to look like "just another struct" member to someone, and at least a named union differentiates it from that. There's also the option of using defines, which is what some actors in OoT do, but I'd rather wait until we've settled what that project's going to do before making a decision here.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.h
EllipticEllipsis: I think we'll have to live with a named union for now. As I said below, defines are an option as well, but best to make a project-wide decision on that.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
```diff
@@ -642,14 +642,14 @@ void EnSnowman_Melt(EnSnowman* this, GlobalContext* globalCtx) {
         }
     }
 
-    this->actor.scale.y = this->timer * 0.0002f * this->eenoScale;
-    this->actor.scale.x = (this->eenoScale * 0.0139999995f) - (0.4f * this->actor.scale.y);
-    this->actor.scale.z = (this->eenoScale * 0.0139999995f) - (0.4f * this->actor.scale.y);
+    this->actor.scale.y = this->work.timer * 0.0002f * this->eenoScale;
+    this->actor.scale.x = (this->eenoScale * (1.4f / 100.0f)) - (0.4f * this->actor.scale.y);
+    this->actor.scale.z = (this->eenoScale * (1.4f / 100.0f)) - (0.4f * this->actor.scale.y);
```
EllipticEllipsis, line 647: I suspect that 0.01f is going to me more in line with the rest of these eventually, but it doesn't really matter.

EllipticEllipsis (approved):

src/overlays/actors/ovl_En_Snowman/z_en_snowman.c
tom-overton: Yeah, good point. I'll switch it to 0.01f.

src/overlays/actors/ovl_En_Snowman/z_en_snowman.h
tom-overton: Yeah, it's unfortunate that OoT hasn't decided on a solution for this yet, but whatever it is, we should be consistent. I'll leave this (and the other one) as-is for now.

hensldm (approved):

Kenix3 (approved):

PR #819, opened 2022-05-08
Change `Rand_Next` to `u32`, document `rand.c` a bit more

EllipticEllipsis:
---

Given the amount of times this is used as u32 vs s32, this seems more likely to me. It's used as s32 in specific patterns, for certain shifts, usually when used as an angle, and when comparing to 0 (presumably to get a straight 0.5 probability split), whereas basically everything else has it as u32 or doesn't mind.

I also added some documentation based on reading the relevant section in *Numerical Recipes in C*, and tweaked the code in the last couple of functions to be more like OoT's current version. Also worth noting is that the `-1.5f` does not mess with the instructions farther up, but the `-1.0f` does!


src/overlays/actors/ovl_En_Kusa/z_en_kusa.c
```diff
@@ -400,11 +400,11 @@ void EnKusa_Init(Actor* thisx, GlobalContext* globalCtx) {
...
-        D_80936CD8 = ((u32)Rand_Next() >> 0x10);
-        D_80936CDA = ((u32)Rand_Next() >> 0x10);
-        D_80936CDC = ((u32)Rand_Next() >> 0x10);
-        D_80936CDE = ((u32)Rand_Next() >> 0x10);
-        D_80936CE0 = ((u32)Rand_Next() >> 0x10);
+        D_80936CD8 = (Rand_Next() >> 0x10);
+        D_80936CDA = (Rand_Next() >> 0x10);
+        D_80936CDC = (Rand_Next() >> 0x10);
+        D_80936CDE = (Rand_Next() >> 0x10);
+        D_80936CE0 = (Rand_Next() >> 0x10);
```
hensldm, line 407:
```suggestion
        D_80936CD8 = Rand_Next() >> 0x10;
        D_80936CDA = Rand_Next() >> 0x10;
        D_80936CDC = Rand_Next() >> 0x10;
        D_80936CDE = Rand_Next() >> 0x10;
        D_80936CE0 = Rand_Next() >> 0x10;
```

src/boot_O2/rand.c
```diff
@@ -62,12 +65,19 @@ u32 Rand_Next_Variable(u32* rndNum) {
...
  * Generates the next pseudo-random floating-point number between 0.0f and
  * 1.0f from the provided rndNum.
+ *
+ * @remark This is also recommended by Numerical Recipes, pp. 284-5:
+ * > If you need floating-point values instead of 32-bit integers, and want to avoid a divide by floating-point 2^{32},
+ * > a dirty trick is to mask in an exponent that makes the value lie between 1 and 2, then subtract 1.0.
+ * > [...]
+ * > Your authors have tried very hard to make almost all of the material in this book machine and compiler independent
+ * > - indeed, even programming language independent. This subsection is a rare aberration. Forgive us. Once in a great
+ * > while the temptation to be really dirty is just irresistible."
```
hensldm, line 75: Probably don't need to quote the book. I think the reference is enough.

src/boot_O2/rand.c
```diff
@@ -77,8 +87,8 @@ f32 Rand_ZeroOne_Variable(u32* rndNum) {
  */
 f32 Rand_Centered_Variable(u32* rndNum) {
     u32 next = (*rndNum * RAND_MULTIPLIER) + RAND_INCREMENT;
-    // clang-format off
-    *rndNum = next; sRandFloat = (next >> 9) | 0x3F800000;
-    // clang-format on
+
+    *rndNum = next;
+    sRandFloat = (next >> 9) | 0x3F800000;
```
hensldm, line 92: I think I would prefer this to have the same style as the one above (assign `next` inside the calculation), though not a huge deal.

src/boot_O2/rand.c
EllipticEllipsis: Oh, I quite liked the quote, I think it adds a bit of flavour and context.

src/boot_O2/rand.c
EllipticEllipsis: Yeah, I suppose that makes sense, it's just IDO being weird.

src/overlays/actors/ovl_Obj_Hakaisi/z_obj_hakaisi.c
```diff
@@ -386,10 +386,10 @@ void func_80B15264(ObjHakaisi* this) {
     s32 pad;
     s16 sp32 = Rand_Next();
 
-    Matrix_InsertRotation(Rand_Next(), Rand_Next(), Rand_Next(), MTXMODE_NEW);
+    Matrix_InsertRotation((s32)Rand_Next(), (s32)Rand_Next(), (s32)Rand_Next(), MTXMODE_NEW);
```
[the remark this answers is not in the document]
EllipticEllipsis: I didn't add any of these automatically, I'm guessing not recasting does something weird when it does the u32 -> s16.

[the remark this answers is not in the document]
EllipticEllipsis: Hmm, checked again, it is okay either way. However, I think we probably should cast since this function takes s16s, what do others think?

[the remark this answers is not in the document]
hensldm: You can already guess my thoughts, but personally implicit casts here are fine. The function itself will tell you what the actual type is.

src/boot_O2/rand.c
fig02: fwiw i also think the quote from the book is kinda out of place

hensldm (approved):

src/boot_O2/rand.c
```diff
@@ -1,20 +1,23 @@
...
 
-// The latest generated random number, used to generate the next number in the sequence.
+//! The latest generated random number, used to generate the next number in the sequence.
 static u32 sRandInt = 1;
 
-// Space to store a value to be re-interpreted as a float.
-// This can't be static because it is used in z_kankyo
+//! Space to store a value to be re-interpreted as a float.
+//! This can't be static because it is used in z_kankyo
 u32 sRandFloat;
```
engineer124, line 8:
```suggestion
//! This can't be static because it is used in z_kankyo
u32 gRandFloat;
```
If it's used externally, should it have the `g` prefix to indicate that it's global and used across files?

AngheloAlf (approved):

Kenix3 (approved):
Its also more consistent with the other random functions.

PR #825, opened 2022-05-26
EnBaba Documented (Bomb Shop Lady)

hensldm:
Baba PR:
Documents the Bomb Shop Lady, EnBaba. Still a WIP, mainly opening to help with #823 and is based off that branch, so this PR looks larger than normal.

This is the first documentation of an actor using z_schedule.c so would like to know if what I've done is fine and something we would like to standardize. In particular not sure on the enum for schedule results.


src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,212 +334,219 @@ void func_80BA8DF4(EnBaba* this, GlobalContext* globalCtx) {
...
 }
 
-s32 func_80BA8F88(EnBaba* this, GlobalContext* globalCtx, ScheduleResult* arg2) {
-    u16 sp26 = (u16)(gSaveContext.save.time - 0x3FFC);
-    u16 temp;
-    u8 sp23 = ENBABA_GET_3F00(&this->actor);
+s32 EnBaba_InitTimePath(EnBaba* this, GlobalContext* globalCtx, ScheduleResult* scheduleResult) {
+    u16 now = SCHEDULE_TIME_NOW;
+    u16 startTime;
+    u8 path = ENBABA_GET_PATH(&this->actor);
```
engineer124, line 347:
```suggestion
    u8 pathIndex = ENBABA_GET_PATH_INDEX(&this->actor);
```
I think `path` and `pathIndex` should be distinguished

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,212 +334,219 @@ void func_80BA8DF4(EnBaba* this, GlobalContext* globalCtx) {
...
     }
 
-    if (this->unk_410 == NULL) {
+    if (this->timePath == NULL) {
         return false;
     }
 
-    if ((this->unk_434 != 0) && (this->unk_436 >= 0)) {
-        temp = sp26;
+    if ((this->scheduleResult != 0) && (this->timePathTimeSpeed >= 0)) {
```
engineer124, line 357:
```suggestion
    if ((this->scheduleResult != BABA_SCHEDULE_RESULT_NONE) && (this->timePathTimeSpeed >= 0)) {
```
I find it very unfortunate that `scheduleResult` is the variable name used to represent both `ScheduleResult* scheduleResult` and `u8 scheduleResult` i.e. `scheduleResult.result`. These are two different variables with overlying names (one is the struct itself, one is the return value from running the script), but I haven't thought of anything better. The best thing I've thought of is `this->scheduleResultValue` to indicate that it's the result/return value instead of the `ScheduleResult` struct. Maybe some name needs to change in the schedule system perhaps?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -107,19 +128,19 @@ static DamageTable sDamageTable = {
...
+    /* 0x12 */ SCHEDULE_CMD_RET_VAL_L(BABA_SCHEDULE_RESULT_KNOCKED_OVER),
     /* 0x15 */ SCHEDULE_CMD_RET_NONE(),
-    /* 0x16 */ SCHEDULE_CMD_RET_TIME(0, 0, 0, 30, 2),
+    /* 0x16 */ SCHEDULE_CMD_RET_TIME(0, 0, 0, 30, BABA_SCHEDULE_RESULT_FOLLOW_TIME_PATH),
     /* 0x1C */ SCHEDULE_CMD_RET_NONE(),
     /* 0x1D */ SCHEDULE_CMD_RET_NONE(),
 };
 
-s32 D_80BAA4A8[] = { -1, -1, 0 };
+static s32 sAdditionalTimePaths[] = { -1, -1, 0 };
```
engineer124, line 143:
```suggestion
static s32 sSearchTimePathLimit[] = {
    -1, // Get the last pathIndex in the path's connected by nextPathIndex (unused)
    -1, // Get the last pathIndex in the path's connected by nextPathIndex (unused)
    0,  // Get the first pathIndex in the path's connected by nextPathIndex
};
```
What about something like this? These are used as the last argument in `SubS_GetAdditionalPath(` which looking at that function (sorry for not reviewing those before, these path functions really interest me), That argument is the number of paths to search through. By setting that to 0, it means get the first path in the chained path list connected by unk1/nextPathIndex. But by setting that to 0, that means there is no limit and it grabs the last path in the chained path list connected by unk1/nextPathIndex

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,212 +334,219 @@ void func_80BA8DF4(EnBaba* this, GlobalContext* globalCtx) {
...
-    this->unk_430 = sp26 - temp;
-    temp = this->unk_410->count - 2;
-    this->unk_428 = this->unk_424 / temp;
-    this->unk_438 = 0;
-    this->unk_42C = (this->unk_430 / this->unk_428) + 2;
-    this->unk_43C = 0;
+    this->timePathElapsedTime = now - startTime;
+
+    startTime = this->timePath->count - 2; // startTime resused to store total waypoints
+    this->timePathWaypointTime = this->timePathTotalTime / startTime;
```
engineer124, line 372:
```suggestion
    numPointsMinusTwo = startTime = this->timePath->count - 2;
    this->timePathWaypointTime = this->timePathTotalTime / numPointsMinusTwo;
```
This is dumb but clearer? You can introduce `u16 numPointsMinusTwo;` at the bottom of the stack without affecting stack, then, use this variable like this and just slide `startTime` in there, and you have a matching and clearer docs?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,212 +334,219 @@ void func_80BA8DF4(EnBaba* this, GlobalContext* globalCtx) {
...
+    this->actor.world.pos.x = timePathPoint.x;
+    this->actor.world.pos.z = timePathPoint.z;
 
     if (SubS_InCsMode(globalCtx)) {
-        sp54 = this->unk_430;
-        sp50 = this->unk_42C;
-        sp58 = this->actor.world.pos;
+        timePathElapsedTime = this->timePathElapsedTime;
+        timePathWaypoint = this->timePathWaypoint;
+        timePathPoint = this->actor.world.pos;
```
engineer124, line 430:
```suggestion
        prevTimePathElapsedTime = this->timePathElapsedTime;
        prevTimePathWaypoint = this->timePathWaypoint;
        timePathPoint = this->actor.world.pos;
```
because these variables are just storing values if in a cutscene, I think `prev` is an indication that something is being stored and restored later on. I didn't add `prev` to `timePathPoint` since that's used in other ways as well. Thoughts?

engineer124 (commented):
Very nicely done with this `TimePathing` system, I've always been curious how npc's moving in a time-dependent way works! I've done a deep dive of this system, and am giving my review of `TimePathing` here as some of my comments are related to how individual actors will be doc'd as well

src/code/z_sub_s.c
```diff
@@ -214,17 +213,252 @@ void SubS_UpdateFlags(u16* flags, u16 setBits, u16 unsetBits) {
...
+/**
+ * Fills the weightArray to be used with time paths
+ *
+ * @param weightArray an array of values that are used to compute the weightVal and the individual weights
+ * @param numPoints the number of points considered with weights
+ * @param len the length to fill the array, generally the path count + numPoints
+ *
+ * @note `SubS_TimePathing_Update` assumes numPoints to be 3
+ */
+void SubS_TimePathing_FillWeightArray(f32 weightArray[], s32 numPoints, s32 len) {
```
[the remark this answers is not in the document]
engineer124: Sorry, this comment seemed to be dropped. What about a macro for this hardcoded 3 that represents the number of previous points to use? Something like:
`#define SUBS_TIME_PATHING_ORDER 3`? such that:
```
state = SubS_TimePathing_ComputeWeightVal(weightVal, *elapsedTime, waypointTime, totalTime, path->count, SUBS_TIME_PATHING_ORDER, weightArray);
SubS_TimePathing_ComputePointXZ(&point->x, &point->z, *weightVal, SUBS_TIME_PATHING_ORDER, *waypoint, points, weightArray);
SubS_TimePathing_FillWeightArray(weightArray, SUBS_TIME_PATHING_ORDER, this->timePath->count + SUBS_TIME_PATHING_ORDER);
```

src/code/z_sub_s.c
```diff
@@ -781,8 +1194,8 @@ s32 SubS_CopyPointFromPathList(Path* paths, s32 pathIndex, s32 pointIndex, Vec3f
     return false;
 }
 
-u8 SubS_GetPathCount(Path* paths, s32 index) {
-    Path* path = &paths[index];
+u8 SubS_GetPathCountFromPathList(Path* paths, s32 pathIndex) {
+    Path* path = &paths[pathIndex];
 
     return path->count;
```
engineer124, line 1200: I also think in the `Path` struct, `count` -> `numPoints` since `count` makes me think of an incrementer that can change, while `num` makes me think more of "total"

src/code/z_sub_s.c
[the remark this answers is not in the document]
hensldm: I had thought about a macro, but wasn't sure. It does seem nice to have though.

src/code/z_sub_s.c
```diff
@@ -214,17 +213,252 @@ void SubS_UpdateFlags(u16* flags, u16 setBits, u16 unsetBits) {
...
+ * @param waypointTime how much time per each waypoint
+ * @param totalTime how much time the path should take to travel
+ * @param pathCount the path count
+ * @param numPoints the number of points considered when computing the next point to move to
+ * @param weightArray see `SubS_TimePathing_FillWeightArray`
+ *
+ * @return s32 0 for error, 1 if still on the path, and 2 if the end of the path should be reached
+ */
+s32 SubS_TimePathing_ComputeWeightVal(f32* weightVal, s32 elapsedTime, s32 waypointTime, s32 totalTime, s32 pathCount,
+                                      s32 numPoints, f32 weightArray[]) {
```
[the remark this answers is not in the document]
hensldm: That makes sense. I think progress seems best since it represents how far along, the progress, along the given path.
This also seems to make sense for the `WeightPathing` systems, though that seems more like progress towards a given waypoint rather than the path as a whole.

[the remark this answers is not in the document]
hensldm: Agree with order.
pathCount should probably be changed with a bigger discussion/different PR about the pathing struct.

src/code/z_sub_s.c
```diff
@@ -214,17 +213,252 @@ void SubS_UpdateFlags(u16* flags, u16 setBits, u16 unsetBits) {
...
+/**
+ * Computes the weights to be used with time paths
+ *
+ * @param numPoints the number of points considered when computing the next point to move to, max is 11
+ * @param weightVal see `SubS_TimePathing_ComputeWeightVal`
+ * @param waypoint the current time relative to the start time of the path
+ * @param weightArray see `SubS_TimePathing_FillWeightArray`
+ * @param weights how much to weight each point considered
+ */
+void SubS_TimePathing_ComputeWeights(s32 numPoints, f32 weightVal, s32 waypoint, f32 weightArray[], f32 weights[]) {
```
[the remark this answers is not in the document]
hensldm: That does seem to verbose, I can add a comment in the function header though about this being interpolation.

src/code/z_sub_s.c
```diff
@@ -214,17 +213,252 @@ void SubS_UpdateFlags(u16* flags, u16 setBits, u16 unsetBits) {
...
+            break;
+    }
+
+    *elapsedTime += timeSpeed;
+    if (*elapsedTime >= totalTime) {
+        *elapsedTime = totalTime;
+    } else if (*elapsedTime < 0) {
+        *elapsedTime = 0;
+    }
+    *waypoint = (*elapsedTime / waypointTime) + 2;
```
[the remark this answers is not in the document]
hensldm: I think it makes sense. Its accounting for the order, but needs to start indexing at 0 rather than 1, which is where the minus 1 comes from.

src/code/z_sub_s.c
```diff
@@ -214,17 +213,252 @@ void SubS_UpdateFlags(u16* flags, u16 setBits, u16 unsetBits) {
...
+        return 0;
+    }
+    weightValMultiplier = 1.0f / waypointTime;
+    k = 0;
+    for (i = numPoints - 1; i < pathCount; i++) {
+        for (j = 0; j < waypointTime; j++) {
+            if (k == elapsedTime) {
+                break;
+            }
+            *weightVal += (weightArray[i + 1] - weightArray[i]) * weightValMultiplier;
```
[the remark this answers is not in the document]
hensldm: As a small test, I tried just replacing the loops with just `(f32)elapsedTime / (f32)waypointTime`, and it does seem to behave normally (albeit just a simple eye test).

src/code/z_sub_s.c
```diff
@@ -214,17 +213,252 @@ void SubS_UpdateFlags(u16* flags, u16 setBits, u16 unsetBits) {
...
+void SubS_TimePathing_FillWeightArray(f32 weightArray[], s32 numPoints, s32 len) {
+    s32 i;
+    f32 val = 0.0f;
+
+    for (i = 0; i < len; i++) {
+        if ((i >= numPoints) && (i < (len - numPoints + 1))) {
+            val += 1.0f;
+        }
+        weightArray[i] = val;
+    }
```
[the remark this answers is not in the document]
hensldm: I'm not sure about renaming `weightArray` to `offsetIndices`. I agree that represents the data well, but the same thing is used in calculating the interpolation weights which I don't think makes sense using something called `offsetIndices` (I could be wrong I don't know too much about interpolations and spline). Personally I think it makes more sense to base it off the interpolation usage, rather than from this function.
Also of note is that the time pathing doesn't necessarily require this particular `array`, it is just the default one and the only one that eventually got used.

src/code/z_sub_s.c
```diff
@@ -362,9 +596,190 @@ Path* SubS_GetDayDependentPath(GlobalContext* globalCtx, u8 pathIndex, u8 max, s
     return path;
 }
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_sub_s/func_8013C068.s")
+/**
+ * Computes the point to move toward using a weight based algorithm that considers 4 points along the path
```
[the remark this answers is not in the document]
hensldm: Let me know if you find anything

src/overlays/actors/ovl_Bg_Ingate/z_bg_ingate.c
```diff
@@ -188,7 +190,7 @@ void func_80953F9C(BgIngate* this, GlobalContext* globalCtx) {
 
         if (this->unk160 & 2) {
 
-            if (this->unk164->unk1 != 0xFF) {
+            if (this->timePath->unk1 != 0xFF) {
```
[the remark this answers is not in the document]
hensldm: Agreed, though should probably be done in a separate PR (along with count? and unk2). I had planned to change this in an early `SubS`, but decided not to for a couple reasons.
1. It would be a large diff and distract from the real purpose of the PR
2. It would cause a lot of conflicts (not as much a worry anymore with less PRs) especially since it would be nearly impossible to use namefixer for it.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: Yeah, I thought of that to. Personally I think is should probably be something in the schedule system, but I'm not sure. Probably the struct member `result` needs to change, but I have no idea what to.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: I like the name change, but not sure about the comments. Some actors have quite a few so it seems silly to specify it would be the `first`, `second`, etc path.

hensldm: Note I also like the idea of changing that func's arg to limit instead of max.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: Hmm, it also seems to work everywhere else, though not all actors that use this have been fully decomped. I think I would be fine with it, though I think `numWaypoints` would be better.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: Perhaps change it to `action` to mirror cutscenes, and how the schedule is communicating what action to take?

src/code/z_sub_s.c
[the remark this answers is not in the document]
engineer124: I see, yeah that's fine then. I wonder if it's worth adding something in the function itself then, just a comment explaining the for-loop and how it's creating an array of indices?

src/code/z_sub_s.c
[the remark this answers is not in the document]
engineer124: I was also curious about running a testing changing:
`#define SUBS_TIME_PATHING_ORDER 3`
into a different number, like `4` or even `10` and see the result. It'll definitely break for all the undecompiled actors who still use 3 from their asm, but just focusing on actors that have been decompiled and changed

src/overlays/actors/ovl_En_Baba/z_en_baba.c
engineer124: `action` isn't too bad, especially if it mirrors cutscenes. I'd be curious to see more actors documented though before knowing if `action` is right though. Another name I was thinking of was just `output`?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
engineer124: Yeah that's fine, it was more for my own understanding, and I'm pretty sure the docs I have above is wrong anyway taking a second look, because the first check in `SubS_GetAdditionalPath(`:
`i >= limit`
will always return true for the first path when `i = 0` and either `limit = 0` or `limit = -1`, so actually I'm not quite sure the difference in using `0` or `-1`?

src/code/z_sub_s.c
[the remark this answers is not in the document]
hensldm: Yeah, I can add a more descriptive description to the function itself explaining what the array looks like.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: the `-1` seems to only have meaning in the actor itself. Like the bodyParts arrays where `-1` just means none.
`if (sSearchTimePathLimit[scheduleResult->result] >= 0)`

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: `output` seems fine for now. Will probably add a comment about `action` as a possible name.

hensldm: Actually, I like changing the struct name to `ScheduleOutput` instead.

engineer124: I like that too actually, changing the struct to `ScheduleOutput` and leaving `result as is`. so:
`scheduleOutput.result`
`scheduleOutput.time0`
`scheduleOutput.time1`
`scheduleOutput.hasResult`

hensldm:
With the playstate rename PR ready to go, going to undraft this.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -128,142 +149,145 @@ static TrackOptionsSet sTrackOptions = {
...
     }
 }
 
-void func_80BA8C4C(PlayState* play, u16 nextEntrance) {
+void EnBaba_SetNextEntrance(PlayState* play, u16 nextEntrance) {
     play->nextEntranceIndex = nextEntrance;
     play->unk_1887F = 0x40;
     gSaveContext.nextTransition = 0x40;
     play->sceneLoadFlag = 0x14;
 }
```
engineer124, line 288:
```suggestion
void EnBaba_SetNextEntrance(PlayState* play, u16 nextEntrance) {
    play->nextEntranceIndex = nextEntrance;
    play->transitionType = TRANS_TYPE_64;
    gSaveContext.nextTransitionType = TRANS_TYPE_64;
    play->transitionTrigger = TRANS_TRIGGER_START;
}
```
This is what this function looks like after my transition PR, so I don't think it's just setting the next entrance, it's also triggering the next transition. Maybe something simple like `EnBaba_TransitionScene` or `EnBaba_TriggerTransition`?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,357 +336,364 @@ void func_80BA8DF4(EnBaba* this, PlayState* play) {
...
+        this->timePathPoint = timePathPoint;
     }
 
     return false;
 }
 
-void func_80BA93AC(EnBaba* this, PlayState* play) {
-    if (this->unk_434 != 1) {
-        if (this->unk_434 == 2) {
+void EnBaba_HandleScheduleOutput(EnBaba* this, PlayState* play) {
```
engineer124, line 457: I wonder if `EnBaba_HandleScheduleOutput` and `EnBaba_ProcessScheduleOutput` should be `EnBaba_HandleScheduleResult` and `EnBaba_ProcessScheduleResult` respectively? Both are really only using the "result" part of the output. But then also if we want consistency between different actors that use the scheduling system, do they tend to use all of "scheduleOutput" in handling/processing?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,357 +336,364 @@ void func_80BA8DF4(EnBaba* this, PlayState* play) {
...
                 play->msgCtx.msgMode = 0x43;
                 play->msgCtx.unk12023 = 4;
-                if (this->unk_40A & 8) {
+                if (this->flags & ENBABA_AUTOTALK) {
                     if (CHECK_QUEST_ITEM(QUEST_BOMBERS_NOTEBOOK)) {
                         if (play->msgCtx.unk120B1 == 0) {
                             gSaveContext.save.weekEventReg[81] |= 2;
-                            func_80BA8C4C(play, 0xD670);
+                            EnBaba_SetNextEntrance(play, 0xD670);
                             return;
```
engineer124, line 585: Is this return needed? Seems like the rest is just else and if-else cases

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: Makes sense. I like `EnBaba_TriggerTransition` of the two you suggested.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: Unfortunately it is required.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: I think that makes sense for `EnBaba_HandleScheduleOutput`, and I would even shorten it more to `EnBaba_HandleSchedule` since it actually doesn't deal with the `ScheduleOutput` strcut. For `EnBaba_ProcessScheduleOutput`, on the other hand I think it makes more sense to keep `ScheduleOutput` mainly because the full `ScheduleOutput` struct is passed in, and while that function only deals with the `ScheduleResult`, it calls functions that makes use of the other fields.

src/overlays/actors/ovl_En_Gk/z_en_gk.c
```diff
@@ -1027,7 +1027,7 @@ void EnGk_Init(Actor* thisx, PlayState* play) {
                 Actor_MarkForDeath(&this->actor);
             } else {
                 this->unk_318 = this->actor.cutscene;
-                this->path = SubS_GetPathByIndex(play, ENGK_GET_F0(&this->actor), 15);
+                this->path = SubS_GetPathByIndex(play, ENGK_GET_F0(&this->actor), 0xF);
```
EllipticEllipsis, line 1030: Is there a particular reason to put the maximum in hex?

src/overlays/actors/ovl_En_Ig/z_en_ig.c
```diff
@@ -616,9 +617,9 @@ s32 func_80BF1FA8(EnIg* this, PlayState* play, ScheduleResult* arg2) {
 
         this->timePathElapsedTime = sp2E - phi_v1;
 
-        phi_v1 = this->timePath->count - (SUBS_TIME_PATHING_ORDER - 1);
+        tmp = phi_v1 = this->timePath->count - (SUBS_TIME_PATHING_ORDER - 1);
 
-        this->timePathWaypointTime = this->timePathTotalTime / phi_v1;
+        this->timePathWaypointTime = this->timePathTotalTime / tmp;
```
EllipticEllipsis, line 622: Why is this an improvement?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -5,29 +5,50 @@
...
-#include "overlays/actors/ovl_En_Ossan/z_en_ossan.h"
-#include "objects/object_bba/object_bba.h"
 
 #define FLAGS (ACTOR_FLAG_1 | ACTOR_FLAG_8 | ACTOR_FLAG_10)
 
 #define THIS ((EnBaba*)thisx)
 
+#define ENBABA_END_CONVERSATION (1 << 0)
+#define ENBABA_VISIBLE (1 << 1)
+#define ENBABA_KNOCKED_OVER (1 << 2) // Don't track player
```
EllipticEllipsis, line 15: Notably these first three do not fit the subS pattern.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,355 +334,366 @@ void func_80BA8DF4(EnBaba* this, GlobalContext* globalCtx) {
...
                 this->actor.flags |= ACTOR_FLAG_10000;
             }
             func_800B8614(&this->actor, globalCtx, 100.0f);
         }
     }
 }
 
-void func_80BA9848(EnBaba* this, GlobalContext* globalCtx) {
-    u8 temp_v0 = Message_GetState(&globalCtx->msgCtx);
+void EnBaba_FollowScheduleConverse(EnBaba* this, GlobalContext* globalCtx) {
```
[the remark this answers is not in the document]
hensldm: Didn't think of it tbh, but yeah `Talk` is better.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: `EnBaba` doesn't make use of that system at all.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -311,357 +336,364 @@ void func_80BA8DF4(EnBaba* this, PlayState* play) {
...
                 this->actor.flags |= ACTOR_FLAG_10000;
             }
             func_800B8614(&this->actor, play, 100.0f);
         }
     }
 }
 
-void func_80BA9848(EnBaba* this, PlayState* play) {
-    u8 temp_v0 = Message_GetState(&play->msgCtx);
+void EnBaba_FollowScheduleConverse(EnBaba* this, PlayState* play) {
```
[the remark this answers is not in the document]
hensldm: So this is different and actually related to the third question in your main comment. The thing is that `EnBaba` has a particular type, given by the params, where the actor just follows the schedule given by the script. In particular I named this the `ENBABA_TYPE_FOLLOW_SCHEDULE` in the enum. This particular function is used only by that type, so the `FollowSchedule` has more to do with the type of the actor, rather than the actual schedule itself. I'll agree that this is a bit confusing, but I couldn't think of a better name for the `type`. If you got any suggestions feel free to let me know.

src/overlays/actors/ovl_En_Baba/z_en_baba.h
```diff
@@ -2,58 +2,59 @@
...
+    /* 0x2DE */ Vec3s turnTarget;
+    /* 0x2E4 */ Vec3s headRot;
+    /* 0x2EA */ Vec3s torsoRot;
+    /* 0x2F0 */ UNK_TYPE1 unk2F0[0x12];
+    /* 0x302 */ s16 limbRotTableY[18];
+    /* 0x326 */ s16 limbRotTableZ[18];
+    /* 0x34A */ UNK_TYPE1 unk34A[0xBA];
+    /* 0x404 */ s16 sakonDeadTimer;
+    /* 0x406 */ UNK_TYPE1 unk406[4];
+    /* 0x40A */ u16 flags;
```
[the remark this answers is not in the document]
hensldm: Hmm, I'm not sure I completely agree with this 100%. I feel like there are/will be usages of these kinds of variables that wouldn't necessarily be called states, so `stateFlags` might be kind of confusing.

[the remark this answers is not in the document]
hensldm: Ok, for lack of a better term I'll go with `stateFlags`, but I am curious if we can find a better term for these.

hensldm:
To answer your other two questions.
1. `SCHD_RES`?
2. As far as I know yes. Schedule scripts only seem to return 0 when using the `SCHEDULE_CMD_RET_NONE` command.

EllipticEllipsis:
Or even SCH, it's not like there'd be any other schedule-related enums in an actor?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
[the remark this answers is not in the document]
hensldm: Hmm, I quite like `FollowSchedule` for the actual action function name, as that is the action the actor is doing, following their schedule. If at all possible I'd rather find a way to help distinguish `HandleSchedule` which isn't an action function, but in this case holds most of the logic of how to follow the schedule. I'm also not sure if this will be a more used pattern. For example Bjt doesn't seem to use this pattern, but just has all the logic in the `Wait/Idle` function.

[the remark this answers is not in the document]
EllipticEllipsis: Yeah, although I suspect that ??? is atypical here since it's so simple. Could look at something like the Postman, or Anju's Mother.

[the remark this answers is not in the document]
hensldm: So it does seem to be a more general pattern. Looks like Sakon, the Postman, Dampei, and the milk bar bartender follow this pattern.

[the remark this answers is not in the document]
hensldm: I would say it does. What you call `EnNb_Wait`, I would call `EnNb_FollowSchedule` (it is the action function that gets set when the actor wants to follow the schedule and is what calls RunScript), than `func_80BC0D1C` would be what I would call `EnNb_HandleSchedule` (it holds the logic of actually following the schedule based on the processed output)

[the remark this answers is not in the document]
hensldm: Oh I forgot to mention (also don't know if you already saw it above), but I originally had `HandleSchedule` as `HandleScheduleOutput` https://github.com/zeldaret/mm/pull/825#discussion_r912521654 . Though I agree with engineer that it doesn't really need the Output, and I think shortening it to `HandleSchedule` was best.

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -128,142 +149,145 @@ static TrackOptionsSet sTrackOptions = {
     { 0x1770, 4, 1, 6 },
 };
 
-s32 func_80BA8820(EnBaba* this, PlayState* play) {
-    this->unk_144 = (EnOssan*)SubS_FindActor(play, &this->unk_144->actor, ACTORCAT_NPC, ACTOR_EN_OSSAN);
+s32 EnBaba_FindBombShopkeeper(EnBaba* this, PlayState* play) {
+    //! The bomb shopkeeper is an EnSob1, but initalizes itself with id `ACTOR_EN_OSSAN`
```
tom-overton, line 153: Does this deserve a @bug in front of it?

src/overlays/actors/ovl_En_Baba/z_en_baba.c
```diff
@@ -128,142 +149,145 @@ static TrackOptionsSet sTrackOptions = {
...
     Player* player = GET_PLAYER(play);
 
-    switch (this->unk_1E0) {
+    switch (this->textId) {
         case 0:
-            if (this->unk_40A & 8) {
+            if (this->stateFlags & BOMB_SHOP_LADY_STATE_AUTOTALK) {
                 if (gSaveContext.save.weekEventReg[33] & 8) {
-                    this->unk_1E0 = 0x2A34;
+                    this->textId = 0x2A34;
```
tom-overton, line 171: Can you provide a short summary of the text? You can find a list of all text here: https://docs.google.com/spreadsheets/d/1cbFFzoKaCgqMybAfahhDYPHq1vaCo2pc8v6BfZKhk2M/edit#gid=1312253055

tom-overton (approved):

src/overlays/actors/ovl_En_Baba/z_en_baba.c
hensldm: I don't really consider it a bug. It is more an unsafe design choice that I want to bring attention to.

EllipticEllipsis (approved):

Kenix3 (approved):
This includes some changes that could've been split out. 

Nonetheless, good job on this documentation!

PR #826, opened 2022-05-28
Gfxprint OK

hensldm:
Matches the last two functions in `gfxprint.c`. Also a little documentation/cleanup based on OoT.


src/boot_O2/gfxprint.c
```diff
@@ -1,19 +1,12 @@
 #include "global.h"
 
+//! TODO: Need to extract
 extern u16 sGfxPrintFontTLUT[64];
-extern u16 sGfxPrintUnkTLUT[16];
-extern u8 sGfxPrintUnkData[8];
+extern u16 sGfxPrintRainbowTLUT[16];
+extern u8 sGfxPrintRainbowData[8];
 extern u8 sGfxPrintFontData[2048];
```
EllipticEllipsis, line 7: Note to self: must finish the extraction program for this.

src/boot_O2/gfxprint.c
```diff
@@ -93,49 +86,95 @@ void GfxPrint_PrintCharImpl(GfxPrint* this, u8 c) {
...
+                break;
+            case '\n':
+                this->posY += 32;
+            case '\r':
+                this->posX = this->baseX;
+                break;
+            case '\t':
+                do {
+                    GfxPrint_PrintCharImpl(this, ' ');
+                } while ((this->posX - this->baseX) % 256);
```
EllipticEllipsis, line 130: I feel like some of these constants might be better in hex, although it depends how we want to do them across the board.

EllipticEllipsis (approved):
Good job on this, just the voice stuff left in boot now. Couple of minor suggestions, but I'm happy to get this in in its current state of documentation and finish it up when we extract the font etc.

src/boot_O2/gfxprint.c
```diff
@@ -34,12 +27,12 @@ void GfxPrint_InitDlist(GfxPrint* this) {
         gDPSetTileSize(this->dlist++, i * 2, 0, 0, 60, 1020);
     }
 
-    gDPSetPrimColorMod(this->dlist++, 0, 0, this->color.rgba);
+    gDPSetColor(this->dlist++, G_SETPRIMCOLOR, this->color.rgba);
```
engineer124, line 30:
```suggestion
    gDPSetColor(this->dList++, G_SETPRIMCOLOR, this->color.rgba);
```
could you change all the `dlist` -> `dList` in this file? It seems it's not 100% consistent, but `dList` is currently used 10x more than `dlist` and it's an abbreviation of two words, so I think this makes sense

src/boot_O2/gfxprint.c
```diff
@@ -145,34 +184,34 @@ void GfxPrint_Init(GfxPrint* this) {
...
     this->color.rgba = 0;
-    this->flag &= ~GFXPRINT_FLAG1;
-    this->flag &= ~GFXPRINT_USE_RGBA16;
-    this->flag |= GFXPRINT_FLAG4;
-    this->flag |= GFXPRINT_UPDATE_MODE;
+    this->flags &= ~GFXP_FLAG_HIRAGANA;
+    this->flags &= ~GFXP_FLAG_RAINBOW;
+    this->flags |= GFXP_FLAG_SHADOW;
+    this->flags |= GFXP_FLAG_UPDATE;
 }
```
engineer124, line 191: This is probably too nitpicky, but could you make the spacing consistent with OoT (or even just make a space between color.rgba and this->flags

image

src/boot_O2/gfxprint.c
```diff
@@ -93,49 +86,95 @@ void GfxPrint_PrintCharImpl(GfxPrint* this, u8 c) {
...
     }
 }
 
-GfxPrint* GfxPrint_Callback(GfxPrint* this, const char* str, size_t size) {
+void* GfxPrint_Callback(void* arg, const char* str, size_t size) {
+    GfxPrint* this = arg;
+
     GfxPrint_PrintStringWithSize(this, str, sizeof(char), size);
     return this;
 }
```
engineer124, line 174:
```suggestion
    GfxPrint_PrintStringWithSize(this, str, sizeof(char), size);

    return this;
}
```
Same silly nitpick comparing to OoT, I like the extra space. Not required though

include/z64.h
```diff
@@ -416,27 +416,36 @@ typedef struct {
...
+#define GFXP_RAINBOW_OFF_CHAR 0x8A
+
+#define GFXP_FLAG_HIRAGANA (1 << 0)
+#define GFXP_FLAG_RAINBOW  (1 << 1)
+#define GFXP_FLAG_SHADOW   (1 << 2)
+#define GFXP_FLAG_UPDATE   (1 << 3)
+#define GFXP_FLAG_ENLARGE  (1 << 6)
+#define GFXP_FLAG_OPEN     (1 << 7)
+
 typedef struct GfxPrint {
```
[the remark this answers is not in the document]
hensldm: I'm always for splitting headers. Note that I've also split off the functions from `functions.h`.

src/boot_O2/gfxprint.c
hensldm: Not sure, they might be better as defines. Probably best to wait for further docs.

include/gfxprint.h
```diff
@@ -0,0 +1,56 @@
...
+#define GFXP_RAINBOW_ON_CHAR 0x8B
+#define GFXP_RAINBOW_OFF "\x8A"
+#define GFXP_RAINBOW_OFF_CHAR 0x8A
+
+#define GFXP_FLAG_HIRAGANA (1 << 0)
+#define GFXP_FLAG_RAINBOW  (1 << 1)
+#define GFXP_FLAG_SHADOW   (1 << 2)
+#define GFXP_FLAG_UPDATE   (1 << 3)
+#define GFXP_FLAG_ENLARGE  (1 << 6)
+#define GFXP_FLAG_OPEN     (1 << 7)
```
engineer124, line 25: Looking at OoT and the complete repo, these flags look like they're internal to the `gfxprint` system while the flags above (ex. `GFXP_RAINBOW_ON`) are more external. I think these flags can be moved internally to `gfxprint.c`?

src/boot_O2/gfxprint.c
EllipticEllipsis: That's fair enough, just mentioning it in case other people want to disagree now.

include/gfxprint.h
hensldm: Good call.

engineer124 (approved):

Kenix3 (approved):

PR #831, opened 2022-05-30
Yasi descriptive names

MrPolymorph:
This was the first actor I done, so I'm just finishing this actor off by naming the contents of the object and then use more the descriptive symbols.

### DIFF
--- a/assets/xml/objects/object_obj_yasi.xml
+++ b/assets/xml/objects/object_obj_yasi.xml
@@ -1,11 +1,11 @@
 ﻿<Root>
     <File Name="object_obj_yasi" Segment="6">
-        <DList Name="gYasiDL" Offset="0x360" />
-        <Texture Name="gYasiTLUT" OutName="yasi_tlut" Format="rgba16" Width="4" Height="4" Offset="0x630" />
-        <Texture Name="gYasiFrondTex" OutName="yasi_frond_tex" Format="ci4" Width="32" Height="128" Offset="0x650" />
-        <Texture Name="gDekuNutTex" OutName="deku_nut_tex" Format="rgba16" Width="16" Height="16" Offset="0xE50" />
-        <Texture Name="gYasiScarredWoodTex" OutName="yasi_scarred_wood_tex" Format="rgba16" Width="16" Height="16" Offset="0x1050" />
-        <Texture Name="gYasiWoodTex" OutName="yasi_wood_tex" Format="rgba16" Width="8" Height="8" Offset="0x1250" />
-        <Collision Name="gYasiCol" Offset="0x1428" />
+        <DList Name="gPalmTreeDL" Offset="0x360" />
+        <Texture Name="gPalmTreeTLUT" OutName="palm_tree_tlut" Format="rgba16" Width="4" Height="4" Offset="0x630" />
+        <Texture Name="gPalmTreeFrondTex" OutName="palm_tree_frond_tex" Format="ci4" Width="32" Height="128" Offset="0x650" />
+        <Texture Name="gPalmTreeDekuNutTex" OutName="palm_tree_deku_nut_tex" Format="rgba16" Width="16" Height="16" Offset="0xE50" />
+        <Texture Name="gPalmTreeScarredWoodTex" OutName="palm_tree_scarred_wood_tex" Format="rgba16" Width="16" Height="16" Offset="0x1050" />
+        <Texture Name="gPalmTreeWoodTex" OutName="palm_tree_wood_tex" Format="rgba16" Width="8" Height="8" Offset="0x1250" />
+        <Collision Name="gPalmTreeCol" Offset="0x1428" />
     </File>
 </Root>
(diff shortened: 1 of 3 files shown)


assets/xml/objects/object_obj_yasi.xml
```diff
@@ -1,11 +1,11 @@
...
-        <Texture Name="object_obj_yasi_TLUT_000630" OutName="tlut_000630" Format="rgba16" Width="4" Height="4" Offset="0x630" />
-        <Texture Name="object_obj_yasi_Tex_000650" OutName="tex_000650" Format="ci4" Width="32" Height="128" Offset="0x650" />
-        <Texture Name="object_obj_yasi_Tex_000E50" OutName="tex_000E50" Format="rgba16" Width="16" Height="16" Offset="0xE50" />
-        <Texture Name="object_obj_yasi_Tex_001050" OutName="tex_001050" Format="rgba16" Width="16" Height="16" Offset="0x1050" />
-        <Texture Name="object_obj_yasi_Tex_001250" OutName="tex_001250" Format="rgba16" Width="8" Height="8" Offset="0x1250" />
-        <Collision Name="object_obj_yasi_Colheader_001428" Offset="0x1428" />
+        <DList Name="gYasiDL" Offset="0x360" />
+        <Texture Name="gYasiTLUT" OutName="yasi_tlut" Format="rgba16" Width="4" Height="4" Offset="0x630" />
+        <Texture Name="gYasiFrondTex" OutName="yasi_frond_tex" Format="ci4" Width="32" Height="128" Offset="0x650" />
+        <Texture Name="gDekuNutTex" OutName="deku_nut_tex" Format="rgba16" Width="16" Height="16" Offset="0xE50" />
```
EllipticEllipsis, line 6:
```suggestion
        <Texture Name="gYasiDekuNutTex" OutName="yasi_deku_nut_tex" Format="rgba16" Width="16" Height="16" Offset="0xE50" />
```
Since this is not the only Deku Nut texture knocking around and could easily clash with something in gameplay_keep

assets/xml/objects/object_obj_yasi.xml
```diff
@@ -1,11 +1,11 @@
 ﻿<Root>
     <File Name="object_obj_yasi" Segment="6">
-        <DList Name="object_obj_yasi_DL_000360" Offset="0x360" />
-        <Texture Name="object_obj_yasi_TLUT_000630" OutName="tlut_000630" Format="rgba16" Width="4" Height="4" Offset="0x630" />
-        <Texture Name="object_obj_yasi_Tex_000650" OutName="tex_000650" Format="ci4" Width="32" Height="128" Offset="0x650" />
-        <Texture Name="object_obj_yasi_Tex_000E50" OutName="tex_000E50" Format="rgba16" Width="16" Height="16" Offset="0xE50" />
-        <Texture Name="object_obj_yasi_Tex_001050" OutName="tex_001050" Format="rgba16" Width="16" Height="16" Offset="0x1050" />
-        <Texture Name="object_obj_yasi_Tex_001250" OutName="tex_001250" Format="rgba16" Width="8" Height="8" Offset="0x1250" />
-        <Collision Name="object_obj_yasi_Colheader_001428" Offset="0x1428" />
+        <DList Name="gYasiDL" Offset="0x360" />
```
EllipticEllipsis, line 3: If this is the whole tree, I'd say name it as such
```suggestion
        <DList Name="gYasiTreeDL" Offset="0x360" />
```

EllipticEllipsis (commented):
I'd prefer if we used `PalmTree` for the variable names in this object, since most people are not going to know about https://en.wiktionary.org/wiki/椰子#Japanese

assets/xml/objects/object_obj_yasi.xml
MrPolymorph: That's fair RE:translations, how far do we tend to go with this? is it just variable names within an an actor? or does the whole thing get translated eventually (inc, filenames?)

EllipticEllipsis: It's basically just object symbols, params macros and enums/defines for now. Everything will get renamed eventually when it's no longer useful to have the original names around, but for the ones I just listed we have no original names and so can do what we like immediately.

MrPolymorph: okay that makes sense. thanks for the clarification :)

assets/xml/objects/object_obj_yasi.xml
```diff
@@ -1,11 +1,11 @@
 ﻿<Root>
     <File Name="object_obj_yasi" Segment="6">
```
hensldm, line 2: Can you add a comment explaining what this object is?

EllipticEllipsis (approved):

hensldm (approved):
Thanks for the cleanup.

tom-overton (approved):

Kenix3 (approved):

PR #834, opened 2022-06-02
`DmZl` OK and some docs of `object_zl4` (Child Zelda used in one cutscene)

isghj5:
I don't have the energy to document every animation in this object, there are over 40, but only 7 are used by MM, and a lot of them appear to be replica of OoT animations but... If I want naming parity with OoT I have to find the right match, Child Zelda is broken up into multiple objects in OoT. Sitting there comparing slightly different animations from two games in two different z64util windows (they like to crash for me too) is just painful.

But I found she has a backflip and a flute, so that's neat.

unknown


src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+    ZELDA_ANIM_TURNING_TOWARD_PLAYER,
+    ZELDA_ANIM_FACING_PLAYER,
+    ZELDA_ANIM_GIVING_OCARINA_START,
+    ZELDA_ANIM_GIVING_OCARINA,
+    ZELDA_ANIM_PLAYING_OCARINA_START,
+    ZELDA_ANIM_PLAYING_OCARINA,
+    ZELDA_ANIM_MAX,
+} DmZlAnimations;
+
+static u64* sMouthTextures[4] = {
```
hensldm, line 54:
```suggestion
static TexturePtr sMouthTextures[4] = {
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+};
+
+typedef enum {
+    ZELDA_MOUTH_NEUTRAL,
+    ZELDA_MOUTH_SMILING,
+    ZELDA_MOUTH_FROWNING,
+    ZELDA_MOUTH_OPEN,
+} DmZlMouthTextures;
+
+static u64* sEyeTextures[0xA] = {
```
hensldm, line 68:
```suggestion
static TexturePtr  sEyeTextures[] = {
```
Also are the `NULL`s needed. Look like they might be file padding?

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+void DmZl_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    DmZl* this = THIS;
+
+    this->animationIndex = 0;
+    this->unused2BA = 0;
+    this->actor.targetArrowOffset = 1000.0f;
+    ActorShape_Init(&this->actor.shape, 0.0f, ActorShadow_DrawCircle, 24.0f);
+    // NULL jointTable, NULL morphTable, limbcount == 0 ? surprised this works
+    SkelAnime_InitFlex(globalCtx, &this->skelAnime, &gZl4Skeleton, NULL, NULL, NULL, 0);
```
hensldm, line 100: This signifies to the SkelAnime system to just dynamically allocate the joint and morph tables.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+    DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], 0);
```
hensldm, line 101:
```suggestion
    DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], ZELDA_ANIM_FACING_AWAY);
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+    Actor_SetScale(&this->actor, 0.01f);
+    this->actionFunc = DmZl_DoNothing;
+}
+
+void DmZl_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    DmZl* this = THIS;
```
hensldm, line 107:
```suggestion
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+}
+
+void DmZl_DoNothing(DmZl* this, GlobalContext* globalCtx) {
+}
+
+void DmZl_UpdateCutscene(DmZl* this, GlobalContext* globalCtx) {
+    s32 actionIndex;
+
+    if (Cutscene_CheckActorAction(globalCtx, 0x66) != 0) {
```
hensldm, line 116:
```suggestion
    if (Cutscene_CheckActorAction(globalCtx, 0x66)) {
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+            }
+        }
+
+        Cutscene_ActorTranslateAndYaw(&this->actor, globalCtx, actionIndex);
+    }
+
+    if ((Animation_OnFrame(&this->skelAnime, this->skelAnime.endFrame) != 0) &&
+        (((actionIndex = this->animationIndex) == 1) || (actionIndex == 3) || (actionIndex == 5))) {
+        this->animationIndex++;
+        DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], ZELDA_ANIM_FACING_AWAY);
```
hensldm, line 148: Is it possible to break the if condition up so that the `actionIndex` assignment doesn't occur in the if?

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,243 @@ const ActorInit Dm_Zl_InitVars = {
...
+    DmZl* this = THIS;
+
+    DmZl_UpdateFace(this);
+    SkelAnime_Update(&this->skelAnime);
+    DmZl_UpdateCutscene(this, globalCtx);
+    this->actionFunc(this, globalCtx);
+}
+
+s32 DmZl_OverrideLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, Actor* thisx) {
+    return 0;
```
hensldm, line 239:
```suggestion
    return false;
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -9,11 +9,24 @@ typedef void (*DmZlActionFunc)(struct DmZl*, GlobalContext*);
 
 typedef struct DmZl {
     /* 0x0000 */ Actor actor;
```
hensldm, line 11:
```suggestion
    /* 0x000 */ Actor actor;
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -9,11 +9,24 @@ typedef void (*DmZlActionFunc)(struct DmZl*, GlobalContext*);
...
-    /* 0x0144 */ char unk_144[0x11C];
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ char pad188[0xD8]; // unused by DmZl
     /* 0x0260 */ DmZlActionFunc actionFunc;
```
hensldm, line 14:
```suggestion
    /* 0x260 */ DmZlActionFunc actionFunc;
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -9,11 +9,24 @@ typedef void (*DmZlActionFunc)(struct DmZl*, GlobalContext*);
 
 typedef struct DmZl {
     /* 0x0000 */ Actor actor;
-    /* 0x0144 */ char unk_144[0x11C];
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ char pad188[0xD8]; // unused by DmZl
```
hensldm, line 13:
```suggestion
    /* 0x188 */ UNK_TYPE1 pad188[0xD8]; // unused by DmZl
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -9,11 +9,24 @@ typedef void (*DmZlActionFunc)(struct DmZl*, GlobalContext*);
...
     /* 0x0260 */ DmZlActionFunc actionFunc;
-    /* 0x0264 */ char unk_264[0x70];
+    /* 0x264 */ char pad264[0x4C]; // unused by DmZl
```
hensldm, line 15:
```suggestion
    /* 0x264 */ UNK_TYPE1 pad264[0x4C]; // unused by DmZl
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -9,11 +9,24 @@ typedef void (*DmZlActionFunc)(struct DmZl*, GlobalContext*);
...
+    /* 0x2B0 */ s16 animationIndex;
+    /* 0x2B2 */ u8 eyeTextureIndex1;
+    /* 0x2B3 */ u8 eyeTextureIndex2;
+    /* 0x2B4 */ u8 mouthTextureIndex;
+    /* 0x2B5 */ u8 nextEyeState;   // used to control eye state, but not set by our actor, outside of actor?
+    /* 0x2B6 */ u8 nextMouthMouth; // used to control mouth state, but not set by our actor, outside of actor?
+    /* 0x2B8 */ s16 blinkTimer;
+    /* 0x2BA */ s16 unused2BA; // init to 0, unread/unchanged by DmZl after
+    /* 0x2BC */ char pad2BC[0x18]; // unused by DmZl
```
hensldm, line 24:
```suggestion
    /* 0x2BC */ UNK_TYPE1 pad2BC[0x18]; // unused by DmZl
```

hensldm (commented):
Nice work. Some minor things.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
isghj5: Weird? That sounds useful but I dont think I've ever seen that done before, was it added in MM or something I wonder?

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
isghj5: Er... is it? what is being checked here? That the action is started? That the cutscene is possible?

hensldm: It is verifying that it is currently a cutscene, and the given action is valid.
If your curious the function is pretty small at the bottom of `z_demo.c`.

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
-        <Texture Name="object_zl4_Tex_005DB0" OutName="tex_005DB0" Format="rgba16" Width="32" Height="32" Offset="0x5DB0" />
-        <DList Name="object_zl4_DL_00A630" Offset="0xA630" />
-        <DList Name="object_zl4_DL_00AFC8" Offset="0xAFC8" />
-        <DList Name="object_zl4_DL_00B8D8" Offset="0xB8D8" />
-        <DList Name="object_zl4_DL_00BB40" Offset="0xBB40" />
-        <DList Name="object_zl4_DL_00BC90" Offset="0xBC90" />
-        <DList Name="object_zl4_DL_00BE70" Offset="0xBE70" />
-        <DList Name="object_zl4_DL_00C0E0" Offset="0xC0E0" />
-        <DList Name="object_zl4_DL_00C230" Offset="0xC230" />
+        <!-- Child Zelda: from the cutscene where Deku Link recalls how to play Song of Time after getting his Ocarina back -->
```
tom-overton, line 3: This should go above the \<File\> node, not below it.

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+
+        <Animation Name="gDmZl4IdleAnim" Offset="0x654" />
```
tom-overton, line 5: What do you think about prefixing these with `gChildZelda` instead of `gDmZl4`? I tend to prefix objects with what it actually *is* rather than the filename.

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <Texture Name="gDmZl4BlueUndershirtGradient1Tex" OutName="blue_undershirt_gradient" Format="rgba16" Width="8" Height="8" Offset="0x59F0" />
+        <Texture Name="gDmZl4BlueUndershirtGradient2Tex" OutName="blue_undershirt_gradient_2" Format="rgba16" Width="16" Height="16" Offset="0x5A70" />
+        <Texture Name="gDmZl4StripedShoulderTex" OutName="striped_shoulder_patern" Format="ci8" Width="8" Height="8" Offset="0x5C70" />
+        <Texture Name="gDmZl4UndershirtWrinklesTex" OutName="undershirt_wrinkles" Format="ci8" Width="16" Height="16" Offset="0x5CB0" />
+        <Texture Name="gDmZl4PendantTex" OutName="pendant" Format="rgba16" Width="32" Height="32" Offset="0x5DB0" />
+
+        <DList Name="gDmZl4ChestDl" Offset="0xA630" />
+        <DList Name="gDmZl4HeadDl" Offset="0xAFC8" />
+        <DList Name="gDmZl4RightUpperArmDl" Offset="0xB8D8" />
+        <DList Name="gDmZl4RightForArmDl" Offset="0xBB40" />
```
tom-overton, line 60: Did you mean "forearm" here?

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <DList Name="gDmZl4RightHandDl" Offset="0xBC90" />
+        <DList Name="gDmZl4LightUpperArmDl" Offset="0xBE70" />
```
tom-overton, line 62: Did you mean "left" here?

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+
+        <Texture Name="gDmZl4FingersTex" OutName="fingers" Format="rgba16" Width="16" Height="16" Offset="0x56F0" />
+        <Texture Name="gDmZl4BraceletTex" OutName="bracelet" Format="ci8" Width="16" Height="16" Offset="0x58F0" />
+        <Texture Name="gDmZl4BlueUndershirtGradient1Tex" OutName="blue_undershirt_gradient" Format="rgba16" Width="8" Height="8" Offset="0x59F0" />
+        <Texture Name="gDmZl4BlueUndershirtGradient2Tex" OutName="blue_undershirt_gradient_2" Format="rgba16" Width="16" Height="16" Offset="0x5A70" />
+        <Texture Name="gDmZl4StripedShoulderTex" OutName="striped_shoulder_patern" Format="ci8" Width="8" Height="8" Offset="0x5C70" />
+        <Texture Name="gDmZl4UndershirtWrinklesTex" OutName="undershirt_wrinkles" Format="ci8" Width="16" Height="16" Offset="0x5CB0" />
+        <Texture Name="gDmZl4PendantTex" OutName="pendant" Format="rgba16" Width="32" Height="32" Offset="0x5DB0" />
+
+        <DList Name="gDmZl4ChestDl" Offset="0xA630" />
```
tom-overton, line 57: Usually we suffix DisplayLists with "DL" (all caps) instead of "Dl"
```suggestion
        <DList Name="gDmZl4ChestDL" Offset="0xA630" />
```

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <Animation Name="gDmZl4Anim_024510" Offset="0x24510" />
+        <!-- arms behind back head down -->
+        <Animation Name="gDmZl4Anim_024B68" Offset="0x24B68" />
+
+        <!-- Flute, unused? -->
+        <DList Name="gDmZl4FluteDl" Offset="0x24D80" />
+        <DList Name="gDmZl4FlutePtrDl" Offset="0x24E40" />
+        <Texture Name="gDmZl4FluteHoleTex" OutName="flute_hole" Format="rgba16" Width="16" Height="8" Offset="0x24E50" />
+        
+        <TextureAnimation Name="gDmZlEmptyTextAnim" Offset="0x24F50" />
```
tom-overton, line 175: We use "TexAnim" as the suffix for these.
```suggestion
        <TextureAnimation Name="gDmZlEmptyTexAnim" Offset="0x24F50" />
```

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <Animation Name="gDmZl4Anim_023A50" Offset="0x23A50" />
+        <!-- shock? you chose no at the wrong time? -->
+        <Animation Name="gDmZl4Anim_0241F4" Offset="0x241F4" />
+        <Animation Name="gDmZl4Anim_024510" Offset="0x24510" />
+        <!-- arms behind back head down -->
+        <Animation Name="gDmZl4Anim_024B68" Offset="0x24B68" />
+
+        <!-- Flute, unused? -->
+        <DList Name="gDmZl4FluteDl" Offset="0x24D80" />
+        <DList Name="gDmZl4FlutePtrDl" Offset="0x24E40" />
```
tom-overton, line 172: What does "Ptr" mean here?

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <!-- unused, left over from OoT I think  -->
+        <Animation Name="gDmZl4TurningAroundAnim" Offset="0xF250" />
+        <Animation Name="gDmZl4IdleHandsInFrontAnim" Offset="0xFDC8" />
+        <Animation Name="gDmZl4RaiseFluteToPlayAnim" Offset="0x10334" />
+        <Animation Name="gDmZl4LowerFluteAfterPlayAnim" Offset="0x105C8" />
+        <Animation Name="gDmZl4PlayingFluteAnim" Offset="0x10D50" />
+        <Animation Name="gDmZl4FacingAwayIdle" Offset="0x11294" />
+
+        <!-- Used in MM -->
+        <Animation Name="gDmZl4GivingItemToLinkStartAnim" Offset="0x11BB8" />
```
tom-overton, line 111: Just thinking about shortening this a bit; does the name need to reference Link? Couldn't it just be `gDmZl4GivingItemStartAnim`? If she was giving an item to a different character, would she use a different animation?

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <Animation Name="gDmZl4GivingItemToLinkLoopAnim" Offset="0x122E0" />
+        <Animation Name="gDmZl4RaisingOcarinaToPlayAnim" Offset="0x12A84" />
+        <Animation Name="gDmZl4PlayingOcarinaAnim" Offset="0x1303C" />
+
+        <!-- She does a Backflip -->
+        <Animation Name="gDmZl4BackFlipAnim" Offset="0x13934" />
```
tom-overton, line 117: Since "backflip" is one word:
```suggestion
        <Animation Name="gDmZl4BackflipAnim" Offset="0x13934" />
```

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <Limb Name="gDmZl4RightFootLimb" Type="Standard" EnumName="ZL4_LIMB_RIGHT_FOOT" Offset="0xDF64" />
+        <Limb Name="gDmZl4Dress1Limb" Type="Standard" EnumName="ZL4_LIMB_DRESS_1" Offset="0xDF70" />
+        <Limb Name="gDmZl4Dress2Limb" Type="Standard" EnumName="ZL4_LIMB_DRESS_2" Offset="0xDF7C" />
+        <Limb Name="gDmZl4Dress3Limb" Type="Standard" EnumName="ZL4_LIMB_DRESS_3" Offset="0xDF88" />
+        <Limb Name="gDmZl4TorsoLimb" Type="Standard" EnumName="ZL4_LIMB_TORSO" Offset="0xDF94" />
+        <Limb Name="gDmZl4LeftUpperArmLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_UPPER_ARM" Offset="0xDFA0" />
+        <Limb Name="gDmZl4LeftForarmLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_FORARM" Offset="0xDFAC" />
+        <Limb Name="gDmZl4LeftHandLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_HAND" Offset="0xDFB8" />
+        <Limb Name="gDmZl4RightUpperArmLimb" Type="Standard" EnumName="ZL4_LIMB_RIGHT_UPPER_ARM" Offset="0xDFC4" />
+        <Limb Name="gDmZl4RightForarmLimb" Type="Standard" EnumName="ZL4_LIMB_RIGHT_FORARM" Offset="0xDFD0" />
```
tom-overton, line 97:
```suggestion
        <Limb Name="gDmZl4RightForearmLimb" Type="Standard" EnumName="ZL4_LIMB_RIGHT_FOREARM" Offset="0xDFD0" />
```

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <Limb Name="gDmZl4LeftLegLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_LEG" Offset="0xDF40" />
+        <Limb Name="gDmZl4LeftFootLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_FOOT" Offset="0xDF4C" />
+        <Limb Name="gDmZl4RightLegLimb" Type="Standard" EnumName="ZL4_LIMB_RIGHT_LEG" Offset="0xDF58" />
+        <Limb Name="gDmZl4RightFootLimb" Type="Standard" EnumName="ZL4_LIMB_RIGHT_FOOT" Offset="0xDF64" />
+        <Limb Name="gDmZl4Dress1Limb" Type="Standard" EnumName="ZL4_LIMB_DRESS_1" Offset="0xDF70" />
+        <Limb Name="gDmZl4Dress2Limb" Type="Standard" EnumName="ZL4_LIMB_DRESS_2" Offset="0xDF7C" />
+        <Limb Name="gDmZl4Dress3Limb" Type="Standard" EnumName="ZL4_LIMB_DRESS_3" Offset="0xDF88" />
+        <Limb Name="gDmZl4TorsoLimb" Type="Standard" EnumName="ZL4_LIMB_TORSO" Offset="0xDF94" />
+        <Limb Name="gDmZl4LeftUpperArmLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_UPPER_ARM" Offset="0xDFA0" />
+        <Limb Name="gDmZl4LeftForarmLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_FORARM" Offset="0xDFAC" />
```
tom-overton, line 94:
```suggestion
        <Limb Name="gDmZl4LeftForearmLimb" Type="Standard" EnumName="ZL4_LIMB_LEFT_FOREARM" Offset="0xDFAC" />
```

include/z64animation.h
```diff
@@ -25,7 +25,7 @@ typedef enum {
     /* 3 */ ANIMMODE_ONCE_INTERP,
     /* 4 */ ANIMMODE_LOOP_PARTIAL,
     /* 5 */ ANIMMODE_LOOP_PARTIAL_INTERP
-} AnimationModes;
+} AnimationModes; // AnimationInfo mode
```
tom-overton, line 28: Is there a particular reason this comment was added?

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,241 @@ const ActorInit Dm_Zl_InitVars = {
...
+
+void DmZl_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    DmZl* this = THIS;
+
+    this->animationIndex = 0;
+    this->unused2BA = 0;
+    this->actor.targetArrowOffset = 1000.0f;
+    ActorShape_Init(&this->actor.shape, 0.0f, ActorShadow_DrawCircle, 24.0f);
+    // these three set to NULL should mean they are dynamically allocated
```
tom-overton, line 96: Not sure if this deserves a comment in this actor specifically; other actors do this too, so it feels like this comment belongs on `SkelAnime_InitFlex` if anything.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,241 @@ const ActorInit Dm_Zl_InitVars = {
...
+
+    if (this->blinkTimer < 3) {
+        this->eyeTextureIndex2 = this->blinkTimer;
+        this->eyeTextureIndex1 = this->blinkTimer;
+    }
+
+    switch (this->nextEyeState) {
+        case 0:
+            if (this->blinkTimer == 0) {
+                this->blinkTimer = Rand_S16Offset(0x1E, 0x1E);
```
tom-overton, line 169: Probably easier to understand when written these in decimal:
```suggestion
                this->blinkTimer = Rand_S16Offset(30, 30);
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,241 @@ const ActorInit Dm_Zl_InitVars = {
...
+            }
+            break;
+        case 1:
+            if (this->blinkTimer == 0) {
+                this->eyeTextureIndex1 = this->eyeTextureIndex2 = 2;
```
tom-overton, line 174: Do you think it makes sense to make an enum for these eye texture indices in the same way you made one for the mouth ones?

assets/xml/objects/object_zl4.xml
isghj5: ah yeah, sometimes writing in camel case for a long time my hands cramp up, will fix

assets/xml/objects/object_zl4.xml
isghj5: The display list consists of
```
Gfx gDmZl4FlutePtrDl[] = {
    gsSPDisplayList(gDmZl4FluteDl),
    gsSPEndDisplayList(),
};
``` It calls the DL above it, then ends. Kinda pointerish, not sure if there is a better term for doing this in a DL

File: assets/xml/objects/object_zl4.xml (hunk 21 above)
Comment (isghj5) [reply assets/xml/objects/object_zl4.xml:111]: makes sense, will do

File: src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c (hunk 28 above)
Comment (isghj5) [reply src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c:174]: Oh there should absolutely (would help navigate the eye state func) just didnt get around to it.

File: assets/xml/objects/object_zl4.xml (hunk 20 above)
Comment (tom-overton) [reply assets/xml/objects/object_zl4.xml:172]: Hmm, I'm assuming it's unused then? I'm not sure what to call this either...maybe a "wrapper"?

File: assets/xml/objects/object_zl4.xml (hunk 18 above)
Comment (isghj5) [reply assets/xml/objects/object_zl4.xml:57]: For some reason in the back of my mind its backwards, like Dl was used instead of DL and I thought I had to correct it, brain rot.

File: include/z64animation.h (hunk 25 above)
Comment (isghj5) [reply include/z64animation.h:28]: Got tired of searching for that enum and not finding it, but yeah its not a proper tag and I'll remove

File: src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c (hunk 26 above)
Comment (isghj5) [reply src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c:96]: I'm not generally fond of finding weird code and having no local documentation explaining the weirdness.

"Documentation exists in another file somewhere else" is kinda annoying (especially if I don't know it even exists). I personally prefer redundant documentation to absent.

**Comment by isghj5** [comment]
On second thought, I'm diverging from OoT documentation for eyes because I think they are in error.

`eye_open_looking_up_right.ci8.png` are the eyes she uses to look straight forward, they are the default eyes, she is not looking to the right or upwards.

And between the two halfs, `half 2` is used for her blinking animation, and while she doesn't use it in MM I think the other eye is what she uses when smiling when happy at link's responses in OoT.

zeyes

File: src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c [hunk 29]
```diff
@@ -171,29 +176,29 @@ void DmZl_UpdateFace(DmZl* this) {
...
-                this->eyeTextureIndex1 = this->eyeTextureIndex2 = 2;
+                this->eyeTextureIndex1 = this->eyeTextureIndex2 = ZELDA_EYE_CLOSED;
             }
             break;
         case 2:
             if (this->blinkTimer == 0) {
-                this->eyeTextureIndex1 = 5;
-                this->eyeTextureIndex2 = 6;
+                this->eyeTextureIndex1 = ZELDA_EYE_OPEN_LOOKING_LEFT;
+                this->eyeTextureIndex2 = ZELDA_EYE_OPEN_LOOKING_RIGHT;
```
tom-overton, line 185: These enums make this much more readable, thanks. Can we can figure out what these two indices are? Are they left and right eye?

isghj5: I'm still not sure which is which, is that zelda looking left or zelda looking right? I still struggle to understand how DL handle textures, so not sure which is flipped and which is not, both say NOMIRROR...?
(also why does it load the same tlut twice...?)
from the HeadDL
```c
   gsDPSetTextureLUT(G_TT_RGBA16),
    gsDPLoadTLUT_pal256(gDmZl4EyeTLUT),
    gsDPLoadTextureBlock(0x08000000, G_IM_FMT_CI, G_IM_SIZ_8b, 32, 32, 0, G_TX_NOMIRROR | G_TX_CLAMP, G_TX_NOMIRROR |
                         G_TX_CLAMP, 5, 5, G_TX_NOLOD, G_TX_NOLOD),
    gsDPSetTileSize(G_TX_RENDERTILE, 0, 0, 0x007C, 0x007C),
    gsSPVertex(&object_zl4Vtx_0065B0[572], 19, 0),
    gsSP2Triangles(0, 1, 2, 0, 3, 2, 4, 0),
    gsSP2Triangles(3, 5, 0, 0, 0, 2, 3, 0),
    gsDPPipeSync(),
    gsDPSetTextureLUT(G_TT_RGBA16),
    gsDPLoadTLUT_pal256(gDmZl4EyeTLUT),
    gsDPLoadTextureBlock(0x09000000, G_IM_FMT_CI, G_IM_SIZ_8b, 32, 32, 0, G_TX_NOMIRROR | G_TX_CLAMP, G_TX_NOMIRROR |
                         G_TX_CLAMP, 5, 5, G_TX_NOLOD, G_TX_NOLOD),
    gsDPSetTileSize(G_TX_RENDERTILE, 0, 0, 0x007C, 0x007C),
    gsSP2Triangles(6, 7, 8, 0, 8, 9, 6, 0),
    gsSP2Triangles(10, 7, 6, 0, 7, 11, 8, 0),
    gsDPPipeSync(),
    gsDPSetTextureLUT(G_TT_RGBA16),

```

tom-overton: I'm fairly sure that what we call eyeTextureIndex2 is the right eye, and the eyeTextureIndex1 is the left one. You can see the them use load the texture for eyeTextureIndex2 into segment 8 here:
```c
gSPSegment(POLY_OPA_DISP++, 0x08, Lib_SegmentedToVirtual(sEyeTextures[this->eyeTextureIndex2]));
```

And when I put a texture into that segment in Z64Utils, her right eye gets "filled in" while her left eye remains "empty":
image

A similar thing happens to the left eye if you put a texture in segment 9, which the code uses for eyeTextureIndex1. I'm not 100% certain how this DList is handling the texture either, but just experimentally, I think we can safely call these `leftEyeIndex` and `rightEyeIndex`, or something similar.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
tom-overton: Well, in this case, there is a comment about it directly above `SkelAnime_InitFlex`:
```c
/**
 * Initializes a flex skeleton to a looping animation, dynamically allocating the frame tables if not given.
 */
 ```
In my IDE, just hovering my mouse over `SkelAnime_InitFlex` shows this comment, making it pretty clear what's going on. Not everyone uses an IDE, though, so perhaps others will find this comment helpful.

File: src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h [hunk 30]
```diff
@@ -8,12 +8,25 @@ struct DmZl;
...
-    /* 0x0260 */ DmZlActionFunc actionFunc;
-    /* 0x0264 */ char unk_264[0x70];
+    /* 0x000 */ Actor actor;
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ UNK_TYPE1 pad188[0xD8]; // unused by DmZl
+    /* 0x260 */ DmZlActionFunc actionFunc;
+    /* 0x264 */ UNK_TYPE1 pad264[0x4C]; // unused by DmZl
+    /* 0x2B0 */ s16 animationIndex;
+    /* 0x2B2 */ u8 eyeTextureLeft;
+    /* 0x2B3 */ u8 eyeTextureRight;
```
tom-overton, line 18: I'd prefer if these two referenced "index" or something like that. The current names make it seem like these are textures in their own right, whereas they are actually indices into `sEyeTextures`.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -8,12 +8,25 @@ struct DmZl;
...
+    /* 0x2B4 */ u8 mouthTextureIndex;
+    /* 0x2B5 */ u8 nextEyeState;   // used to control eye state, but not set by our actor, outside of actor?
+    /* 0x2B6 */ u8 nextMouthMouth; // used to control mouth state, but not set by our actor, outside of actor?
```
tom-overton, line 21: Just noticed how weird this name is. Did you mean "nextMouthState"?

isghj5: oh yeah thats odd, probably was a search replace falsepositive.

tom-overton (approved):

hensldm (approved):

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
EllipticEllipsis: No, OoT has it too: https://github.com/zeldaret/oot/blob/master/src/code/z_skelanime.c#L1379-L1388

assets/xml/objects/object_zl4.xml
EllipticEllipsis: I'd rather have `ChildZelda` myself.

assets/xml/objects/object_zl4.xml
EllipticEllipsis: It's like a wrapper function in C, "wrapper" is what we usually use.

include/z64animation.h
EllipticEllipsis: You can add a comment above the enum about what it's used for, or better is probably to add a documentation comment above the structs that usually use it.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
EllipticEllipsis: The other way would have been to just lock this in an emulator.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
EllipticEllipsis: It's not the three being set to NULL anyway, it's only two of them. And yes, this is the point of having the documentation comment above the function in the first place. (It's not used that much in MM, yet, but that's probably mostly because there's a lot of bosses left, which tend to use it.)

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,177 @@
...
+        <Animation Name="gDmZl4Anim_022840" Offset="0x22840" />
+        <Animation Name="gDmZl4Anim_022BE8" Offset="0x22BE8" />
+        <!-- showing off something to her side? -->
+        <Animation Name="gDmZl4Anim_0235D4" Offset="0x235D4" />
+        <Animation Name="gDmZl4Anim_023A50" Offset="0x23A50" />
+        <!-- shock? you chose no at the wrong time? -->
+        <Animation Name="gDmZl4Anim_0241F4" Offset="0x241F4" />
+        <Animation Name="gDmZl4Anim_024510" Offset="0x24510" />
+        <!-- arms behind back head down -->
+        <Animation Name="gDmZl4Anim_024B68" Offset="0x24B68" />
```
EllipticEllipsis, line 167: These are identical to OoT (of course the pointers are different since the object is shifted). Essentially this object is 
 - identical to OoT down to 0xE050
 - new animations
 - 0x15DB0–0x24B80 = OoT's 0xE050–end
 - New flute DLs

and this is IMO worth noting in comments (this object does not change at all between versions in OoT).

You'll find that most, if not all, are documented in OoT's z_en_zl4.c: https://github.com/zeldaret/oot/blob/master/src/overlays/actors/ovl_En_Zl4/z_en_zl4.c#L132-L177 which you can refer to for preliminary documentation if you want; all you have to do is shift the addresses. Otherwise there's not a lot of point in documenting these in a less useful way, we may as well wait for the OoT names.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
     (ActorFunc)DmZl_Draw,
 };
 
-#endif
+static AnimationInfo sAnimations[7] = {
```
EllipticEllipsis, line 33:
```suggestion
static AnimationInfo sAnimations[] = {
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+
+typedef enum {
+    /* 0 */ ZELDA_MOUTH_NEUTRAL,
+    /* 1 */ ZELDA_MOUTH_SMILING,
+    /* 2 */ ZELDA_MOUTH_FROWNING,
+    /* 3 */ ZELDA_MOUTH_OPEN,
+} DmZlMouthTextures;
+
+static TexturePtr sEyeTextures[] = {
+    gDmZl4EyeOpenNormalTex, gDmZl4EyeBlinkingTex,        gDmZl4EyeClosedTex,           gDmZl4EyeWideTex,
```
EllipticEllipsis, line 69: You can't show blinking with one texture, is it just half-closed? We usually call those `EyeHalf`

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+    /* 0 */ ZELDA_EYE_OPEN_NORMAL,
+    /* 1 */ ZELDA_EYE_BLINKING,
+    /* 2 */ ZELDA_EYE_CLOSED,
+    /* 3 */ ZELDA_EYE_WIDE,
+    /* 4 */ ZELDA_EYE_HAPPY,
+    /* 5 */ ZELDA_EYE_OPEN_LOOKING_LEFT,
+    /* 6 */ ZELDA_EYE_OPEN_LOOKING_RIGHT,
+} DmZlEyeTextures;
+
+void DmZl_ChangeAnimation(SkelAnime* skelAnime, AnimationInfo animation[], u16 index) {
```
EllipticEllipsis, line 83: I remain unconvinced that these array arguments are actually helpful. The argument should be an `animationInfo` or `animInfo` anyway, since the actual animation is just one of the fields.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+    Animation_Change(skelAnime, animation->animation, animation->playSpeed, animation->startFrame, endFrame,
+                     animation->mode, animation->morphFrames);
+}
+
+void DmZl_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    DmZl* this = THIS;
+
+    this->animationIndex = 0;
+    this->unused2BA = 0;
```
EllipticEllipsis, line 98: If it's unused it's usually just `unk` with a comment. And shouldn't `animationIndex` use the enum?

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+    this->actor.targetArrowOffset = 1000.0f;
+    ActorShape_Init(&this->actor.shape, 0.0f, ActorShadow_DrawCircle, 24.0f);
+    // these three set to NULL should mean they are dynamically allocated
+    SkelAnime_InitFlex(globalCtx, &this->skelAnime, &gZl4Skeleton, NULL, NULL, NULL, 0);
+    DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], ZELDA_ANIM_FACING_AWAY);
```
EllipticEllipsis, line 103: This doesn't seem like the right use given that the first thing the ChangeAnimation function does is add the last argument on? I.e. the behaviour would be unexpected if `this->animationIndex` was not `0`.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+}
+
+void DmZl_DoNothing(DmZl* this, GlobalContext* globalCtx) {
+}
+
+void DmZl_UpdateCutscene(DmZl* this, GlobalContext* globalCtx) {
+    s32 actionIndex;
+
+    if (Cutscene_CheckActorAction(globalCtx, 0x66)) { // currently in cutscene
+        actionIndex = Cutscene_GetActorActionIndex(globalCtx, 0x66);
```
EllipticEllipsis, line 118: Please remove the comment, there are currently 122 uses of this function in the codebase, and it's clear from the name itself what the function does: it checks if the current cs action is the one in argument.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+            if (nextAnimationIndex != this->animationIndex) {
+                this->animationIndex = nextAnimationIndex;
+                DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], ZELDA_ANIM_FACING_AWAY);
+            }
+        }
+
+        Cutscene_ActorTranslateAndYaw(&this->actor, globalCtx, actionIndex);
+    }
+
+    if (Animation_OnFrame(&this->skelAnime, this->skelAnime.endFrame) != 0) {
```
EllipticEllipsis, line 146: This is boolean, as you can again tell from the name.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+        actionIndex = this->animationIndex;
+        if ((actionIndex == 1) || (actionIndex == 3) || (actionIndex == 5)) {
+            this->animationIndex++;
+            DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], ZELDA_ANIM_FACING_AWAY);
```
EllipticEllipsis, line 150: Again, the enum usage is not correct here: the way this is used, the last argument is an increment to the index rather than an index itself, so this just says "use the current animationIndex".

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+
+typedef enum {
+    ZELDA_ANIM_FACING_AWAY,
+    ZELDA_ANIM_TURNING_TOWARD_PLAYER,
+    ZELDA_ANIM_FACING_PLAYER,
+    ZELDA_ANIM_GIVING_OCARINA_START,
+    ZELDA_ANIM_GIVING_OCARINA,
+    ZELDA_ANIM_PLAYING_OCARINA_START,
+    ZELDA_ANIM_PLAYING_OCARINA,
+    ZELDA_ANIM_MAX,
```
EllipticEllipsis, line 51: Don't forget numeric comments like in the others.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+} DmZlAnimations;
+
+static TexturePtr sMouthTextures[4] = {
```
EllipticEllipsis, line 54:
```suggestion
static TexturePtr sMouthTextures[] = {
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+        case 1:
+            this->mouthTextureIndex = ZELDA_MOUTH_SMILING;
+            break;
+        case 2:
+            this->mouthTextureIndex = ZELDA_MOUTH_FROWNING;
+            break;
+        case 3:
+            this->mouthTextureIndex = ZELDA_MOUTH_OPEN;
+            break;
+    }
```
EllipticEllipsis, line 224: Both switches in this function can use the appropriate enums for their cases.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,246 @@ const ActorInit Dm_Zl_InitVars = {
...
+}
 
-extern UNK_TYPE D_0600DE08;
-extern UNK_TYPE D_0600E038;
+void DmZl_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx) {
+    DmZl* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_Dm_Zl/func_80A38190.s")
+    if (limbIndex == ZL4_LIMB_RIGHT_HAND) {
+        if ((this->animationIndex >= ZELDA_ANIM_GIVING_OCARINA_START) && (this->animationIndex < ZELDA_ANIM_MAX)) {
```
EllipticEllipsis, line 249: This would be clearer with a `<=` IMO, and better for modding if someone wanted to add animations after the last one that didn't draw the Ocarina.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -8,12 +8,25 @@ struct DmZl;
...
+    /* 0x260 */ DmZlActionFunc actionFunc;
+    /* 0x264 */ UNK_TYPE1 pad264[0x4C]; // unused by DmZl
+    /* 0x2B0 */ s16 animationIndex;
+    /* 0x2B2 */ u8 eyeTextureIndexLeft;
+    /* 0x2B3 */ u8 eyeTextureIndexRight;
+    /* 0x2B4 */ u8 mouthTextureIndex;
+    /* 0x2B5 */ u8 nextEyeState;   // used to control eye state, but not set by our actor, outside of actor?
+    /* 0x2B6 */ u8 nextMouthState; // used to control mouth state, but not set by our actor, outside of actor?
+    /* 0x2B8 */ s16 blinkTimer;
+    /* 0x2BA */ s16 unused2BA; // init to 0, unread/unchanged by DmZl after
```
EllipticEllipsis, line 23:
```suggestion
    /* 0x2BA */ s16 unk_2BA; // set but not used by this actor.
```
Until we find it used for something else. Of course you can watch it in an emulator to see if it ever does get read.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -8,12 +8,25 @@ struct DmZl;
...
 typedef struct DmZl {
-    /* 0x0000 */ Actor actor;
-    /* 0x0144 */ char unk_144[0x11C];
-    /* 0x0260 */ DmZlActionFunc actionFunc;
-    /* 0x0264 */ char unk_264[0x70];
+    /* 0x000 */ Actor actor;
+    /* 0x144 */ SkelAnime skelAnime;
+    /* 0x188 */ UNK_TYPE1 pad188[0xD8]; // unused by DmZl
+    /* 0x260 */ DmZlActionFunc actionFunc;
+    /* 0x264 */ UNK_TYPE1 pad264[0x4C]; // unused by DmZl
```
EllipticEllipsis, line 15: I bet these are the joint- and morphTables and a ColliderCylinder, based on what OoT has. I don't see any harm in typing them as such, even if they're unused.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
```diff
@@ -8,12 +8,25 @@ struct DmZl;
...
+    /* 0x2B6 */ u8 nextMouthState; // used to control mouth state, but not set by our actor, outside of actor?
+    /* 0x2B8 */ s16 blinkTimer;
+    /* 0x2BA */ s16 unused2BA; // init to 0, unread/unchanged by DmZl after
+    /* 0x2BC */ UNK_TYPE1 pad2BC[0x18]; // unused by DmZl
 } DmZl; // size = 0x2D4
 
 extern const ActorInit Dm_Zl_InitVars;
 
+// vanilla params are 0x0000, never checked or used
+#define DMZL_TYPE_SOT_CUTCSENE 0
```
EllipticEllipsis, line 30: Since the params are never used, it's probably not worth having this... does anyone feel strongly either way?

EllipticEllipsis (commented):
Good job on this so far, I have a number of suggestions for improving the documentation.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
hensldm: Don't feel strongly, though I would probably lean towards removing it.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
isghj5: I'm not convinced, if it was a one to one conversion there would be no need for a switch, they could just assign `this->mouthTextureIndex = this->nextMouthState;`

But sure I can change it.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
isghj5: I was leaning the other way, I think for modding its important to know what parameters are actually used in the vanilla game to know what changes you made to params input would be affected if the vanilla actor was encountered.

But I can remove it.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
isghj5: How is it clear? What is "ActorAction?" It's the same thing as a "cs action" I guess?

At first glance I assumed it was checking what the actor or player was busy doing, what their "Action" was. Maybe they are busy in an animation, picking something up, which would make sense for a function in the Cutscene File to check if it can **start** a cutscene. Instead its checking if a cutscene is already started, which is different then my first assumption. 

And then we pass a mystery number `0x66` to it, that doesn't help explain what is happening either. Are there a finite list of Cutscenes accessible from global? A small list of limited player actions sounds more believable. 

Nothing about the name of this function alone is intuitive from this end, but I'll remove it.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
isghj5: That doesn't explain what the third value is doing, if its getting an AnimationInfo passed, what does `<AnimationInfo> + 1` evaluate to? This C makes no sense either way. 

Right now, I have to assume the function was designed so that there would be double animations in the array, where there is an offset of one for a different age or different... something, some modifier that would get evaluated last second, but it was unused by this actor because its a 3 minute cutscene actor cutdown from a bigger original design.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,255 @@ const ActorInit Dm_Zl_InitVars = {
...
+                DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], 0);
+            }
+        }
+
+        Cutscene_ActorTranslateAndYaw(&this->actor, globalCtx, actionIndex);
+    }
+
+    if (Animation_OnFrame(&this->skelAnime, this->skelAnime.endFrame)) {
+        actionIndex = this->animationIndex;
+        if ((actionIndex == 1) || (actionIndex == 3) || (actionIndex == 5)) {
```
hensldm, line 156:
```suggestion
        if ((actionIndex == ZELDA_ANIM_TURNING_TOWARD_PLAYER) || (actionIndex == ZELDA_ANIM_FACING_PLAYER) || (actionIndex == ZELDA_ANIM_GIVING_OCARINA)) {
```
Based on the line above, these are suppose to be animations.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,255 @@ const ActorInit Dm_Zl_InitVars = {
...
+                this->animationIndex = nextAnimationIndex;
+                DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], 0);
+            }
+        }
+
+        Cutscene_ActorTranslateAndYaw(&this->actor, globalCtx, actionIndex);
+    }
+
+    if (Animation_OnFrame(&this->skelAnime, this->skelAnime.endFrame)) {
+        actionIndex = this->animationIndex;
```
hensldm, line 155: I'm assuming the variable reuse is needed for matching.

hensldm (commented):
One more thought.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.h
EllipticEllipsis: Not addressed yet?

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,255 @@ const ActorInit Dm_Zl_InitVars = {
...
+        case ZELDA_MOUTH_FROWNING:
+            this->mouthTextureIndex = ZELDA_MOUTH_FROWNING;
+            break;
+        case ZELDA_MOUTH_OPEN:
+            this->mouthTextureIndex = ZELDA_MOUTH_OPEN;
+            break;
+    }
+
+    if (this->animationIndex == ZELDA_ANIM_PLAYING_OCARINA) {
+        // override blinking: eyes closed while playing ocarina
```
EllipticEllipsis, line 235: Comment unnecessary now IMO, since it's now covered by the enums.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +30,255 @@ const ActorInit Dm_Zl_InitVars = {
...
+                    break;
+                case 3:
+                    nextAnimationIndex = ZELDA_ANIM_GIVING_OCARINA_START;
+                    break;
+                case 4:
+                    nextAnimationIndex = ZELDA_ANIM_PLAYING_OCARINA_START;
+                    break;
+                case 1:
+                default:
+                    break;
```
EllipticEllipsis, line 142: Can these go as the first case? Sometimes in switches like this IDO lets you reorder them, be good if they could actually go in numerical order.

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
isghj5: `s/actionIndex/this->animationIndex`
Screenshot_2022-06-15_11-48-43

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -153,7 +163,10 @@ void DmZl_UpdateCutscene(DmZl* this, GlobalContext* globalCtx) {
 
     if (Animation_OnFrame(&this->skelAnime, this->skelAnime.endFrame)) {
         actionIndex = this->animationIndex;
-        if ((actionIndex == 1) || (actionIndex == 3) || (actionIndex == 5)) {
+
+        if ((actionIndex == ZELDA_ANIM_TURNING_TOWARD_PLAYER) || (actionIndex == ZELDA_ANIM_GIVING_OCARINA_START) ||
+            (actionIndex == ZELDA_ANIM_PLAYING_OCARINA_START)) {
+            // these animations dont loop at the end, they lead into the next animation
```
EllipticEllipsis, line 169:
```suggestion
            // these animations don't loop at the end, they lead into the next animation
```

src/overlays/actors/ovl_Dm_Zl/z_dm_zl.c
```diff
@@ -30,27 +29,268 @@ const ActorInit Dm_Zl_InitVars = {
...
+            this->animationIndex++;
+            DmZl_ChangeAnimation(&this->skelAnime, &sAnimations[this->animationIndex], 0);
+        }
+    }
+}
+
+/**
+ * Updates the eye blinking and state, and mouth textures.
+ *  (Copy-Paste from OoT, Unused in MM)
```
EllipticEllipsis, line 178: The function's used, it's called in Update, it's just that most of the control flow isn't touched since `nextEyeState` is never set. You could comment on that above the switch.

assets/xml/objects/object_zl4.xml
```diff
@@ -1,125 +1,163 @@
...
-        <Animation Name="object_zl4_Anim_0235D4" Offset="0x235D4" />
-        <Animation Name="object_zl4_Anim_023A50" Offset="0x23A50" />
-        <Animation Name="object_zl4_Anim_0241F4" Offset="0x241F4" />
-        <Animation Name="object_zl4_Anim_024510" Offset="0x24510" />
-        <Animation Name="object_zl4_Anim_024B68" Offset="0x24B68" />
-        <DList Name="object_zl4_DL_024D80" Offset="0x24D80" />
-        <DList Name="object_zl4_DL_024E40" Offset="0x24E40" />
-        <Texture Name="object_zl4_Tex_024E50" OutName="tex_024E50" Format="rgba16" Width="16" Height="8" Offset="0x24E50" />
-        <!-- <Blob Name="object_zl4_Blob_024F50" Size="0x10" Offset="0x24F50" /> -->
+        <Animation Name="gDmZl4IdleAnim" Offset="0x654" />
```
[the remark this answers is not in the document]
isghj5: Odd way to format that suggestion github, the animation line is removed then re-added without changes?

EllipticEllipsis (approved):

Kenix3 (approved):

PR #838, opened 2022-06-05
Effect Icesmoke OK

hensldm:
Mostly the same as OoT. Main differences are the object is now in `gameplaykeep` instead of `object_fz`, added logic to smoothly decrease alpha to zero when life is below 10, and providing a negative scale results in slightly different `TexScroll` parameters.


engineer124 (approved):

isghj5 (approved):

jpburnett (approved):
🧊

EllipticEllipsis (approved):

Kenix3 (approved):

PR #862, opened 2022-06-22
Adding Git information to tutorial

jpburnett:
Adding an intro to git with helpful commands for someone unfamiliar with git.


docs/CONTRIBUTING.md
```diff
@@ -84,7 +80,8 @@ void CollisionCheck_SpawnWaterDroplets(GlobalContext* globalCtx, Vec3f* v);
 ```
 
 Before PRing with a `NON_MATCHING`, you can try
-- using the [decomp-permuter](tools/decomp-permuter) to find a closer match,
+
+- using the [decomp-permuter](../tools/decomp-permuter) to find a closer match,
```
hensldm, line 84: We no longer have `decomp-permuter` in tools. Probably better to link to the github repo

docs/tutorial/intro_to_git.md
```diff
@@ -0,0 +1,402 @@
...
+             master
+               v
+---o---o---o---o
+            \
+             o---o
+                 ^
+                 A
+```
+
+Each `o` is a commit. The lines show the commit that each commit mas based on: we can see in the local there is one commit with two commits based on it. The local has two branches, `master` and `A`, and they currently both point to the commits furthest along their respective chains.
```
hensldm, line 26:
```suggestion
Each `o` is a commit. The lines show the commit that each commit was based on: we can see in the local there is one commit with two commits based on it. The local has two branches, `master` and `A`, and they currently both point to the commits furthest along their respective chains.
```

EllipticEllipsis (approved):

hensldm (approved):

Kenix3 (approved):

PR #877, opened 2022-06-27
z_map_exp OK and somewhat documented

tom-overton:
I tried to do my best to name things, but it's hard without the other map files being decompiled. I ended up taking a lot from OoT here.


EllipticEllipsis (commented):
Good job on this, I agree that there's not much that can be done until the rest of the map files are at least decompiled.

src/code/z_map_exp.c
```diff
@@ -1,31 +1,231 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_map_exp/func_8010A238.s")
+    for (i = 0; i < ARRAY_COUNT(sDungeonAndBossScenes); i++) {
+        if (Play_GetOriginalSceneNumber(play->sceneNum) == sDungeonAndBossScenes[i]) {
+            return i;
+        }
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_map_exp/func_8010A2AC.s")
+    return -1;
+}
```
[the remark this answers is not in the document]
tom-overton: Just to double-check, you mean document with comments above the function, right? That's what I interpreted, and I'm happy to do it, but I just want to make sure we're on the same page.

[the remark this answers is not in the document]
EllipticEllipsis: Yeah, same as the other files with multiline per-function Doxygen comments, although when the only argument is `play` there's not much reason to document the arguments.

src/code/z_map_exp.c
```diff
@@ -1,31 +1,231 @@
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_map_exp/func_8010A164.s")
+static s32 D_801BF5A4[] = {
+    SCENE_22DEKUCITY,
+    SCENE_KOEPONARACE,
+    SCENE_F01,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_map_exp/func_8010A194.s")
+static s16 sLastRoomNum = 99;
```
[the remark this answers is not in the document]
tom-overton: I'm a bit confused by that last sentence. Do you mean "move them inside the functions"? Because they don't match in the functions unless they're `static`. Or do you mean "remove the `static` from them, and then also move them around so they're closer to the functions that use them"?

src/code/z_map_exp.c
```diff
@@ -1,31 +1,231 @@
...
+
+    return true;
+}
+
+void Map_SavePlayerInitialInfo(PlayState* play) {
+    Player* player = GET_PLAYER(play);
+
+    sPlayerInitialPosX = player->actor.world.pos.x;
+    sPlayerInitialPosZ = player->actor.world.pos.z;
+    sPlayerInitialDirection = ((0x7FFF - player->actor.shape.rot.y) / 0x400);
```
hensldm, line 124:
```suggestion
    sPlayerInitialDirection = (0x7FFF - player->actor.shape.rot.y) / 0x400;
```

hensldm (approved):
1 small nit

src/code/z_map_exp.c
```diff
@@ -1,31 +1,231 @@
...
+
+    func_80105B34(play);
+
+    if ((play->pauseCtx.state == 0) && (play->pauseCtx.debugState == 0)) {
+        if (Map_IsInDungeonArea(play)) {
+            temp_v0_2 = func_80109124(player->actor.world.pos.y);
+            if (temp_v0_2 != -1) {
+                gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].unk_14 |=
+                    gBitFlags[4 - temp_v0_2];
+                XREG(94) = 4 - temp_v0_2;
```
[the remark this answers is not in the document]
tom-overton: The 4 here screams something related to dungeons, though what exactly I still don't know.

tom-overton:
In addition to responding to reviews, the last commit renames `gSaveContext.unk_48C8` to `gSaveContext.dungeonIndex`, and it also documents Elforg a *tiny* bit more now that things are more obvious. Let me know if you'd rather some or all of that is reverted and placed in a separate PR.

include/z64save.h
```diff
@@ -273,6 +273,13 @@ typedef enum SunsSongState {
     /* 3 */ SUNSSONG_SPECIAL // time does not advance, but signals the song was played. used for freezing redeads
 } SunsSongState;
 
+typedef enum DungeonIndex {
+    /* 0 */ DUNGEON_INDEX_WOODFALL,
+    /* 1 */ DUNGEON_INDEX_SNOWHEAD,
+    /* 2 */ DUNGEON_INDEX_GREAT_BAY,
+    /* 3 */ DUNGEON_INDEX_STONE_TOWER,
```
EllipticEllipsis, line 280: I'd note that this also applies to Inverted Stone Tower Temple even though it's a different scene. We should also distinguish between Stone Tower (the outside) and the Temple itself.

tom-overton: I thought just labeling it "**Dungeon** Index" was sufficient, no? Stone Tower is not a dungeon, whereas Stone Tower Temple is.

EllipticEllipsis: Yeah, I'm thinking more generally, what you get if you search for one or the other.

src/code/z_map_exp.c
```diff
@@ -75,6 +75,9 @@ s32 Map_IsInDungeonArea(PlayState* play) {
     return true;
 }
 
+/**
+ * The values in this array get directly mapped onto the DungeonIndex enum.
+ */
```
EllipticEllipsis, line 80: Ah, I meant put // comments like OoT has all over player_lib now.

include/z64save.h
tom-overton: Hmm, then should they *all* have TEMPLE at the end? I don't see how that same line of logic doesn't apply for Woodfall, for example. I still see this as unnecessary verbosity, but I could do it for all of them.

EllipticEllipsis: Actually it's a fair point for all of them, come to think of it. I think we can manage without, then. One could consider whether we should call them temples instead of dungeons thoughout, but it's probably not worth it.

tom-overton: Whoops, well, I already added TEMPLE, expecting you to say that was a good idea, lol. Let me know if I should remove it.

EllipticEllipsis: I'm not that bothered until it becomes obvious that it's stupidly long in use.

tom-overton: Sure, let's just keep TEMPLE for now then and see how it gets used. Can always drop it later.

EllipticEllipsis (approved):
Note: @engineer124 wants to look at this before merge.

src/code/z_map_exp.c
```diff
@@ -1,31 +1,255 @@
...
+            interfaceCtx->unk_27A = mapIndex;
+        }
+    } else {
+        play->interfaceCtx.mapRoomNum = 0;
+    }
+
+    if (gSaveContext.sunsSongState != SUNSSONG_SPEED_TIME) {
+        gSaveContext.sunsSongState = SUNSSONG_INACTIVE;
+    }
+}
```
engineer124, line 167:
```suggestion
void Map_InitRoomData(PlayState* play, s16 room) {
    s32 mapIndex = gSaveContext.mapIndex;
    InterfaceContext* interfaceCtx = &play->interfaceCtx;

    func_80105C40(room);
    
    if (room >= 0) {
        if (Map_IsInDungeonOrBossArea(play)) {
            gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].unk_18 |=
                gBitFlags[room];
            interfaceCtx->mapRoomNum = room;
            interfaceCtx->unk_27A = mapIndex;
        }
    } else {
        interfaceCtx->mapRoomNum = 0;
    }

    if (gSaveContext.sunsSongState != SUNSSONG_SPEED_TIME) {
        gSaveContext.sunsSongState = SUNSSONG_INACTIVE;
    }
}
```
This function should be structured like this (and it matches). `interfaceCtx` should be loaded at the top, and used where applicable

src/code/z_map_exp.c
```diff
@@ -1,31 +1,255 @@
...
+    s32 mapIndex;
+    InterfaceContext* interfaceCtx;
+
+    mapIndex = gSaveContext.mapIndex;
+    func_80105C40(room);
+    if (room >= 0) {
+        if (Map_IsInDungeonOrBossArea(play)) {
+            interfaceCtx = &play->interfaceCtx;
+            gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].unk_18 |=
+                gBitFlags[room];
```
engineer124, line 156:
```suggestion
            gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].rooms |=
                gBitFlags[room];
```
This permanent flags looks to be storing which rooms have been visited, which I presume would be used in drawing the map to highlight visited rooms. This name also lines up with OoT.

src/code/z_map_exp.c
```diff
@@ -1,31 +1,255 @@
...
+    return true;
+}
+
+void Map_SavePlayerInitialInfo(PlayState* play) {
+    Player* player = GET_PLAYER(play);
+
+    sPlayerInitialPosX = player->actor.world.pos.x;
+    sPlayerInitialPosZ = player->actor.world.pos.z;
+    sPlayerInitialDirection = (0x7FFF - player->actor.shape.rot.y) / 0x400;
+}
```
engineer124, line 144: I know this is what this is called in OoT, but I'm not sure I like this name very much. "PlayerInitial" very much makes me think of when "player" is initialized, which is when the scene is initialized. However, this function is called every single room transition, which happens many times after player is initialized. I think it could be something like:

`Map_SavePlayerInitialInfo` -> `Map_SavePlayerRoomInitInfo`
`sPlayerInitialPosX` -> `sPlayerRoomInitPosX`
`sPlayerInitialPosZ` -> `sPlayerRoomInitPosZ`
`sPlayerInitialDirection` -> `sPlayerRoomInitDirection` (or maybe `sPlayerRoomInitDir`)

I even wonder if this could reflect that these are the coordinates and direction of the red arrow/marker on the minimap that's present for each room:
`sMinimapInitPosX` or `sMinimapPlayerInitPosX` or `sMiniMapInitEntrancePosX` just to brainstorm a few names. But maybe just `RoomInit` is fine
image

src/code/z_map_exp.c
```diff
@@ -1,31 +1,255 @@
...
+
+    mapIndex = gSaveContext.mapIndex;
+    func_80105C40(room);
+    if (room >= 0) {
+        if (Map_IsInDungeonOrBossArea(play)) {
+            interfaceCtx = &play->interfaceCtx;
+            gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].unk_18 |=
+                gBitFlags[room];
+            interfaceCtx->mapRoomNum = room;
+            interfaceCtx->unk_27A = mapIndex;
```
engineer124, line 158:
```suggestion
            interfaceCtx->dungeonMapIndex = mapIndex;
```
This seems to be the map index for dungeons/boss areas specifically. Maybe a name `dungeonOrBossAreaMapIndex` for consistency with `Map_IsInDungeonOrBossArea`

src/code/z_map_exp.c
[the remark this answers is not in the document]
engineer124:
```suggestion
                gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].floors |=
                    gBitFlags[4 - floor];
                XREG(94) = 4 - floor;
```
I tested this in-game and `temp_v0_2 = func_80109124(player->actor.world.pos.y);` is getting the floor number. The index is relative to each of the 4 dungeons, such that:

Woodfall Temple:
1F -> 0
2F -> 1

Snowhead Temple:
B1 -> 0
1F -> 1
2F -> 2
3F -> 3
4F -> 4

Great Bay Temple:
B2 -> 0
B1 -> 1
1F -> 2

Stone Tower Temple:
B1 -> 0
1F -> 1

The "4" here represent the max index for the floor available (a dungeon in MM can have only 5 floors, indexed 0-4), so they're storing floors visiting backwards. That's what the `4 - floor` is. Even if you look at Snowhead Temple Map, which has the most floors, the Map can only fit 5 floors on the left. So I wonder if it's worth having a define, something like:
`#define FLOOR_INDEX_MAX 4` or something

image

src/code/z_map_exp.c
```diff
@@ -1,31 +1,255 @@
...
+    }
+
+    return true;
+}
+
+s32 D_801BF5A4[] = {
+    SCENE_22DEKUCITY,
+    SCENE_KOEPONARACE,
+    SCENE_F01,
+};
```
engineer124, line 116: Any thoughts yet as to what connects these 3 scenes?

engineer124 (commented):
Nice work! Some comments/docs suggestions

src/code/z_map_exp.c
tom-overton: I just tested with some random values, and indeed these are for the red arrow.

src/code/z_map_exp.c
tom-overton: Nope, we don't know. It's not "scenes that have moving dots on the map", because Stone Tower and Road to Ikana have dots for the boulders. I've also tried just removing all of the things in this array, but there doesn't seem to be any actual effect on the game.

tom-overton:
I didn't know where to stick `#define FLOOR_INDEX_MAX 4`, so I just put it right above the function that uses it. Open to suggestions, though.

src/code/z_map_exp.c
engineer124: bump, nothing was changed here. Would you prefer this to be unnamed for now?

src/code/z_map_exp.c
```diff
@@ -1,31 +1,259 @@
...
+/**
+ * When a room is loaded, this function is used to save the player's position and rotation
+ * so that the red arrow on the minimap can be drawn correctly.
+ */
+void Map_SavePlayerRoomInitInfo(PlayState* play) {
+    Player* player = GET_PLAYER(play);
+
+    sMinimapInitPosX = player->actor.world.pos.x;
+    sMinimapInitPosZ = player->actor.world.pos.z;
+    sMinimapInitDir = (0x7FFF - player->actor.shape.rot.y) / 0x400;
```
engineer124, line 147:
```suggestion
void Minimap_SavePlayerRoomInitInfo(PlayState* play) {
    Player* player = GET_PLAYER(play);

    sMinimapInitPosX = player->actor.world.pos.x;
    sMinimapInitPosZ = player->actor.world.pos.z;
    sMinimapInitDir = (0x7FFF - player->actor.shape.rot.y) / 0x400;
```
Because this is specifically for the minimap, I wonder if this prefix is appropriate?

src/code/z_map_exp.c
```diff
@@ -1,31 +1,259 @@
...
+    }
+
+    func_80105B34(play);
+
+    if ((play->pauseCtx.state == 0) && (play->pauseCtx.debugState == 0)) {
+        if (Map_IsInDungeonArea(play)) {
+            floor = func_80109124(player->actor.world.pos.y);
+            if (floor != -1) {
+                gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].unk_14 |=
+                    gBitFlags[FLOOR_INDEX_MAX - floor];
```
engineer124, line 248:
```suggestion
                gSaveContext.save.permanentSceneFlags[Play_GetOriginalSceneNumber(play->sceneNum)].floors |=
                    gBitFlags[FLOOR_INDEX_MAX - floor];
```
Same here, this was missed.

engineer124:
> I didn't know where to stick `#define FLOOR_INDEX_MAX 4`, so I just put it right above the function that uses it. Open to suggestions, though.

That's fine, I believe there's a `z64map.h` being introduced with `z_map_disp.c` so we can move the define there once that's decompiled

src/code/z_map_exp.c
tom-overton: Oops, I just completely missed it, sorry. Let me name it.

src/code/z_map_exp.c
tom-overton: I think the prefixes are still WIP considering that map_disp and map_exp are not decompiled, but sure, might as well prefix it with that.

tom-overton:
Oh, I didn't realize "just make a `z64map.h`" was an option. Why don't I just do that now, and future map PRs can expand on it.

src/code/z_map_exp.c
tom-overton: This gets used in quite a few actors for non-floor-like things, seemingly. I've made the change so you can see for yourself, but I'm curious if you're still up for it once you see it.

engineer124: Interesting, so they do seem to reuse `unk_14` based on scene. For all dungeon scenes, it's used to store floors bits. But for non-dungeons scenes, it's used for various other bits depending on the scene. So yeah, it'll either need a more generic name or a union of flags. Definitely outside the scope of this PR, so yeah can keep it as `unk_14` for now.

include/z64map.h
```diff
@@ -0,0 +1 @@
+#define FLOOR_INDEX_MAX 4
```
engineer124, line 1:
```suggestion
#ifndef Z64MAP_H
#define Z64MAP_H

#define FLOOR_INDEX_MAX 4

#endif
```
Don't forget the header-guards

src/code/z_map_exp.c
tom-overton: Sounds good, I'll change it back.

include/z64map.h
tom-overton: Whoops, haven't made a C header in over an age. Thanks for the reminder.

include/z64save.h
```diff
@@ -79,7 +79,7 @@ typedef struct PermanentSceneFlags {
     /* 0x0C */ u32 clearedRoom;
     /* 0x10 */ u32 collectible;
     /* 0x14 */ u32 unk_14;
```
engineer124, line 81:
```suggestion
    /* 0x14 */ u32 unk_14; // varies based on scene. For dungeons, floors visited. 
```
Maybe we can record what we worked out? Optional though

include/z64save.h
```diff
@@ -273,6 +273,13 @@ typedef enum SunsSongState {
     /* 3 */ SUNSSONG_SPECIAL // time does not advance, but signals the song was played. used for freezing redeads
 } SunsSongState;
 
+typedef enum DungeonIndex {
+    /* 0 */ DUNGEON_INDEX_WOODFALL_TEMPLE,
+    /* 1 */ DUNGEON_INDEX_SNOWHEAD_TEMPLE,
+    /* 2 */ DUNGEON_INDEX_GREAT_BAY_TEMPLE,
+    /* 3 */ DUNGEON_INDEX_STONE_TOWER_TEMPLE, // Also applies to Inverted Stone Tower Temple
+} DungeonIndex;
```
engineer124, line 281:
```suggestion
typedef enum {
    /* 0 */ DUNGEON_INDEX_WOODFALL_TEMPLE,
    /* 1 */ DUNGEON_INDEX_SNOWHEAD_TEMPLE,
    /* 2 */ DUNGEON_INDEX_GREAT_BAY_TEMPLE,
    /* 3 */ DUNGEON_INDEX_STONE_TOWER_TEMPLE, // Also applies to Inverted Stone Tower Temple
} DungeonIndex;
```
Don't need this I believe. Also, this probably makes more sense in `z64save.h` since it is an enum for `gSaveContext.dungeonIndex`. But I should at least ask if it's worth moving to `z64map.h`. Don't think so though

tom-overton: I think it belongs in `z64save.h,` yeah. In a future where we don't just combine everything into one big header, someone who only interacts with `gSaveContext` probably expects to get this enum when they include `z64save.h`.

engineer124 (approved):

include/z64save.h
```diff
@@ -257,7 +257,7 @@ typedef struct SaveContext {
     /* 0x3F60 */ u8 screenScaleFlag;                    // "framescale_flag"
     /* 0x3F64 */ f32 screenScale;                       // "framescale_scale"
     /* 0x3F68 */ CycleSceneFlags cycleSceneFlags[120];  // Scene flags that are temporarily stored over the duration of a single 3-day cycle
-    /* 0x48C8 */ u16 unk_48C8;                          // "scene_id_mix"
+    /* 0x48C8 */ u16 dungeonIndex;                          // "scene_id_mix"
     /* 0x48CA */ u8 maskMaskBit[27];                    // "mask_mask_bit", masks given away on the Moon
```
engineer124, line 261:
```suggestion
    /* 0x48C8 */ u16 dungeonIndex;                       // "scene_id_mix"
    /* 0x48CA */ u8 maskMaskBit[27];                    // "mask_mask_bit", masks given away on the Moon
```
Could you fix the spacing on this comment?

Kenix3 (approved):

PR #888, opened 2022-07-01
set LANG variable in Makefile

SailorSnoW:
On non-english system, the `diff.py` script could not be able to work due to the map file being created using the system localization.

As the `search_map_file()` function of the script is searching for "_load address_" words, map files created using non-english localization will not be read correctly, causing the script to fail with `Not able to find .o file for function.`.

A workaround is then to set the `LANG` environnement variable to `en_US` or `C` to ensure the map files being created in english.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,9 @@ N_THREADS ?= $(shell nproc)
 
 #### Setup ####
 
+# Ensure the map file being created using English localization
+export LANG := C
+
 ifeq ($(NON_MATCHING),1)
   CFLAGS := -DNON_MATCHING
   CPPFLAGS := -DNON_MATCHING


Dragorn421:
Hi, do you have documentation on binutil's usage of `LANG`? I can't find anything about it in the docs. I only found a mention of it in gcc docs :smile: 
Also I know there is an option when building binutils to make it locale-aware or not, or something similar, do you know if all distribution-provided packages are built with that enabled?

AngheloAlf:
Huh, this is a weird an unexpected feature...

Using `LANG=es_CL.utf8 make `
```
..ovl_En_Ma4    0x0000000080abdca0     0x2b90 dirección de carga 0x0000000000f23710
```

Using `LANG=fr_FR.utf8 make`:
```
..ovl_En_Ma4    0x0000000080abdca0     0x2b90 address de chargement 0x0000000000f23710
```

I guess a proper fix would be to change the behaviour of `diff.py` to not rely on `load address`, but I guess isn't a bad workaround for now.
We should probably investigate about the locale-awareness of binutils as dragorn said.

Makefile
```diff
@@ -28,6 +28,8 @@ N_THREADS ?= $(shell nproc)
 
 #### Setup ####
 
+export LANG := C
```
AngheloAlf, line 31: Could you add a comment explaining why this is needed?

SailorSnoW: done :)

SailorSnoW:
I don't know a lot about binutils, and from what I've searched, I still haven't found anything about the localization stuffs in binutils.

As Dragorn said, theses localization variables seems to be mentionned only in the gcc doc:
https://gcc.gnu.org/onlinedocs/gcc-4.4.0/gcc/Environment-Variables.html#Environment-Variables/

Dragorn421:
:crying_cat_face: no secret docs

Snooping around some more, I realized `LANG` is a more generic env variable than just binutils/gcc, see `man locale`

Also from https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_02
> If the locale value is "C" or "POSIX", the POSIX locale shall be used and the standard utilities behave in accordance with the rules in [POSIX Locale](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html#tag_07_02) for the associated category.

So `LANG=C` is good I guess and binutils Just Works :tm: with it

--------------

Another thing, should the locale be set for the whole makefile or just when linking specifically for the map file?:

```diff
-	$(LD) -T build/undefined_syms.txt -T build/ldscript.txt --no-check-sections --accept-unknown-input-arch --emit-relocs -Map build/mm.map -o $@
+	LANG=C $(LD) -T build/undefined_syms.txt -T build/ldscript.txt --no-check-sections --accept-unknown-input-arch --emit-relocs -Map build/mm.map -o $@
```

AngheloAlf (approved):

hensldm (approved):

Kenix3 (approved):

PR #899, opened 2022-07-04
Document object_uch (Aliens) and remove a dummy label match in EnInvadepoh

tom-overton:
Just felt like doing this. Note that the "official" name of these things are "Ghosts", but I have absolutely never heard anyone in my life call them that. I called out the official name in a comment, but I named all the assets after "Alien", which is what I usually hear them called.

While I was in Invadepoh, I took another look at one of the few remaining dummy label matches and managed to solve it, so I included it in this PR.


src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
```diff
@@ -618,8 +618,8 @@ unkStruct80B50350 D_80B50350[10];
 EnInvadepoh* D_80B503F0;
 EnInvadepoh* D_80B503F4;
 EnInvadepoh* D_80B503F8;
-AnimatedMaterial* D_80B503FC;
-AnimatedMaterial* D_80B50400;
+AnimatedMaterial* alienEyeBeamTexAnim;
+AnimatedMaterial* alienEmptyTexAnim;
```
AngheloAlf, line 622:
```suggestion
AnimatedMaterial* sAlienEyeBeamTexAnim;
AnimatedMaterial* sAlienEmptyTexAnim;
```

assets/xml/objects/object_uch.xml
```diff
@@ -1,46 +1,64 @@
 ﻿<Root>
+    <!-- Assets for the Ghosts, more-commonly known as the aliens -->
```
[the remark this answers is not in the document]
tom-overton: It comes from The Legend of Zelda: Encyclopedia, which while I am not an expert on the "Zelda canon", I think we pull some other names from it. Additionally, their model is called "ghost.cmb" in MM3D; skeletal models are seemingly almost-always renamed by Grezzo to match "official" terminology:
image

With these two pieces of evidence combined, I'm fairly certain that "Ghost" is the official name for these things...which, as you noted, is insanely confusing. There are tons of ghosts in this game, many more "ghost-like" than this, and the English-speaking community has called them "aliens" for at least two decades now. That's why I used `Alien` in every single asset, but I figured I'd toss the official name in as a comment. I can remove it, though, if it adds unnecessary confusion.

src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
hensldm: Yeah they probably should. In general `bss` should also be static since they are file specific, we just don't because IDO + bss is a nightmare, let alone static bss that wouldn't show up in the mapfile.

hensldm (commented):
Also if you want, on line 3686 there is a `(this->unk3BC < 0) ^ 1` that can be changed to `(this->unk3BC >= 0)` and works fine.

src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
tom-overton: Sure, that makes sense. Perhaps the style guide should be updated to mention that *all* file-specific variables should start with `s`.

hensldm (approved):

assets/xml/objects/object_uch.xml
```diff
@@ -1,46 +1,64 @@
...
+        <DList Name="gAlienEyeBeamDL" Offset="0x80" />
+        <Texture Name="gAlienEyeBeamMaskTex" OutName="alien_eye_beam_mask" Format="i8" Width="32" Height="32" Offset="0x130" />
+        <TextureAnimation Name="gAlienEyeBeamTexAnim" Offset="0x550" />
+
+        <!-- An empty texture animation, seemingly planned to animate something on the alien's body. -->
+        <TextureAnimation Name="gAlienEmptyTexAnim" Offset="0x560" />
+
+        <!-- Alien Animations -->
+        <Animation Name="gAlienDeathAnim" Offset="0x608" /> <!-- Original name is "uch_die_body" -->
+        <Animation Name="gAlienDamageAnim" Offset="0x6C8" /> <!-- Original name is "uch_gaku_body" (probably short for the onomatopoeia for "jerking") -->
```
EllipticEllipsis, line 14: Ideally we'd describe these by what they look like rather than how they're used.

AngheloAlf (approved):

assets/xml/objects/object_uch.xml
tom-overton: These two animations look almost identical (these are both animated GIFs):
damage
death

So describing them based on how they look seems counterproductive.

(We've also used die/death and damage/damaged as animation names in the past without issue, so it'd be pretty weird/almost kind of funny to switch away from that for the object where they are almost completely identical).

tom-overton: Talked it over in discord and I'm swapping to `gAlienJerkingAnim`. The death one is staying though.

Kenix3 (approved):

PR #914, opened 2022-07-07
Adding Docker Support

jpburnett:
This commit adds support for building the repository with Docker. Instructions for how to build are added to `BUILDING_DOCKER.md` as well as explaining briefly how it works.


EllipticEllipsis (commented):
Can't really comment on the technical aspects, but I think it is worth adding a note that since none of our regular contributors use this setup, it's not guaranteed to work on latest master without adjustment (think OoT has a similar note?).

isghj5:
Screenshot_2022-07-07_11-39-49

Well this is gonna take awhile to test on my 6mbps internet connection (shared with other people)...

docs/BUILDING_DOCKER.md
```diff
@@ -0,0 +1,47 @@
...
+
+## 4. Setup and Build the ROM
+
+Once inside the container, you can follow step will need to run the following command:
+
+```bash
+git config --global --add safe.directory /mm
+```
+
+This command will add the mm repo as a trusted repository. After that continue with step [5. Make and Build the ROM](../README.md#5-make-and-build-the-rom) of the Linux instructions to setup and build the ROM, or run any other command you need.
```
isghj5, line 47: Was this line meant to be split into parts? `[5.` kinda looks like it was meant to be another double hash header

Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM ubuntu:21.10 as build
```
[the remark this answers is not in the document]
jpburnett: I think it was from documentation I was using and probably just ctrl+c, ctrl+v'd it...I think changing it to any LTS version makes perfect sense.

docs/BUILDING_DOCKER.md
jpburnett: I think it was a remnant of formatting, I can fix it 👍

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN python3 -m pip install --no-cache-dir -r requirements.txt
+
+ENV LANG C.UTF-8
+
+RUN mkdir /mm
+
+WORKDIR /mm
+
+ENTRYPOINT ["/bin/bash", "-c"]
```
hensldm, line 32: newline?

hensldm (approved):
Seems good to me, but should have somebody who can actually test approve as well.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    unzip \
+    clang-tidy-11 \
+    clang-format-11 \
+    libpng-dev && \
+    apt clean && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY requirements.txt requirements.txt
+
+RUN python3 -m pip install --no-cache-dir -r requirements.txt
```
isghj5, line 24:
```suggestion
RUN python3 -m pip install --no-cache-dir colorama watchdog python-Levenshtein cxxfilt \
                                              pycparser -r requirements.txt
```
Requested change: more dependencies installed for tools

These were required to get mips2c and diff.py working.
Formating might be weird, no preference if you want to change it.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
+    apt-get update && apt-get install -y \
+    build-essential \
+    binutils-mips-linux-gnu \
+    pkg-config \
+    python3 \
+    python3-pip \
+    git \
+    wget \
+    unzip \
```
isghj5, line 15:
```suggestion
    unzip \
    vbindiff \
    vim \
```
vbindiff is so important when you aren't sure what isn't matching, Vim I want because of a script of mine that dumps details into a editable window.

isghj5 (commented):
I went through all the tools I could think of that anyone would want to run in this VM:
`mips2c` sometimes has an odd error, posted in `#mm-misc`, not sure what it is but even if we don't figure it out the webui still exists so workaround exists.
It might be beyond the scope of this PR but I feel like `gfxdis.f2dex2` should be installed in this docker, since it's so important for Draw function decomp, but getting that installed looks more involved, might be worth a different PR instead.
Didn't test permuter, I use a script that needs more work before I can even attempt to run it.

Minor thing I would like fixed ~but not sure if possible~:
the `git config --global --add safe.directory /mm` line has to be run every time you spin up the docker as the image doesn't save it. ~If that could be added to the Dockerfile so its not required if I want to check git that would be nice to fix (again, not sure if this is possible, not sure what Dockerfile can even handle).~

just tested, just adding RUN in front of that line and adding to the Dockerfile works (neat)... uh, is there a reason not to add it?
```suggestion
RUN git config --global --add safe.directory /mm
```

Dockerfile
AngheloAlf: PR #918 adds those to the `requirements.txt` file, so if that one gets in there should be no need to add those explicitly here

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ubuntu:22.04 as build
+ENV TZ=UTC
+
+LABEL Cammoguy <jpburnett@gwu.edu>
```
AngheloAlf, line 4: What is `LABEL` intended for? Can it have a bit of a more generic name?

jpburnett: I think it is just fluff. I was using other files as reference and used as the "maintainer" so I put it in there so people knew who to yell at if it wasn't working 😅

isghj5 (approved):

EllipticEllipsis (approved):

AngheloAlf (approved):

Kenix3 (approved):

PR #918, opened 2022-07-08
Yet another disasm.py speedup

AngheloAlf:
Change `disasm.py` to use [rabbitizer](https://pypi.org/project/rabbitizer/) as the instruction decoder instead of the bundled `mips_isa.py` file.

The advantage of using this package instead of what we are doing now is this package is a fast MIPS instruction decoder fully written in C, so runtime and memory usage are greatly improved.
For example, running the disasm.py script on current master:

```
$ time python3 tools/disasm/disasm.py -j 1 --reg-names=o32 --all

...

real    1m1.855s
user    1m1.200s
sys    0m4.860s
```

This PR:
```
$ time python3 tools/disasm/disasm.py -j 1 --reg-names=o32 --all

...

real    0m32.606s
user    0m30.492s
sys    0m4.293s
```

(I couldn't get proper memory usage measurements because the multithreading stuff messed up with them.  lmk if anybody knows a proper way to measure it)

I cleaned up and de-duplicated some code. It is a bit more readable imo.

I also took this opportunity to clean up the `requirements.txt` file and remove what seems to be dead dependencies and add `diff.py` dependencies and `rabbitizer`


tools/disasm/disasm.py
```diff
@@ -2642,4 +2467,4 @@ def do_splits(out_dict, info):
 
         out_path /= label + ".s"
         with open(out_path, "w") as outfile:
-            outfile.write(all_out.strip() + "\n")
+            outfile.write(all_out.strip() + "\n")
```
hensldm, line 2470: newline?

requirements.txt
```diff
@@ -1,7 +1,7 @@
 libyaz0>=0.5
```
hensldm, line 1: Just in general for this file, it looks like requirements.txt are allowed to have comments (general python comments # work). I think it would be helpful to section the requirements off based on what they are needed for (Building, disasm, diff, etc.).

hensldm (commented):
I'm also checking it out to do some testing

tools/disasm/disasm.py
AngheloAlf: I have no idea how this happened

hensldm (approved):
Seems to work fine for me. Always enjoy the speed ups :) .

Jenkinsfile
```diff
@@ -20,6 +20,12 @@ pipeline {
                 sh 'bash -c "tools/reloc_spec_check.sh"'
             }
         }
+        stage('Install Python dependencies') {
+            steps {
+                echo 'Installing Python dependencies'
+                sh 'python3 -m pip install -r requirements.txt'
+            }
```
EllipticEllipsis, line 27: I wish we could just subrepo it instead, this basically counters all new timesave.

tools/asm-processor/prelude.inc
```diff
@@ -41,3 +41,6 @@
 .set $fs4f,         $f29
 .set $fs5,          $f30
 .set $fs5f,         $f31
+
+# Floating-point status register
+.set $FpcCsr,       $31
```
EllipticEllipsis, line 46: This is a subrepo, right? So this should be PR'd to that repo. And the permuter might need it too. And maybe decomp.me?

EllipticEllipsis (commented):
Works on Mac, code looks good from what I can see, but there's a few issues with the surroundings.

Jenkinsfile
AngheloAlf: The package is written in C, so people would require to build it themselves. They would require to install the python dev package to build it too. People on Windows (cygwin, not wsl) would require the VisualStudio toolchain just to build it. Too much trouble imo.

Also, this step only adds 2 extra seconds on the jenkins side

tools/asm-processor/prelude.inc
AngheloAlf: No. As every time this topic comes up, the asm processor is not a subrepo... for reasons I can never remember.

If using this register alias is an issue I can disable it on rabbitizer's configuration

hensldm: https://github.com/zeldaret/mm/pull/278#pullrequestreview-735244353

Jenkinsfile
hensldm: Yeah subrepo doesn't really help since Jenkins would have to build it anyways, which is almost certainly going to be longer than just installing a few python packages.

jpburnett (approved):

tools/asm-processor/prelude.inc
EllipticEllipsis: I think we can probably live without this, I expect mips2c wouldn't like it either.

AngheloAlf: Removed the use of `$FpcCsr` for now instead

EllipticEllipsis (approved):

Kenix3 (approved):

PR #939, opened 2022-07-13
Effect Gspk OK

hensldm:


src/overlays/effects/ovl_Effect_Ss_G_Spk/z_eff_ss_g_spk.c
```diff
@@ -1,30 +1,145 @@
...
     EffectSsGSpk_Init,
 };
 
-#endif
+static TexturePtr sSparkTextures[] = {
+    gElectricSpark1Tex,
+    gElectricSpark2Tex,
+    gElectricSpark3Tex,
+    gElectricSpark4Tex,
+};
```
isghj5, line 39: So these are drawn sequentially because the index is incremented?

Might want to add that as a comment that this is meant to be a sort of manual animation

isghj5 (approved):

src/overlays/effects/ovl_Effect_Ss_G_Spk/z_eff_ss_g_spk.c
EllipticEllipsis: And since it's an array, maybe start at 0?

EllipticEllipsis (approved):

src/overlays/effects/ovl_Effect_Ss_G_Spk/z_eff_ss_g_spk.c
hensldm: I'd prefer to keep the asset names synced with OoT for now which start at 1, though I do see the merit of starting at 0. I can come back to it when all the effects are done/in. Same sentiment with the comment, as it would apply to more then just this effect, so kind of want to wait to see how it goes. (Don't worry I'll add this to my todo list so it won't get lost).

Kenix3 (approved):

PR #949, opened 2022-07-14
Some minor cleanup involving stray fairies

tom-overton:
I noticed this was another example of having the limbs in an external enum rather than in the object, so I went and did that. I also noticed there was no params-packing macro, so I did that too. And while I was working with elforg, I did some very minor cleanup.

### DIFF
--- a/src/overlays/actors/ovl_En_Elforg/z_en_elforg.c
+++ b/src/overlays/actors/ovl_En_Elforg/z_en_elforg.c
@@ -5,7 +5,6 @@
  */
 
 #include "z_en_elforg.h"
-#include "objects/gameplay_keep/gameplay_keep.h"
 
 #define FLAGS (ACTOR_FLAG_10)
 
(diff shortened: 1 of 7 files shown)


AngheloAlf (approved):

src/overlays/actors/ovl_En_Elforg/z_en_elforg.c
```diff
@@ -70,34 +69,37 @@ void EnElforg_Init(Actor* thisx, PlayState* play) {
     s32 pad;
     EnElforg* this = THIS;
 
-    Actor_SetScale(&this->actor, 0.01f);
-    this->flags = 0;
+    Actor_SetScale(thisx, 0.01f);
```
EllipticEllipsis, line 72: What's the improvement from using thisx here?

tom-overton: Elforg was done back before we decided that, so I only used `thisx` for the one macro below. Things have changed since then, so that's why I used `thisx` in Init.

hensldm (approved):

Kenix3 (approved):

PR #951, opened 2022-07-14
Document DmChar02 and its animations, rename one struct var in DmStk

tom-overton:
I documented the four animations that this actor used and finished documenting the actor itself (it was mostly already done, thanks Chloe!). While I was doing that, I noticed that Skull Kid used `animationId`, which is something that almost nothing else in the codebase used, so I switched it over to `animIndex` like how DmChar02 does it.


src/overlays/actors/ovl_Dm_Char02/z_dm_char02.c
```diff
@@ -130,15 +145,16 @@ void DmChar02_Update(Actor* thisx, PlayState* play) {
     DmChar02* this = THIS;
 
     SkelAnime_Update(&this->skelAnime);
-    this->unk2F0 = this->unk2F0;
+    this->unk_2F0 = this->unk_2F0;
```
isghj5, line 148: This is required to match? Wouldn't IDO recognize this doesn't do anything and omit it anyway?

isghj5 (approved):

src/overlays/actors/ovl_Dm_Char02/z_dm_char02.c
ChloeBangBang: It is required to match, and is also the only usage of the field. Very weird.

tom-overton: Yeah this in a few actors like this. I have zero idea why it's necessary, but it is. Thanks IDO!

Kenix3 (approved):

PR #966, opened 2022-07-20
EnRat (Real Bombchu) OK and Documented, object_rat documented 🐀💣 

tom-overton:
I'm not 100% on these names (and I also left some unnamed Vec3f temps in one of the functions), but at this point I think the fastest way to improve these names is to get some eyes on them.


src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +44,910 @@ const ActorInit En_Rat_InitVars = {
...
+            return;
+        }
+    } else if (((this->collider.base.ocFlags1 & OC1_HIT) && (((this->collider.base.oc->category == ACTORCAT_ENEMY)) ||
+                                                             (this->collider.base.oc->category == ACTORCAT_BOSS) ||
+                                                             (this->collider.base.oc->category == ACTORCAT_PLAYER))) ||
+               ((this->actionFunc == EnRat_ChasePlayer) && (this->timer == 0))) {
+        this->collider.base.ocFlags1 &= ~OC1_HIT;
+        EnRat_Explode(this, play);
+        return;
+    }
```
isghj5, line 804: Seems slightly odd to me, actors with this much collision logic usually get their own CheckCollision function instead of update being this big, wonder if this dev was different, new to the project?

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +44,910 @@ const ActorInit En_Rat_InitVars = {
...
+
+    this->actionFunc(this, play);
+
+    if ((this->actionFunc != EnRat_PostDetonation) && (this->actionFunc != EnRat_Revive)) {
+        if (this->stunTimer > 0) {
+            this->stunTimer--;
+        } else if (this->stunTimer < 0) {
+            if (this->stunTimer == -2) {
+                this->stunTimer = -1;
```
isghj5, line 813: Hmm, non timer behavior on timer. Union?

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +44,910 @@ const ActorInit En_Rat_InitVars = {
...
+s32 EnRat_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, Actor* thisx) {
+    EnRat* this = THIS;
+
+    if (limbIndex == REAL_BOMBCHU_LIMB_BODY) {
+        pos->y -= this->revivePosY;
+    }
+
+    if (limbIndex == REAL_BOMBCHU_LIMB_TAIL_END) {
+        *dList = NULL;
+    }
```
isghj5, line 877: Weird. It's clear from the next function they just want this limb to draw the fuse, but why would they nullify the DL here 100% of the time when they could just make that limb null then add their thing next function? Does the skeleton in z64Utils actually show something on that tail limb?

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +44,910 @@ const ActorInit En_Rat_InitVars = {
...
+/**
+ * Fly backwards until the Real Bombchu touches a wall, ceiling, or floor, then explode.
+ */
+void EnRat_Bounced(EnRat* this, PlayState* play) {
+    this->actor.shape.rot.x -= 0x700;
+    Math_StepToF(&this->actor.shape.yOffset, 1700.0f, 170.0f);
+    this->timer--;
+    if (this->timer == 0) {
+        this->timer = 30;
+    }
```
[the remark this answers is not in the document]
tom-overton: They use the timer to control the bomb flashes. I think the idea here is that if the Bombchu got bounced, the bomb should flash, but it shouldn't automatically detonate, or really even do anything. You see this pattern with the Dungeon-type Bombchu, who constantly go 30 -> 0 -> 30 -> 0 -> etc to keep the bomb flashing. This particular one you highlighted just does it for both types of Bombchu.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
tom-overton: Yeah this was definitely a weird actor in terms of implementation. I think large chunks of it got copied from EnBomChu without a ton of modification.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
tom-overton: Ehhh, not really convinced. When you hookshot a Bombchu, it sets `stunTimer` to -2, which then *immediately* goes to -1 (this is the code you highlighted). However, they don't *do* anything with that value; they never check it later in the actor, making the whole thing sort of pointless. I didn't feel like making a union for something that might as well not exist.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
tom-overton: Yeah, it's a triangle:
image

Your guess is as good as mine as to why they actually put something here.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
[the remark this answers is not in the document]
tom-overton: To add to what I said before, Overworld-type Bombchu *do* use the timer for something; they explode when it reaches 0.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
isghj5: I can think of at least two possibilities:

They originally wanted a static bomb but it looked bad so they decided to make it dynamic after the fact.

Or maybe they wanted the bomb/fuse to be removable/disarmable with some item/form move, and they were going to give him some frizy tail left over after you remove the bomb, but they ran out of time and had to work on other things.

Well, a blue triangle isnt as cool but at least there is something there after all to nullify.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
isghj5: If you are sure -1/-2 does nothing should you include a comment here saying the values aren't used by the code for anything? My opinion here is to leave some sort of comment to stop the reader from thinking those values actually have purpose and understanding them isn't necessary nearby to prevent confusion, although I have been asked to remove comments like that before.

tom-overton: Reading it again and cross-referencing with OoT's flags documentation, it actually does *something*, but it's a bit silly. They start by setting it to -2, and I think after this point, the Bombchu gets hooked (aka, ACTOR_FLAG_2000 is set). Then, it gets set to -1, and any time *after* this, if ACTOR_FLAG_2000 is unset, the Bombchu blows up.

I guess you could call this something like `hookedState` or something, with values like `HOOK_STARTED` and `HOOKED`. But I'm still not 100% sure it deserves a union. What do you think?

isghj5: Yeah I would union at that point, with the two states, I don't have a better name for the two states although I like hookedState.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+        angle = this->actor.yawTowardsPlayer - this->actor.shape.rot.y;
+        if (this->axisUp.y < -0.25f) {
+            angle -= 0x8000;
+        }
+    } else {
+        if (Actor_DistanceToPoint(&this->actor, &this->actor.home.pos) > 50.0f) {
+            Vec3f sp64;
+            Vec3f sp58;
+            Vec3f sp4C;
+            Vec3f sp40;
```
hensldm, line 302: Not quite sure what to name these/`sp74` myself.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+/**
+ * Returns true if the Real Bombchu is on a collision poly or is on the water's surface.
+ */
+s32 EnRat_IsOnCollisionPoly(PlayState* play, Vec3f* posA, Vec3f* posB, Vec3f* posResult, CollisionPoly** poly,
+                            s32* bgId) {
+    WaterBox* waterBox;
+    s32 isOnWater;
+    f32 waterSurface;
+
+    if ((WaterBox_GetSurface1(play, &play->colCtx, posB->x, posB->z, &waterSurface, &waterBox)) &&
```
hensldm, line 340:
```suggestion
    if (WaterBox_GetSurface1(play, &play->colCtx, posB->x, posB->z, &waterSurface, &waterBox) &&
```

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+        }
+    }
+
+    if ((this->animLoopCounter == 0) && (Rand_ZeroOne() < 0.05f)) {
+        Actor_PlaySfxAtPos(&this->actor, NA_SE_EN_BOMCHU_VOICE);
+        this->animLoopCounter = 5;
+    }
+
+    if (!(player->stateFlags3 & PLAYER_STATE3_100) && (this->actor.xzDistToPlayer < this->attackRange) &&
+        (Player_GetMask(play) != PLAYER_MASK_STONE) && (Actor_IsFacingPlayer(&this->actor, 0x3800))) {
```
hensldm, line 592:
```suggestion
        (Player_GetMask(play) != PLAYER_MASK_STONE) && Actor_IsFacingPlayer(&this->actor, 0x3800)) {
```

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+    this->animLoopCounter = 3;
+    this->actor.speedXZ = 0.0f;
+    this->actionFunc = EnRat_SpottedPlayer;
+}
+
+/**
+ * Play the "spotted" animation a few times, then start chasing the player.
+ */
+void EnRat_SpottedPlayer(EnRat* this, PlayState* play) {
+    if ((this->animLoopCounter == 3) && (Animation_OnFrame(&this->skelAnime, 5.0f))) {
```
hensldm, line 609:
```suggestion
    if ((this->animLoopCounter == 3) && Animation_OnFrame(&this->skelAnime, 5.0f)) {
```

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+ */
+void EnRat_ChasePlayer(EnRat* this, PlayState* play) {
+    Player* player = GET_PLAYER(play);
+    Vec3f blureP1;
+    Vec3f blureP2;
+
+    this->actor.speedXZ = 6.1f;
+    if (this->hasLostTrackOfPlayer) {
+        if (!(player->stateFlags3 & PLAYER_STATE3_100) && (Player_GetMask(play) != PLAYER_MASK_STONE) &&
+            (Actor_IsFacingPlayer(&this->actor, 0x3000))) {
```
hensldm, line 657:
```suggestion
            Actor_IsFacingPlayer(&this->actor, 0x3000)) {
```

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+        this->timer = 30;
+    }
+
+    if (this->actor.bgCheckFlags & (0x10 | 0x8 | 0x1)) {
+        EnRat_Explode(this, play);
+    }
+}
+
+void EnRat_Explode(EnRat* this, PlayState* play) {
+    EnBom* bomb;
```
hensldm, line 734:
```suggestion
    EnBom* bomb = (EnBom*)Actor_Spawn(&play->actorCtx, play, ACTOR_EN_BOM, this->actor.world.pos.x, this->actor.world.pos.y,
                               this->actor.world.pos.z, 0, 0, 0, ENBOM_0);
```

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+        Matrix_ReplaceRotation(&play->billboardMtxF);
+        Matrix_MultZero(&this->smokePos);
+        this->smokePos.y += 15.0f;
+        currentMatrixState = Matrix_GetCurrent();
+
+        if (this->actionFunc == EnRat_ChasePlayer) {
+            s32 i;
+
+            gDPSetPrimColor(POLY_XLU_DISP++, 0, 0, 255, 255, 150, 255);
+            gDPSetEnvColor(POLY_XLU_DISP++, 0xFF, 0x00, 0x00, 0x00);
```
hensldm, line 907:
```suggestion
            gDPSetEnvColor(POLY_XLU_DISP++, 255, 0, 0, 0);
```

hensldm (approved):
Just a bunch of minor things.

hensldm:
Also forgot 🐀 :)

assets/xml/objects/object_rat.xml
```diff
@@ -1,28 +1,38 @@
 ﻿<Root>
+    <!-- Assets for Real Bombchus -->
```
engineer124, line 2: Where does the name "Real Bombchu" come from, I would picture it more as something like "Enemy Bombchu" perhaps?

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+    normDotUp = DOTXYZ(normal, this->axisUp);
+
+    if (fabsf(normDotUp) >= 0.999f) {
+        return false;
+    }
+
+    angle = func_80086C48(normDotUp);
+    if (angle < 0.001f) {
+        return false;
+    }
```
engineer124, line 237:
```suggestion
    normDotUp = DOTXYZ(normal, this->axisUp);
    if (fabsf(normDotUp) >= 0.999f) {
        return false;
    }

    angle = func_80086C48(normDotUp);
    if (angle < 0.001f) {
        return false;
    }
```
Just for consistency

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+            Matrix_MultVec3f(&this->actor.home.pos, &sp64);
+            Matrix_MultVec3f(&this->actor.world.pos, &sp58);
+            angle = Math_Vec3f_Yaw(&sp58, &sp64) - Math_Vec3f_Yaw(&sp58, &sp4C);
+            if (sp40.y < -0.25f) {
+                angle -= 0x8000;
+            }
+
+            angle += (s16)randPlusMinusPoint5Scaled(2048.0f);
+        } else {
+            angle = (Rand_ZeroOne() < 0.1f) ? (s16)randPlusMinusPoint5Scaled(2048.0f) : 0;
```
engineer124, line 319:
```suggestion
            angle += (s16)randPlusMinusPoint5Scaled(0x800);
        } else {
            angle = (Rand_ZeroOne() < 0.1f) ? (s16)randPlusMinusPoint5Scaled(0x800) : 0;
```
Especially with the clamp below

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+        }
+    }
+
+    angle = CLAMP(angle, -0x800, 0x800);
+    Matrix_RotateAxisF(angle * 0.0000958738f, &this->axisUp, MTXMODE_NEW);
```
engineer124, line 324:
```suggestion
    Matrix_RotateAxisF(angle * (M_PI / 0x8000), &this->axisUp, MTXMODE_NEW);
```
We really should consider an `BINANG_TO_RAD_ALT` or something for cases like this. It'd be worth comparing the possible usage of that to the current `#define BINANG_TO_RAD(binang) (((f32)binang / 0x8000) * M_PI)`

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+        return true;
+    }
+
+    return false;
+}
+
+/**
+ * "Floor" in this case means floors, walls, and ceilings, since the Real Bombchu can run on any of them.
+ */
+s32 EnRat_IsTouchingFloor(EnRat* this, PlayState* play) {
```
engineer124, line 368: Isn't all surfaces floors, walls, ceilings? So could this be `EnRat_IsTouchingSurface`? Or if there's surfaces that aren't included, could you emphasize those in the description?

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+    }
+
+    return true;
+}
+
+/**
+ * Transform coordinates from actor coordinate space to world space, according to current orientation.
+ * `offset` is expected to already be at world scale.
+ */
+void EnRat_ActorCoordsToWorld(EnRat* this, Vec3f* offset, Vec3f* pos) {
```
engineer124, line 458: Could you make an note about what actor coordinate space is? It's distinct from model space right? Which is what I originally thought it was.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+        this->axisUp.z = (cos * this->axisUp.z) - (sin * tempX);
+
+        tempX = this->axisLeft.x;
+        this->axisLeft.x = (sin * this->axisLeft.z) + (cos * tempX);
+        this->axisLeft.z = (cos * this->axisLeft.z) - (sin * tempX);
+    }
+}
+
+Vec3f sSmokeAccel = { 0.0f, 0.6f, 0.0f };
+Color_RGBA8 sSmokeColor = { 255, 255, 255, 255 };
```
engineer124, line 512:
```suggestion
static Vec3f sSmokeAccel = { 0.0f, 0.6f, 0.0f };
static Color_RGBA8 sSmokeColor = { 255, 255, 255, 255 };
```
I feel these are generic enough that they might clash with other names. I would assume many actors could spawn smoke

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+    Animation_MorphToLoop(&this->skelAnime, &gRealBombchuRunAnim, -5.0f);
+    this->actor.speedXZ = 6.1f;
+    EnRat_UpdateSparkOffsets(this);
+    this->actionFunc = EnRat_ChasePlayer;
+}
+
+Vec3f sBlureP1Offset = { 0.0f, 10.5f, -9.0f };
+Vec3f sBlureP2LeftOffset = { 18.0f, 0.0f, -7.5f };
+Vec3f sBlureP2RightOffset = { -18.0f, 0.0f, -7.5f };
+Vec3f sDustVelocity = { 0.0f, 3.0f, 0.0f };
```
engineer124, line 643: Same comment as above, these could probably use `static` since they're generic names

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+
+    DECR(this->timer);
+
+    if ((this->timer == 0) && (EN_RAT_GET_TYPE(&this->actor) == EN_RAT_TYPE_DUNGEON)) {
+        this->timer = 30;
+    }
+
+    EnRat_SpawnSmoke(this, play);
+    this->visualJitter =
+        (5.0f + (Rand_ZeroOne() * 3.0f)) * Math_SinS(((Rand_ZeroOne() * 512.0f) + 12288.0f) * this->timer);
```
engineer124, line 680:
```suggestion
        (5.0f + (Rand_ZeroOne() * 3.0f)) * Math_SinS(((Rand_ZeroOne() * 0x200) + 0x3000) * this->timer);
```

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+
+    if (EN_RAT_GET_TYPE(&this->actor) == EN_RAT_TYPE_DUNGEON) {
+        EnRat_ActorCoordsToWorld(this, &sBlureP1Offset, &blureP1);
+
+        EnRat_ActorCoordsToWorld(this, &sBlureP2LeftOffset, &blureP2);
+        EffectBlure_AddVertex(Effect_GetByIndex(this->blure1Index), &blureP1, &blureP2);
+
+        EnRat_ActorCoordsToWorld(this, &sBlureP2RightOffset, &blureP2);
+        EffectBlure_AddVertex(Effect_GetByIndex(this->blure2Index), &blureP1, &blureP2);
+    } else if ((this->actor.floorPoly != NULL) && !(play->gameplayFrames & 3)) {
```
engineer124, line 690:
```suggestion
    } else if ((this->actor.floorPoly != NULL) && ((play->gameplayFrames & 3) == 0)) {
```
Possible `% 4`?

src/overlays/actors/ovl_En_Rat/z_en_rat.c
```diff
@@ -28,123 +47,913 @@ const ActorInit En_Rat_InitVars = {
...
+                currentMatrixState->mf[3][0] = this->smokePos.x + ptr->x;
+                currentMatrixState->mf[3][1] = this->smokePos.y + ptr->y;
+                currentMatrixState->mf[3][2] = this->smokePos.z + ptr->z;
+                gSPMatrix(POLY_XLU_DISP++, Matrix_NewMtx(play->state.gfxCtx),
+                          G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+                gSPSegment(POLY_XLU_DISP++, 0x08, sSparkTextures[(play->gameplayFrames + i) & 3]);
+                gSPDisplayList(POLY_XLU_DISP++, gEffSparkDL);
+            }
+
+            Matrix_Scale(0.022222223f, 0.022222223f, 0.022222223f, MTXMODE_APPLY);
```
engineer124, line 921: Could these be 1 / 45?

src/overlays/actors/ovl_En_Rat/z_en_rat.h
```diff
@@ -2,12 +2,57 @@
...
 
 #include "global.h"
+#include "objects/object_rat/object_rat.h"
+
+#define EN_RAT_IS_OVERWORLD_TYPE(thisx) ((thisx)->params & 0x8000)
+#define EN_RAT_GET_TYPE(thisx) ((thisx)->params)
+#define EN_RAT_GET_ATTACK_RANGE(thisx) ((thisx)->params & 0xFF)
+
+/**
+ * There are quite a few differences between the two types of Real Bombchu:
```
engineer124, line 12:
```suggestion
 * There are many differences between the two types of Real Bombchu:
```

assets/xml/objects/object_rat.xml
tom-overton: Straight from the game itself:

NKZE01_2022-07-24_17-05-25

src/overlays/actors/ovl_En_Rat/z_en_rat.c
tom-overton: Yeah, I'm sure there's just some math thing I'm unfamiliar with here. I wish I could name them, but I'm not going to pretend like I know what these are.

src/overlays/actors/ovl_En_Rat/z_en_rat.c
tom-overton: Yeah, %4 works

src/overlays/actors/ovl_En_Rat/z_en_rat.c
tom-overton: Nope, that doesn't work.

hensldm: Hmm, this surprises me. My script says `1.0f/45.0f` should produce the same hex as in the rodata.

tom-overton: Ah, apparently it works, but only if you replace all three. I replaced one, and it didn't work.

tom-overton:
Fixed the title too.

AngheloAlf (approved):

src/overlays/actors/ovl_En_Rat/z_en_rat.c
Kenix3: here may be a better name for axisUp, but it seems to be like a cameras up vector. It's the norm of the floor poly this is standing on.

So first a new matrix is created matrix is rotated based on the negative angles on the actor.
Then axisUp is multiplied into the current matrix and placed in sp40. 
Then the current position has the axis forwarded to it, which would be the tangent of the collision poly. This result is placed in sp64.
Then sp64 is multiplied into the current matrix and placed in sp4c.
Then the current home position is multiplied into the current matrix and placed in sp64 (seems reused)
Then the current world position is multiplied into the current matrix and placed into sp58.
Then it looks for the yaw between sp58 and sp64 then subtract the yaw between sp58 and sp4c from the original yaw and sets that to angle.
The angle is then randomized a little.

The angle is the output of this calculation.

Does this help you name it at all?

tom-overton: If I'm being honest, not particularly, no. I don't really understand what the matrix at the start is even supposed to represent.

Kenix3: This can be added later. Can you put my comments in at least?

tom-overton: As we talked about in Discord, I managed to sort of figure out what's going on here with debug prints. Feel free to critique the names, however, since I'm not going to claim I have a perfect understanding of this or anything.

Kenix3 (approved):

PR #967, opened 2022-07-20
Document powder keg assets

tom-overton:
I wanted these assets for something, so I figured I'd label them. I also named `EnBomStruct`, since I was messing around with the fuse anyway to document the asset.


assets/xml/overlays/ovl_En_Bom.xml
```diff
@@ -1,9 +1,20 @@
...
+
+        <!-- Textures and DisplayList for the Powder Keg -->
+        <Texture Name="gPowderKegTLUT" OutName="powder_keg_tlut" Format="rgba16" Width="16" Height="16" Offset="0x22B0"/>
+        <Texture Name="gPowderKegTex" OutName="powder_keg" Format="ci8" Width="32" Height="32" Offset="0x24B0"/>
+        <DList Name="gPowderKegDL" Offset="0x2EF0"/>
+
+        <!-- Textures and DisplayList for the Goron Skull on the side of the Powder Keg -->
+        <Texture Name="gPowderKegGoronSkullTLUT" OutName="powder_keg_goron_skull_tlut" Format="rgba16" Width="16" Height="16" Offset="0x3048"/>
+        <Texture Name="gPowderKegGoronSkullTex" OutName="powder_keg_goron_skull" Format="ci8" Width="16" Height="32" Offset="0x3248"/>
+        <DList Name="gPowderKegGoronSkullDL" Offset="0x3548"/>
```
engineer124, line 18: Is it actually a Goron Skull? I thought it was a generic skull. We should make sure these names are synced with the `GI` version in my `z_draw` PR, although it looks pretty similar.
https://github.com/engineer124/mm/blob/z_draw/assets/xml/objects/object_gi_bigbomb.xml

assets/xml/overlays/ovl_En_Bom.xml
```diff
@@ -1,9 +1,20 @@
...
+        <!-- Texture and DisplayLists for the Powder Keg's fuse -->
+        <Texture Name="gPowderKegFuseTex" OutName="powder_keg_fuse" Format="i8" Width="2" Height="2" Offset="0x21E0"/>
+        <DList Name="gPowderKegFuseMaterialDL" Offset="0x21E8"/>
+        <DList Name="gPowderKegFuseModel1DL" Offset="0x2270"/>
+        <DList Name="gPowderKegFuseModel2DL" Offset="0x2290"/>
+
+        <!-- Textures and DisplayList for the Powder Keg -->
+        <Texture Name="gPowderKegTLUT" OutName="powder_keg_tlut" Format="rgba16" Width="16" Height="16" Offset="0x22B0"/>
+        <Texture Name="gPowderKegTex" OutName="powder_keg" Format="ci8" Width="32" Height="32" Offset="0x24B0"/>
+        <DList Name="gPowderKegDL" Offset="0x2EF0"/>
```
engineer124, line 13: I think I used `gPowderKegBarrelDL` instead of just a generic `gPowderKegDL` here, but it would be good double check if they're the same or different

assets/xml/overlays/ovl_En_Bom.xml
tom-overton: Yeah, it's subtly different from the GI version, but it's similar enough.
GI version:
image

En_Bom version:
image

As for why I called it a "Goron Skull", it doesn't look really human. It has eyes that are really far apart and a pointed head. It's even more obviously a Goron (in my opinion) in MM3D:
image

assets/xml/overlays/ovl_En_Bom.xml
tom-overton: GI version:
image

En_Bom version:
image

The scale is *way* different, and the asset itself is subtly different, but yeah, it's the same thing. I'll put "barrel" in mine too.

engineer124 (approved):

hensldm (approved):

src/overlays/actors/ovl_En_Bom/z_en_bom.c
```diff
@@ -668,19 +668,19 @@ void func_808726DC(PlayState* play, Vec3f* arg1, Vec3f* arg2, Vec3f* arg3, s32 a
     f32 temp_f20;
     Vec3f spCC;
     Vec3f spC0;
-    EnBomStruct* ptr = &D_80874650[0];
-    EnBomStruct* ptr2 = &D_80874650[1];
+    PowderKegFuseSegment* ptr = &sPowderKegFuseSegments[0];
+    PowderKegFuseSegment* ptr2 = &sPowderKegFuseSegments[1];
```
EllipticEllipsis, line 672: Can probably just do a quick name on these variables too.

tools/disasm/variables.txt
```diff
@@ -5740,7 +5740,7 @@
     0x808743E8:("D_808743E8","f32","",0x4),
     0x808743EC:("D_808743EC","f32","",0x4),
     0x808743F0:("D_808743F0","f32","",0x4),
-    0x80874650:("D_80874650","s80874650","[16]",0x1c0),
+    0x80874650:("sPowderKegFuseSegments","s80874650","[16]",0x1c0),
```
EllipticEllipsis, line 5743: Doesn't really matter, but the type is wrong now.

EllipticEllipsis (approved):

src/overlays/actors/ovl_En_Bom/z_en_bom.c
tom-overton: To what? We use `ptr` for these sorts of things all the time.

EllipticEllipsis: Only when we don't know what they are.

tom-overton: This isn't true. A collection of stuff I found from the codebase:
En_Tite:
```c
ptr = &this->limbPos[i];
```
Bg_Spdweb:
```c
ptr = &this->collider.elements[i];
```
En_Kame:
```c
ptr = &this->limbPos[5];
```
En_Kusa (admittedly a bit different because it references `unk_194`, but we know that's an array of `EnKusa2`):
```c
        ptr = this->unk_194;
        actor = (EnKusa2*)Actor_SpawnAsChildAndCutscene(
            &play->actorCtx, play, ACTOR_EN_KUSA2, this->actor.world.pos.x, this->actor.world.pos.y,
            this->actor.world.pos.z, 0, Rand_Next() >> 0x10, 0, 1, this->actor.cutscene, this->actor.unk20, NULL);
        *ptr = actor;
```

tom-overton: But to cut to the point, I don't know why there are two pointers here. And given that, I'd rather just keep the names we have, because something like `segmentPtr` and `segmentPtr2` don't really seem better than the current names.

tom-overton: Slept on this one and I changed my mind on it. I swapped them all over to `fuseSegmentPtr`. Wish I knew why they frequently have two pointers but oh well.

Kenix3 (approved):

PR #990, opened 2022-07-26
General Cleanup 10

hensldm:
Its been mentioned in a couple PRs, but Elliptic and I agree that the limb enum names should be singular, so this PR converts these from using plural to singular (Limbs->Limb). It also changes the auto-genned names to PascalCase to match our style for enum names.


assets/xml/objects/object_boss01.xml
```diff
@@ -113,7 +113,7 @@
         <Limb Name="object_boss01_Standardlimb_00EFB8" Type="Standard" EnumName="OBJECT_BOSS01_1_LIMB_31" Offset="0xEFB8" />
         <Limb Name="object_boss01_Standardlimb_00EFC4" Type="Standard" EnumName="OBJECT_BOSS01_1_LIMB_32" Offset="0xEFC4" />
         <Limb Name="object_boss01_Standardlimb_00EFD0" Type="Standard" EnumName="OBJECT_BOSS01_1_LIMB_33" Offset="0xEFD0" />
-        <Skeleton Name="object_boss01_Skel_00F0A8" Type="Flex" LimbType="Standard" LimbNone="OBJECT_BOSS01_1_LIMB_NONE" LimbMax="OBJECT_BOSS01_1_LIMB_MAX" EnumName="object_boss01_1_Limbs" Offset="0xF0A8" />
+        <Skeleton Name="object_boss01_Skel_00F0A8" Type="Flex" LimbType="Standard" LimbNone="OBJECT_BOSS01_1_LIMB_NONE" LimbMax="OBJECT_BOSS01_1_LIMB_MAX" EnumName="ObjectBoss011Limbs" Offset="0xF0A8" />
```
jpburnett, line 116: Do we care for an exception for the pascal case here or is the `Boss011` okay?

jpburnett: Actually coming back to this it's not a huge deal, it'll be "fixed" once the actor is documented more

jpburnett (approved):
Gonna go out on a limb here and say good job ;)

tom-overton (approved):

assets/xml/objects/object_boss01.xml
tom-overton: Yeah, I'm not a huge fan of this, but at the end of the day this is a temporary name. Once the object is actually documented, this will go away.

hensldm: For what it is worth the same thought went through my head as I was fixing these. However it is temporary, and I can't really think of a good solution other then just wait until they renamed.

AngheloAlf: Speaking of which, shouldn't this be `ObjectBoss011Limb` instead? (without the "s")

hensldm: Yes, yes it should. Fixed.

AngheloAlf (approved):

assets/xml/objects/gameplay_field_keep.xml
```diff
@@ -19,7 +19,7 @@
         <Limb Name="gameplay_field_keep_Standardlimb_002F60" Type="Standard" EnumName="GAMEPLAY_FIELD_KEEP_LIMB_05" Offset="0x2F60" />
         <Limb Name="gameplay_field_keep_Standardlimb_002F6C" Type="Standard" EnumName="GAMEPLAY_FIELD_KEEP_LIMB_06" Offset="0x2F6C" />
         <Limb Name="gameplay_field_keep_Standardlimb_002F78" Type="Standard" EnumName="GAMEPLAY_FIELD_KEEP_LIMB_07" Offset="0x2F78" />
-        <Skeleton Name="gameplay_field_keep_Skel_002FA0" Type="Normal" LimbType="Standard" LimbNone="GAMEPLAY_FIELD_KEEP_LIMB_NONE" LimbMax="GAMEPLAY_FIELD_KEEP_LIMB_MAX" EnumName="gameplay_field_keep_Limbs" Offset="0x2FA0" />
+        <Skeleton Name="gameplay_field_keep_Skel_002FA0" Type="Normal" LimbType="Standard" LimbNone="GAMEPLAY_FIELD_KEEP_LIMB_NONE" LimbMax="GAMEPLAY_FIELD_KEEP_LIMB_MAX" EnumName="GameplayFieldKeepLimb" Offset="0x2FA0" />
```
EllipticEllipsis, line 22: Could you name the stuff that's used in EnButte? `gButterflySkel` etc. I make the exception here because it's the only unnamed skeleton in a keep object.

hensldm: I can, though I really do not want to do it in this PR. Seems better to just make a PR documenting `EnButte` instead (I'll finish what is in OoT and port it over). 
Though regarding your comment about this being the only unnamed skeleton in a keep object, there are actually 4 in gameplay_keep without autogenned enum names so are not  showing up in this PR. Out of curiosity @AngheloAlf was this intentional or an oversight? If an oversight I can add them to this PR real quick.

hensldm: So I went through and figured out what all the unnamed skeletons. I still feel naming everything associated with them is out of scope for this PR, instead I'll update the `EnumName` prefix for the buttefly, and add enums with the proper prefix for the gameplay_keep ones (as well as for the door skel which is named but has no enum).

Kenix3 (approved):

PR #992, opened 2022-07-27
General Cleanup 11

hensldm:
I should be getting close to done with these for the time being :) .
Does 5 things:
1. Changes char to UNK_TYPE1 for decomped actors with struct padding
2. Fixes a couple usages of SubS functions
3. Uses the EXCHANGE_ITEMs as the last arg to func_800B8500
4. Cleans up some actor macros
5. A few misc stuff I saw


isghj5:
While you are swapping around my old GET_ACTOR_PARAM macros back to ACTOR_GET_PARAM, EnKakasi also should be fixed.

Dragorn421:
Is `char` -> `UNK_TYPE1` something that has been discussed for MM?

For OoT iirc last time we talked about these UNK_ types the consensus seemed to be "we won't ever know what these are/were so char is okay as a placeholder"

AngheloAlf:
Yes we have been using `UNK_TYPE1` as a placeholder instead of `char`

AngheloAlf (approved):

src/overlays/actors/ovl_En_Knight/z_en_knight.h
```diff
@@ -12,16 +12,16 @@ typedef struct EnKnight {
...
+    /* 0x152 */ char unk_152[0x1];
     /* 0x153 */ u8 unk_153;
     /* 0x154 */ u8 unk_154;
-    /* 0x155 */ char pad155[0x2CB];
+    /* 0x155 */ char unk_155[0x2CB];
     /* 0x420 */ EnKnightActionFunc actionFunc;
     /* 0x424 */ char unk_424[0x25C];
     /* 0x680 */ u8 unk_680;
-    /* 0x681 */ char pad681[9];
+    /* 0x681 */ char unk_681[0x9];
```
isghj5, line 24: Wait are we removing pads for unk_ values now? Do we know they are values used by another actor in the Igos fight?

edit: also they are char

src/overlays/actors/ovl_En_Ma_Yto/z_en_ma_yto.h
```diff
@@ -40,7 +40,7 @@ typedef enum {
     /* 4 */ MA_YTO_TYPE_4  // HugCutscene? Doesn't seem to work properly in-game
 } EnMaYtoType;
 
-#define EN_MA_YTO_PARSE_TYPE(params) (((params)&0xF000) >> 12)
+#define EN_MA_YTO_PARSE_TYPE(thisx) (((thisx)->params & 0xF000) >> 12)
```
isghj5, line 43: How do you feel about removing `PARSE` from params macros?

src/overlays/actors/ovl_En_Knight/z_en_knight.h
hensldm: This actor has yet to be decomped, so unsure if they really are pads so would prefer `unk` for the time being. Same reason why they stay `char`

EllipticEllipsis: We have no idea what these are since this actor has not been decompiled yet. The struct is therefore the same as the original but with some of the variables identified. `pad` means explicitly that it is known not to be used, that has always been how it's been used in these projects.

src/overlays/actors/ovl_En_Ma_Yto/z_en_ma_yto.h
hensldm: Sure, `GET` seems more common.

isghj5 (approved):

src/overlays/actors/ovl_Obj_Um/z_obj_um.h
```diff
@@ -72,7 +72,7 @@ typedef enum {
     /* 4 */ OBJ_UM_TYPE_POST_MILK_RUN // milk road, post-minigame
 } ObjUmType;
 
-#define OBJ_UM_PARSE_TYPE(thisx) (((thisx)->params & 0xFF00) >> 8)
+#define OBJ_UM_GET_TYPE(thisx) (((thisx)->params & 0xFF00) >> 8)
 #define OBJ_UM_PARSE_PATH_INDEX(thisx) ((thisx)->params & 0xFF)
```
isghj5, line 76: oh one more

Kenix3 (approved):

PR #1012, opened 2022-08-04
Fado, part 2

EllipticEllipsis:
Several improvements to the way that Fado works that were not pushed to the version that got merged, whoops 😓 . Also deleted all the cfg files, apologies for the size of the diff, although that was pretty inevitable. Formatted and rewrote some of the tools folder too, which will also look a much larger change than it really is.


hensldm (approved):
As long as Jenkins is happy, I'm happy.

tom-overton (approved):
Stuff that isn't .cfg files looks fine to me; I did not bother to look at .cfg files even a little bit.

AngheloAlf (commented):
I think we don't use either `yaz0.c`, `yaz0tool.c` nor `makeromfs.c` anywhere, so I think this is a great opportunity to yeet them

EllipticEllipsis:
I think I'd rather just clean the whole folder out in one go once we have a proper tools repo ready.

AngheloAlf (approved):

PR #1022, opened 2022-08-11
Fix `EnHorseAction` enum

AngheloAlf:
While decomping player_actor I noticed the `EnHorseAction` is wrong, it uses the same exact values as OoT but it is a bit shifted.

I also added the `EN_HORSE_CHECK` macros from OoT and made a bit of cleanup


src/overlays/actors/ovl_En_Horse/z_en_horse.h
```diff
@@ -46,33 +46,33 @@ typedef void (*EnHorseCsFunc)(struct EnHorse*, PlayState*, CsCmdActorAction*);
...
+typedef enum EnHorseAction {
+    /*  0 */ ENHORSE_ACT_FROZEN,
+    /*  1 */ ENHORSE_ACT_INACTIVE,
+    /*  2 */ ENHORSE_ACT_IDLE,
+    /*  3 */ ENHORSE_ACT_FOLLOW_PLAYER,
+    /*  4 */ ENHORSE_ACT_INGO_RACE,
+    /*  5 */ ENHORSE_ACT_5,
+    /*  6 */ ENHORSE_ACT_6,
+    /*  7 */ ENHORSE_ACT_MOUNTED_IDLE,
+    /*  8 */ ENHORSE_ACT_MOUNTED_IDLE_WHINNEYING,
```
EllipticEllipsis, line 58: It's actually spelt "whinny": https://en.wiktionary.org/wiki/whinny

src/overlays/actors/ovl_En_Horse/z_en_horse.h
```diff
@@ -46,33 +46,33 @@ typedef void (*EnHorseCsFunc)(struct EnHorse*, PlayState*, CsCmdActorAction*);
...
+    /* 17 */ ENHORSE_ACT_HIGH_JUMP,
+    /* 18 */ ENHORSE_ACT_CS_UPDATE,
+    /* 19 */ ENHORSE_ACT_HBA,
+    /* 20 */ ENHORSE_ACT_FLEE_PLAYER,
     /* 21 */ ENHORSE_ACT_21,
     /* 22 */ ENHORSE_ACT_22,
     /* 23 */ ENHORSE_ACT_23,
     /* 24 */ ENHORSE_ACT_24,
-    /* 25 */ ENHORSE_ACT_25,
+    /* 25 */ ENHORSE_ACT_25
```
EllipticEllipsis, line 75: I'd prefer not to abbreviate `ACTION`, it's only 3 more letters, and we might want `ACT` for something else.

EllipticEllipsis (commented):
Well that makes a lot more sense, but I have a couple of naming quibbles.

isghj5 (approved):
Yeah a lot of that does look more accurate

EllipticEllipsis (approved):

Kenix3 (approved):

PR #1026, opened 2022-08-14
Purge `ExchangeItemID` enum

AngheloAlf:
From current knowledge from player_actor the `ExchangeItemID` was merged with the `PlayerActionParam` enum. 
The values from `exchangeItemId` and action param can be used interchangeably, as seen in [this function](https://github.com/Decompollaborate/mm/blob/b5d95892cd19094a0041c3205be1c95d8c715364/src/overlays/actors/ovl_player_actor/z_player.c#L19839) from player. `exchangeItemId` is even used to [index the get item table](https://github.com/Decompollaborate/mm/blob/b5d95892cd19094a0041c3205be1c95d8c715364/src/overlays/actors/ovl_player_actor/z_player.c#L16597).

Edit: Also I renamed the FISHING_POLE to FISHING_ROD since that's how it is called in-game


EllipticEllipsis (approved):

src/code/z_actor.c
```diff
@@ -1850,10 +1850,10 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
 
 // Actor_PickUpExchange? Seems to be called with exchangeItemId -1 if the same actor used Actor_PickUp
 // This function is also used to toggle the "Speak" action on the A button
-s32 func_800B8500(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, s32 exchangeItemId) {
+s32 func_800B8500(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, PlayerActionParam exchangeItemId) {
```
hensldm, line 1853: Should we also rename the args/temps that are named `exchangeItemId`?

hensldm (commented):
1 question

src/code/z_actor.c
EllipticEllipsis: I would say no, because it's not acting as actionParam (see the Player struct, it's a separate field that happens to use the same enum.

Could we typedef one enum as the other?

hensldm (approved):

src/code/z_actor.c
AngheloAlf: We could make a typedef like
```c
typedef PlayerActionParam ExchangeItemID;
```
But it feels a bit overkill and it may get confusing at some times.

I think we shouldn't make this typedef right now but revisit the idea in the future.

EllipticEllipsis: Maybe leave a note above the actionparam enum that it's also used for exchangeitemid (and that we might consider a typedef in future).

Kenix3 (approved):

PR #1043, opened 2022-08-20
EnGrasshopper (Dragonflies) OK and documented, object_grasshopper documented

tom-overton:
This actor is objectively unfinished, and boy can you tell it! I tried by best to untangle some of the weirder parts of this, but if anything is confusing, just let me know. Also went back and changed something in EnPp based on knowledge that I gained from this actor, but it should be pretty small.

There's this stubborn fake-match in here if anyone wants to take a look: https://decomp.me/scratch/4wJBW


src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.h
```diff
@@ -2,16 +2,78 @@
...
     /* 0x0000 */ Actor actor;
-    /* 0x0144 */ char unk_144[0x164];
+    /* 0x0144 */ SkelAnime skelAnime;
+    /* 0x0188 */ Vec3s jointTable[DRAGONFLY_LIMB_MAX];
+    /* 0x0218 */ Vec3s morphTable[DRAGONFLY_LIMB_MAX];
     /* 0x02A8 */ EnGrasshopperActionFunc actionFunc;
-    /* 0x02AC */ char unk_2AC[0x15C4];
+    /* 0x02AC */ u8 decision;
+    /* 0x02AD */ u8 shouldTurn;
+    /* 0x02AE */ char unk_2AE[2];
```
hensldm, line 41:
```suggestion
    /* 0x02AE */ UNK_TYPE1 unk_2AE[2];
```
Same for the other 2 please :)

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -15,41 +16,114 @@ void EnGrasshopper_Destroy(Actor* thisx, PlayState* play);
...
+typedef enum {
+    /* 0 */ EN_GRASSHOPPER_DECISION_ATTACK,
+    /* 1 */ EN_GRASSHOPPER_DECISION_FLY,
+    /* 2 */ EN_GRASSHOPPER_DECISION_ROAM_IN_CIRCLES, // Never used in the final game
+} EnGrasshopperNextAction;
+
+typedef enum {
+    /* 0 */ EN_GRASSHOPPER_BANK_STATE_BANKING,
+    /* 1 */ EN_GRASSHOPPER_BANK_STATE_DONE,
+} EnGrasshopperbankState;
```
hensldm, line 67:
```suggestion
} EnGrasshopperBankState;
```

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Grasshopper/func_809A71CC.s")
+        if (this->shouldTurn) {
+            this->baseFlyHeight = randPlusMinusPoint5Scaled(50.0f) + this->flyingHomePos.y;
+            this->targetRot.y = Math_Atan2S(diffX, diffZ);
+            this->timer = Rand_S16Offset(30, 30);
+        }
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Grasshopper/func_809A7494.s")
+    if ((Player_GetMask(play) != PLAYER_MASK_STONE) && !(gSaveContext.eventInf[4] & 2) && (!this->shouldTurn) &&
```
hensldm, line 381:
```suggestion
    if ((Player_GetMask(play) != PLAYER_MASK_STONE) && !(gSaveContext.eventInf[4] & 2) && !this->shouldTurn &&
```

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
+            this->timer = Rand_S16Offset(30, 30);
+        }
+
+        Math_ApproachF(&this->actor.world.pos.y, this->targetPosY, 0.1f, 10.0f);
+        collisionCheckPos.x = (Math_SinS(this->actor.shape.rot.y) * 100.0f) + this->actor.world.pos.x;
+        collisionCheckPos.y = this->actor.world.pos.y;
+        collisionCheckPos.z = (Math_CosS(this->actor.shape.rot.y) * 100.0f) + this->actor.world.pos.z;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Grasshopper/func_809A78EC.s")
+        if ((this->actor.bgCheckFlags & 8) || (BgCheck_SphVsFirstPoly(&play->colCtx, &collisionCheckPos, 10.0f))) {
```
hensldm, line 441:
```suggestion
        if ((this->actor.bgCheckFlags & 8) || BgCheck_SphVsFirstPoly(&play->colCtx, &collisionCheckPos, 10.0f)) {
```

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
+    if ((BgCheck_SphVsFirstPoly(&play->colCtx, &this->tailTipPos, 40.0f)) &&
+        (((this->drawDmgEffType == ACTOR_DRAW_DMGEFF_FROZEN_SFX)) ||
+         (this->drawDmgEffType == ACTOR_DRAW_DMGEFF_FROZEN_NO_SFX)) &&
+        (this->drawDmgEffTimer != 0)) {
+        Actor_SpawnIceEffects(play, &this->actor, this->bodyPartsPos, ARRAY_COUNT(this->bodyPartsPos), 2, 0.3f, 0.2f);
+        this->drawDmgEffTimer = 0;
+        this->drawDmgEffType = ACTOR_DRAW_DMGEFF_FIRE;
+    }
+
+    if ((isUnderWater) || (BgCheck_SphVsFirstPoly(&play->colCtx, &this->tailTipPos, 10.0f))) {
```
hensldm, line 805:
```suggestion
    if (isUnderWater || BgCheck_SphVsFirstPoly(&play->colCtx, &this->tailTipPos, 10.0f)) {
```

hensldm (approved):

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
+    Math_SmoothStepToS(&this->actor.shape.rot.y, this->actor.yawTowardsPlayer, 0xA, 0xFA0, 0xA);
+    Math_ApproachF(&this->actor.world.pos.x, this->targetApproachPos.x, 0.3f, this->approachSpeed);
+    Math_ApproachF(&this->actor.world.pos.y, this->targetApproachPos.y, 0.1f, this->approachSpeed);
+    Math_ApproachF(&this->actor.world.pos.z, this->targetApproachPos.z, 0.3f, this->approachSpeed);
+    Math_ApproachF(&this->approachSpeed, 7.0f, 0.3f, 1.0f);
+
+    if (WaterBox_GetSurface1(play, &play->colCtx, this->tailTipPos.x, this->tailTipPos.z, &this->waterSurface,
+                             &waterBox)) {
+        if (this->tailTipPos.y < this->waterSurface) {
+            if ((this->splashCount < 3) || !(play->gameplayFrames & 7)) {
```
AngheloAlf, line 570: This can't be `% 8` right?

AngheloAlf (approved):

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -15,41 +16,114 @@ void EnGrasshopper_Destroy(Actor* thisx, PlayState* play);
...
+    -1, -1, 0, 1, 2, 3, 4, 5, -1, -1, 6, 7, -1, 8, 9, -1, 10, 11, -1, 12, 13, -1, -1, -1,
+};
+
+static s8 sParentBodyParts[] = {
+    -1, -1, -1, 0, 1, 2, 3, -1, -1, -1, -1, 6, -1, -1, 8, -1, -1, 10, -1, -1, 12, -1, -1, -1,
+};
+
+static u8 sShadowSizes[] = {
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+};
```
EllipticEllipsis, line 83: Do all of these arrays have to be the same size, dependent on something? If so might be best to initialise with a specified size.

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -15,41 +16,114 @@ void EnGrasshopper_Destroy(Actor* thisx, PlayState* play);
...
+
+typedef enum {
+    /* 0x0 */ EN_GRASSHOPPER_DMGEFF_NO_DAMAGE,  // Deals no damage
```
EllipticEllipsis, line 86: If it has no effect we usually say `IMMUNE`, since damage is generally handled by the actual number.

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -15,41 +16,114 @@ void EnGrasshopper_Destroy(Actor* thisx, PlayState* play);
...
+    /* 0x2 */ EN_GRASSHOPPER_DMGEFF_FIRE = 0x2, // Damages and sets the Dragonfly on fire
+    /* 0x3 */ EN_GRASSHOPPER_DMGEFF_FREEZE,     // Damages and freezes the Dragonfly in ice
+    /* 0x4 */ EN_GRASSHOPPER_DMGEFF_LIGHT_ORB,  // Damages and surrounds the Dragonfly with light orbs
+    /* 0xE */ EN_GRASSHOPPER_DMGEFF_HOOK = 0xE, // If hit by the Hookshot, it pulls the Dragonfly towards the player
+    /* 0xF */ EN_GRASSHOPPER_DMGEFF_DAMAGE,     // Deals regular damage
```
EllipticEllipsis, line 91: Likewise, this just seems to be `NONE` (as in no effect).

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -15,41 +16,114 @@ void EnGrasshopper_Destroy(Actor* thisx, PlayState* play);
...
+
+static u8 sShadowSizes[] = {
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+};
+
+typedef enum {
+    /* 0x0 */ EN_GRASSHOPPER_DMGEFF_NO_DAMAGE,  // Deals no damage
+    /* 0x2 */ EN_GRASSHOPPER_DMGEFF_FIRE = 0x2, // Damages and sets the Dragonfly on fire
+    /* 0x3 */ EN_GRASSHOPPER_DMGEFF_FREEZE,     // Damages and freezes the Dragonfly in ice
+    /* 0x4 */ EN_GRASSHOPPER_DMGEFF_LIGHT_ORB,  // Damages and surrounds the Dragonfly with light orbs
```
EllipticEllipsis, line 89: I think most of the enemies we've named it after the thing that causes the effect, i.e. fire/ice/light, rather than the actual thing it does. At least that's how I've been doing it.

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
+
+static AnimationHeader* sAnimations[] = {
+    &gDragonflyRaiseTailAnim, &gDragonflyLowerTailAnim, &gDragonflyFlyAnim,  &gDragonflyAttackAnim,
+    &gDragonflyHoverAnim,     &gDragonflyDamageAnim,    &gDragonflyDeadAnim, &gDragonflyFallAnim,
+};
+
+static u8 sAnimationModes[] = {
+    ANIMMODE_ONCE, ANIMMODE_ONCE, ANIMMODE_LOOP, ANIMMODE_ONCE,
+    ANIMMODE_LOOP, ANIMMODE_ONCE, ANIMMODE_ONCE, ANIMMODE_ONCE,
+};
```
EllipticEllipsis, line 198: This looks like it and the enum would go better above (or in?) the animation-changing function. Might also want to put enum comments in, at least on the modes.

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Grasshopper/func_809A700C.s")
+        if (BgCheck_SphVsFirstPoly(&play->colCtx, &collisionCheckPos, 10.0f)) {
+            this->shouldTurn = true;
+        } else {
+            this->shouldTurn = false;
+        }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Grasshopper/func_809A7134.s")
+        //! @bug: unreachable code. To get here, the type must NOT be EN_GRASSHOPPER_TYPE_WOODFALL
```
EllipticEllipsis, line 367:
```suggestion
        //! @bug Unreachable code. To get here, the type must NOT be EN_GRASSHOPPER_TYPE_WOODFALL
```
IIRC this is the usual format, although it's not very consistent.

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
EllipticEllipsis: Seems like this should be possible?

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
+    Vec3f splashPos;
+
+    for (i = 0; i < 6; i++) {
+        EnGrasshopper_InitializeEffect(this, &this->tailTipPos);
+    }
+
+    if (WaterBox_GetSurface1(play, &play->colCtx, this->tailTipPos.x, this->tailTipPos.z, &this->waterSurface,
+                             &waterBox)) {
+        if (this->tailTipPos.y < this->waterSurface) {
+            if ((this->splashCount < 3) || !(play->gameplayFrames & 7)) {
```
EllipticEllipsis, line 620: Same here?

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
+
+    Collider_UpdateSpheres(limbIndex, &this->collider);
+    if ((limbIndex == DRAGONFLY_LIMB_ROOT) || (limbIndex == DRAGONFLY_LIMB_TAIL_SEGMENT_2) ||
+        (limbIndex == DRAGONFLY_LIMB_TAIL_TIP) || (limbIndex == DRAGONFLY_LIMB_BACK_LEFT_UPPER_LEG) ||
+        (limbIndex == DRAGONFLY_LIMB_BACK_RIGHT_UPPER_LEG) || (limbIndex == DRAGONFLY_LIMB_FRONT_LEFT_UPPER_LEG) ||
+        (limbIndex == DRAGONFLY_LIMB_FRONT_LEFT_FOOT) || (limbIndex == DRAGONFLY_LIMB_FRONT_RIGHT_UPPER_LEG) ||
+        (limbIndex == DRAGONFLY_LIMB_HEAD) ||
+        //! @bug: These do not check for valid limbs. Limb index 0x18 is DRAGONFLY_LIMB_MAX, so it (and any index
+        //! larger than it) is not tied to an actual limb.
+        (limbIndex == 0x18) || (limbIndex == 0x19) ||
```
EllipticEllipsis, line 966: Oh what a mess, this may be connected to the weird size of the shadow array too. Also they're indices, so should be in dec.

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
```diff
@@ -64,92 +138,990 @@ const ActorInit En_Grasshopper_InitVars = {
...
+void EnGrasshopper_PostLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx) {
+    EnGrasshopper* this = THIS;
+    Vec3f sEffectOffsetFromTailTop = { 500.0f, 0.0f, 0.0f };
+    Vec3f sZeroVec3f = { 0.0f, 0.0f, 0.0f };
+
+    Matrix_Push();
+
+    if ((this->type != EN_GRASSHOPPER_TYPE_WOODFALL_TEMPLE_FINAL_ROOM) &&
+        (this->type != EN_GRASSHOPPER_TYPE_WOODFALL)) {
+        if ((s8)(sLimbIndexToShadowBodyPartsIndex[limbIndex] >= 0)) {
```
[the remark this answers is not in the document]
tom-overton: Yes, otherwise you don't generate these instructions:
image

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
tom-overton: You gave me completely the opposite guidance on EnSnowman: https://github.com/zeldaret/mm/pull/818#discussion_r864344360
I've been following that guidance for every actor I've done from then to now, including this one. So which is actually preferred?

EllipticEllipsis: Both have advantages and disadvantages, let's leave it and decide when we have all of them.

tom-overton: Sounds good, and there will be inconsistencies anyways. Prior to Snowman, I named them after the *cause*. Then, from Snowman to present day, I named them after the *thing they do*. I never went back and changed my old actors to match how I've been doing things post-Snowman, so there will definitely be inconsistencies there. I guess the best thing to do now is "just give them a logical name and sort out the inconsistencies later".

src/overlays/actors/ovl_En_Grasshopper/z_en_grasshopper.c
tom-overton: `sLimbIndexToShadowBodyPartsIndex` has to be `DRAGONFLY_LIMB_MAX`-sized. The other two I'm not sure about.

tom-overton: Oh, never mind, I remember the other two now. This is a weird one because they get used like this:
```c
SubS_GenShadowTex(this->shadowBodyPartsPos, &this->actor.world.pos, shadowTex, i / 5.0f,
    ARRAY_COUNT(this->shadowBodyPartsPos), sShadowSizes, sParentBodyParts);
```
The thing is that `this->shadowBodyPartsPos` is defined like this:
```c
/* 0x03D8 */ Vec3f shadowBodyPartsPos[14]; // Was almost certainly declared with 24 entries, but only the first 14 are used.
/* 0x0480 */ UNK_TYPE1 unk_480[0x78];
```
So the other two arrays are length 24 now, which theoretically *should* be the size of `this->shadowBodyPartsPos`. But for some bizarre reason, they only use the first 14 entries.

EllipticEllipsis: I'd be tempted to beef that up to the theoretical extent and use that. Up to you, this is such a messy actor it's hard to decide.

tom-overton: Yeah, I asked Anon about this, as the subs expert, and he suggested the current approach. But it could just be:
```c
SubS_GenShadowTex(this->shadowBodyPartsPos, &this->actor.world.pos, shadowTex, i / 5.0f,
    14, sShadowSizes, sParentBodyParts);
```
With `shadowBodyPartsPos` being length 24.

hensldm: When we discussed in Discord, I don't think I realized those other arrays were also length 24. So yeah, I think it would be better to have the shadow body parts array the full 24, than just hardcode the 14, probably with a comment.

tom-overton: If you want to see the original conversation with Anon, it starts here: https://discord.com/channels/688807550715560050/688851976746041391/1009998644637872238

tom-overton: Ah, I think we commented at the same time. I'll switch to the length 24 approach.

EllipticEllipsis (approved):

PR #1044, opened 2022-08-20
Cleanup some "raw" actor flags

tom-overton:
I was looking at ArmsHook recently, and I noticed that instead of using `ACTOR_FLAG_2000`, they used 0x2000 in some places. This PR tries to clean up cases like this and use `ACTOR_FLAG_X` whenever possible. I may have missed something, but at least it's a start.

### DIFF
--- a/src/overlays/actors/ovl_Arms_Hook/z_arms_hook.c
+++ b/src/overlays/actors/ovl_Arms_Hook/z_arms_hook.c
@@ -164,7 +164,7 @@ void ArmsHook_Shoot(ArmsHook* this, PlayState* play) {
 
         grabbed = this->grabbed;
         if (grabbed != NULL) {
-            if ((grabbed->update == NULL) || (grabbed->flags & ACTOR_FLAG_2000) != ACTOR_FLAG_2000) {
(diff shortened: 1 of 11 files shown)


isghj5 (approved):

src/overlays/actors/ovl_Arms_Hook/z_arms_hook.c
```diff
@@ -164,7 +164,7 @@ void ArmsHook_Shoot(ArmsHook* this, PlayState* play) {
 
         grabbed = this->grabbed;
         if (grabbed != NULL) {
-            if ((grabbed->update == NULL) || (grabbed->flags & 0x2000) != 0x2000) {
+            if ((grabbed->update == NULL) || (grabbed->flags & ACTOR_FLAG_2000) != ACTOR_FLAG_2000) {
```
AngheloAlf, line 167:
```suggestion
            if ((grabbed->update == NULL) || !CHECK_FLAG_ALL(grabbed->flags, ACTOR_FLAG_2000)) {
```

AngheloAlf (approved):

jpburnett (approved):

EllipticEllipsis (approved):

PR #1055, opened 2022-08-26
Document BgInibsMoveBg (Twinmold's Arena) and object_inibs_object

tom-overton:
Since I was documenting Twinmold, I figured I should document the related actors too. This is the simplest one, so I started here. I should note that the names for the collision headers should be taken with a decent grain of salt; I loaded them in-game and used the red dot of the Hookshot to try and map out what they look like.


jpburnett (approved):
🪱 🔥  🪱 🧊

src/overlays/actors/ovl_Bg_Inibs_Movebg/z_bg_inibs_movebg.c
```diff
@@ -41,9 +41,9 @@ void BgInibsMovebg_Init(Actor* thisx, PlayState* play) {
     Actor_ProcessInitChain(&this->dyna.actor, sInitChain);
     DynaPolyActor_Init(&this->dyna, 1);
 
-    this->unk_15C = D_80B96560[BGINIBSMOVEBG_GET_F(thisx)];
-    this->unk_160 = D_80B96568[BGINIBSMOVEBG_GET_F(thisx)];
-    this->unk_164 = D_80B96570[BGINIBSMOVEBG_GET_F(thisx)];
+    this->opaDList = sOpaDLists[BG_INIBS_MOVEBG_GET_SIZE(thisx)];
```
hensldm, line 44: Small nit, but since you call the assets `modes` I think it would make more sense to call the macro `GET_MODE` instead of `GET_SIZE`, though not a big thing.

hensldm (approved):
1 small suggestion

EllipticEllipsis (approved):

AngheloAlf (approved):

PR #1057, opened 2022-08-28
Obj_Kendo_Kanban OK and Documented (Decomp was mostly done already)

StickyThwomp:
This PR addresses matching decompilation of the last remaining function in Kendo Kanban, and documentation/naming of the object.


src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +233,251 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
+        this->rootCornerPos = this->cornerPoints[idxRootCornerPos];
 
         Matrix_Push();
         Matrix_SetTranslateRotateYXZ(this->actor.world.pos.x, this->actor.world.pos.y, this->actor.world.pos.z,
                                      &this->actor.shape.rot);
         Matrix_Scale(this->actor.scale.x, this->actor.scale.y, this->actor.scale.z, MTXMODE_APPLY);
-        Matrix_MultVec3f(&this->unk_2E4, &this->actor.world.pos);
-        this->actor.world.pos = sp5C;
+        Matrix_MultVec3f(&this->rootCornerPos, &this->actor.world.pos);
+        this->actor.world.pos = rootCornerPos;
```
StickyThwomp, line 370: Would like another set of eyes on this. Unless there's a side-effect of the Matrix_* operations I'm not understanding, it would appear that actor.world.pos could simply be set to rootCornerPos and the entire bundle of Matrix_* calls could be eliminated?

include/z64player.h
```diff
@@ -585,7 +585,7 @@ typedef struct Player {
     /* 0xAB2 */ Vec3s unk_AB2;
     /* 0xAB8 */ f32 unk_AB8;
     /* 0xABC */ f32 unk_ABC;
-    /* 0xAC0 */ f32 unk_AC0;
+    /* 0xAC0 */ f32 displacementY;
```
hensldm, line 588: This would probably be better left for `player` decomp/documentation which is currently ongoing.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -1,28 +1,40 @@
 /*
  * File: z_obj_kendo_kanban.c
  * Overlay: ovl_Obj_Kendo_Kanban
- * Description: Swordsman's School - Cuttable Board
+ * Description: Swordsman's School - Cuttable Board at back of school
  */
 
 #include "z_obj_kendo_kanban.h"
 #include "objects/object_dora/object_dora.h"
+#include "z64math.h"
```
hensldm, line 9: This should already be included from "global.h" in the actors header, no need to include it again.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -36,29 +48,32 @@ const ActorInit Obj_Kendo_Kanban_InitVars = {
...
+
+// Centerpoint of the piece
+static Vec3f sCenterPointLeftHalf = { -150.0f, 425.0f, 40.0f };
+static Vec3f sCenterPointRightHalf = { 150.0f, 425.0f, 40.0f };
+static Vec3f sCenterPointBottomHalf = { 0.0f, 140.0f, 40.0f };
+static Vec3f sCenterPointTopHalf = { 0.0f, 565.0f, 40.0f };
+
+// Displaylists for the 4 pieces.
+static Gfx* sDisplayLists[] = { gKendoKanbanTopRightDL, gKendoKanbanTopLeftDL, gKendoKanbanBottomRightDL,
+                                gKendoKanbanBottomLeftDL };
```
hensldm, line 65:
```suggestion
static Gfx* sDisplayLists[] = { gKendoKanbanTopRightDL, gKendoKanbanTopLeftDL, gKendoKanbanBottomRightDL,
                                gKendoKanbanBottomLeftDL, };
```
Should format better with a trailing comma.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -155,13 +170,14 @@ static DamageTable sDamageTable = {
 
 static CollisionCheckInfoInit2 sColChkInfoInit = { 8, 0, 0, 0, MASS_HEAVY };
 
-Vec3f D_80B6681C = { 0.0f, 0.0f, 0.0f };
-Vec3f D_80B66828 = { 1.0f, 0.0f, 0.0f };
+static Vec3f sNullVec3F = { 0.0f, 0.0f, 0.0f };
```
hensldm, line 173:
```suggestion
static Vec3f sZeroVec = { 0.0f, 0.0f, 0.0f };
```
This is usually what we call vectors like this.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -155,13 +170,14 @@ static DamageTable sDamageTable = {
...
+static Vec3f sUnitVectorX = { 1.0f, 0.0f, 0.0f };
```
hensldm, line 174:
```suggestion
static Vec3f sUnitVecX = { 1.0f, 0.0f, 0.0f };
```
Small nit, but I think `Vector` can just be shortened to `Vec`

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +233,251 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
-void func_80B65DA8(ObjKendoKanban* this, PlayState* play) {
-    Vec3f sp5C;
+void ObjKendoKanban_HandlePhysics(ObjKendoKanban* this, PlayState* play) {
+    Vec3f rootCornerPos = this->cornerPos[0];
     s32 pad[2];
-    Vec3f sp48;
-    s32 pad2;
-    s32 index = 0;
+    Vec3f vecCenterOut = this->actor.world.pos;
+    const s16 MAX_BOUNCE_COUNT = 7;
```
hensldm, line 335: Hmm, I feel like this should stay as a pad. Personally I'd probably just use 7 directly as a magic number, though could also make a define at the top. Not sure on other people's thoughts.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +233,251 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
+    s32 idxRootCornerPos = 0;
```
hensldm, line 336:
```suggestion
    s32 indexRootCornerPos = 0;
```
We prefer `index` over `idx` (`idx` can sometimes get confused with `id`)

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +233,251 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
         }
     }
 
-    if (index != this->unk_2FC) {
-        this->unk_300 = 1;
-        this->unk_2FC = index;
-        this->unk_2E4 = this->unk_29C[index];
+    // When the lowest point changes, re-initialize the actor position to that point
+    if (idxRootCornerPos != this->idxLastRootCornerPos) {
+        this->bHasNewRootCornerPos = true;
```
hensldm, line 361: Out of curiosity, what is the `b` for?

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
hensldm: I know at least `Matrix_Scale` with modify the current matrix on the stack (however it also gets popped of later, so I'm not sure). @EllipticEllipsis would probably know more.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +233,251 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
-    if (this->unk_300 == 1) {
-        if (this->unk_308 >= 7) {
-            s16 temp_v0_3 = this->unk_302 & 0x3FFF;
-
-            if (temp_v0_3 >= 0x2000) {
-                temp_v0_3 -= 0x4000;
+    if (this->bHasNewRootCornerPos == true) {
+        if (this->numBounces >= MAX_BOUNCE_COUNT) {
+            s16 deltaRotationAngle = this->rotationAngle & 0x3FFF; // 90 degrees
+            if (deltaRotationAngle >= 0x2000) {                    // 45 degrees
```
hensldm, line 386:
```suggestion
            s16 deltaRotationAngle = this->rotationAngle & 0x3FFF; // 90 degrees
            
            if (deltaRotationAngle >= 0x2000) {                    // 45 degrees
```

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +233,251 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
             }
         }
     }
 }
 
-s32 func_80B6618C(ObjKendoKanban* this, PlayState* play) {
+s32 ObjKendoKanban_IsPlayerOnTop(ObjKendoKanban* this, PlayState* play) {
+    const s32 UNDETERMINED = 0;
+    const s32 UP = 1;
+    const s32 DOWN = -1;
```
hensldm, line 431: This would probably be better as an enum, rather than three distinct constants.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +233,251 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
-    for (i = 0; i < ARRAY_COUNT(this->unk_26C); i++) {
-        if (i != 3) {
-            phi_v0 = i + 1;
+    s32 j;
+    Vec2f playerToCornerA;
+    Vec2f playerToCornerB;
+    s32 priorDir = UNDETERMINED;
+
+    for (i = 0; i < ARRAY_COUNT(this->cornerPos); i++) {
+        j = i != 3 ? (i + 1) : 0;
```
hensldm, line 441:
```suggestion
        j = (i != 3) ? (i + 1) : 0;
```

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.h
```diff
@@ -7,34 +7,28 @@ struct ObjKendoKanban;
...
+    /* 0x2FC */ s32 idxLastRootCornerPos;
+    /* 0x300 */ s16 bHasNewRootCornerPos;
+    /* 0x302 */ s16 rotationAngle;
+    /* 0x304 */ s16 rotationVelocity;
     /* 0x306 */ UNK_TYPE1 pad306[2];
-    /* 0x308 */ s16 unk_308;
-    /* 0x30A */ s16 unk_30A;
-    /* 0x30C */ u8 unk_30C;
+    /* 0x308 */ s16 numBounces;
+    /* 0x30A */ UNK_TYPE2 unk_30A;
```
hensldm, line 30:
```suggestion
    /* 0x30A */ s16 unk_30A;
```
This can stay an `s16` as we know what it is.

hensldm (commented):
Really nice work, especially matching the NON_MATCHING

include/z64player.h
EllipticEllipsis: Agreed, what seems to usually be the case with Player is that something is used for 6 different things in different contexts. I also don't find this a very meaningful name anyway.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
EllipticEllipsis: Agreed.

include/z64player.h
hensldm: I'd say kind of a bit of both. Player should almost certainly be avoided, not because it is ongoing, but player is so big it is difficult to know if that is the actual purpose of the variable. It seems reasonable in this context, but it could be used differently in other contexts that might shed more light. Crossover is fine, as long as their is enough context to figure out that is actually the function of the struct member.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
StickyThwomp: My style consistency is leaking. Some code I interact with makes heavy use of hungarian notation, and that slipped through. It notes that it's used as a bool. "Is"/"has" works perfectly fine for that and is more consistent, so I can drop the "b" no problem.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
EllipticEllipsis: This is very strange-looking. It creates a new scale-rotate-translate that sets up the transformation for the current actor coordinates. It applies it to the lowest corner's position, writes that to world.pos. And then overwrites that with the `rootCornerPos` almost immediately, and drops the matrix. The whole function looks quite redundant already, too.

So yes, afaics this is a waste of time.

EllipticEllipsis (commented):
Couple of additional things.

include/z64player.h
```diff
@@ -256,203 +255,200 @@ typedef struct struct_80122744_arg1 {
...
-// 
-#define PLAYER_STATE1_800000     (1 << 23)
-// 
-#define PLAYER_STATE1_1000000    (1 << 24)
-// 
-#define PLAYER_STATE1_2000000    (1 << 25)
-// 
-#define PLAYER_STATE1_4000000    (1 << 26)
+//
+#define PLAYER_STATE1_1 (1 << 0)
```
EllipticEllipsis, line 259: Please don't format the headers, we prefer to keep them more consistent than the formatter manages.

include/z64player.h
```diff
@@ -602,12 +598,12 @@ typedef struct Player {
...
     /* 0xADF */ s8 unk_ADF[4];
     /* 0xAE3 */ s8 unk_AE3[4];
-    /* 0xAE7 */ s8 unk_AE7; // a timer
+    /* 0xAE7 */ s8 unk_AE7;  // a timer
     /* 0xAE8 */ s16 unk_AE8; // multipurpose timer
     /* 0xAEC */ f32 unk_AEC;
     /* 0xAF0 */ Vec3f unk_AF0[2];
     /* 0xB08 */ f32 unk_B08[8];
-    /* 0xB28 */ s16 unk_B28; //Burning stick timer?
+    /* 0xB28 */ s16 unk_B28; // Burning stick timer?
```
EllipticEllipsis, line 606: Stuff like this should stay, though.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -1,28 +1,48 @@
 /*
  * File: z_obj_kendo_kanban.c
  * Overlay: ovl_Obj_Kendo_Kanban
- * Description: Swordsman's School - Cuttable Board
+ * Description: Swordsman's School - Cuttable Board at back of school
```
EllipticEllipsis, line 4: I don't really think this clarification is necessary, what do others think?

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -1,28 +1,48 @@
...
+
+// Number of bounces the board takes before settling.
+#define MAX_BOUNCE_COUNT (7)
+
 #define THIS ((ObjKendoKanban*)thisx)
 
+typedef enum {
+    /* -1 */ DOWN = -1,
+    /*  0 */ UNDETERMINED,
+    /*  1 */ UP,
```
EllipticEllipsis, line 31: At least put a prefix on it to decrease ambiguity.

include/z64player.h
StickyThwomp: My bad, vscode set to autoformat on save. After undoing the name change, I had no intent for any modifications in this file to remain.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +245,248 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
+        // Initialize the newly spawned left half
+        this->rotationVelocity = 0x71C; // 10 degrees
+        this->actor.velocity = sVelocityLeftHalf;
+        this->centerPoint = sCenterPointLeftHalf;
+
+        this->cornerPoints[0] = sPointTL;
+        this->cornerPoints[1] = sPointTC;
+        this->cornerPoints[2] = sPointBC;
+        this->cornerPoints[3] = sPointBL;
+
```
EllipticEllipsis, line 304: We don't usually put the extra space at the end of the ifs, but it doesn't really matter.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +245,248 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
         this->actor.velocity.x *= 0.8f;
         this->actor.velocity.z *= 0.8f;
     }
 
-    if (this->unk_300 == 1) {
-        if (this->unk_308 >= 7) {
-            s16 temp_v0_3 = this->unk_302 & 0x3FFF;
+    if (this->hasNewRootCornerPos == true) {
+        if (this->numBounces >= MAX_BOUNCE_COUNT) {
+            s16 deltaRotationAngle = this->rotationAngle & 0x3FFF; // 90 degrees
```
EllipticEllipsis, line 397: This is more like % 0x4000 than actually 90 degrees, though. I'm not saying try to match it like that, just that that's what it's actually doing.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -217,238 +245,248 @@ void ObjKendoKanban_Destroy(Actor* thisx, PlayState* play) {
...
-            if (temp_v0_3 >= 0x2000) {
-                temp_v0_3 -= 0x4000;
+            if (deltaRotationAngle >= 0x2000) { // 45 degrees
+                deltaRotationAngle -= 0x4000;   // 90 degrees
             }
-            this->unk_302 -= temp_v0_3;
-            this->unk_304 = 0;
-            func_80B65D54(this);
+            this->rotationAngle -= deltaRotationAngle;
+            this->rotationVelocity = 0;
```
EllipticEllipsis, line 403: Could abbreviate all of these `rotation`s to `rot`

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -457,54 +495,41 @@ void ObjKendoKanban_Update(Actor* thisx, PlayState* play) {
...
     this->actionFunc(this, play);
-
-    func_80B66304(this, play);
+    ObjKendoKanban_UpdateCollision(this, play);
 }
 
-#ifdef NON_MATCHING
 void ObjKendoKanban_Draw(Actor* thisx, PlayState* play) {
     ObjKendoKanban* this = THIS;
+
```
EllipticEllipsis, line 503: Seems to be an extra blank line here.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -457,54 +495,41 @@ void ObjKendoKanban_Update(Actor* thisx, PlayState* play) {
...
-    if (this->unk_30C == OBJKENDOKANBAN_F_0) {
+    if (this->boardFragments == PART_FULL) {
         gSPMatrix(POLY_OPA_DISP++, Matrix_NewMtx(play->state.gfxCtx), G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
         gSPDisplayList(POLY_OPA_DISP++, gKendoKanbanDL);
     } else {
-        Matrix_RotateAxisS(this->unk_302, &this->unk_2F0, MTXMODE_APPLY);
-        Matrix_Translate(-this->unk_2E4.x, -this->unk_2E4.y, -this->unk_2E4.z, MTXMODE_APPLY);
-
+        // Fragments of boards are movable, so apply their rotation/translation, and display only the DLs which
+        // apply.
```
EllipticEllipsis, line 515: This comment isn't really required, this is basically always what these matrix calls do in Draw functions.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
StickyThwomp: I added it since it was a cause of confusion for myself. Alongisde the practice log, the context is probably unnecessary, but if viewed in isolation, at least I confused it with the practice log since I wasn't aware of any other cuttable thing in the swordsman's school.

That said, figured it would have helped me for it to have been there, so it probably will help someone else at some point, but I won't argue for it past this comment. If you (or others) confirm opposition to this change, will undo.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
StickyThwomp: Meant to contextualize the value, not the statement, but I can see how that can be ambiguous. I preferred DEG_TO_BINANG(90) instead of 0x3FFF but saw that was frowned upon in the style documentation, so added the degree comment instead. I'll seek a more clear comment or remove entirely.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -1,28 +1,48 @@
...
+// Number of bounces the board takes before settling.
+#define MAX_BOUNCE_COUNT (7)
+
 #define THIS ((ObjKendoKanban*)thisx)
 
+typedef enum {
+    /* -1 */ DIR_DOWN = -1,
+    /*  0 */ DIR_UNDETERMINED,
+    /*  1 */ DIR_UP,
+} BoardDirection;
```
hensldm, line 32:
```suggestion
typedef enum {
    /* -1 */ OBJKENDOKANBAN_DIR_DOWN = -1,
    /*  0 */ OBJKENDOKANBAN_DIR_UNDETERMINED,
    /*  1 */ OBJKENDOKANBAN_DIR_UP,
} ObjKendoKanbanDirection;
```
Prefixing enums/structs with the actor's name is preferred to help avoid name collisions. Could you also do the same for the `BoardFragmentFlags` up above?

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
```diff
@@ -155,13 +182,14 @@ static DamageTable sDamageTable = {
...
-Vec3f D_80B66828 = { 1.0f, 0.0f, 0.0f };
+static Vec3f sZeroVec = { 0.0f, 0.0f, 0.0f };
+static Vec3f sUnitVecX = { 1.0f, 0.0f, 0.0f };
 
 void ObjKendoKanban_Init(Actor* thisx, PlayState* play) {
     s32 pad[2];
     ObjKendoKanban* this = THIS;
-    Vec3f sp70[3];
+
+    Vec3f vertices[3];
```
hensldm, line 192:
```suggestion
    Vec3f vertices[3];
```

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
hensldm: One thing you could do is remove the matrix calls, and see if the actor behaves differently.

src/overlays/actors/ovl_Obj_Kendo_Kanban/z_obj_kendo_kanban.c
hensldm: Personally I don't see a reason not to have, and if it is beneficial for some we can probably just keep it.

hensldm (approved):

EllipticEllipsis (approved):

include/z64player.h
```diff
@@ -598,12 +602,12 @@ typedef struct Player {
...
     /* 0xADF */ s8 unk_ADF[4];
     /* 0xAE3 */ s8 unk_AE3[4];
-    /* 0xAE7 */ s8 unk_AE7;  // a timer
+    /* 0xAE7 */ s8 unk_AE7; // a timer
     /* 0xAE8 */ s16 unk_AE8; // multipurpose timer
     /* 0xAEC */ f32 unk_AEC;
     /* 0xAF0 */ Vec3f unk_AF0[2];
     /* 0xB08 */ f32 unk_B08[8];
-    /* 0xB28 */ s16 unk_B28; // Burning stick timer?
+    /* 0xB28 */ s16 unk_B28; //Burning stick timer?
```
EllipticEllipsis, line 610:
```suggestion
    /* 0xB28 */ s16 unk_B28; // Burning stick timer?
```

StickyThwomp: Left this one out, as the PR no longer touches z64player.h

PR #1061, opened 2022-09-01
code_801780F0 OK and documented, EnPaper documented

EllipticEllipsis:
The culmination of 11 months of work, on and off. Huge thanks are due to @petrie911 , @engineer124 , and @Synray for their help on this, I feel like I didn't contribute much to the matching at all in the end despite how long I spent on these 3 functions. The last two can probably still be improved, should anyone want to try: the temps are initialised in rather strange places, for one.

This file was absolutely horrible to decompile, match, document, etc., and we all now know way more than we ever wished to about constructing RSP matrices. This is probably the worst file in the entire game, or at least the most dense with badness: after all this nonsense, the last two functions don't even work properly! As such I have not held back on the documentation, there's likely more comments in this file than code now. As ever all comments and suggestions gratefully received.

On the bright side, EnPaper, which uses this, has some mildly interesting physics going on, although the effect it has in-game is completely negligible and only visible in the wedding scene in the credits.

Also a minor improvement in `Actor_UpdateActor`, thanks @petrie911 .


src/code/code_801780F0.c
```diff
@@ -0,0 +1,352 @@
+/**
+ * @file code_801780F0.c
```
hensldm, line 2: Any thoughts on a name for the file?

src/overlays/actors/ovl_Dm_Bal/z_dm_bal.c
```diff
@@ -59,18 +59,18 @@ void DmBal_Init(Actor* thisx, PlayState* play) {
     Actor_UpdateBgCheckInfo(play, &this->actor, 0.0f, 0.0f, 0.0f, 4);
     this->timer = 60;
     this->eyeIndex = 0;
-    this->unk_336 = 0;
-    func_80C1EAC4(this);
+    this->keepEyesShut = 0;
```
hensldm, line 62:
```suggestion
    this->keepEyesShut = false;
```

src/overlays/actors/ovl_Dm_Bal/z_dm_bal.c
```diff
@@ -98,34 +98,34 @@ void func_80C1EAE8(DmBal* this, PlayState* play) {
...
-        this->unk_336 = 0;
+        this->keepEyesShut = false;
         this->eyeIndex = 0;
         D_80C1F2C0 = 0x63;
     }
 }
 
 void func_80C1EC60(DmBal* this, PlayState* play) {
-    f32 temp_fv1_2;
+    f32 temp;
```
hensldm, line 117:
```suggestion
    f32 scaleX;
```
Seems reasonable name, even if it does get squared to be the real scaleX

src/overlays/actors/ovl_En_Paper/z_en_paper.c
```diff
@@ -35,114 +39,155 @@ const ActorInit En_Paper_InitVars = {
     (ActorFunc)EnPaper_Draw,
 };
 
-Vec3f D_80C1FC60 = { 0.0f, 0.0f, 1.0f };
+static Vec3f sUnitVectorZ = { 0.0f, 0.0f, 1.0f };
```
hensldm, line 42:
```suggestion
static Vec3f sUnitVecZ = { 0.0f, 0.0f, 1.0f };
```
Personally I think it is better to shorten it to `Vec`, but I'll leave it up to you.

src/overlays/actors/ovl_En_Paper/z_en_paper.c
```diff
@@ -151,8 +196,8 @@ void EnPaper_Update(Actor* thisx, PlayState* play) {
     EnPaper* this = THIS;
 
     this->actionFunc(this, play);
+    EnPaper_UpdateWind(this);
 
-    func_80C1F87C(this);
     if (this->timer == 0) {
         Actor_MarkForDeath(&this->actor);
     } else {
```
hensldm, line 203:
```suggestion
        Actor_MarkForDeath(&this->actor);
        return;
    }
```
Since it is an `Actor_MarkForDeath`

src/overlays/actors/ovl_En_Paper/z_en_paper.c
```diff
@@ -184,14 +230,16 @@ void EnPaper_Draw(Actor* thisx, PlayState* play) {
...
         }
 
-        if ((this->actor.floorHeight - 40.0f) < ptr->unk_18.y) {
-            MtxF* mf = GRAPH_ALLOC(play->state.gfxCtx, sizeof(MtxF));
+        if ((this->actor.floorHeight - 40.0f) < piece->pos.y) {
+            Mtx* mtx = GRAPH_ALLOC(play->state.gfxCtx, sizeof(Mtx));
 
-            func_8017842C(mf, this->actor.scale.x, this->actor.scale.y, this->actor.scale.z, ptr->unk_30, ptr->unk_0C.x,
-                          ptr->unk_0C.y, ptr->unk_0C.z, ptr->unk_18.x, ptr->unk_18.y, ptr->unk_18.z);
+            //! @bug This function does not necessarily form a correct rotation, see its documentation for details.
```
hensldm, line 236: Personally I don't think the bug tag is warranted. It is enough to have the bug tag on the actual function itself.

src/code/code_801780F0.c
EllipticEllipsis: In theory it lies between `speed_meter` and `sys_cfb`, so it might be something like `speed_mtx` or something. Name should probably mention mtx?

src/overlays/actors/ovl_En_Paper/z_en_paper.c
EllipticEllipsis: Probably fair.

src/code/code_801780F0.c
hensldm: `speed_mtx` seems fine to me.

hensldm (approved):

AngheloAlf (approved):
Good work on the documentation

src/overlays/actors/ovl_Dm_Bal/z_dm_bal.c
```diff
@@ -138,38 +138,36 @@ void func_80C1ED0C(DmBal* this) {
     }
 }
 
-void func_80C1ED64(DmBal* this, PlayState* play, Vec3f* arg2, Vec3f* arg3, f32 arg4) {
-    Actor* paper = Actor_Spawn(&play->actorCtx, play, ACTOR_EN_PAPER, arg2->x, arg2->y, arg2->z, 0, 0, 0, 0);
+// Throw confetti
+void DmBal_SpawnPaper(DmBal* this, PlayState* play, Vec3f* pos, Vec3f* vel, f32 gravity) {
```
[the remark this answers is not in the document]
EllipticEllipsis: I moved the comment instead since the routine is animation-based.

AngheloAlf (approved):

hensldm (approved):

src/code/su_mtx.c
```diff
@@ -0,0 +1,361 @@
...
+ * The final result is
+ *
+ * \f[
+ *  \begin{bmatrix}
+ *      s_x & 0 & 0 & 0 \\
+ *      0 & s_y & 0 & 0 \\
+ *      0 & 0 & s_z & 0 \\
+ *      t_x & t_y & t_z & 1
+ *  \end{bmatrix}
+ * \f]
```
tom-overton, line 66: Perhaps it's just me, but I really don't like LaTeX comments. One of the major reasons why is simply because I don't know LaTeX, and I don't think it should be a requirement to understand what this code is doing. We have an expectation that people coming into the codebase have some degree of understanding with C, which is a totally fair expectation. What I think is a less fair expectation is assuming that people coming into the codebase understand LaTeX syntax, especially when nothing within the code actually requires that you know it. If you want to do anything meaningful with this codebase, you can't escape C, but you can absolutely do meaningful work without being able to read LaTeX.

Furthermore, LaTeX is simply less readable within standard code editors than alternative ways of writing the same information. Unless you are specifically going out of your way to parse and display the LaTeX correctly, something like this is more readable:
```
s_x,   0,   0, 0
  0, s_y,   0, 0
  0,   0, s_z, 0
t_x, t_y, t_z, 1
```
Maybe in the future, there will be generated docs that *do* parse and display the LaTeX correctly, but for now, they seemingly don't exist. And even when they *do* exist, people will still want to be able to read and understand the documentation within a standard code editor.

Lastly, LaTeX syntax breaks things like VSCode hover definitions. This is less of an important point than the other two, but I think it's still relevant. Here's what happens when you hover over `Matrix_Translate` today:
image
And here's what it looks like with my attempt to remove the LaTeX formatting from the comment. VSCode still doesn't put the line breaks in correctly, but you can sort of see the matrix this time:
image
It works correctly if you drop `@brief`, `@param`, and `@remark` from the comment. I'm not sure this is really worth it, but I figured I'd put it out just to show it working:
image

Ultimately, I'd really prefer to leave specialized formatting out of the codebase entirely, especially in cases like this where I'm really not sure what is gained from using it. But LaTeX formatting already exists in comments already in places like sys_matrix (hence my VSCode thing above), so maybe everyone else is fine with it and this is just a "me problem".

src/code/su_mtx.c
```diff
@@ -0,0 +1,361 @@
...
+ * @param[in] scaleX Scale in the X direction
+ * @param[in] scaleY Scale in the Y direction
+ * @param[in] scaleZ Scale in the Z direction
+ * @param[in] translateX X component of translation
+ * @param[in] translateY Y component of translation
+ * @param[in] translateZ Z component of translation
+ *
+ * @remark Original name: "suMtxMakeTS"
+ */
+void Mtx_SetTranslateScaleMtx(Mtx* mtx, f32 scaleX, f32 scaleY, f32 scaleZ, f32 translateX, f32 translateY,
```
Kenix3, line 78: Trying to think of another name for this, but most of the mathematical definitions for translate and scale also include rotate.

src/code/su_mtx.c
```diff
@@ -0,0 +1,361 @@
...
+ * @param[in] axisY Y component of axis to rotate about
+ * @param[in] axisZ Z component of axis to rotate about
+ * @param[in] translateX X component of translation
+ * @param[in] translateY Y component of translation
+ * @param[in] translateZ Z component of translation
+ *
+ * @remark Original name: probably something like "suMtxMakeSRT", although Animal Forest function is a Tait-Bryan
+ * rotation rather than axis-angle.
+ */
+void Mtx_SetTranslationRotationScaleMtx(Mtx* mtx, f32 scaleX, f32 scaleY, f32 scaleZ, s32 angle, f32 axisX, f32 axisY,
```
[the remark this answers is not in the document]
EllipticEllipsis: I don't think it's possible to do that while maintaining consistency with the others (as, indeed, you note in the other comment). I'm also not concerned about it since this function is used once and so badly written that we actively want to discourage people from using it.

PR #1065, opened 2022-09-08
sys_cmpdata OK

mzxrules:
affected by pr #1025 since there is a missing file split


src/code/sys_cmpdma.c
```diff
@@ -19,14 +19,152 @@
...
 
+typedef union {
+    struct {
+        u8 r;
+        u8 g;
+        u8 b;
+        u8 a;
+    };
+    s32 rgba;
+} Struct1_80178AC0;
```
hensldm, line 30: Would Color_RGBA8_u32 work instead?

src/code/sys_cmpdma.c
```diff
@@ -19,14 +19,152 @@
...
+
+typedef union {
+    u16 a, b;
+    s32 ab;
+} Struct2_80178AC0;
+
+typedef struct {
+    /* 0x00 */ u32 data1;
+    /* 0x00 */ u32 data2;
```
hensldm, line 39:
```suggestion
    /* 0x00 */ u32 start;
    /* 0x04 */ u32 end;
```
Based on usage, I think `start` and `end` would make sense.

src/code/sys_cmpdma.c
```diff
@@ -19,14 +19,152 @@
...
+
+        *(phi_v0++) = spC.rgba;
+    } while ((u32)phi_a0 < (u32)dst);
+}
+#else
+void func_80178AC0(u16* src, void* dst, u32 size);
 #pragma GLOBAL_ASM("asm/non_matchings/code/sys_cmpdma/func_80178AC0.s")
+#endif
+
+void func_80178C80(u8* vrom, s32 id, s32* start, s32* size, s32* flag) {
```
hensldm, line 100:
```suggestion
void func_80178C80(u8* rom, s32 id, s32* start, s32* size, s32* flag) {
```
Looks like it is the first arg to `DmaRomToRam` so this would be rom.

src/code/sys_cmpdma.c
```diff
@@ -19,14 +19,152 @@
...
+
+void func_80178DAC(u32 vrom, s32 id, void* dst, s32 size) {
+    s32 romStart;
+    s32 compressedSize;
+    s32 flag;
+
+    func_80178C80(vrom, id, &romStart, &compressedSize, &flag);
+    if (flag & 1) {
+        void* temp = SystemArena_Malloc(0x1000);
+        func_80178D7C(romStart, compressedSize, temp);
```
hensldm, line 141:
```suggestion
        void* temp = SystemArena_Malloc(0x1000);
        
        func_80178D7C(romStart, compressedSize, temp);
```

src/code/sys_cmpdma.c
```diff
@@ -19,14 +19,152 @@
...
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cmpdma/func_80178C80.s")
+void func_80178E3C(uintptr_t vrom, s32 id, void* dst, s32 size) {
+    func_80178DAC(DmaMgr_TranslateVromToRom(vrom), id, dst, size);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cmpdma/func_80178D7C.s")
+void func_80178E7C(uintptr_t vrom, void* dst, s32 arg2) {
+    u32 temp_s2 = DmaMgr_TranslateVromToRom(vrom);
```
hensldm, line 154:
```suggestion
    uintptr_t rom = DmaMgr_TranslateVromToRom(vrom);
```

src/code/sys_cmpdma.c
```diff
@@ -19,14 +19,152 @@
...
+    u32 i;
+    u32 end;
+    s32 phi_s1;
+    u32 var;
```
hensldm, line 158:
```suggestion
    u32 dataStart;
```

hensldm (commented):
I'd also say to probably just implement the file split here, especially since the PR that actually implements it is still drafted.

src/code/sys_cmpdma.c
```diff
@@ -19,14 +19,152 @@
...
+typedef struct {
+    /* 0x00 */ u32 data1;
+    /* 0x00 */ u32 data2;
+} CmpDmaBssStruct;
+
+// static CmpDmaBssStruct D_801FBBE0;
+extern CmpDmaBssStruct D_801FBBE0;
+
+#ifdef NON_MATCHING
+// Some constants re-ordered at the top
```
hensldm, line 46: Forgot in my other review, but would you be able to post a scratch for this? Unless you still want to work on it.

src/code/sys_cmpdma.c
mzxrules: Can't say I agree.

Starting with`func_80178C80`, the function begins by reading in and storing into data1 the first word of a yaz archive, which stores the offset to the start of the data segment, relative to the file. Then in the refOff == 0 case, you read in `rom + 4` into data1, which stores the offset to the start of the second file relative to the _data segment_, which effectively is equivalent to the end of the first file/size of the first file. Only the last case do you have data1/data2 storing the start/end relative to the data segment. 

In func_80178E7C, I don't see how you could say that data1 is storing the start of anything; in the code it's being used to calculate how many times the loop should iterate for.

src/code/sys_cmpdma.c
mzxrules: mentioned in the other comment, but I don't see the logic for making this dataStart

src/code/sys_cmpdma.c
hensldm: Ah so it seems this just acts as a buffer to grab some kind of data, so yeah `start` and `end` don't make sense since the data is different. Not saying it would be better, but does just an array of 2 `u32`s work as well instead of defining a new struct?

src/code/sys_cmpdma.c
mzxrules: https://decomp.me/scratch/OSthZ

src/code/sys_cmpdma.c
mzxrules: never mind, this was actually correct; it's calculating the number of files in the segment

src/code/sys_cmpdma.c
mzxrules: I've reworked this into a union

src/code/sys_cmpdma.c
```diff
@@ -1,32 +1,117 @@
...
+    union {
+        u32 dmaWord[2];
+        u32 dataStart;
+        u32 dataSize;
+        struct {
+            u32 start;
+            u32 end;
+        } offset;
+    };
+} CmpDmaBuffer;
```
hensldm, line 14: To be honest I'm not sure how I feel about the union. While it does make things clear in the code below, it just adds a lot of complexity that I'm not sure is worth it. I think I still prefer an array of length 2 with comments explaining what the buffer is being used for. However, I am willing to keep the union.

src/code/sys_cmpdma.c
```diff
@@ -1,32 +1,117 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cmpdma/func_80178978.s")
+void func_80178AC0(u16* src, void* dst, size_t size) {
+    Color_RGBA8_u32 spC;
+    Color_RGBA16_2 tc;
+    Color_RGBA14 tc2;
+    u32* dstCur = dst;
+    u16* src16 = src;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cmpdma/func_801789D4.s")
+    while (((u32)dstCur) - size < ((u32)dst)) {
```
hensldm, line 25:
```suggestion
    while (((uintptr_t)dstCur) - size < ((uintptr_t)dst)) {
```

hensldm (approved):

src/code/sys_cmpdma.c
mzxrules: eh, I think it's fine as the complexity is well contained within the file.

include/color.h
```diff
@@ -51,4 +51,24 @@ typedef union {
...
 
+typedef union {
+    struct {
+        u32 r : 5;
+        u32 g : 5;
+        u32 b : 5;
+        u32 a : 1;
+    };
+    u16 rgba;
+} Color_RGBA16_2;
```
AngheloAlf, line 62: What's the difference between this and `Color_RGBA16`? Can't we merge those two?

src/code/sys_cmpdma.c
```diff
@@ -1,32 +1,117 @@
...
+        u32 dataSize;
+        struct {
+            u32 start;
+            u32 end;
+        } offset;
+    };
+} CmpDmaBuffer;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cmpdma/func_80178818.s")
+static CmpDmaBuffer sDmaBuffer;
```
AngheloAlf, line 16: Could you remove the `static` on this bss variable so it appears on the map file?
```suggestion
CmpDmaBuffer sDmaBuffer;
```

include/color.h
mzxrules: Because division is involved, the r,g,b needs to either be typed as or cast to u32 in sys_cmpdma to generate the correct codegen in func_80178AC0, otherwise there will be extra checks introduced. 

Color_RGBA16 is only used in PreRender_AntiAliasAlgorithm, and changing the type from u16 -> u32 mostly works except that the pxIn right change to logical instead of arithmetic, and require an s32 cast to fix this.

So yes they can be combined, but we would need to introduce a cast either on the sys_cmpdma side or PreRender_AntiAliasAlgorithm side.

AngheloAlf (approved):

include/color.h
Kenix3: Is the one file/function that this type is used in used anywhere outside of it's own file? Might make sense to put this struct in the C file where it's used, unless the function is public.

src/code/sys_cmpdma.c
Kenix3: Personally I'm fine with the union.

Kenix3 (approved):

PR #1074, opened 2022-09-13
Document object_market_obj (West Clock Town Market)

tom-overton:
It's still sort of strange to me that this is an object in the first place, but I documented the whole thing.

I called the walls "gold", "silver", and "bronze" because they vaguely looked like that color, plus it's a nice cohesive theme.
image
image
image

I called this the "floor with edge" to distinguish it from the regular floor near the South Clock Town entrance:
west_clock_town_market_floor_with_edge rgba16
It gets used both for the stairs and for the edge of this slope:
GLideN64_ZELDA_MAJORA'S_MASK_008

Lastly, the little bobble head cow's body has two textures: one for most of the body, and one for the part of the cow where the bobble head is attached. I just called these "body" and "front", respectively.


assets/xml/objects/object_market_obj.xml
```diff
@@ -1,65 +1,69 @@
...
+        <Texture Name="gWestClockTownMarketRugTex" OutName="west_clock_town_market_rug" Format="rgba16" Width="64" Height="16" Offset="0x135E0" />
+        <Texture Name="gWestClockTownMarketBankCurtainTex" OutName="west_clock_town_market_bank_curtain" Format="ia8" Width="32" Height="32" Offset="0x13DE0" />
+        <Texture Name="gWestClockTownMarketCowFrontTex" OutName="west_clock_town_market_cow_front" Format="ci8" Width="8" Height="8" Offset="0x141E0" />
+        <Texture Name="gWestClockTownMarketCowBodyTex" OutName="west_clock_town_market_cow_body" Format="i8" Width="8" Height="8" Offset="0x14220" />
+        <Texture Name="gWestClockTownMarketBankAdvertisementsTex" OutName="west_clock_town_market_bank_advertisements" Format="i4" Width="64" Height="32" Offset="0x14260" />
+        <Texture Name="gWestClockTownMarketSafeWheelAndHandleTex" OutName="west_clock_town_market_safe_wheel_and_handle" Format="rgba16" Width="16" Height="16" Offset="0x14660" />
+
+        <!-- West Clock Town Market DisplayLists -->
+        <DList Name="gWestClockTownMarketBankAdvertisementsAndFakeDoorNightDL" Offset="0x18C60" />
```
hensldm, line 64: More so just out of curiosity, what do you mean by fake door?

hensldm (approved):

assets/xml/objects/object_market_obj.xml
AngheloAlf: I have the same question

AngheloAlf (approved):

EllipticEllipsis:
I'd say something like `brick-edged ground` or something. I don't think it's really `floor` since it's outside.

assets/xml/objects/object_market_obj.xml
tom-overton: Behind the bank vaults, there's this fake door:
image

I call it "fake" because it's not an actual door; it's just a picture of one. You can't interact with it in any way, and in fact you can't even see it without going out of bounds. I didn't want people mistaking this for a texture for a real, working door, hence the name. However, I'm happy to change it, since I kind of don't care very much about the name of a texture you can't even see under normal circumstances.

EllipticEllipsis: Oh, it's off to the side in the banker's kiosk?

tom-overton: Correct. You can't see it because the vaults are in front of it. But it's just off to the side behind the banker.

tom-overton:
Sure, and I also switched `Floor` with `Ground`

assets/xml/objects/object_market_obj.xml
EllipticEllipsis: I think I specifically don't want `fake door` because that sounds like DoorKiller in OoT. Possibly just door would be fine, with a note that it's hidden in-game.

EllipticEllipsis (commented):
I'm not completely sold on the gold/silver/bronze thing, but I can't think of anything that's definitely better for now, I'd consider yellow/grey/something, and maybe call it plaster? (Is it plaster? Seems like it from the brick poking out the bottom in some places.)

PR #1077, opened 2022-09-16
Document object_dblue_object (Great Bay Temple objects)

tom-overton:
This one was a really tough one. Not only does it handle a lot of different components, but it also has a bunch of unused content in it too, including 33(!) TLUTs that aren't used for anything at all.

One extremely minor naming thing I'd like to call out; in MM3D, the unused collision is named `m2_seauzuuzu2_bgdatainfo`, all in lowercase. However, all the other components have the three letters after `m2` written in ALL CAPS, so I wrote it as such in the XML too. This is further supported by "uzuuzu" being a word, but "seauzuuzu" *not* being a word.


hensldm (approved):

AngheloAlf (approved):

src/overlays/actors/ovl_Bg_Dblue_Balance/z_bg_dblue_balance.c
```diff
@@ -54,35 +54,35 @@ typedef struct {
 
 BgDblueBalanceStruct2 D_80B83A20[] = {
```
EllipticEllipsis, line 55: Could you name this one too?

EllipticEllipsis (approved):

PR #1083, opened 2022-09-19
Document object_az (Beaver Bros.)

tom-overton:
Pretty straightforward, though figuring out the vortex stuff was kind of annoying.


Kenix3 (approved):

hensldm (approved):

src/overlays/actors/ovl_En_Az/z_en_az.c
```diff
@@ -1680,26 +1756,26 @@ static struct_80A98F94 D_80A9930C[5][4] = {
...
 };
-static TexturePtr D_80A99404[3] = {
-    object_az_Tex_016018,
-    object_az_Tex_016818,
-    object_az_Tex_017018,
+static TexturePtr D_80A99404[] = {
+    gBeaverYoungerBrotherBelt1Tex,
+    gBeaverYoungerBrotherBelt2Tex,
+    gBeaverYoungerBrotherBelt3Tex,
 };
```
EllipticEllipsis, line 1779: Can name these arrays, then. Is there no way to distinguish these textures?

tom-overton: For the eyes, almost certainly not. It's just a repeating sequence:
beaver_younger_brother_eye_1 rgba16
beaver_younger_brother_eye_2 rgba16
beaver_younger_brother_eye_3 rgba16
beaver_younger_brother_eye_4 rgba16

For the belt, maybe you could name them based on the colors? But it's an infinitely repeating sequence just like the eyes:
beaver_younger_brother_belt_1 rgba16
beaver_younger_brother_belt_2 rgba16
beaver_younger_brother_belt_3 rgba16

EllipticEllipsis: Oh right, it's specifically the propeller thing. Hmm, yes, I'd say red,green,blue for those in that order. Nothing we can do about the eyes, though, except start at 0.

tom-overton: I can change the propeller thing, but I don't think it's common for us to zero-index texture names like that.

EllipticEllipsis: I think some of us discussed it before and it was considered a good idea, but there's plenty of effects that will be done later anyway so I'm not going to worry about it for now.

PR #1084, opened 2022-09-19
Give Deku Princess/King/Butler's objects the "correct" name

tom-overton:
I was trying to document the Deku Princess's object, which is currently named `object_dnp`, but I noticed that I couldn't find any file with the name `dnp` in MM3D. Instead, I found her assets in a file named `zelda2_dnq.gar.lzs`. You can also see that all of her animations are prefixed with `dnq`:
image

This was pretty confusing, because we use `object_dnq` for the Deku *King's* object. So I poked around in MM3D and found his assets in `zelda2_dno.gar.lzs`. Similar to what happened with the princess, his animations are also prefixed with `dno`:
image

This only added to the confusion, because we use `object_dno` for the Deku *Butler's* object! I eventually found his assets in `zelda2_dnj.gar.lzs`:
image

Given the animation name prefixes, I'm almost certain that the MM3D names for these objects were their original name. Thus, I went and changed the names in the repo to match. This *does* introduce a mismatch between the names of the actors and the objects they use, but I'm 99% confident that mismatch was there to begin with.

Here's a table of the names, in case you got confused by any part of the explanation above:

| Character | Actor Name | Current Object Name | Proposed Object Name | MM3D Name|
|-----------|--------------|-----------------------|-----------------------|-----------------|
| Princess   | `ovl_En_Dnp`  | `object_dnp`            | `object_dnq`          | `zelda2_dnq.gar.lzs`|
| King         | `ovl_En_Dnq`  | `object_dnq`            | `object_dno`          | `zelda2_dno.gar.lzs`|
| Butler       | `ovl_En_Dno`  | `object_dno`            | `object_dnj`          | `zelda2_dnj.gar.lzs`|


hensldm (approved):
As annoying as it is, this seems to make the most sense with the evidence at hand and I agree with this change.

AngheloAlf (approved):
Annoying, but it is better this way

EllipticEllipsis (commented):
I'd recommend putting notes in these three that explain that the names are correct but confusing, otherwise I think this is for the best, it's the same as when we have to write counterintuitive code to match stuff, it's the nature of trying to get the project accurate.

tom-overton:
I've been thinking about it more, and I think it's the *actors* that have the confusing names, not the objects. `dnq` is short for `dekunuts queen`, as shown in the screenshot. `dno` is probably short for `dekunuts 王 (king)`. As for `dnj`, I can't quite figure out what this one is short for, but it doesn't change my next point.

For the actors, `ovl_En_Dnp` is probably short for `dekunuts princess`, which still makes sense. But then the Deku King is `ovl_En_Dnq`, which as we established before means `dekunuts queen`. This doesn't make much sense. And the Deku Butler is `ovl_En_Dno`, which *probably* doesn't make sense, though maybe there's a Japanese word that starts with "o" that would fit here.

I'm happy to put notes explaining the discrepancy *somewhere* in the repo, but in my opinion, its the actors that are misnamed. Should I put those comments within the actors?

EllipticEllipsis:
Makes sense to me.

tom-overton:
Done, thanks to @emilybrooks for helping me with my Japanese questions.

EllipticEllipsis (approved):
Nice, like the detail in the comments.

PR #1098, opened 2022-09-30
Object Tru Docs (Koume)

hensldm:


assets/xml/objects/object_tru.xml
```diff
@@ -1,108 +1,117 @@
...
+        <DList Name="gKoumeLeftSleeveEndDL" Offset="0xDB50" />
+        <DList Name="gKoumeRightThighDL" Offset="0xDC40" /> <!-- Renders a single triangle instead of a normal-looking limb -->
+        <DList Name="gKoumeRightShinDL" Offset="0xDC88" />
+        <DList Name="gKoumeRightFootDL" Offset="0xDD38" />
+        <DList Name="gKoumeLeftThighDL" Offset="0xDDD0" /> <!-- Renders a single triangle instead of a normal-looking limb -->
+        <DList Name="gKoumeLeftShinDL" Offset="0xDE18" />
+        <DList Name="gKoumeLeftFootDL" Offset="0xDEC8" />
+
+        <Animation Name="gKoumeInjuredTalkingAnim" Offset="0xEEDC" />
+        <Animation Name="gKoumeInjuredLyingDownAnim" Offset="0xF9A0" />
```
tom-overton, line 49: Slightly different verb tense than the other animations. Maybe `gKoumeInjuredTalkAnim` and `gKoumeInjuredLieDownAnim`?

src/overlays/actors/ovl_En_Tru/z_en_tru.c
```diff
@@ -24,6 +24,26 @@ s32 func_80A87DC0(Actor* thisx, PlayState* play);
 void func_80A87FD0(EnTru* this, PlayState* play);
 void func_80A881E0(EnTru* this, PlayState* play);
 
+typedef enum {
+    /* 0x00 */ KOUME_ANIM_INJURED_LYING_DOWN1,
+    /* 0x01 */ KOUME_ANIM_INJURED_LYING_DOWN2,
```
tom-overton, line 29: Would want to change these two if you change the animation name. Also, can you tell why one of these would be used instead of the other?

src/overlays/actors/ovl_En_Tru/z_en_tru.c
```diff
@@ -24,6 +24,26 @@ s32 func_80A87DC0(Actor* thisx, PlayState* play);
...
+    /* 0x02 */ KOUME_ANIM_TRY_GET_UP,
+    /* 0x03 */ KOUME_ANIM_INJURED_RAISE_HEAD,
+    /* 0x04 */ KOUME_ANIM_INJURED_TALKING,
```
tom-overton, line 32: Same thing here; you'd want to change this one if you change the animation name.

src/overlays/actors/ovl_En_Tru/z_en_tru.c
```diff
@@ -24,6 +24,26 @@ s32 func_80A87DC0(Actor* thisx, PlayState* play);
...
+    /* 0x05 */ KOUME_ANIM_INJURED_HEAD_UP1,
+    /* 0x06 */ KOUME_ANIM_INJURED_HEAD_UP2,
```
tom-overton, line 34: Similar to above; is there any difference between these two?

src/overlays/actors/ovl_En_Tru_Mt/z_en_tru_mt.c
```diff
@@ -20,6 +20,26 @@ void func_80B76A64(EnTruMt* this, PlayState* play);
...
+    /* 0x08 */ KOUME_ANIM_SHAKE,
+    /* 0x09 */ KOUME_ANIM_DRINK,
+    /* 0x0A */ KOUME_ANIM_FINISHED_DRINKING,
+    /* 0x0B */ KOUME_ANIM_HEALED,
+    /* 0x0C */ KOUME_ANIM_HOVER1,
+    /* 0x0D */ KOUME_ANIM_TAKE_OFF,
+    /* 0x0E */ KOUME_ANIM_FLY,
+    /* 0x0F */ KOUME_ANIM_HOVER2,
+    /* 0x10 */ KOUME_ANIM_MAX
+} KoumeAnimation;
```
tom-overton, line 41: Are all of these used? If not, can you comment which ones aren't?

assets/xml/objects/object_tru.xml
hensldm: I agree with the `talk` but I'd prefer `InjuredLyingDown`. `InjuredLieDown` seems more like the animation is of her lying down (i.e. the process to lie down), rather than its the animation when she is lying down injured.

src/overlays/actors/ovl_En_Tru_Mt/z_en_tru_mt.c
hensldm: Between the two actors, there is one unused anim. I marked that one for both.

src/overlays/actors/ovl_En_Tru/z_en_tru.c
hensldm: The difference between these two would be the morph frames (renamed them so the ones with morph frames end with the word morph), which AFAIK is for a smoother transition between animations, but I'm not 100% sure.

src/overlays/actors/ovl_En_Tru/z_en_tru.c
hensldm: See https://github.com/zeldaret/mm/pull/1098#discussion_r985168795

assets/xml/objects/object_tru.xml
tom-overton: Yeah, that's fine. I think I did something like that in `object_ma1` too for that exact reason.

src/overlays/actors/ovl_En_Jc_Mato/z_en_jc_mato.c
```diff
@@ -145,15 +145,15 @@ void EnJcMato_Update(Actor* thisx, PlayState* play) {
     }
 }
 
-static Vec3f movement = { 0.0f, -2500.0f, 0.0f };
+static Vec3f sOffset = { 0.0f, -2500.0f, 0.0f };
 
 void EnJcMato_Draw(Actor* thisx, PlayState* play) {
```
EllipticEllipsis, line 150: Might as well go in the function.

src/overlays/actors/ovl_En_Tru_Mt/z_en_tru_mt.h
```diff
@@ -16,15 +16,15 @@ typedef struct EnTruMt {
...
+    /* 0x286 */ Vec3s morphTable[KOUME_LIMB_MAX];
     /* 0x328 */ u16 unk_328;
     /* 0x32A */ UNK_TYPE1 unk32A[0xC];
     /* 0x336 */ Vec3s unk_336;
     /* 0x33C */ Vec3f unk_33C;
     /* 0x348 */ s16 unk_348;
     /* 0x34A */ s16 unk_34A;
-    /* 0x34C */ s16 unk_34C;
+    /* 0x34C */ s16 eyeTexIndex;
     /* 0x34E */ s16 unk_34E;
```
EllipticEllipsis, line 28: And this is the blinkTimer.

src/overlays/actors/ovl_En_Tru/z_en_tru.h
```diff
@@ -45,12 +46,12 @@ typedef struct EnTru {
     /* 0x368 */ s16 unk_368;
     /* 0x36A */ s16 unk_36A;
     /* 0x36C */ s16 unk_36C;
```
EllipticEllipsis, line 48: And this is the blinkTimer for this one.

src/overlays/actors/ovl_En_Tru/z_en_tru.c
```diff
@@ -24,6 +24,26 @@ s32 func_80A87DC0(Actor* thisx, PlayState* play);
...
+    /* 0x08 */ KOUME_ANIM_SHAKE, // Unused
+    /* 0x09 */ KOUME_ANIM_DRINK,
+    /* 0x0A */ KOUME_ANIM_FINISHED_DRINKING,
+    /* 0x0B */ KOUME_ANIM_HEALED,
+    /* 0x0C */ KOUME_ANIM_HOVER1,
+    /* 0x0D */ KOUME_ANIM_TAKE_OFF,
+    /* 0x0E */ KOUME_ANIM_FLY,
+    /* 0x0F */ KOUME_ANIM_HOVER2,
+    /* 0x10 */ KOUME_ANIM_MAX
+} KoumeAnimation;
```
EllipticEllipsis, line 45: Since this seems to be the main actor of the two, this should probably go in the header and that header be included in both. There's obviously not much else shared since even the eyeIndex and blinkTimer are in different orders.

hensldm: Since the full animation array is duplicated, I think it makes more sense for the enum to be duplicated as well. (Though I should have changed the name/prefix to add an MT which I think I just forgot)

AngheloAlf: Agreed. It would be better if we had a way to move both the enum and the array to a common place, but since we don't have any way to do that then this is fine.

src/overlays/actors/ovl_En_Tru/z_en_tru.c
```diff
@@ -307,8 +326,8 @@ void func_80A85F84(EnTruUnkStruct* arg0, PlayState* play) {
 
             gSPMatrix(POLY_XLU_DISP++, Matrix_NewMtx(play->state.gfxCtx), G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
             idx = ((f32)arg0->unk_02 / arg0->unk_01) * 8.0f;
-            gSPSegment(POLY_XLU_DISP++, 0x08, Lib_SegmentedToVirtual(sDustTextures[idx]));
-            gSPDisplayList(POLY_XLU_DISP++, object_tru_DL_01A830);
+            gSPSegment(POLY_XLU_DISP++, 0x08, Lib_SegmentedToVirtual(sDustTextures[(idx)]));
```
AngheloAlf, line 329: why the extra `()`?

AngheloAlf (approved):

src/overlays/actors/ovl_En_Tru/z_en_tru.c
hensldm: Whoops. I was messing with the function, and must have forgot to remove the () again.

EllipticEllipsis (approved):

tom-overton (approved):

PR #1104, opened 2022-10-02
Document leftover pause joint tables

tom-overton:
In OoT, there exist multiple joint tables used for the pause menu; they're located right above the player animations within `gameplay_keep`. I noticed that in MM, we had an unknown blob above the player animations, and some it suspiciously looked like joint table data.

What was strange here is that there wasn't enough room for all five joint tables from OoT. I did some more digging and I found out that, despite what the Deku Stick DL is asking for, the Deku Stick texture is *actually* an 8x8 texture. Defining it as such makes it so that all give OoT pause menu joint tables can be extracted. This means that the Deku Stick, in-game, will render some garbage based on the Child Link joint tables. I wasn't sure how to document that, though; I would slap a `@bug` above it normally, but the DList gets extracted.

### DIFF
--- a/assets/xml/objects/gameplay_keep.xml
+++ b/assets/xml/objects/gameplay_keep.xml
@@ -39,7 +39,7 @@
         <Texture Name="gameplay_keep_Tex_002E08" OutName="tex_002E08" Format="rgba16" Width="16" Height="16" Offset="0x2E08" />
         <Texture Name="gameplay_keep_Tex_003008" OutName="tex_003008" Format="i8" Width="8" Height="8" Offset="0x3008" />
         <Blob Name="gameplay_keep_Blob_003050" Size="0x20" Offset="0x3050" />
-        <DList Name="gDekuStickDL" Offset="0x32B0" /> <!-- Supplies the wrong size to gsDPLoadTextureBlock, meaning it uses part of gLinkPauseChildJointTable as if it were a texture. -->
+        <DList Name="gDekuStickDL" Offset="0x32B0" /> <!-- @bug Supplies the wrong size to gsDPLoadTextureBlock, meaning it uses part of gLinkPauseChildJointTable as if it were a texture. -->
         <Blob Name="gameplay_keep_Blob_003400" Size="0x9D0" Offset="0x3400" />
         <DList Name="gameplay_keep_DL_005360" Offset="0x5360" />
         <DList Name="gameplay_keep_DL_0056C0" Offset="0x56C0" />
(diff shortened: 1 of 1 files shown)


assets/xml/objects/gameplay_keep.xml
```diff
@@ -69,8 +69,26 @@
         <Texture Name="gameplay_keep_Tex_00C7B0" OutName="tex_00C7B0" Format="rgba16" Width="8" Height="8" Offset="0xC7B0" />
         <Texture Name="gameplay_keep_Tex_00C830" OutName="tex_00C830" Format="rgba16" Width="16" Height="16" Offset="0xC830" />
         <Texture Name="gameplay_keep_Tex_00CA30" OutName="tex_00CA30" Format="rgba16" Width="16" Height="16" Offset="0xCA30" />
-        <Texture Name="gDekuStickTex" OutName="tex_00CC30" Format="i8" Width="16" Height="16" Offset="0xCC30" />
-        <Blob Name="gameplay_keep_Blob_00CD30" Size="0x210" Offset="0xCD30" />
+        <Texture Name="gDekuStickTex" OutName="deku_stick" Format="i8" Width="8" Height="8" Offset="0xCC30" />
```
EllipticEllipsis, line 72: Think you'd better add a note to the DL too explaining the problem (and, I would suggest, how to fix it, although some may regard that as out of scope).

AngheloAlf (approved):

assets/xml/objects/gameplay_keep.xml
```diff
@@ -39,7 +39,7 @@
         <Texture Name="gameplay_keep_Tex_002E08" OutName="tex_002E08" Format="rgba16" Width="16" Height="16" Offset="0x2E08" />
         <Texture Name="gameplay_keep_Tex_003008" OutName="tex_003008" Format="i8" Width="8" Height="8" Offset="0x3008" />
         <Blob Name="gameplay_keep_Blob_003050" Size="0x20" Offset="0x3050" />
-        <DList Name="gDekuStickDL" Offset="0x32B0" />
+        <DList Name="gDekuStickDL" Offset="0x32B0" /> <!-- Supplies the wrong size to gsDPLoadTextureBlock, meaning it uses part of gLinkPauseChildJointTable as if it were a texture. -->
```
hensldm, line 42: I'd probably add a `@bug` tag.

hensldm (approved):

assets/xml/objects/gameplay_keep.xml
EllipticEllipsis: Agreed, I'll approve and merge when that's done.

EllipticEllipsis (approved):

PR #1109, opened 2022-10-06
Play (2 NON_MATCHINGS)

hensldm:
Wanted to PR this sooner rather than later to start getting others eyes on this.

Play_UpdateMain: https://decomp.me/scratch/aBH98
Play_DrawMain: https://decomp.me/scratch/YQMUX


include/regs.h
```diff
@@ -105,4 +110,11 @@
 #define R_FB_FILTER_A                     SREG(84)
 #define R_FB_FILTER_ENV_COLOR(c)          SREG(85 + c)
 
+
+#define R_PICTOGRAPH_PHOTO_STATE          SREG(89)
+#define R_MOTION_BLUR_ALPHA               SREG(90)
+#define R_MOTION_BLUR_ENABLED             SREG(91)
+#define R_MOTION_BLUR_PRIORITY_ALPHA      SREG(92)
+#define R_MOTION_BLUR_PRIORITY_ENABLED    SREG(93)
```
engineer124, line 118: It looks like `SREG` should be at the top based on ordering

src/code/z_message.c
```diff
@@ -329,7 +329,7 @@ void func_80151938(PlayState* play, u16 textId) {
     }
     msgCtx->unk1203C = msgCtx->unk1203A;
 
-    if (play->pauseCtx.unk_1F0 != 0) {
+    if (play->pauseCtx.bombersNotebookOpen != 0) {
```
engineer124, line 332:
```suggestion
    if (play->pauseCtx.bombersNotebookOpen) {
```

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+    ShrinkWindow_Destroy();
+    TransitionFade_Destroy(&this->unk_18E48);
+    VisMono_Destroy(&sVisMono);
+    func_80140EA0(D_801F6D4C);
+    D_801F6D4C = NULL;
+
+    if (gSaveContext.save.weekEventReg[0x5C] & 0x80) {
+        Actor_CleanupContext(&this->actorCtx, this);
+    }
+    gSaveContext.save.weekEventReg[0x5C] &= (u8)~0x80;
```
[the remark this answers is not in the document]
AngheloAlf: This is a very curious flag, since the only place it is set seems to be on `Actor_InitContext`, which I guess it makes sense considering what it does, but it doesn't make sense it is a weekEventReg

src/code/z_play_hireso.c
```diff
@@ -16,12 +16,12 @@
...
 #pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/func_8016F4EC.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/func_8016F5A8.s")
+#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/Hireso_Update.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/func_8016FC78.s")
+#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/Hireso_Init.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/func_8016FC98.s")
+#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/Hireso_Destroy.s")
```
engineer124, line 27: Since you're working on Hireso, I wonder if we could consider using `Notebook_` prefixes for functions? So:
`Notebook_Init`
`Notebook_Update`
...
Maybe something to consider

src/overlays/actors/ovl_Bg_Crace_Movebg/z_bg_crace_movebg.c
```diff
@@ -328,8 +328,8 @@ void BgCraceMovebg_ClosingDoor_Close(BgCraceMovebg* this, PlayState* play) {
     if (Math_StepToF(&this->doorHeight, 0.0f, 1.0f)) {
         if (!(this->stateFlags & BG_CRACE_MOVEBG_FLAG_PLAYER_IS_BEYOND_DOOR) &&
             !Flags_GetSwitch(play, BG_CRACE_MOVEBG_GET_SWITCH_FLAG(&this->dyna.actor) + 1)) {
-            play->unk_18845 = 1;
-            func_80169FDC(&play->state);
+            play->haltAllActors = 1;
```
engineer124, line 331:
```suggestion
            play->haltAllActors = true;
```

src/overlays/actors/ovl_En_Warp_tag/z_en_warp_tag.c
```diff
@@ -209,13 +209,13 @@ void EnWarpTag_RespawnPlayer(EnWarptag* this, PlayState* play) {
...
-                    play->setupEntranceList[playerSpawnIndex].room, playerParams, &newRespawnPos,
+                    &play->state, 0, entrance, play->setupEntranceList[playerSpawnIndex].room, playerParams,
+                    &newRespawnPos,
                     ((((playerActorEntry->rot.y >> 7) & 0x1FF) / 180.0f) * 32768.0f)); // DEG_TO_BINANG ?
 
-                func_80169EFC(&play->state);
+                Play_TriggerVoidOut(&play->state);
                 gSaveContext.respawnFlag = ~0x4;
-                func_80165690();
+                Play_DisableMotionBlur();
```
engineer124, line 218: There's probably quite a few functions that can be added to namefixer.py. Might even be worth including the `TRANS_TYPE_` and `TRANS_MODE_` enums in there

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+    s32 i;
+    s32 j;
+    u32 pixel;
+    u32 r;
+    u32 g;
+    u32 b;
+
+    switch (type) {
+        case 4: {
+            u8* destI4 = (u8*)destI;
```
engineer124, line 122: I wonder if this could either use an enum, or `type` could be more descriptive of being the "size" of the `I` (`I4`/`I5`/`I8`/`I16`), I feel there should be a term for that. Might be a @Thar0 question

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+    /* 0 */ PICTOGRAPH_PHOTO_STATE_OFF,
+    /* 1 */ PICTOGRAPH_PHOTO_STATE_START,
+    /* 2 */ PICTOGRAPH_PHOTO_STATE_TAKE,
+    /* 3 */ PICTOGRAPH_PHOTO_STATE_FINISH
+} PictographPhotoState;
+
+// Will take the photograph, but doesn't compress and save it
+void Play_StartTakingPictographPhoto(void) {
+    R_PICTOGRAPH_PHOTO_STATE = PICTOGRAPH_PHOTO_STATE_START;
+}
```
engineer124, line 217:
```suggestion
// Will take the photograph, but doesn't compress and save it
void Play_TriggerPictographPhoto(void) {
    R_PICTOGRAPH_PHOTO_STATE = PICTOGRAPH_PHOTO_STATE_START;
}
```
Maybe as a thought? Or `RequestPictographPhoto`?

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+s32 Play_ChooseDynamicTransition(PlayState* this, s32 transitionType) {
+    s32 nextTransitionType = transitionType;
+
+    if (transitionType == TRANS_TYPE_FADE_DYNAMIC) {
+        if (!gSaveContext.save.isNight) {
+            nextTransitionType = TRANS_TYPE_FADE_WHITE;
+        } else {
+            nextTransitionType = TRANS_TYPE_FADE_BLACK;
+        }
+    }
```
engineer124, line 233: I wonder if this could be called `TRANS_TYPE_FADE_DAYTIME` or something, to indicate that it changes based on day or night?

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+const char D_801DFA34[] = "all";
+const char D_801DFA38[] = "a";
+const char D_801DFA3C[] = "a";
+const char D_801DFA40[] = "b";
+const char D_801DFA44[] = "b";
+const char D_801DFA48[] = "c";
+const char D_801DFA4C[] = "c";
+const char D_801DFA50[] = "d";
+const char D_801DFA54[] = "d";
+const char D_801DFA58[] = "e";
```
engineer124, line 334: What do we do with this mess? Just leave it here? Don't suppose this can be made into an array to compact it somehow

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+                     ((this->nextEntrance == ENTRANCE(PATH_TO_MOUNTAIN_VILLAGE, 1)) &&
+                      !(gSaveContext.save.weekEventReg[0x21] & 0x80)) ||
+                     ((this->nextEntrance == ENTRANCE(ROAD_TO_SOUTHERN_SWAMP, 1)) &&
+                      !(gSaveContext.save.weekEventReg[0x14] & 2)) ||
+                     ((this->nextEntrance == ENTRANCE(TERMINA_FIELD, 2)) &&
+                      !(gSaveContext.save.weekEventReg[0x37] & 0x80)) ||
+                     ((this->nextEntrance == ENTRANCE(ROAD_TO_IKANA, 1)) &&
+                      !(gSaveContext.save.weekEventReg[0x34] & 0x20))) &&
+                    (!func_800FE590(this) || (Entrance_GetSceneId(this->nextEntrance + sceneLayer) < 0) ||
+                     (Audio_GetActiveSequence(0) != NA_BGM_FINAL_HOURS))) {
```
engineer124, line 621:
```suggestion
                     (Audio_GetActiveSequence(SEQ_PLAYER_BGM_MAIN) != NA_BGM_FINAL_HOURS))) {
```

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play/func_80165E1C.s")
+                if (func_800FD768()) {
+                    func_801A4058(20);
+                    gSaveContext.seqId = 0xFF;
+                    gSaveContext.ambienceId = 0xFF;
+                }
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play/func_80165E7C.s")
+                if (func_800FE590(this) && (Entrance_GetSceneId(this->nextEntrance + sceneLayer) >= 0) &&
+                    (Audio_GetActiveSequence(0) == NA_BGM_FINAL_HOURS)) {
```
engineer124, line 634:
```suggestion
                    (Audio_GetActiveSequence(SEQ_PLAYER_BGM_MAIN) == NA_BGM_FINAL_HOURS)) {
```

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+    GraphicsContext* gfxCtx = this->state.gfxCtx;
+
+    {
+        GraphicsContext* gfxCtx2 = this->state.gfxCtx;
+
+        if (sBombersNotebookOpen) {
+            if (D_801FBBD4 != 1) {
+                MsgEvent_SendNullTask();
+                func_80178818();
+                gfxCtx2->curFrameBuffer = SysCfb_GetFbPtr(gfxCtx2->framebufferIdx % 2);
```
engineer124, line 1458:
```suggestion
                gfxCtx2->curFrameBuffer = SysCfb_GetFbPtr(gfxCtx2->framebufferIndex % 2);
```
Since it's in this function?

src/code/z_play.c
```diff
@@ -540,13 +2060,280 @@ void Play_AssignPlayerActorCsIdsFromScene(GameState* thisx, s32 startActorCsId)
...
+
+    PreRender_Init(&this->pauseBgPreRender);
+    PreRender_SetValuesSave(&this->pauseBgPreRender, D_801FBBCC, D_801FBBCE, NULL, NULL, NULL);
+    PreRender_SetValues(&this->pauseBgPreRender, D_801FBBCC, D_801FBBCE, NULL, NULL);
+
+    this->unk_18E64 = D_801FBB90;
+    this->pictoPhotoI8 = gPictoPhotoI8;
+    this->unk_18E68 = D_80784600;
+    this->unk_18E58 = D_80784600;
+    this->unk_18E60 = D_80784600;
```
engineer124, line 2253: I'd like another pair of eyes on this, but these all seem to be various framebuffer pointers. Mainly I'm curious if `pictoPhotoI8` should have `fb` in its name somehow, or if it should even be a `TextuerPtr` or if its treated more as an image. I don't fully understand framebuffer stuff

engineer124 (commented):
Great to see this open, thanks for organizing this! I can start with some cleanup and discussion questions

src/overlays/actors/ovl_Bg_Crace_Movebg/z_bg_crace_movebg.c
AngheloAlf, line 331:
```suggestion
            play->haltAllActors = true;
```
Unless this is not a bool

src/overlays/actors/ovl_En_Warp_tag/z_en_warp_tag.c
```diff
@@ -224,7 +224,7 @@ void EnWarpTag_RespawnPlayer(EnWarptag* this, PlayState* play) {
         if (new15E < 0) {
             new15E = 0;
         }
-        func_80165658(new15E * 0.04f); // unknown Play_ function
+        Play_SetMotionBlurAlpha(new15E * 0.04f);
```
AngheloAlf, line 227: Would `/25.0f` match? or maybe `* (1.0f / 25.0f)`
```suggestion
        Play_SetMotionBlurAlpha(new15E / 25.0f);
```

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+void Play_DisableMotionBlurPriority(void) {
+    R_MOTION_BLUR_PRIORITY_ENABLED = false;
+}
+
+typedef enum {
+    /* 0 */ PICTOGRAPH_PHOTO_STATE_OFF,
+    /* 1 */ PICTOGRAPH_PHOTO_STATE_START,
+    /* 2 */ PICTOGRAPH_PHOTO_STATE_TAKE,
+    /* 3 */ PICTOGRAPH_PHOTO_STATE_FINISH
+} PictographPhotoState;
```
engineer124, line 212: Because `R_PICTOGRAPH_PHOTO_STATE` also appears in `z_parameter.c`, could you move this to `z64.h`? I also introduce a `z64interface.h` in my parameter branch, so maybe it could end up there? Either way, it's not local to this file

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+void Play_EnableMotionBlur(u32 alpha) {
+    R_MOTION_BLUR_ALPHA = alpha;
+    R_MOTION_BLUR_ENABLED = true;
+}
+
+void Play_DisableMotionBlur(void) {
+    R_MOTION_BLUR_ENABLED = false;
+}
+
+void Play_ConvertBufferToI(void* destI, u16* srcBuf, s32 bufWidth, s32 pixelLeft, s32 pixelTop, s32 pixelRight,
```
Thar0, line 111: This is a bit confusing as a name if you don't know it's dealing with "Intensity images", and it doesn't mention the format of the input buffer which is expected to be RGBA16 only. I suggest something more like `Play_ConvertImageRgba16ToIntensity`, perhaps shuffling `Image` elsewhere

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+
+    Matrix_MtxFToMtx(&this->billboardMtxF, this->billboardMtx);
+    Matrix_RotateYF(((s16)(Camera_GetCamDirYaw(GET_ACTIVE_CAM(this)) + 0x8000)) * 0.0000958738f, MTXMODE_NEW);
+    Matrix_ToMtx(this->billboardMtx + 1);
+
+    gSPSegment(POLY_OPA_DISP++, 0x01, this->billboardMtx);
+    gSPSegment(POLY_XLU_DISP++, 0x01, this->billboardMtx);
+    gSPSegment(OVERLAY_DISP++, 0x01, this->billboardMtx);
+
+    if (1) {
```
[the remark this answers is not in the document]
engineer124: I just checked the scratch for `Play_Draw` and doesn't seem to affect regalloc

src/code/z_play.c
engineer124: Alright yeah, played around with this a bit and this matches:
```c
const char D_801DFA18[] = "来た!!!!!!!!!!!!!!!!!!!!!";
const char D_801DFA34[][4] = {
    "all", "a",  "a",  "b",  "b",  "c",  "c",  "d",   "d",   "e",  "e",  "f",  "fa", "fa", "fb", "fb",
    "fc",  "fc", "fd", "fd", "fe", "fe", "fg", "fg",  "fh",  "fh", "fi", "fi", "fj", "fj", "fk", "fk",
    "f",   "g",  "g",  "h",  "h",  "i",  "i",  "all", "all", "a",  "b",  "c",  "d",  "e",  "f",  "g",
    "h",   "i",  "f",  "fa", "fb", "fc", "fd", "fe",  "ff",  "fg", "fh", "fi", "fj", "fk",
};
```
so that's probably better

include/variables.h
```diff
@@ -3479,7 +3479,7 @@ extern GfxPool gGfxPools[2];
 extern u8 gAudioHeap[0x138000];
 extern u8 gSystemHeap[UNK_SIZE];
 
-extern u8 D_80780000[0x4600];
+extern u8 gPictoPhotoI8[0x4600];
```
engineer124, line 3482: Maybe something else to think about: we could introduce defines for the resolution of the pictophoto here:
```
#define PICTO_RESOLUTION_X 160
#define PICTO_RESOLUTION_Y 112
```

But `z64snap.h` already uses this terminology for a different set of values, but that seems to be specifically for the range that a pictoactor will "register" as being in the photo, so I think that should be named after "CAPTURE_REGION"  or something.

Here's my first attempt at some of these names and macros:

https://github.com/engineer124/mm/commit/2d64b28e8ef3892f6ebe327d60b83441cb05d084

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+                break;
+        }
+
+        if (this->envCtx.sandstormState != 0) {
+            Environment_DrawSandstorm(this, this->envCtx.sandstormState);
+        }
+
+        if (this->unk_18876 != 0) {
+            Environment_FillScreen(gfxCtx, 0, 0, 0, this->unk_18876, 3);
+        }
```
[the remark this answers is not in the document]
fig02: its placement (and usage, if there are any usages) are important for naming this.
Its functionally the same thing as `bgCoverAlpha`, but the bg cover happens after the scene/room is drawn and before actors. hence "bg cover"

This seems to happen after actors and other environmental things are drawn, but before interface elements and some other things

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+            }
 
+            gSPEndDisplayList(sp74++);
+            Graph_BranchDlist(sp70, sp74);
+            POLY_OPA_DISP = sp74;
+            this->unk_18B49 = 2;
+            SREG(33) |= 1;
+        } else {
+        Play_Draw_PostWorldDraw:
+            Play_PostWorldDraw(this);
```
fig02, line 1420: the main point of this grouping of tasks in this function here is that they are things that should NOT appear in the screenshot taken by the pause menu for the background. 

oot calls this "overlay elements" currently, which is not the best name, but more gets the point across that its the hud, textboxes and other things that should not be drawn in the screenshot frame of the pause menu (or oots game over screen)

we can possibly come up with a good term for this and use it in both games. I dont think post world draw conveys that purpose as well as it can. but i dont have any ideas for a new name yet

include/regs.h
```diff
@@ -47,6 +47,11 @@
 #define R_RUN_SPEED_LIMIT                 REG(45)
 #define R_UPDATE_RATE                     SREG(30)
 #define R_PAUSE_MENU_MODE                 SREG(94)
+#define R_PLAY_FILL_SCREEN_ON             MREG(64)
+#define R_PLAY_FILL_SCREEN_R              MREG(65)
+#define R_PLAY_FILL_SCREEN_G              MREG(66)
+#define R_PLAY_FILL_SCREEN_B              MREG(67)
+#define R_PLAY_FILL_SCREEN_ALPHA          MREG(68)
```
fig02, line 54: does mm use this feature at all? in oot this is used only by king dodongo. the "real" screenfill that should be used is the environment context variables that can set color and enable it and stuff. 

if it ended up getting more use in this game then this seems fine, but from oots perspective this feature is more of a hack than anything lmao

src/code/z_construct.c
```diff
@@ -1,4 +1,5 @@
 #include "global.h"
 
-void func_801AAAA0(PlayState* play) {
+// Remnant from OoT
+void Construct_InitRegs(PlayState* play) {
 }
```
fig02, line 5: would prob be good to wait for oot to decide on a name for this, and then name it accordingly. 
I for example am not sure of "Construct" being the system name for it, even tho it is the name of the file. Seems like something that will need more discussion on that end

src/code/z_lights.c
```diff
@@ -261,12 +261,12 @@ void LightContext_SetAmbientColor(LightContext* lightCtx, u8 r, u8 g, u8 b) {
...
 }
 
-void func_80102544(LightContext* lightCtx, u8 a1, u8 a2, u8 a3, s16 numLights, s16 sp16) {
-    lightCtx->unk7 = a1;
-    lightCtx->unk8 = a2;
-    lightCtx->unk9 = a3;
-    lightCtx->unkA = numLights;
-    lightCtx->unkC = sp16;
+void LightContext_SetFogProperties(LightContext* lightCtx, u8 r, u8 g, u8 b, s16 near, s16 far) {
+    lightCtx->fogColor.r = r;
```
fig02, line 265: oot has `LightContext_SetFog` and I think this is fine to use here too. its implied that you are changing properties of the thing

src/code/z_scene_proc.c
```diff
@@ -22,7 +22,7 @@ static Gfx sSceneDrawDefaultDL[] = {
 /**
  * Executes the current scene draw config handler.
  */
-void Scene_ExecuteDrawConfig(PlayState* play) {
+void Scene_Draw(PlayState* play) {
     static void (*sceneDrawConfigHandlers[])(PlayState*) = {
```
fig02, line 26: im not too excited about this one personally. it makes it sound like its drawing more than it actually is. its more of a setup for the scene as a whole while the rooms carry the weight of drawing the actual scene/room. so I think being explicit that its drawing the config in some way is more beneficial

src/code/z_play.c
fig02: I was confused reading this comment cause "daytime" sounded to me like just "day". I do understand your point now though is "day time based"

I think dynamic is fine though

src/code/z_play.c
fig02: does debug say what this is for

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
+                    if ((pad60 > 0) && ((pad60 % 2) != 0)) {
+                        this->envCtx.fillScreen = true;
+                        this->envCtx.screenFillColor[0] = this->envCtx.screenFillColor[1] =
+                            this->envCtx.screenFillColor[2] = 150;
+                        this->envCtx.screenFillColor[3] = 80;
+                    } else {
+                        this->envCtx.fillScreen = false;
+                    }
+                } else {
+                    Room_HandleLoadCallbacks(this, &this->roomCtx);
```
fig02, line 1006: this isnt a change from this PR, but I am curious on this choice of name if anyone knows. theres no callbacks occurring anywhere in this afaik, but may just be misunderstanding the intent

src/code/z_play.c
```diff
@@ -1,81 +1,1608 @@
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play/Play_Main.s")
+    {
+        GraphicsContext* gfxCtx = this->state.gfxCtx;
+
+        if (1) {
+            this->state.gfxCtx = NULL;
+        }
+        Play_SetupUpdate(this);
+        this->state.gfxCtx = gfxCtx;
```
fig02, line 1503: these "setup" functions for update and draw seem to really be about handling the bombers notebook while its open, and if its not it just passes it onto the normal update and draw functions. Im wondering if we can name it more based on that, instead of implying that its setting things up for the main update and draw processes

part of me wants these "setup" functions to actually be named update and draw, because they are the direct functions called by main. but I can understand why that may not be desirable

include/regs.h
engineer124: So far, this system is used by:
- Twinmold
- Song of Soaring Effect
- Koume in Woods of Mystery
- Grandma when telling storytime

src/code/z_play.c
engineer124: Name came from eons ago, before `z_room.c` even existed it looks like, let alone decomped
https://github.com/zeldaret/mm/commit/9bc4990814d2e953df294bc57f7bca1918bd3bac

Thar0: I guess the rationale behind the name is that it doesn't really do anything unless the DMA is complete, so it's a "callback" with respect to DMA completion. But yeah there are better names for sure

src/code/z_play.c
engineer124: Looks like `"来た!!!!!!!!!!!!!!!!!!!!!"` can be inferred from OoT:

```c
case TRANS_MODE_SANDSTORM_END_INIT:
    if (this->transitionTrigger == TRANS_TRIGGER_END) {
        this->envCtx.sandstormState = SANDSTORM_DISSIPATE;
        this->envCtx.sandstormPrimA = 255;
        this->envCtx.sandstormEnvA = 255;
        // "It's here!!!!!!!!!"
        LOG_STRING("来た!!!!!!!!!!!!!!!!!!!!!", "../z_play.c", 3471);
        this->transitionMode = TRANS_MODE_SANDSTORM_END;
    } else {
        this->transitionMode = TRANS_MODE_SANDSTORM_INIT;
    }
    break;
```

I'm not seeing these letter string in debug at a first glance (at least what's been put into `z_play`, the things I'm seeing are:

```c
((void)"kankyo changed water, sound on\n");

((void)"kankyo changed water off, sound off\n");

// Sound initials are here. 111
((void)"\n\n\nサウンドイニシャル来ました。111\n");

((void)"\nSEQ_ID_TIME_LIMIT=[%d] Na_CheckPlayingBgmFlag = [%d]");

// Sound initials are here.
((void)"\n\n\nサウンドイニシャル来ました。\n");

// The sound initials for hijacking BGM have arrived.
((void)"\n\n\n乗っ取りＢＧＭ用サウンドイニシャル来ました。\n");

((void)"来た!!!!!!!!!!!!!!!!!!!!!");

// Call fbdemo_init failed!
((void)"fbdemo_init呼出し失敗！\n");
```

`fbdemo_init` based on OoT is `TransitionUnk_Init`
```c
TransitionUnk* TransitionUnk_Init(TransitionUnk* this, s32 row, s32 col) {
    osSyncPrintf("fbdemo_init(%08x, %d, %d)\n", this, row, col);
```
So in MM, that string is probably somewhere around here:
```c
if (TransitionUnk_Init(&sTrnsnUnk, 10, 7) == NULL) {
    gTrnsnUnkState = 0;
} else {
    D_801F6D0C = gZBufferPtr;
    gTrnsnUnkState = 3;
    Game_SetFramerateDivisor(&this->state, 1);
}
```

But nothing on those batch of letters. Maybe those are related to all transition types?

src/code/z_construct.c
Dragorn421: if this is the function I think it is, https://github.com/zeldaret/oot/pull/1354 (not merged) renames it from `func_80112098` to `Regs_InitData`

src/code/z_play.c
hensldm: AFAIK it is not a framebuffer as it isn't storing frame data in the general sense, so doesn't need `fb`. I could see using TexturePtr, but it also is a known texture type so I think `u8*` works as well.

src/code/z_play.c
```diff
@@ -540,13 +2060,280 @@ void Play_AssignPlayerActorCsIdsFromScene(GameState* thisx, s32 startActorCsId)
...
+    D_801F6D4C->primColor.a = 0;
+    D_801F6D4C->envColor.r = 0;
+    D_801F6D4C->envColor.g = 0;
+    D_801F6D4C->envColor.b = 0;
+    D_801F6D4C->envColor.a = 0;
+    EnvFlags_UnsetAll(this);
+    THA_GetSize(&this->state.heap);
+    zAllocSize = THA_GetSize(&this->state.heap);
+    zAlloc = (uintptr_t)THA_AllocEndAlign16(&this->state.heap, zAllocSize);
+    ZeldaArena_Init(((zAlloc + 8) & ~0xF), (zAllocSize - ((zAlloc + 8) & ~0xF)) + zAlloc);
```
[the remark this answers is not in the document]
hensldm: I'm not sure (it is the same in OoT). It does work as an `ALIGN16(zAlloc - 7)` though not sure if that would be preferred.

src/code/z_play.c
hensldm: Personally I like the more general dynamic. In theory you could add more dynamic transitions that are not based on the time of day.

src/code/z_play.c
hensldm: I'm leaning more and more this way, does anyone have any objections?

src/code/z_play.c
EllipticEllipsis: IIRC z_room looks very similar between the games, so it may have a direct OoT equivalent anyway.

src/code/z_play.c
EllipticEllipsis: That seems more like a side effect than its intended purpose, especially since some of the interface does still draw when kaleido's open. The point for me was that neither the light adjustment nor the screen shrinking effect is really an overlay, and the distinguishing feature seemed to be that it drew after everything else.

src/code/z_play.c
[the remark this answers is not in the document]
EllipticEllipsis: Doesn't init realign the pointer anyway? Not that this seems very sensible, mind you.

src/code/z_play.c
EllipticEllipsis: `TIME_DEPENDENT` is the only alternative I can think of, which is longer but more precise.

src/code/z_play.c
fig02: it is 100% its purpose, not a side effect. They draw in the pause menu yes, but the point is the SCREENSHOT image that is in the background. All of the things contained within this function are skipped during the prerender screenshot step so that it wont be in the picture. screen shrinking effect included.

fig02: agree on overlay not being the correct term, but this is particular grouping of these drawing elements is soley for the purpose of the prerender image and should have a better name to reflect that. this is all controlled by 
`R_PAUSE_MENU_MODE` as it runs through its different states while taking the screenshot and processing the image

fig02: also just noticed it skips this stuff for taking a pictograph photo, so the hud etc isnt in the picture. same concept though

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+        }
+
+        if (this->unk_18876 != 0) {
+            Environment_FillScreen(gfxCtx, 0, 0, 0, this->unk_18876, 3);
+        }
+
+        DebugDisplay_DrawObjects(this);
+        Play_DrawMotionBlur(this);
+
+        if (((R_PAUSE_MENU_MODE == 1) || (gTrnsnUnkState == 1)) || (R_PICTOGRAPH_PHOTO_STATE == 1)) {
```
engineer124, line 1322:
```suggestion
        if (((R_PAUSE_MENU_MODE == 1) || (gTrnsnUnkState == 1)) || (R_PICTOGRAPH_PHOTO_STATE == PICTOGRAPH_PHOTO_STATE_START)) {
```
There's a few of these missed for `R_PICTOGRAPH_PHOTO_STATE`, could you double check these?

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+                this->pauseBgPreRender.cvgSave = this->unk_18E58;
+            } else if (R_PICTOGRAPH_PHOTO_STATE == 1) {
+                R_PICTOGRAPH_PHOTO_STATE = 2;
+                this->pauseBgPreRender.fbufSave = (u16*)gfxCtx->zbuffer;
+                this->pauseBgPreRender.cvgSave = this->unk_18E58;
+            } else {
+                gTrnsnUnkState = 2;
+                this->pauseBgPreRender.fbufSave = (u16*)gfxCtx->zbuffer;
+                this->pauseBgPreRender.cvgSave = NULL;
+            }
```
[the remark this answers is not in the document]
fig02: they are very similar yeah. main difference is that `PAUSE_BG_PRERENDER_DONE` for the pause menu has more of a drastic effect on the game, in that it will allow the pause menu and other "overlay elements" as oot currently calls them to draw after the screenshot/prerender has already been taken.

The "setup" step for both avoids the call to the "overlay elements" or "post world elements" as both games call it currently, so that the hud and other elements arent shown in the picto picture or pause background. 
Process for both processes the image with prerender filters via `PreRender_ApplyFilters`, and the picto has the extra conversion step as well.

It seems like its prob a good idea that they line up. But I would take "bg" out of your current suggestion. its only called bg for pause because its in the background of the pause menu. its not a background for the picto picture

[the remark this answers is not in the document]
hensldm: Hmm, looking at it, even the Motion Blur status mirrors it as well
```c
typedef enum {
    /* 0 */ MOTION_BLUR_STOPPED,
    /* 1 */ MOTION_BLUR_INITALIZING,
    /* 2 */ MOTION_BLUR_RUNNING
} MotionBlurStatus;
```
Kind of seems like how play processes multistep process systems.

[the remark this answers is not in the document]
hensldm: Except possibly `gTrnsnUnkState` which seems close to this pattern but not quite.

src/code/z_play.c
```diff
@@ -540,13 +1997,280 @@ void Play_AssignPlayerActorCsIdsFromScene(GameState* thisx, s32 startActorCsId)
...
+        if ((gSaveContext.save.weekEventReg[52] & 0x20) && (scene == ENTR_SCENE_IKANA_CANYON)) {
+            gSaveContext.nextCutsceneIndex = 0xFFF2;
+        }
+
+        if ((gSaveContext.save.weekEventReg[55] & 0x80) &&
+            ((scene == ENTR_SCENE_GREAT_BAY_COAST) || (scene == ENTR_SCENE_ZORA_CAPE))) {
+            gSaveContext.nextCutsceneIndex = 0xFFF0;
+        }
+
+        if (INV_CONTENT(ITEM_OCARINA) != ITEM_OCARINA) {
```
EllipticEllipsis, line 2079:
```suggestion
        // Not "First cycle" Termina Field
        if (INV_CONTENT(ITEM_OCARINA) != ITEM_OCARINA) {
```
maybe. Indeed, this entire section is for loading the alternatives based on conditions being satisfied.

src/code/z_play.c
```diff
@@ -540,13 +1997,280 @@ void Play_AssignPlayerActorCsIdsFromScene(GameState* thisx, s32 startActorCsId)
...
+    EnvFlags_UnsetAll(this);
+    THA_GetSize(&this->state.heap);
+    zAllocSize = THA_GetSize(&this->state.heap);
+    zAlloc = (uintptr_t)THA_AllocEndAlign16(&this->state.heap, zAllocSize);
+    ZeldaArena_Init(((zAlloc + 8) & ~0xF), (zAllocSize - ((zAlloc + 8) & ~0xF)) + zAlloc);
+    Actor_InitContext(this, &this->actorCtx, this->linkActorEntry);
+
+    while (!Room_HandleLoadCallbacks(this, &this->roomCtx)) {
+        // Empty
+    }
```
EllipticEllipsis, line 2248: Comment not needed (obviously this is a "wait for something" loop).

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+    u32 b;
+
+    switch (bitDepth) {
+        case 4: {
+            u8* destI4 = (u8*)destI;
+            u32 upper;
+            u32 lower;
+
+            for (i = pixelTop; i <= pixelBottom; i++) {
+                for (j = pixelLeft; j <= pixelRight; j += 2) {
```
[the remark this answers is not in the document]
EllipticEllipsis: @Thar0 reminds me that it's `G_IM_SIZ_16b_BYTES`.

src/code/z_play.c
hensldm: Fig suggested `Play_UpdateGame/Play_DrawGame` in the discord, which I agree with.

src/code/z_play.c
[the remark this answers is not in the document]
hensldm: So I've synced all three enums, but haven't combined any yet. I'm not opposed to it, but I'm kind of intersted in which should be combined, and would like to know more about `gTrnsnUnkState` as well. Other's opinions would also be appreciated.

src/code/z_play.c
fig02: it was more of an example name, Im not 100% for it cause it implies that the other gamestates arent the game as a whole, which they are. so its kinda weird. 
And one may argue that "game" has closer meaning to gamestate cause of its original name, but I think we are passed that point by now 

"gameplay" is also not great imo, but at least in oot we do have things using that term like `Player_DrawGameplay` `Player_OverrideLimbDrawGameplayDefault` `Player_OverrideLimbDrawGameplayFirstPerson`
etc, which is conveying the same meaning of "gameplay" as we are trying to here

src/code/z_play.c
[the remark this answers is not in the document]
fig02: atm I think the blur one is not as closely related to the picto and pause menu prerender states. they are similar in that they have "inactive, init, running", whatever terms you want to use for them. but the picto and pause menu ones are very close in purpose in terms of directing control flow of gameplay draw as it takes the screenshots and processes it etc

src/code/z_play.c
EllipticEllipsis: It's a fair question to what extent we want to sync those up, since for the Player they're explicitly "when kaleido isn't open", which still counts as the non–Bombers Notebook draw here; they're similar, but not quite the same.

fig02: even though the non-gameplay state being referred to here is different, the same things are meant by gameplay. so I think it makes sense. But yea idk if this terminology is going to stick around long term

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+                    pixel = srcRgba16[i * rgba16Width + j];
+                    r = (pixel >> 11) & 0x1F;
+                    g = (pixel >> 6) & 0x1F;
+                    b = (pixel >> 1) & 0x1F;
+                    upper = ((r * 2 + g * 4 + b) * 0xF) / 217;
```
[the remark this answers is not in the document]
hensldm: Not from what I can tell.

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+    R_MOTION_BLUR_PRIORITY_ENABLED = true;
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play/func_80166060.s")
+void Play_DisableMotionBlurPriority(void) {
+    R_MOTION_BLUR_PRIORITY_ENABLED = false;
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play/func_801660B8.s")
+// Will take the photograph, but doesn't compress and save it
```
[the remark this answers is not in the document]
hensldm: I think and is more appropriate here, as the actual saving of the photo is with compression. (Technically this function would save the new photo in `gPictoPhotoI8`, it just doesn't compress and save it to saveContext, so will just get overwritten.
As for Photo, it is to differentiate from the actual `Pictograph` item.

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+    this->projectionMtxFDiagonal.z = -this->viewProjectionMtxF.zz;
+
+    SkinMatrix_MtxFMtxFMult(&this->viewProjectionMtxF, &this->billboardMtxF, &this->viewProjectionMtxF);
+
+    this->billboardMtxF.mf[3][2] =
+        (this->billboardMtxF.mf[3][1] =
+             (this->billboardMtxF.mf[3][0] =
+                  (this->billboardMtxF.mf[2][3] =
+                       (this->billboardMtxF.mf[1][3] = (this->billboardMtxF.mf[0][3] = 0.0f)))));
+
```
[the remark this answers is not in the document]
hensldm: Looks like permuter output

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+    this->unk_18E58 = NULL;
+    this->pictoPhotoI8 = NULL;
+    this->unk_18E60 = NULL;
+    this->unk_18E64 = NULL;
+    this->unk_18E68 = NULL;
+    Effect_DestroyAll(this);
+    EffectSS_Clear(this);
+    CollisionCheck_DestroyContext(this, &this->colChkCtx);
+
+    if (gTrnsnUnkState == 3) {
```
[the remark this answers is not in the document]
hensldm: Agree overall. However since this name is temporary anyways, I think it would be better to keep the same terminology as OoT so it would be easier to document them together.

src/code/z_play.c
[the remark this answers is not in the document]
EllipticEllipsis: Yes, it's processing 2 at a time because it's processing to make one byte of the output at a time, right? If you were squashing an RGBA32 in you'd want to do it 4 at a time, etc.

src/code/z_play.c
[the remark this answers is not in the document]
EllipticEllipsis: The item is the Pictograph Box, though.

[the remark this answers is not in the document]
hensldm: Tbh, I always forget it is called the `Pictograph Box`. Hmm, @engineer124 we talked a little about this in Discord, perhaps it would be better if the parameter functions/enums used `PictographBox`, while this photo related things in play would just use `Pictograph`?

src/code/z_play.c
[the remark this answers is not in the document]
hensldm: I'm sure that is quite right? Both I5 and I8 are 1 byte or less, but they still only process 1 pixel at a time.

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+    s32 fbdemoType;
+
+    bzero(transitionCtx, sizeof(TransitionContext));
+
+    fbdemoType = -1;
+    if (transitionType & 0x40) {
+        fbdemoType = FBDEMO_WIPE3;
+    } else if ((transitionType & 0x78) == 0x20) {
+        fbdemoType = FBDEMO_WIPE4;
+    } else if (!(transitionType & 0x60)) {
```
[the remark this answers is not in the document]
hensldm: Went down the rabbit hole, and think I understand what is going on. Decided to go with bit defines. Let me know what you think.

src/code/z_play.c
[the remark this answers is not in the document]
hensldm: True, I forgot about the control flow in draw. So it makes more sense that those need to actually stay in sync. Will combine those into one enum `PictoPreRenderState`.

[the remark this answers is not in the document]
fig02: im not totally sure on combining enums. and at least with that naming scheme, it makes it sound like its only about the picto prerender and not the pause prerender.
imo currently its safe to say that they should be synced up in naming scheme, but can probably be different enums. but thats just imo

src/code/z_play.c
```diff
@@ -393,9 +1857,7 @@ void func_80169ECC(PlayState* this) {
     }
 }
 
-// Gameplay_TriggerVoidOut ?
-// Used by Player, Ikana_Rotaryroom, Bji01, Kakasi, LiftNuts, Test4, Warptag, WarpUzu, Roomtimer
-void func_80169EFC(GameState* thisx) {
+void Play_TriggerVoidOut(GameState* thisx) {
```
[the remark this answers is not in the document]
EllipticEllipsis: Think my preference would be to leave them unnamed for now, if we name them it looks like we are more certain about what they are than we actually are.

src/code/z_play.c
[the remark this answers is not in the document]
hensldm: Maybe `worldCoverAlpha` since like how Elliptic described `PostWorldDraw` this covers everything involving the game world.

src/code/z_play.c
hensldm: Still considering different names, but went ahead and implemented this using just `UpdateGame` and `DrawGame` for now.

src/code/z_play.c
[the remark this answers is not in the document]
AngheloAlf: I vote for using `Pictograph` and `PictographBox` to differentiate them

AngheloAlf (approved):

src/code/z_play.c
EllipticEllipsis: I don't want `Game` for the same reason we decided not to use it for gamestates: the whole thing is the game, so it doesn't really make sense to use it to describe a subset.

src/code/z_play.c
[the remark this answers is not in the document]
engineer124: If I was looking at this code fresh and saw `pictograph`, I would immediately think of the item and the system as a whole. I don't think I would see `pictograph` as the photo only, excluding the item, whereas `pictoPhoto` is much clearer

src/code/z_play.c
```diff
@@ -1,81 +1,1545 @@
...
+void Play_ConvertRgba16ToIntensityImage(void* destI, u16* srcRgba16, s32 rgba16Width, s32 pixelLeft, s32 pixelTop,
+                                        s32 pixelRight, s32 pixelBottom, s32 bitDepth) {
+    s32 i;
+    s32 j;
+    u32 pixel;
+    u32 r;
+    u32 g;
+    u32 b;
+
+    switch (bitDepth) {
```
EllipticEllipsis, line 120: This should specify that it's the destination's bit depth if you're not going to do a full Doxygen comment, I think.

src/code/z_play.c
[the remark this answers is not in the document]
EllipticEllipsis: I meant in that the loops in this function operate one byte of the image at a time, so you get
- 4: 2 -> 1
- 5,8: 1 -> 1
- 16: 2 -> 2 (treated as 1)

and these factors can be expressed as ratios, although I suppose the issue here is that the ratios are the wrong way up and so don't really work as integers: if `G_IM_SIZ_4b_BYTES` was 1/2 it would work.

src/code/z_play.c
EllipticEllipsis: Hang on, wrong way round: not having Ocarina *is* First cycle...

src/code/z_play.c
EllipticEllipsis: I think best to wait on this one until we know we have all of the information.

src/code/z_play.c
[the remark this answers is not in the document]
hensldm: Okay I can see that. But as you mention `G_IM_SIZ_4b_BYTES` would be the correct define, but only if it was `1/2`. For that reason, and the fact it would be weird to do it for the other ones as well, I think it is just best to leave it as 2.

src/code/z_play.c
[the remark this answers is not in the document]
hensldm: Posted in [discord](https://discord.com/channels/688807550715560050/873211000122384415/1029582125718065283), but I like the idea of using `PICTO` as a general system that each part shares, than differentiating them with `PHOTO` for the actual photo, `BOX` for the item, and `CAPTURE` for the capture (what is currently in z64snap.h)

src/code/z_play.c
hensldm: I'm not sure it is really needed, since the function name already specifics RGBA16, so the bitDepth can only apply to the I image. Though I should add a doxygen comment, I just got lazy :sweat_smile:

hensldm:
Pushed some clean up regarding the two nonmatchings, along with a closer match for `DrawGame`.

include/z64light.h
```diff
@@ -74,11 +74,9 @@ typedef struct LightsBuffer {
...
     /* 0x0 */ LightNode* listHead;
     /* 0x4 */ Color_RGB8 ambient;
-    /* 0x7 */ u8 unk7;
-    /* 0x8 */ u8 unk8;
-    /* 0x9 */ u8 unk9;
-    /* 0xA */ s16 unkA;
-    /* 0xC */ s16 unkC;
+    /* 0x7 */ Color_RGB8 fogColor;
+    /* 0xA */ s16 fogNear;
+    /* 0xC */ s16 fogFar;
```
Dragorn421, line 79: in oot at least this is actually `zFar` (was merged 2 weeks ago)
```suggestion
    /* 0xC */ s16 zFar;
```

src/code/z_play.c
```diff
@@ -1,81 +1,1575 @@
...
+            u32 upper;
+            u32 lower;
+
+            for (i = pixelTop; i <= pixelBottom; i++) {
+                for (j = pixelLeft; j <= pixelRight; j += 2) {
+                    pixel = srcRgba16[i * rgba16Width + j];
+                    r = RGBA16_GET_R(pixel);
+                    g = RGBA16_GET_G(pixel);
+                    b = RGBA16_GET_B(pixel);
+                    upper = ((r * 2 + g * 4 + b) * 0xF) / 217;
```
Dragorn421, line 144: 217 is `0x1F * (2+4+1)`, it's for normalization

engineer124: Nice, makes sense. I wonder if this whole expression deserves a macro. 

Also, does it make sense to think of this as `(r << 1) + (g << 2) + (b << 0)`? Although I don't fully follow this conversion.

So in full:
`#define COLOR_UPPER(r, g, b) ( (((r << 1) + (g << 2) + (b << 0)) * 0xF) / (0x1F * (1 <<1 + (1 << 2) + (1 << 0))) )`?
except with a better macro name and double checking brackets? Maybe even a separate macro for the normalization constant? Or maybe it's all too much. Just throwing out ideas.

src/code/z_play.c
```diff
@@ -1,81 +1,1575 @@
...
+FbDemoStruct sTrnsnUnk;           // 0x801F6C30
+u16* D_801F6D0C;                  // 0x801F6D0C
+s32 gTrnsnUnkState;               // 0x801F6D10
+VisMono sVisMono;                 // 0x801F6D18
+Color_RGBA8_u32 gVisMonoColor;    // 0x801F6D30
+Struct_80140E80 D_801F6D38;       // 0x801F6D38
+Struct_80140E80* D_801F6D4C;      // 0x801F6D4C
+BombersNotebook sBombersNotebook; // 0x801F6D50
+u8 sBombersNotebookOpen;          // 0x801F6DFC
+u8 sMotionBlurStatus;             // 0x801F6DFD
```
engineer124, line 26: Do we need these addresses? You can just look in `variables.txt` or the map file to get these

src/code/z_play.c
Dragorn421: why use shifts? the 2, 4, 1 are just weights
for example look at https://en.wikipedia.org/wiki/YUV#Conversion_to/from_RGB
Luminance is defined in that page as `Y = 0.299 * r + 0.587 * g + 0.114 * b`
which are close-ish to oot's 2,4,1 (after normalization)

This isn't news to me either, I've already seen this appear in oot's vismono
https://github.com/zeldaret/oot/blob/7ecafcfe7d357ff1cbcf49ce4d87a9efedae552e/src/code/z_vismono.c#L16

The reason for the calculation to not be (r+g+b)/3 btw, is the human eye sees green much better than the other colors (and red better than blue I guess (idk how the coefficients from wikipedia were derived))

AngheloAlf: I think the macro idea sounds good. And as Dragorn pointed out it should probably not use shifts

src/code/z_play.c
```diff
@@ -1,81 +1,1575 @@
 #include "global.h"
+#include "z64rumble.h"
 #include "z64quake.h"
```
engineer124, line 3:
```suggestion
#include "z64quake.h"
#include "z64rumble.h"
```
q before r

src/code/z_play.c
```diff
@@ -1,81 +1,1575 @@
...
+                }
+            }
+            break;
+
+        case TRANS_MODE_SANDSTORM_END_INIT:
+            if (this->transitionTrigger == TRANS_TRIGGER_END) {
+                this->envCtx.sandstormState = 4;
+                this->envCtx.sandstormPrimA = 255;
+                this->envCtx.sandstormEnvA = 255;
+                this->transitionMode = TRANS_MODE_SANDSTORM_END;
```
engineer124, line 844:
```suggestion
                this->envCtx.sandstormEnvA = 255;
                // "It's here!!!!!!!!!"
                (void)"来た!!!!!!!!!!!!!!!!!!!!!";
                this->transitionMode = TRANS_MODE_SANDSTORM_END;
```
From OoT, we know this is where this string goes. Should we put it here like this?

src/code/z_play.c
hensldm: https://github.com/zeldaret/mm/pull/1109#discussion_r990641404

include/z64light.h
hensldm: Based on `Play_DrawMain` WIP, that seems to be the case here as well. I'll change it.

src/code/z_play.c
hensldm: Seems good. I copied the macros from vismono into play (they could possibly just be combined and put in the color header, but I'm not 100% sure they should be correlated in that way). I've also added a `MIX` macro, which I like, but would like others opinions.

include/z64save.h
```diff
@@ -133,6 +133,15 @@ typedef enum {
...
 
+#define PICTO_RESOLUTION_WIDTH 160
+#define PICTO_RESOLUTION_HEIGHT 112
+
+#define PICTO_TOPLEFT_X ((SCREEN_WIDTH - PICTO_RESOLUTION_WIDTH) / 2)
+#define PICTO_TOPLEFT_Y ((SCREEN_HEIGHT - PICTO_RESOLUTION_HEIGHT) / 2)
+
+#define PICTO_SIZE (PICTO_RESOLUTION_WIDTH * PICTO_RESOLUTION_HEIGHT)
+#define PICTO_COMPRESSED_SIZE (PICTO_SIZE * 5 / 8)
+
```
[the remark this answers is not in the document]
hensldm: I believe there are three aspects we need to differentiate (please correct me if I'm wrong, and there are actually more).
- The actual photo (Photo)
- The item (Box)
- "The subregion of the picto photo that will set the flag for an actor being in the photo" (see z64snap.h) (Capture?)

src/code/z_play.c
AngheloAlf: Since those variables are not static and their addresses are present on `variables.txt` and the expected map file then imo this is a bit unnecessary

hensldm:
As mentioned in Discord, cleaned up the compression and decompression functions.

include/z64.h
```diff
@@ -869,15 +807,14 @@ typedef struct {
...
-typedef void (*ColChkVsFunc)(PlayState*, CollisionCheckContext*, Collider*, Collider*);
-typedef s32 (*ColChkLineFunc)(PlayState*, CollisionCheckContext*, Collider*, Vec3f*, Vec3f*);
+typedef s32 (*ColChkResetFunc)(struct PlayState*, Collider*);
+typedef void (*ColChkBloodFunc)(struct PlayState*, Collider*, Vec3f*);
+typedef void (*ColChkApplyFunc)(struct PlayState*, CollisionCheckContext*, Collider*);
+typedef void (*ColChkVsFunc)(struct PlayState*, CollisionCheckContext*, Collider*, Collider*);
+typedef s32 (*ColChkLineFunc)(struct PlayState*, CollisionCheckContext*, Collider*, Vec3f*, Vec3f*);
 
-typedef void(*room_draw_func)(PlayState* play, Room* room, u32 flags);
+typedef void(*RoomDrawFunc)(struct PlayState* play, Room* room, u32 flags);
```
engineer124, line 817: This is changed and addressed in https://github.com/zeldaret/mm/pull/1143, so I'd leave this alone to prevent conflicts

Dragorn421 (approved):
I can't speak for everything (in particular MM stuff) but I don't have further comments
(count this as a half-approval :sunglasses:)

include/functions.h
```diff
@@ -2312,37 +2315,38 @@ s32  TransitionOverlay_Load(TransitionOverlay *overlayEntry);
...
-void func_80165DCC(s32 arg0);
-void func_80165DF0(void);
-// void func_80165E1C(void);
-// void func_80165E7C(void);
-// void func_80165EC0(void);
-// void func_80166060(void);
-Gfx* func_801660B8(PlayState* play, Gfx* gfx);
+// void Play_DrawMotionBlur(void);
+// void Play_InitMotionBlur(void);
+// void Play_DestroyMotionBlur(void);
```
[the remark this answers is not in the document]
hensldm: Done, though I did leave some unused functions that look like they were meant to be exposed, but just didn't get used.

include/z64.h
hensldm: Did what I could. Note this line will still conflict since I needed to leave the `struct Playstate` change.

engineer124 (approved):

louist103 (approved):

src/code/z_play.c
```diff
@@ -549,4 +2056,270 @@ void Play_FillScreen(GameState* thisx, s16 fillScreenOn, u8 red, u8 green, u8 bl
...
+        gSaveContext.save.isNight = false;
+    }
+
+    func_800EDDB0(this);
+
+    if (((gSaveContext.gameMode != 0) && (gSaveContext.gameMode != 1)) || (gSaveContext.save.cutscene >= 0xFFF0)) {
+        gSaveContext.unk_3DC0 = 0;
+        Magic_Reset(this);
+        gSaveContext.sceneLayer = (gSaveContext.save.cutscene & 0xF) + 1;
+        gSaveContext.save.cutscene = 0;
```
engineer124, line 2204:
```suggestion
        // The cutscene system carried over from OoT was originally designed to trigger automatically when `>= 0xFFF0`.
        // This hack turns off cutscene here so that the scripted cutscene isn't immediately played.
        // This allows the `CutsceneManager` system to take over to determine when to play the scripted cutscene
        // Not setting this to 0 here will cause issues as each scene layer now has a list of cutscene scripts, so it will be unresolved as to which cutscene script to play
        gSaveContext.save.cutscene = 0;
```
I'm not sure the best way to phrase this, I don't think it needs to be this verbose. But here's a first attempt to get the idea across of why this line is important

hensldm:
```suggestion

        // Set `gSaveContext.save.cutscene` to 0 so that the `CutsceneManager` takes over, instead of automatically playing.
        gSaveContext.save.cutscene = 0;
```
Perhaps something like this?

engineer124: I think what you committed was fine. Thanks

PR #1111, opened 2022-10-06
Import `EnAObj`'s data and a few cleanups

AngheloAlf:
I noted `EnAObj`'s data haven't been imported yet, so I decided to do it.
I also cleaned it up a bit, named the action funcs and made macros for the params access. Also, for some reason the struct of this actor had extra padding for no apparent reason.

Do we want to make a header for this actor instead of having its stuff on `z64actor.h`?


include/z64actor.h
```diff
@@ -288,8 +288,17 @@ typedef struct EnAObj {
     /* 0x000 */ Actor actor;
     /* 0x144 */ EnAObjActionFunc actionFunc;
     /* 0x148 */ ColliderCylinder collision;
-    /* 0x194 */ UNK_TYPE1 pad194[0x14];
-} EnAObj; // size = 0x1A8
+} EnAObj; // size = 0x194
```
isghj5, line 291: ... That was still matching?

AngheloAlf: The only place where the size of an actor's struct is usually used is in its InitVars, but since we haven't imported the InitVars yet then we were using the hardcodded sizes from the asm, so this struct could have any size and still match.
Importing the data required the sizes to be correct so the ROM can match, then I had to correct them

isghj5 (approved):

jpburnett (approved):

src/code/z_en_a_keep.c
```diff
@@ -1,21 +1,65 @@
...
 
-void EnAObj_Update1(EnAObj* this, PlayState* play);
-void EnAObj_Update2(EnAObj* this, PlayState* play);
+void EnAObj_Init(Actor* thisx, PlayState* play);
+void EnAObj_Destroy(Actor* thisx, PlayState* play);
+void EnAObj_Update(Actor* thisx, PlayState* play);
+void EnAObj_Draw(Actor* thisx, PlayState* play);
+
+void EnAObj_WaitTalk(EnAObj* this, PlayState* play);
+void EnAObj_WaitFinishedTalking(EnAObj* this, PlayState* play);
```
hensldm, line 14: They are fine as they are, but thoughts on 
`EnAObj_Idle` and `EnAObj_Talk`.

hensldm (approved):
1 thought

include/z64actor.h
```diff
@@ -288,8 +288,17 @@ typedef struct EnAObj {
...
+
+typedef enum {
+    /* 0 */ AOBJ_SIGNPOST_OBLONG,
+    /* 1 */ AOBJ_SIGNPOST_ARROW,
+} AObjType;
+
+#define AOBJ_GET_TEXTID(thisx) ((((thisx)->params >> 8) & 0xFF) | 0x300)
```
EllipticEllipsis, line 298: Not sure how I feel about hiding the base textId like this, but we can worry about it later.

EllipticEllipsis (approved):

Kenix3 (approved):

PR #1121, opened 2022-10-14
Document Elegy of Emptiness shell assets in gameplay_keep

tom-overton:
To get in the spooky spirit, I decided to take a look at some of MM's creepier assets; the shells left behind when you play the Elegy of Emptiness. Note that the Zora shell's guitar shares textures with Mikau's grave, so the two should share names as well.


AngheloAlf (approved):

EllipticEllipsis (approved):

engineer124 (approved):
Nice work

PR #1128, opened 2022-10-16
Match 4 functions in BGcheck, remove nonequivalency in DynaPoly_ExpandSRT and BgHakaBombwall

petrie911:
Still some nasty stuff left, of course.

### DIFF
--- a/src/code/z_bgcheck.c
+++ b/src/code/z_bgcheck.c
@@ -73,10 +73,19 @@ BgSpecialSceneMaxObjects sCustomDynapolyMem[] = {
 // TODO: All these bss variables are localized to one function and can
 // likely be made into in-function static bss variables in the future
 
+ 
+
 Vec3f D_801ED9F0[3]; // polyVerts
 Vec3f D_801EDA18[3]; // polyVerts
 Vec3f D_801EDA80[3]; // polyVerts
 Vec3f D_801EDB48[3]; // polyVerts
+
+#ifndef NON_MATCHING
+Vec3f D_801EDB70[3]; // polyVerts;
+Plane D_801EDB98; // plane;       
+Sphere16 D_801EDBA8; // sphere;      
+TriNorm D_801EDBB0; // tri;        
+#endif
 
 char D_801ED950[80];
 char D_801ED9A0[80];
@@ -380,6 +389,8 @@ s32 CollisionPoly_CheckZIntersectApprox(CollisionPoly* poly, Vec3s* vtxList, f32
                                             y, zIntersect);
 }
 
+#ifdef NON_MATCHING
+// Matches, but needs in-function static bss
 s32 CollisionPoly_LineVsPoly(BgLineVsPolyTest* a0) {
     static Vec3f polyVerts[3]; // D_801EDB70
     static Plane plane;        // D_801EDB98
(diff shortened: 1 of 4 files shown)


src/code/z_bgcheck.c
```diff
@@ -4425,27 +4427,24 @@ s32 WaterBox_GetSurfaceImpl(PlayState* play, CollisionContext* colCtx, f32 x, f3
...
-                            *outWaterBox = curWaterBox;
-                            *ySurface = curWaterBox->minPos.y;
-                            *bgId = i;
-                            return true;
-                        }
+        if (!(colCtx->dyna.bgActorFlags[i] & 1) || (colCtx->dyna.bgActorFlags[i] & 4)) {
+            continue;
+        }
+        bgActor = &colCtx->dyna.bgActors[i];
+        if (bgActor->colHeader->numWaterBoxes != 0 && bgActor->colHeader->waterBoxes != NULL) {
```
jpburnett, line 4434:
```suggestion
        if ((bgActor->colHeader->numWaterBoxes != 0) && (bgActor->colHeader->waterBoxes != NULL)) {
```

src/code/z_bgcheck.c
```diff
@@ -4425,27 +4427,24 @@ s32 WaterBox_GetSurfaceImpl(PlayState* play, CollisionContext* colCtx, f32 x, f3
...
+            for (curWaterBox = colCtx->dyna.waterBoxList.boxes + bgActor->waterboxesStartIndex;
+                 curWaterBox <
+                 colCtx->dyna.waterBoxList.boxes + bgActor->waterboxesStartIndex + bgActor->colHeader->numWaterBoxes;
+                 curWaterBox++) {
+                if (curWaterBox->properties & 0x80000) {
+                    continue;
+                }
+                if (curWaterBox->minPos.x < x && x < curWaterBox->minPos.x + curWaterBox->xLength) {
```
jpburnett, line 4442:
```suggestion
                if ((curWaterBox->minPos.x < x) && (x < curWaterBox->minPos.x + curWaterBox->xLength)) {
```
Is that the intended way? I don't know if parenthesis can help separate the args to make it clearer

src/code/z_bgcheck.c
```diff
@@ -4425,27 +4427,24 @@ s32 WaterBox_GetSurfaceImpl(PlayState* play, CollisionContext* colCtx, f32 x, f3
...
+                    if (curWaterBox->minPos.z < z && z < curWaterBox->minPos.z + curWaterBox->zLength) {
```
jpburnett, line 4443: same here

jpburnett (commented):
Just some comments on adding some parenthesis to clear up the comparisons

src/code/z_bgcheck.c
```diff
@@ -379,31 +380,41 @@ s32 CollisionPoly_CheckZIntersectApprox(CollisionPoly* poly, Vec3s* vtxList, f32
                                             y, zIntersect);
 }
 
-#ifdef NON_MATCHING
 s32 CollisionPoly_LineVsPoly(BgLineVsPolyTest* a0) {
     static Vec3f polyVerts[3]; // D_801EDB70
     static Plane plane;        // D_801EDB98
```
AngheloAlf, line 385: We don't really have a good way to deal with in-function static bss reordering, which will be a problem in the future, specially with the big amount of bss this file has.
My recommendation is to leave this function as NON_MATCHING with a note saying this function matches but bss

petrie911: Should the same be done with `SphVsPoly`, which already has its static bss in-functioned?

src/code/z_bgcheck.c
petrie911: It's the way I intended it. I thought I caught them all but apparently not.

src/code/z_bgcheck.c
hensldm: > Should the same be done with `SphVsPoly`, which already has its static bss in-functioned?

I think so. I'm kind of surprised it hasn't been troublesome yet.
Note I've usually left the comment `// needs in-function static bss` for functions that match but just need in function bss. That way it is easier to find them later.

EllipticEllipsis (approved):

AngheloAlf (approved):
The bss still worries me, but hopefully it will be fine

src/code/z_bgcheck.c
hensldm: Bump missing, parens.

src/code/z_bgcheck.c
hensldm: Bump on engineer's comment, for these named bss variables, can you prefix `s`

hensldm (commented):
Bump on other's comments

hensldm (approved):

PR #1136, opened 2022-10-22
Introduce `object_table.h`

AngheloAlf:
The follow-up to #985 

This PR also migrates the `gObjectTable` into a C file. OoT seems to have this in the `z_scene.c` file, but MM has a file boundary on the data, so it can't be there. I decided to create a new file `z_scene_dlftbls.c`, inspired by `z_actor_dlftbls.c`


include/tables/object_table.h
```diff
@@ -0,0 +1,652 @@
...
+/* 0x156 */ DEFINE_OBJECT(object_knight, OBJECT_KNIGHT)
+/* 0x157 */ DEFINE_OBJECT(object_icicle, OBJECT_ICICLE)
+/* 0x158 */ DEFINE_OBJECT(object_spdweb, OBJECT_SPDWEB)
+/* 0x159 */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_159)
+/* 0x15A */ DEFINE_OBJECT(object_boss01, OBJECT_BOSS01)
+/* 0x15B */ DEFINE_OBJECT(object_boss02, OBJECT_BOSS02)
+/* 0x15C */ DEFINE_OBJECT(object_boss03, OBJECT_BOSS03)
+/* 0x15D */ DEFINE_OBJECT(object_boss04, OBJECT_BOSS04)
+/* 0x15E */ DEFINE_OBJECT(object_boss05, OBJECT_BOSS05)
+/* 0x15F */ DEFINE_OBJECT_NULL(object_boss07, OBJECT_UNSET_15F)
```
hensldm, line 361: I like OoT's approach here with justi suffixing `UNUSED` and having a 0 size comment, since it gives a little more context.
For example
`/* 0x00FA */ DEFINE_OBJECT_NULL(object_md, OBJECT_MD_UNUSED) // unused duplicate with size 0`

hensldm (approved):
Just 1 thought, though looks great overall :)

include/tables/object_table.h
```diff
@@ -0,0 +1,652 @@
+/**
+ * Object Table
+ *
+ * DEFINE_OBJECT should be used for normal objects
+ *    - Argument 0: Name of the object segment in spec
+ *    - Argument 1: Enum value for this object
+ *
+ * DEFINE_OBJECT_UNSET and DEFINE_OBJECT_NULL are needed to define empty entries from the original game
```
EllipticEllipsis, line 8: IMO this needs to be more specific, especially since one is only used once and it doesn't really do what it sounds like from the name.

include/tables/object_table.h
```diff
@@ -0,0 +1,652 @@
...
+/* 0x06E */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_6E)
+/* 0x06F */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_6F)
+/* 0x070 */ DEFINE_OBJECT(object_efc_star_field, OBJECT_EFC_STAR_FIELD)
+/* 0x071 */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_71)
+/* 0x072 */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_72)
+/* 0x073 */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_73)
+/* 0x074 */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_74)
+/* 0x075 */ DEFINE_OBJECT(object_rd, OBJECT_RD)
+/* 0x076 */ DEFINE_OBJECT(object_yukimura_obj, OBJECT_YUKIMURA_OBJ)
+/* 0x077 */ DEFINE_OBJECT_UNSET(OBJECT_HEAVY_OBJECT)
```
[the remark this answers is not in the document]
AngheloAlf: I think this object exists on jp but not in other releases, could you check?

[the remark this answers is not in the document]
EllipticEllipsis: Ah, it does exist on nj0 and nj1. It is identical to the GC version of OoT's object_heavy_object.

include/tables/object_table.h
AngheloAlf: Added a few extra explanations, is it better?

EllipticEllipsis: Yeah, I think so. I'm a bit torn on NULL: I see that it's for an object that's not in the rom but the name sounds like it would be a null pointer rather than having the beginning and end addresses the same.

AngheloAlf: It is the OoT name. Do you have any better suggestion? maybe `_NO_SIZE`?

EllipticEllipsis: Or `SIZE_ZERO` possibly.

include/tables/object_table.h
```diff
@@ -5,12 +5,12 @@
...
  *    - Argument 1: Enum value for this object
  *
- * DEFINE_OBJECT_UNSET and DEFINE_OBJECT_NULL are needed to define empty entries from the original game
+ * DEFINE_OBJECT_UNSET and DEFINE_OBJECT_SIZE_ZERO are needed to define empty entries from the original game
  * 
  * DEFINE_OBJECT_UNSET is used to declare a placeholder enum value which maps to an entry with zeroes in `gObjectTable`
  *    - Argument 0: Enum value for this object
  * 
- * DEFINE_OBJECT_NULL is used to declare a placeholder enum value which maps to a zero-size entry in `gObjectTable`
+ * DEFINE_OBJECT_SIZE_ZERO is used to declare a placeholder enum value which maps to a zero-size entry in `gObjectTable`
```
EllipticEllipsis, line 13:
```suggestion
 * DEFINE_OBJECT_SIZE_ZERO is used to declare a placeholder enum value which maps to a zero-size entry in `gObjectTable` with no corresponding segment.
```

EllipticEllipsis (approved):

AngheloAlf:
Renamed `z_scene_dlftbls.c` to `object_table.c` following the discussion on Discord

hensldm (approved):

engineer124 (approved):
lgtm

PR #1137, opened 2022-10-22
Add ALIGN16 to GRAPH_ALLOC

hensldm:


src/code/z_scene_proc.c
```diff
@@ -113,7 +113,7 @@ void AnimatedMat_DrawTwoTexScroll(PlayState* play, s32 segment, void* params) {
  * Generates a displaylist that sets the prim and env color, and stores it in the provided segment ID.
  */
 void AnimatedMat_SetColor(PlayState* play, s32 segment, F3DPrimColor* primColorResult, F3DEnvColor* envColor) {
-    Gfx* colorDList = GRAPH_ALLOC(play->state.gfxCtx, sizeof(Gfx) * 4);
+    Gfx* colorDList = GRAPH_ALLOC(play->state.gfxCtx, 3 * sizeof(Gfx));
```
AngheloAlf, line 116: huh

AngheloAlf (approved):
Thanks for doing this

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -658,14 +658,14 @@ void EnBox_PostLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3s* rot,
 }
 
 Gfx* EnBox_SetRenderMode1(GraphicsContext* gfxCtx) {
-    Gfx* dl = GRAPH_ALLOC(gfxCtx, sizeof(Gfx) * 2);
+    Gfx* dl = GRAPH_ALLOC(gfxCtx, sizeof(Gfx));
```
EllipticEllipsis, line 661: Could you change the ones in here to `gfx` and `head` to match elsewhere? Actually should probably make this consistent across the board.

EllipticEllipsis (commented):
Certainly looks a lot better.

src/overlays/actors/ovl_En_Zot/z_en_zot.c
```diff
@@ -1345,7 +1345,7 @@ void EnZot_Update(Actor* thisx, PlayState* play) {
 }
 
 Gfx* func_80B99580(GraphicsContext* gfxCtx) {
-    Gfx* dList = GRAPH_ALLOC(gfxCtx, sizeof(Gfx) * 2);
+    Gfx* dList = GRAPH_ALLOC(gfxCtx, sizeof(Gfx));
```
engineer124, line 1348: In some places, it goes from:
`GRAPH_ALLOC(gfxCtx, sizeof(Gfx) * 2);` -> `GRAPH_ALLOC(gfxCtx, sizeof(Gfx));`
and in others:
`GRAPH_ALLOC(gfxCtx, sizeof(Gfx) * 2);` -> `GRAPH_ALLOC(gfxCtx, 2 * sizeof(Gfx));`
How do you determine which is right? Do they both always match?

EllipticEllipsis: You look at how many displaylists/mtx the rest of the function actually uses to see how much space it needs.

hensldm: Just to expand on what Elliptic said, since the GRAPH_ALLOC now automatically aligns sizes to 16 byte values, and `Gfx` is only 8 Bytes, every odd number of `Gfx`s will also match the next even number above it since they have the same alignment.
i.e 1 -> 2, 3 ->4 etc.

src/code/z_scene_proc.c
hensldm: See https://github.com/zeldaret/mm/pull/1137#discussion_r1002562044, this only uses 3 graphics commands, but due to alignment this is the same as allocating for 4 commands.

src/code/z_eff_spark.c
```diff
@@ -162,7 +162,7 @@ void EffectSpark_Draw(void* thisx, GraphicsContext* gfxCtx) {
         gSPSetGeometryMode(POLY_XLU_DISP++, G_ZBUFFER | G_SHADE | G_SHADING_SMOOTH);
         gDPPipeSync(POLY_XLU_DISP++);
 
-        vertices = GRAPH_ALLOC(gfxCtx, ALIGN16(this->numElements * sizeof(Vtx[4])));
+        vertices = GRAPH_ALLOC(gfxCtx, this->numElements * sizeof(Vtx[4]));
```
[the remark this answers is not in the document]
hensldm: So it actually doesn't match exactly as you have it, it needs to have `4 * sizeof(Vtx)` as part of 1 computation before `this->numElements` gets multiplied. So to do it like this it would have to be
`vertices = GRAPH_ALLOC(gfxCtx, this->numElements * (4 * sizeof(Vtx)));`
Is that fine?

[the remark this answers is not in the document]
EllipticEllipsis: Seems fine to me, certainly better than current.

src/overlays/actors/ovl_En_Box/z_en_box.c
hensldm: Went with `gfx` and `gfxHead`

src/code/z_actor.c
```diff
@@ -1721,16 +1721,14 @@ void func_800B8050(Actor* actor, PlayState* play, s32 flag) {
...
+        Gfx* gfxHead = gfx;
 
         OPEN_DISPS(play->state.gfxCtx);
 
-        gDPSetHilite1Tile(displayListHead++, 1, hilite, 0x10, 0x10);
-        gSPEndDisplayList(displayListHead);
-        gSPSegment(POLY_OPA_DISP++, 0x07, displayList);
+        gDPSetHilite1Tile(gfxHead++, 1, hilite, 0x10, 0x10);
+        gSPEndDisplayList(gfxHead++);
+        gSPSegment(POLY_OPA_DISP++, 0x07, gfx);
```
EllipticEllipsis, line 1731: This looks backwards to me, `head` should be the one that doesn't move and is returned?

hensldm: Yeah that does make more sense.

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -874,20 +874,20 @@ void EnRr_Update(Actor* thisx, PlayState* play) {
...
-    Mtx* matrix;
+    Mtx* mtx;
     Vec3f* vecPtr;
     s32 i;
     EnRrStruct* ptr;
     Vec3f spA4;
     f32 temp_f20;
 
-    matrix = GRAPH_ALLOC(play->state.gfxCtx, sizeof(Mtx) * 4);
+    mtx = GRAPH_ALLOC(play->state.gfxCtx, 4 * sizeof(Mtx));
```
EllipticEllipsis, line 884: Can be defined where declared.

src/overlays/actors/ovl_Boss_02/z_boss_02.c
```diff
@@ -1336,8 +1336,8 @@ void Boss02_Twinmold_Draw(Actor* thisx, PlayState* play2) {
     Boss02* this = THIS;
     s32 i;
     s32 idx;
-    Mtx* mtx;
-    Mtx* mtxIter;
+    Mtx* mtx = GRAPH_ALLOC(play->state.gfxCtx, 23 * sizeof(Mtx));
+    Mtx* mtxIter = mtx;
```
EllipticEllipsis, line 1340: Maybe this should be `mtxHead` and `mtx` too?

EllipticEllipsis (approved):

jpburnett (approved):
I have one word for this change......
approve!

PR #1146, opened 2022-11-13
Add original asset names to all `object_a*` files

tom-overton:
For a long time, most people have just asked me what the original asset names from MM3D are, rather than finding the names themselves. This is, overall, fine, but it has a "bus factor" of 1; if I'm not around, you can't get the names from me. So I've decided to just put the original asset names in for *every* object so that people don't have to rely on me.

This PR does every single object whose name starts with `object_a*`. Thanks, as always, to @emilybrooks for patiently answering my Japanese questions.

### DIFF
--- a/assets/xml/objects/object_an1.xml
+++ b/assets/xml/objects/object_an1.xml
@@ -1,16 +1,16 @@
 ﻿<Root>
     <File Name="object_an1" Segment="6">
-        <Animation Name="object_an1_Anim_001090" Offset="0x1090" />
-        <Animation Name="object_an1_Anim_00144C" Offset="0x144C" />
-        <Animation Name="object_an1_Anim_001E74" Offset="0x1E74" />
-        <Animation Name="object_an1_Anim_0026B4" Offset="0x26B4" />
-        <Animation Name="object_an1_Anim_00341C" Offset="0x341C" />
-        <Animation Name="object_an1_Anim_003EA8" Offset="0x3EA8" />
-        <Animation Name="object_an1_Anim_00544C" Offset="0x544C" />
-        <Animation Name="object_an1_Anim_0065C8" Offset="0x65C8" />
-        <Animation Name="object_an1_Anim_0071E8" Offset="0x71E8" />
-        <Animation Name="object_an1_Anim_007E08" Offset="0x7E08" />
-        <Animation Name="object_an1_Anim_008B6C" Offset="0x8B6C" />
+        <Animation Name="object_an1_Anim_001090" Offset="0x1090" /> <!-- Original name is "an_ajimi" ("tasting; sampling​") -->
+        <Animation Name="object_an1_Anim_00144C" Offset="0x144C" /> <!-- Original name is "an_ajimiend" -->
+        <Animation Name="object_an1_Anim_001E74" Offset="0x1E74" /> <!-- Original name is "an_aogu" ("to look up") -->
+        <Animation Name="object_an1_Anim_0026B4" Offset="0x26B4" /> <!-- Original name is "an_bikkuri" ("to be surprised") -->
+        <Animation Name="object_an1_Anim_00341C" Offset="0x341C" /> <!-- Original name is "an_obonwait" ("obon" = "tray") -->
+        <Animation Name="object_an1_Anim_003EA8" Offset="0x3EA8" /> <!-- Original name is "an_obonwalk" -->
+        <Animation Name="object_an1_Anim_00544C" Offset="0x544C" /> <!-- Original name is "an_ojigit" ("bow; bowing") -->
+        <Animation Name="object_an1_Anim_0065C8" Offset="0x65C8" /> <!-- Original name is "an_ryouri" ("cooking") -->
+        <Animation Name="object_an1_Anim_0071E8" Offset="0x71E8" /> <!-- Original name is "an_suwaru" ("to sit") -->
+        <Animation Name="object_an1_Anim_007E08" Offset="0x7E08" /> <!-- Original name is "an_suzetsubou" ("su" = short for "suwaru", "zetsubou" = "despair; hopelessness​") -->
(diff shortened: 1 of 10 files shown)


hensldm (approved):
Thanks for doing this :)

assets/xml/objects/object_an2.xml
```diff
@@ -1,12 +1,12 @@
 ﻿<Root>
     <File Name="object_an2" Segment="6">
         <DList Name="object_an2_DL_000370" Offset="0x370" />
-        <DList Name="object_an2_DL_000378" Offset="0x378" />
+        <DList Name="object_an2_DL_000378" Offset="0x378" /> <!-- Original name is "an_kasa_model" ("umbrella; parasol") -->
```
AngheloAlf, line 4: interesting the animation is not prefixed with `an2`

AngheloAlf: same for `an3` and `an4`

assets/xml/objects/object_an1.xml
```diff
@@ -52,15 +52,15 @@
         <Texture Name="object_an1_Tex_0103A0" OutName="tex_0103A0" Format="ci8" Width="32" Height="32" Offset="0x103A0" />
         <!-- <Blob Name="object_an1_Blob_0107A0" Size="0xA40" Offset="0x107A0" /> -->
         <DList Name="object_an1_DL_0111E0" Offset="0x111E0" />
-        <DList Name="object_an1_DL_0111E8" Offset="0x111E8" />
+        <DList Name="object_an1_DL_0111E8" Offset="0x111E8" /> <!-- Original name is "an_obon_model" -->
```
AngheloAlf, line 55: I thought we didn't have the original names of dlists

AngheloAlf (approved):
Thanks a lot for doing this

assets/xml/objects/object_an1.xml
tom-overton: Depends on what "type" of DList it is. For DLists that are part of a skeleton? Nope, they're completely gone. So in this object, for example, all of Anju's limb DLists are just part of `ange.cmb`; I'm 99% sure Grezzo came up with this name because it doesn't have the same prefix as her animations. However, DLists that *aren't* part of a skeleton usually *do* have names. Since Anju's tray and chopsticks aren't part of her skeleton, they are named individually; they even match the `an_whatever` naming scheme.

There are some exceptions to this, mainly around DLists that *weren't* part of the skeleton in MM64 but *are* part of the skeleton in MM3D. To give an example, MM64 Skull Kid has swappable hand DLists that he can use to pose his hands in certain ways. In MM3D, these are all just part of his skeleton, so we "lost" these names.

Does that make sense? It's a bit more of an art than a science, I'll admit, but we do have what are seemingly the original names of *some* DLists.

assets/xml/objects/object_an2.xml
tom-overton: I observed this with some objects in OoT too, mainly the various `ganon_anime`s. While they certainly use a "prefix + name" scheme for most assets, that prefix is definitely *not* always based on the object name, or even *unique* to that object! In this case, it looks like they used `anj` as a prefix for the animations that are used when Anju and Kafei finally meet up, and `an` for everything else. I don't know why they decided to make those specific animations have a separate prefix, but they did.

assets/xml/objects/object_an1.xml
AngheloAlf: Ohhh, very interesting. Yes it makes sense.
Thanks for the explanation

assets/xml/objects/object_an2.xml
tom-overton: Oh, and one important thing I forgot to mention; in MM3D, `an2`, `an3`, and `an4` simply don't exist. They are all combined into `an1`. Something similar happens when the `ganon_anime`s in OoT3D, though bizarrely they *did* create a brand-new object called `zelda_ganon_demo` *just* to handle the Ganondorf that appears before the Ganon fight. So I think in cases like Anju, Ganondorf, or Skull Kid, where the assets get spread across multiple objects, they conceptually treated them like one big object, and then split things out afterwards. Grezzo simply came in and combined them again.

assets/xml/objects/object_an1.xml
tom-overton: Sometimes, we get objects like Anju's, where most of the things she "holds onto" in-game are actually separate models. It looks like only her mask was moved to her hand:
image

For other objects, like Human Link, basically *all* of his separate DLists were moved into his skeleton, depriving us of a ton of original names:
image

EllipticEllipsis (approved):

assets/xml/objects/object_ani.xml
```diff
@@ -53,15 +53,15 @@
...
         <!-- Only the first half of the animation is used by EnAni when knocked out of tree -->
-        <Animation Name="gAniLandingThenStandingUpAnim" Offset="0x9220" />
+        <Animation Name="gAniLandingThenStandingUpAnim" Offset="0x9220" /> <!-- Original name is "Ani_oriru" ("to go down; to come down​") -->
         <!-- used by EnOssan -->
-        <Animation Name="gAniHoldUpHandAnim" Offset="0x9734" />
-        <Animation Name="gAniStandingNormalAnim" Offset="0x9D34" />
-        <Animation Name="gAniHandBehindHeadApologyAnim" Offset="0xA460" />
+        <Animation Name="gAniHoldUpHandAnim" Offset="0x9734" /> <!-- Original name is "Ani_talk" -->
+        <Animation Name="gAniStandingNormalAnim" Offset="0x9D34" /> <!-- Original name is "Ani_wait" -->
+        <Animation Name="gAniHandBehindHeadApologyAnim" Offset="0xA460" /> <!-- Original name is "Ani_zorawait" -->
```
isghj5, line 60: Oh huh, neat

PR #1151, opened 2022-11-16
Use pointer types instead of s32

louist103:
Some of these were missed in the original decomp.
I ignored any warnings related to `uintptr_t` -> `void*` without a cast.
I also wasn't sure how to handle the segment table.


AngheloAlf (approved):
It is very annoying that CC_CHECK didn't catch those. I wonder if there's a way to force CC_CHECK to actually catch them

hensldm (approved):

EllipticEllipsis (approved):

PR #1159, opened 2022-12-04
z_syscfb.c, z_vimode.c, code_80140CE0 split and OK

mzxrules:
supercedes #1025


src/code/sys_cfb.c
```diff
@@ -0,0 +1,125 @@
+#include "global.h"
+
+OSViMode sNotebookViMode;
```
engineer124, line 3:
```suggestion
OSViMode sBombersNotebookViMode;
```
As in the Bombers Notebook?

src/code/code_80140CE0.c
```diff
@@ -0,0 +1,86 @@
+#include "global.h"
+
+extern Gfx D_0E0001C8[];
+
+Gfx D_801C5DD0[] = {
+    gsDPSetOtherMode(G_AD_PATTERN | G_CD_MAGICSQ | G_CK_NONE | G_TC_CONV | G_TF_POINT | G_TT_NONE | G_TL_TILE |
+                         G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
+                     G_AC_NONE | G_ZS_PRIM | G_RM_VISCVG | G_RM_VISCVG2),
+    gsSPBranchList(0x0E0002E0),
```
engineer124, line 9:
```suggestion
    gsSPBranchList(D_0E000000.fillRect),
```
All uses of `0x0E0002E0` in this file should be this

src/code/code_80140CE0.c
```diff
@@ -0,0 +1,86 @@
+#include "global.h"
+
+extern Gfx D_0E0001C8[];
```
engineer124, line 3: This extern can go and its usage can be replace by `D_0E000000.setScissor`

src/code/code_80140CE0.c
```diff
@@ -0,0 +1,86 @@
...
+    gsSPDisplayList(0x0E0002E0),
+    gsDPSetOtherMode(G_AD_PATTERN | G_CD_MAGICSQ | G_CK_NONE | G_TC_CONV | G_TF_POINT | G_TT_NONE | G_TL_TILE |
+                         G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
+                     G_AC_NONE | G_ZS_PRIM | IM_RD | CVG_DST_CLAMP | ZMODE_OPA | FORCE_BL |
+                         GBL_c1(G_BL_CLR_IN, G_BL_0, G_BL_CLR_MEM, G_BL_A_MEM) |
+                         GBL_c2(G_BL_CLR_IN, G_BL_0, G_BL_CLR_MEM, G_BL_A_MEM)),
+    gsSPBranchList(0x0E0002E0),
+};
+
+void func_80140CE0(struct_801F8010* arg0) {
```
engineer124, line 44: This entire file looks like `z_viscvg.c` from the OoT PR:
https://github.com/zeldaret/oot/pull/1374/files

So there are good docs available for this file that we can use, might want to coordinate with Elliptic on this

src/code/code_80140CE0.c
```diff
@@ -0,0 +1,86 @@
...
+
+    if (arg0->setScissor == true) {
+        gSPDisplayList(gfx++, D_0E0001C8);
+    }
+
+    switch (arg0->type) {
+        case 1:
+            gSPDisplayList(gfx++, D_801C5DF0);
+            break;
+        case 2:
```
engineer124, line 70:
```suggestion
        case 1:
            gSPDisplayList(gfx++, D_801C5DF0);
            break;
       
        case 2:
```
Spaces after breaks is the style we're going for. Also add a default case (assuming it still matches)

src/code/z_vimode.c
```diff
@@ -1,23 +1,346 @@
...
+                BURST(57, 30, 5, 70),
+                VSYNC(524),
+                HSYNC(3088, 0),
+                LEAP(3100, 3100),
+                HSTART(108, 748),
+                START(37, 511),
+                BURST(4, 2, 14, 0),
+            },
+        };
+        const ViModeStruct* ptr = &D_801DF0C0[tvType];
```
engineer124, line 113: Does this data need to be in-function to match? It breaks of the flow of the function

src/code/z_vimode.c
```diff
@@ -1,23 +1,346 @@
...
+    R_VI_MODE_EDIT_ULX_ADJ = viMode->leftAdjust;
+    R_VI_MODE_EDIT_LRX_ADJ = viMode->rightAdjust;
+
+    if (SREG(58) == 1) {
+        SREG(58) = 0;
+        switch (SREG(59)) {
+            case 1:
+                ViMode_LogPrint(&osViModePalLan1);
+                break;
+            case 2:
```
engineer124, line 175:
```suggestion
            case 1:
                ViMode_LogPrint(&osViModePalLan1);
                break;
               
            case 2:
```

src/code/z_vimode.c
```diff
@@ -1,23 +1,346 @@
...
+                break;
+            default:
+                ViMode_LogPrint(&viMode->customViMode);
+                break;
+        }
+    }
+}
+
+void ViMode_Load(ViMode* viMode) {
+    if ((R_VI_MODE_EDIT_WIDTH & ~3) == 1) {
```
engineer124, line 186:
```suggestion
void ViMode_Load(ViMode* viMode) {
    //! @bug This condition never passes as the lowest bit is always masked out to 0
    if ((R_VI_MODE_EDIT_WIDTH & ~3) == 1) {
```
Seems like this bug carried over to MM

src/code/z_vimode.c
```diff
@@ -1,23 +1,346 @@
...
+        ctrl |= OS_VI_DIVOT;
+    }
+    if (viFeatures & OS_VI_DIVOT_OFF) {
+        ctrl &= ~OS_VI_DIVOT;
+    }
+    viMode->customViMode.comRegs.ctrl = ctrl;
+}
+
+// This function uses controller input (C buttons + D pad) to reconfigure the custom VI mode
+void ViMode_Update(ViMode* viMode, Input* input) {
```
engineer124, line 245:
```suggestion
/** 
 * This function uses controller input (C buttons + D pad) to reconfigure the custom VI mode
 */
void ViMode_Update(ViMode* viMode, Input* input) {
```
This is the style we use for function descriptions. Applies to a few places.

Also, I think OoT comments are good, noticed some are removed when the functions seems identical

src/code/z_view.c
```diff
@@ -18,8 +18,8 @@ void View_ViewportToVp(Vp* dest, Viewport* src) {
     dest->vp.vtrans[2] = 0x01FF;
     dest->vp.vtrans[3] = 0;
 
-    if ((src->leftX == 0) && (src->rightX == SCREEN_WIDTH_HIGH_RES) && (src->topY == 0) &&
-        (src->bottomY == SCREEN_HEIGHT_HIGH_RES)) {}
+    if ((src->leftX == 0) && (src->rightX == SCREEN_WIDTH_NOTEBOOK) && (src->topY == 0) &&
+        (src->bottomY == SCREEN_HEIGHT_NOTEBOOK)) {}
```
engineer124, line 22:
```suggestion
        (src->bottomY == SCREEN_HEIGHT_BOMBERS_NOTEBOOK)) {}
```

src/code/sys_cfb.c
```diff
@@ -0,0 +1,125 @@
...
+    gCfbUpperAdjust = 10;
+    gScreenWidth = gCfbWidth;
+    gScreenHeight = gCfbHeight;
+    if ((gCfbWidth == SCREEN_WIDTH_HIRES) && (gCfbHeight == SCREEN_HEIGHT_HIRES)) {
+        gActiveViMode = &osViModeNtscHpf1;
+    } else {
+        s32 rightAdjust;
+        s32 lowerAdjust;
+
+    l1:
```
engineer124, line 69:
```suggestion
```
Is this needed?

src/code/sys_cfb.c
```diff
@@ -0,0 +1,125 @@
...
+        rightAdjust = gCfbWidth - 610;
+        lowerAdjust = gCfbHeight - 470;
+        ViMode_Configure(&sNotebookViMode, -1, osTvType, 0, 1, 0, 1, gCfbWidth, gCfbHeight, 30, rightAdjust, 10,
+                         lowerAdjust);
+        gActiveViMode = &sNotebookViMode;
+    }
+    gSysCfbHiResEnabled = true;
+}
+
+// TODO Fake match
```
engineer124, line 79:
```suggestion
//! FAKE:
```
This is the label for fake matches

src/code/sys_cfb.c
```diff
@@ -0,0 +1,125 @@
...
+
+u16* SysCfb_GetZBuffer(void) {
+    return gZBufferPtr;
+}
+
+void* SysCfb_GetWorkBuffer(void) {
+    return gWorkBuffer;
+}
+
+u16 func_80178A34(s32 x, s32 y) { // Get ZBuffer pixel
```
engineer124, line 110:
```suggestion
u16 SysCfb_GetPixelDepth(s32 x, s32 y) {
```
Possibly?

src/code/sys_cfb.c
```diff
@@ -0,0 +1,125 @@
...
+        val = (&zBuff[x])[y * gCfbWidth];
+    } else {
+        val = 0;
+    }
+    return val;
+}
+
+s32 func_80178A94(s32 x, s32 y) {
+    return func_800F50D4(func_80178A34(x, y) * 4) >> 3;
+}
```
engineer124, line 125: For reference, `func_800F50D4` is OoT's `Environment_ZBufValToFixedPoint`

src/code/sys_cfb.c
```diff
@@ -0,0 +1,125 @@
...
+    SysCfb_SetLoResMode();
+}
+
+// Unused
+void SysCfb_Reset(void) {
+    gFramebuffers[0] = 0;
+    gFramebuffers[1] = 0;
+}
+
+uintptr_t SysCfb_GetFbPtr(s32 index) {
```
engineer124, line 95: I don't think this has been resolved yet, but I'm in favor of `Fb` -> `framebuffer` and `Cfb` -> `colorframebuffer`. Doesn't have to be in the scope of this PR though

src/code/sys_cfb.c
mzxrules: I think it's also used for Granny's stories. this is more just a placeholder name

src/code/sys_cfb.c
mzxrules: unfortunately yes. Is there a better name for this?

engineer124: I don't think the name is important, but would be good to add a `//! FAKE:` label. These kind of labels in-function usually indicates there's a scoping issue with the function

src/code/z_vimode.c
mzxrules: Yes. It's a refactoring from OoT that eliminates an if statement

src/code/sys_cfb.c
```diff
@@ -0,0 +1,126 @@
...
+    SysCfb_SetLoResMode();
+}
+
+// Unused
+void SysCfb_Reset(void) {
+    gFramebuffers[0] = NULL;
+    gFramebuffers[1] = NULL;
+}
+
+uintptr_t SysCfb_GetFramebuffer(s32 index) {
```
hensldm, line 96:
```suggestion
void* SysCfb_GetFramebuffer(s32 index) {
```
Don't really see a reason why we need to return a `uintptr_t` vs just a `void*`

src/code/sys_cfb.c
```diff
@@ -0,0 +1,126 @@
...
+
+void* SysCfb_GetWorkBuffer(void) {
+    return gWorkBuffer;
+}
+
+u16 SysCfb_GetZBufferPixel(s32 x, s32 y) {
+    u16* zBuff;
+    u16 val;
+
+    zBuff = SysCfb_GetZBuffer();
```
hensldm, line 115:
```suggestion
    u16* zBuff = SysCfb_GetZBuffer();
    u16 pixel;
    
```

include/z64.h
```diff
@@ -1126,4 +1126,29 @@ typedef struct {
...
+    /* 0x0058 */ s32 rightAdjust;
+    /* 0x005c */ s32 leftAdjust;
+    /* 0x0060 */ s32 lowerAdjust;
+    /* 0x0064 */ s32 upperAdjust;
+    /* 0x0068 */ s32 editState;
+    /* 0x006C */ s32 tvType;
+    /* 0x0070 */ u32 loRes;
+    /* 0x0074 */ u32 antialiasOff;
+    /* 0x0078 */ u32 modeN;
+    /* 0x007c */ u32 fb16Bit;
```
hensldm, line 1149:
```suggestion
    /* 0x007C */ u32 fb16Bit;
```
Also can remove the leading `00`s

include/z64.h
```diff
@@ -1126,4 +1126,29 @@ typedef struct {
...
+    /* 2 */ VI_MODE_EDIT_STATE_2, // active, more adjustments
+    /* 3 */ VI_MODE_EDIT_STATE_3  // active, more adjustments, print comparison with NTSC LAN1 mode
+} ViModeEditState;
+
+typedef struct {
+    /* 0x0000 */ OSViMode customViMode;
+    /* 0x0050 */ s32 viHeight;
+    /* 0x0054 */ s32 viWidth;
+    /* 0x0058 */ s32 rightAdjust;
+    /* 0x005c */ s32 leftAdjust;
```
hensldm, line 1141:
```suggestion
    /* 0x005C */ s32 leftAdjust;
```

include/regs.h
```diff
@@ -113,6 +113,14 @@
...
 #define R_DGN_MINIMAP_Y                   WREG(69)
 
+#define R_VI_MODE_EDIT_STATE        SREG(48)
+#define R_VI_MODE_EDIT_WIDTH        SREG(49)
+#define R_VI_MODE_EDIT_HEIGHT       SREG(50)
+#define R_VI_MODE_EDIT_ULY_ADJ      SREG(51)
+#define R_VI_MODE_EDIT_LRY_ADJ      SREG(52)
+#define R_VI_MODE_EDIT_ULX_ADJ      SREG(53)
+#define R_VI_MODE_EDIT_LRX_ADJ      SREG(54)
+
```
engineer124, line 123: These SREG`s can be moved higher in the file, next to the other SREG's

include/functions.h
```diff
@@ -1322,7 +1322,7 @@ void Font_LoadCharNES(PlayState* play, u8 codePointIndex, s32 offset);
 void Font_LoadMessageBoxEndIcon(Font* font, u16 icon);
 void Font_LoadOrderedFont(Font* font);
 // void func_800F5090(void);
-// void func_800F50D4(void);
+s32 Environment_ZBufValToFixedPoint(s32);
```
engineer124, line 1325:
```suggestion
s32 Environment_ZBufValToFixedPoint(s32 zBufferVal);
```

include/z64.h
```diff
@@ -1014,10 +1014,10 @@ typedef struct {
...
 typedef struct {
-    /* 0x0 */ u32 type;
-    /* 0x4 */ u32 setScissor;
-    /* 0x8 */ Color_RGBA8 color;
-    /* 0xC */ Color_RGBA8 envColor;
+    /* 0x00 */ u32 type;
+    /* 0x04 */ u32 setScissor;
+    /* 0x08 */ Color_RGBA8_u32 color;
+    /* 0x0C */ Color_RGBA8 envColor;
 } struct_801F8010; // size = 0x10
```
engineer124, line 1021: The leading 0's can go. I think we agreed on a style for consistency to have the number of 0's be based on the struct members, not the total struct size

include/variables.h
```diff
@@ -2679,31 +2679,31 @@ extern volatile OSTime sRSPOtherTotalTime;
...
-// extern UNK_TYPE1 D_801FBBE0;
+extern void* gGfxSPTaskOutputBufferSize;
+extern void* sCfbLoRes1;
+extern void* sCfbLoRes0;
+extern u16 (*gZBufferLoRes)[SCREEN_WIDTH * SCREEN_HEIGHT];
+extern u16 (*gWorkBufferLoRes)[SCREEN_WIDTH * SCREEN_HEIGHT];
+extern u64 (*gGfxSPTaskOutputBufferLoRes)[0x3000];
+extern void* gGfxSPTaskOutputBufferSizeLoRes;
+extern void* sCfbHiRes1;
+extern void* sCfbHiRes0;
```
engineer124, line 2696: I think some of these externs are local. Could you remove the externs that are local to minimize changes to bss?

src/code/sys_cfb.c
```diff
@@ -0,0 +1,126 @@
...
+    if (zBuff != NULL) {
+        val = (&zBuff[x])[y * gCfbWidth];
+    } else {
+        val = 0;
+    }
+    return val;
+}
+
+s32 SysCfb_GetZBufferInt(s32 x, s32 y) {
```
engineer124, line 124:
```suggestion
s32 SysCfb_GetZBufferToInt(s32 x, s32 y) {
```
?

mzxrules: Felt it didn't make much sense to use `To` here because typically we use "A to B" naming for conversions; here x and y are used strictly for retrieving data.

include/variables.h
mzxrules: Removing the externs results in non-matching, as doing so affects the .bss order of sys_cfb's variables.

src/code/sys_cfb.c
```diff
@@ -1,20 +1,128 @@
...
+void* gGfxSPTaskOutputBufferSizeHiRes; // Actually points to the end of the task buffer
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cfb/SysCfb_GetFbPtr.s")
+s16 gCfbWidth;
+s16 gCfbHeight;
+s16 gCfbLeftAdjust;
+s16 gCfbUpperAdjust;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cfb/SysCfb_GetZBuffer.s")
+u8 gSysCfbHiResEnabled;
```
AngheloAlf, line 33: Considering the high amount of bss in this file we may prefer to not import the bss yet to avoid headaches in the future

src/code/sys_cfb.c
```diff
@@ -1,20 +1,128 @@
 #include "global.h"
 #include "system_malloc.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cfb/func_80178750.s")
+#define D_80780000 gPictoPhotoI8
```
AngheloAlf, line 4: What's the point of this define exactly?

src/code/sys_cfb.c
```diff
@@ -1,20 +1,128 @@
...
+    gSysCfbHiResEnabled = true;
+}
+
+//! FAKE:
+extern u16 gFramebuffer1_[SCREEN_WIDTH_NOTEBOOK][SCREEN_HEIGHT_NOTEBOOK];
+void SysCfb_Init(void) {
+    sCfbLoRes1 = gFramebuffer1;
+    sCfbLoRes0 = gFramebuffer0;
+    sCfbHiRes1 = gFramebuffer1_;
+    sCfbHiRes0 = D_80780000;
```
AngheloAlf, line 89: Do you know what the 0 and 1 are?
If I had to guess it seems like 0 points to the start of the buffer and 1 points to the end of the buffer. Maybe We could  improve symbols here based on that assumption?

include/ultra64/viint.h
```diff
@@ -0,0 +1,37 @@
+#ifndef ULTRA64_VIINT_H
```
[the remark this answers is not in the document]
mzxrules: Copypasta'd from OoT decomp, so the source would be from there.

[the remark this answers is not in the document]
hensldm: Looks like it is in https://github.com/decompals/ultralib/blob/main/src/io/viint.h

src/code/sys_cfb.c
mzxrules: Probably for my own temporary usage. `D_80780000` had gotten renamed but it doesn't make sense to use the name `gPictoPhotoI8` within the context of this file.

engineer124: Then either `gPictoPhotoI8` should be renamed or `gPictoPhotoI8` should be used. Perhaps `gPictoPhotoI8` could be changed to `D_80780000 ` until we see more. But isn't the picto photo is based on frame buffers though?

src/code/sys_cfb.c
mzxrules: I assume you're asking about gFramebuffer0 and gFramebuffer1. Best way to move forward on this is to look at how everything is getting mapped to memory

Quick address map:

* `80000500` is a buffer for sCfbLoRes1 and sCfbHiRes1
* `80025D00` is a buffer for the skybox data
* `80780000` is a buffer for sCfbHiRes0 and gPictoPhotoI8
* `80784600` is a buffer for various effects (lens, maybe the screen shrink effect)
* `807DA800` maps to sCfbLoRes0

The lores color framebuffer is 320 x 240 pixels, 0x25800 bytes
The highres color framebuffer is 576 x 454 pixels, 0x7FB00 bytes
The pictobox photo texture buffer is 160 x 112 pixels, 0x4600 bytes
The buffer following gPictoPhotoI8 is defined to be 0x56200 bytes

If you look at the big picture, essentially you have a 0x7FB00 block of memory at `80000500` to `80080000` and a 0x80000 block of memory at `80780000` to `80800000`. So the names `gFramebuffer0` and `gFramebuffer1` are aren't correct for universal usage like it is atm.

include/ultra64/viint.h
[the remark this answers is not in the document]
AngheloAlf: ohh, the `src` folder, of course

src/code/sys_cfb.c
AngheloAlf: No, I was asking about `sCfbLoRes1` and `sCfbLoRes0`

AngheloAlf: Also, are you saying `D_80780000` should be a `0x7FB00` long buffer? Because if that's the case then that buffer would overlap with `gFramebuffer0`, which is at `0x807DA800`
To avoid overlap that buffer must be at most `0x5A800`. This would allow for a buffer which is at most 497×372 pixels
What exactly makes you think this framebuffer should be 576 x 454 pixels?

mzxrules: Cfb is the abbreviation for color framebuffer, so these are the two framebuffers used during regular play; the 0 and 1 just indicates what buffer index they are assigned to in gFramebuffers.

Low resolution and high resolution modes are mutually exclusive, so to save ram the high res framebuffers overlap the other buffers.

> Also, are you saying D_80780000 should be a 0x7FB00 long buffer?

Not quite, I'm saying it's an 0x80000 long buffer, since when you're in the game normally you have the 0x4600, 0x56200, and 0x25800 buffers reserved there.

> What exactly makes you think this framebuffer should be 576 x 454 pixels?

`SysCfb_SetHiResMode` configures high resolution mode; within it it sets gCfbWidth, gCfbHeight to 576 by 454

AngheloAlf: Then I propose to do this then:
```suggestion
    sCfbLoRes1 = gFramebuffer1;
    sCfbLoRes0 = gFramebuffer0;
    sCfbHiRes1 = gFramebufferHiRes1;
    sCfbHiRes0 = gFramebufferHiRes0;
```
And add those two extra symbols to undefined_syms.txt
```
gFramebufferHiRes1 = 0x80000500;
gFramebufferHiRes0 = 0x80780000;
```

src/code/sys_cfb.c
```diff
@@ -2,20 +2,127 @@
...
+u64* gGfxSPTaskOutputBufferPtr;
+void* gGfxSPTaskOutputBufferSize; // Actually points to the end of the task buffer
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_cfb/func_80178978.s")
+void* sCfbLoRes1;
+void* sCfbLoRes0;
+u16 (*gZBufferLoRes)[SCREEN_WIDTH * SCREEN_HEIGHT];
+u16 (*gWorkBufferLoRes)[SCREEN_WIDTH * SCREEN_HEIGHT];
+u64 (*gGfxSPTaskOutputBufferLoRes)[0x3000];
+void* gGfxSPTaskOutputBufferSizeLoRes; // Actually points to the end of the task buffer
```
hensldm, line 21: Perhaps instead of size just use end (same for the HiRes version)?
`gGfxSPTaskOutputBufferLoResEnd`

src/code/z_play.c
```diff
@@ -2252,10 +2252,10 @@ void Play_Init(GameState* thisx) {
     R_PICTO_PHOTO_STATE = PICTO_PHOTO_STATE_OFF;
 
     PreRender_Init(&this->pauseBgPreRender);
-    PreRender_SetValuesSave(&this->pauseBgPreRender, D_801FBBCC, D_801FBBCE, NULL, NULL, NULL);
-    PreRender_SetValues(&this->pauseBgPreRender, D_801FBBCC, D_801FBBCE, NULL, NULL);
+    PreRender_SetValuesSave(&this->pauseBgPreRender, gCfbWidth, gCfbHeight, NULL, NULL, NULL);
+    PreRender_SetValues(&this->pauseBgPreRender, gCfbWidth, gCfbHeight, NULL, NULL);
 
-    this->unk_18E64 = D_801FBB90;
+    this->unk_18E64 = gWorkBuffer;
```
hensldm, line 2258:
```suggestion
    this->workBuffer = gWorkBuffer;
```
Should be able to rename this now.

src/code/z_viscvg.c
```diff
@@ -0,0 +1,90 @@
...
+    this->color.r = 255;
+    this->color.g = 255;
+    this->color.b = 255;
+    this->color.a = 255;
+}
+
+void VisCvg_Destroy(struct_801F8010* this) {
+}
+
+void VisCvg_Draw(struct_801F8010* this, Gfx** gfxp) {
```
hensldm, line 54: Could rename the struct now, but I'm fine leaving that for when we clean up the 3 Vis's (VisMono, VisCvg, VisZbuf) if you prefer.

mzxrules: I'd rather have someone else take over as I'm a bit out of my depth here

src/code/sys_cfb.c
hensldm: Actually looks like that is a libultra naming thing, so is probably fine as is.

hensldm (approved):
LGTM :)

src/code/sys_cfb.c
hensldm: Tharo mentioned he would prefer `End` over `Size` https://discord.com/channels/688807550715560050/688851976746041391/1098439376561782856 .
I'm kind of indifferent, but do see his point.

include/variables.h
engineer124: You should be able to use the bss scripts to resolve it?

include/regs.h
```diff
@@ -65,6 +65,15 @@ extern RegEditor* gRegEditor;
...
+#define R_VI_MODE_EDIT_STATE              SREG(48)
+#define R_VI_MODE_EDIT_WIDTH              SREG(49)
+#define R_VI_MODE_EDIT_HEIGHT             SREG(50)
+#define R_VI_MODE_EDIT_ULY_ADJ            SREG(51)
+#define R_VI_MODE_EDIT_LRY_ADJ            SREG(52)
+#define R_VI_MODE_EDIT_ULX_ADJ            SREG(53)
+#define R_VI_MODE_EDIT_LRX_ADJ            SREG(54)
+
 #define R_FB_FILTER_TYPE                  SREG(80)
```
engineer124, line 77:
```suggestion
#define R_UPDATE_RATE                     SREG(30)
#define R_VI_MODE_EDIT_STATE              SREG(48)
#define R_VI_MODE_EDIT_WIDTH              SREG(49)
#define R_VI_MODE_EDIT_HEIGHT             SREG(50)
#define R_VI_MODE_EDIT_ULY_ADJ            SREG(51)
#define R_VI_MODE_EDIT_LRY_ADJ            SREG(52)
#define R_VI_MODE_EDIT_ULX_ADJ            SREG(53)
#define R_VI_MODE_EDIT_LRX_ADJ            SREG(54)
#define R_FB_FILTER_TYPE                  SREG(80)
```
Might be fine as is, but I think spacing is to break up the different letter'd sregs instead of by function

include/variables.h
mzxrules: I don't really see how, since it's sys_cfb's own variables that are being shifted

AngheloAlf: Adding `#include "prevent_bss_reordering.h"` to the sys_cfb file does not fix the bss issue? :thinking:

hensldm: The main issue is bss gets ordered based on externing as well, so if you don't extern everything things get ordered differently. Same issue in play, where everything had to be externed even locals.

src/overlays/actors/ovl_En_Test3/z_en_test3.c
```diff
@@ -5,6 +5,7 @@
  */
 
 #include "z_en_test3.h"
+#include "prevent_bss_reordering.h"
```
engineer124, line 8:
```suggestion

#include "prevent_bss_reordering.h"
#include "z_en_test3.h"
```
Could you swap these? We try to keep this on top

include/variables.h
engineer124: Dang, tricky. That's fine then

engineer124 (approved):

src/code/z_play.c
mzxrules: idk about this because I don't actually know what WorkBuffer really is

AngheloAlf (approved):

PR #1032, opened 2022-08-17
z_map decompiled

mzxrules:

PR #1165, opened 2022-12-31
Update MinimapList struct + ZAPD

hensldm:
See #1032 and https://github.com/zeldaret/ZAPD/issues/254

Note this does match when using current ZAPD main along with https://github.com/zeldaret/ZAPD/pull/265 merged
```
f46493eaa0628827dbd6ad3ecd8d65d6  mm.us.rev1.rom_uncompressed.z64
mm.us.rev1.rom_uncompressed.z64: OK
```

Personally I think it would be best to get this the necessary approvals, merge the ZAPD pr, than I can update ZAPD here and merge it to keep ZAPD non-matching time as short as possible. Though if there are any other ideas to more smoothly handle this I'm happy to hear them.


AngheloAlf:
A dumb solution to always have Jenkins happy would be:
- Introduce a new struct here and keep the old one
```c
typedef struct {
    /* 0x0 */ MinimapEntry* entry;
    /* 0x4 */ s16 unk4;
} MinimapListFixed; // size  = 0x8
```
- Make that ZAPD PR output a `MinimapListFixed` instead of a `MinimapList`, then merge that
- Update ZAPD in MM and rename the `MinimapListFixed` to `MinimapList` plus adding a `typedef` to avoid ZAPD not building
- Make another ZAPD PR to rename `MinimapListFixed` back to `MinimapList`

It is a convoluted process which involves many PRs in two repos, but it would make Jenkins happy. Not really a huge fan of it

Kenix3 (approved):

engineer124 (approved):

PR #1172, opened 2023-01-26
ovl_Obj_Grass_Unit OK and some grass documentation

mzxrules:


src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
```diff
@@ -194,82 +196,80 @@ void func_809A983C(ObjGrass* this, PlayState* play) {
...
                 break;
             }
         }
 
-        if (j < ARRAY_COUNT(D_809AADB0)) {
-            for (x = 2; x >= j; x--) {
-                D_809AADB0[x + 1] = D_809AADB0[x];
-                D_809AADA0[x + 1] = D_809AADA0[x];
+        if (j < ARRAY_COUNT(sNearestGrassGroupsDist)) {
+            for (x = ARRAY_COUNT(sNearestGrassGroupsDist) - 2; x >= j; x--) {
```
AngheloAlf, line 232: Why the `x = ARRAY_COUNT(sNearestGrassGroupsDist) - 2`?

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.h
```diff
@@ -7,40 +7,43 @@
...
-    /* 0x10 */ u8 unk_10;
-} ObjGrassStruct1_1; // size = 0x14
+#define OBJ_GRASS_GROUP_DRAW 1
+#define OBJ_GRASS_NUM_COLLIDERS 20
+
+typedef struct ObjGrassElement {
+    /* 0x00 */ Vec3f pos;
+    /* 0x0C */ s16 rotY;
+    /* 0x0E */ s8 dropTable;
+    /* 0x0F */ u8 elem_flags;
```
AngheloAlf, line 17:
```suggestion
    /* 0x0F */ u8 elemFlags;
```
btw could you make defines for them?

src/overlays/actors/ovl_Obj_Grass_Unit/z_obj_grass_unit.c
```diff
@@ -5,34 +5,161 @@
...
+            homePosYSum += grassElem->pos.y;
+        }
+    }
+    if (grassGroup->count > 0) {
+        grassManager->activeGrassGroups++;
+        grassGroup->homePos.x = this->home.pos.x;
+        grassGroup->homePos.y = (homePosYSum / grassGroup->count);
+        grassGroup->homePos.z = this->home.pos.z;
+    }
+    Actor_Kill(this);
```
AngheloAlf, line 164: this actor kills itself after init?

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
mzxrules: So the purpose of this section is to insert (group, dist) into the sorted list `sNearestGrassGroups`/`Dist` at index `j`. If `j` is the last index in the list (length - 1), then we can simply insert it directly into the list. But, if `j` is the second to last index (length - 2) or less, then we need to shift everything over by 1 until we reach index `j`. Or in other words, we copy the second to last element to the last element, decrement, copy third last element to second etc. until the `j`th element has been moved to `j+1`

One thing I will say is that I dislike using `ARRAY_COUNT` here instead of just defining a constant list length. I've only done so because it's the convention, but I feel it would make the code better to tie everything together with a single define.

src/overlays/actors/ovl_Obj_Grass_Unit/z_obj_grass_unit.c
mzxrules: Yep. The setup is pretty weird and nuanced, so perhaps I should put in a comment explaining how things work. 

You have Obj_Grass, which is the manager actor. It's purpose is to reduce the cost of having a lot of grass spawned at once by only processing a maximum 20 grass bushes on any given update cycle. 

Obj_Grass_Unit is the client of Obj_Grass. It initializes Obj_Grass if it hasn't been already, and then passes the spawn information over to the manager. Once this happens, there's nothing else for it to do so it just dies.

The nuanced thing here is scope. `Obj_Grass_Unit` is responsible for initializing and holding a reference to the Obj_Grass instance. But because ovl_Obj_Grass_Unit uses `ALLOCTYPE_NORMAL` allocation, the manager reference will fall out of scope once all instances are killed. In other words, if you spawn a "burst" of `Obj_Grass_Unit` in one frame (e.g. loading a new room), those `Obj_Grass_Unit` will use the same Obj_Grass instance, but subsequent spawns a few frames later will initialize a completely new manager.

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
```diff
@@ -194,82 +196,80 @@ void func_809A983C(ObjGrass* this, PlayState* play) {
...
+        grassGroup = sNearestGrassGroups[i];
 
-        if (ptr != NULL) {
-            for (j = 0; j < ptr->unk_FC; j++) {
-                if (!(ptr->unk_0C[j].unk_0F & 4)) {
-                    temp_f0 = Math3D_Vec3fDistSq(&ptr->unk_0C[j].unk_00, &player->actor.world.pos);
+        if (grassGroup != NULL) {
+            for (j = 0; j < grassGroup->count; j++) {
+                if (!(grassGroup->elements[j].elem_flags & 4)) {
+                    dist = Math3D_Vec3fDistSq(&grassGroup->elements[j].pos, &player->actor.world.pos);
```
engineer124, line 250:
```suggestion
                    distSq = Math3D_Vec3fDistSq(&grassGroup->elements[j].pos, &player->actor.world.pos);
```
Since this is squared, there's a few names that could reflect that:
`sNearestGrassElementsDist` -> `sNearestGrassElementsDistSq`
`2500.0f` -> `SQ(50.0f)`

src/overlays/actors/ovl_Obj_Grass_Unit/z_obj_grass_unit.c
```diff
@@ -5,34 +5,161 @@
...
+    ACTOR_OBJ_GRASS_UNIT,  ACTORCAT_BG,           FLAGS,
+    GAMEPLAY_FIELD_KEEP,   sizeof(ObjGrassUnit),  (ActorFunc)ObjGrassUnit_Init,
+    (ActorFunc)Actor_Noop, (ActorFunc)Actor_Noop, (ActorFunc)NULL,
 };
 
-#endif
+typedef struct {
+    f32 distance;
+    s16 rotY;
+} PolarCoord;
```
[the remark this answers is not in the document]
EllipticEllipsis: It's very similar to the spherical coordinates struct, I'd say stick it in z64math.h as VecPolar or something.

src/overlays/actors/ovl_Obj_Grass_Carry/z_obj_grass_carry.c
```diff
@@ -61,87 +61,89 @@ static ColliderCylinderInit sCylinderInit = {
...
 
-void func_809AAE94(Vec3f* arg0, f32 arg1) {
-    arg1 += ((Rand_ZeroOne() * 0.2f) - 0.1f) * arg1;
+void ObjGrassCarry_RandScaleVecToZero(Vec3f* vel, f32 scale) {
+    scale += ((Rand_ZeroOne() * 0.2f) - 0.1f) * scale;
 
-    arg0->x -= (arg0->x * arg1);
-    arg0->y -= (arg0->y * arg1);
-    arg0->z -= (arg0->z * arg1);
+    vel->x -= (vel->x * scale);
```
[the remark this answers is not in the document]
mzxrules: seems weirdly inconsistent since the other two are abbr.

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
```diff
@@ -58,35 +58,35 @@ static ColliderCylinderInit sCylinderInit = {
...
-s32 func_809A9110(PlayState* play, Vec3f* arg1) {
-    f32 sp2C;
+s32 func_809A9110(PlayState* play, Vec3f* pos) {
+    f32 w;
     Vec3f sp20;
 
-    SkinMatrix_Vec3fMtxFMultXYZW(&play->viewProjectionMtxF, arg1, &sp20, &sp2C);
+    SkinMatrix_Vec3fMtxFMultXYZW(&play->viewProjectionMtxF, pos, &sp20, &w);
 
     if ((play->projectionMtxFDiagonal.z * -130.13191f) < sp20.z) {
```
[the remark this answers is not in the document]
mzxrules: no idea at all, maybe the hardcore math nerds would know

src/overlays/actors/ovl_Obj_Grass_Carry/z_obj_grass_carry.c
```diff
@@ -61,87 +61,89 @@ static ColliderCylinderInit sCylinderInit = {
...
 };
 
-s16 D_809ABBFC = 0;
-s16 D_809ABC00 = 0;
-s16 D_809ABC04 = 0;
-s16 D_809ABC08 = 0;
+s16 sRotSpeedXTarget = 0;
+s16 sRotSpeedX = 0;
+s16 sRotSpeedYTarget = 0;
+s16 sRotSpeedY = 0;
```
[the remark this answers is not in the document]
mzxrules: what if I just made them static vars instead so they won't conflict?

src/overlays/actors/ovl_Obj_Grass_Carry/z_obj_grass_carry.c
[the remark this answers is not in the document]
mzxrules: opted not to make this change since there isn't a hard reason for spelling out velocity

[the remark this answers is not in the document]
AngheloAlf: `vel`:
image
`velocity`:
image

`velocity` is around 8 times more common than `vel`. We should try to avoid increasing the inconsistency in the repository, so please change it to `velocity`

include/z64math.h
```diff
@@ -39,6 +39,11 @@ typedef struct {
     /* 0x8 */ s32 z;
 } Vec3i; // size = 0xC
 
+typedef struct {
+    /* 0x0 */ f32 distance;
+    /* 0x4 */ s16 rotY;
+} VecPolar; // size = 0x8
```
AngheloAlf, line 45: Since this is named after the polar coordinate system, I guess it would be better to use `angle` instead of `rotY`

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
```diff
@@ -95,42 +97,44 @@ void func_809A91FC(MtxF* matrix) {
     }
 }
 
-void func_809A92D0(ObjGrassStruct1_1* ptr, PlayState* play) {
-    if (!(ptr->unk_0E & 0x10)) {
-        Item_DropCollectibleRandom(play, NULL, &ptr->unk_00, ptr->unk_0E * 0x10);
+void ObjGrass_DropCollectible(ObjGrassElement* grassElem, PlayState* play) {
+    if ((grassElem->dropTable & 0x10) == 0) {
```
AngheloAlf, line 101: Why the `== 0` instead of `!`?

src/overlays/actors/ovl_Obj_Grass_Carry/z_obj_grass_carry.c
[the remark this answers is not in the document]
engineer124: I'm not opposed to discussing using the abbreviation `vel` instead of `velocity` since pos/accel are also abbreviated, but I think that's outside the scope of this PR, and consistency is better so it's easier to make that change all at once. So `velocity` seems good just for now

src/overlays/actors/ovl_Obj_Grass_Carry/z_obj_grass_carry.c
```diff
@@ -61,89 +61,91 @@ static ColliderCylinderInit sCylinderInit = {
...
-    arg1 += ((Rand_ZeroOne() * 0.2f) - 0.1f) * arg1;
+void ObjGrassCarry_RandScaleVecToZero(Vec3f* vel, f32 scale) {
+    scale += ((Rand_ZeroOne() * 0.2f) - 0.1f) * scale;
 
-    arg0->x -= (arg0->x * arg1);
-    arg0->y -= (arg0->y * arg1);
-    arg0->z -= (arg0->z * arg1);
+    vel->x -= (vel->x * scale);
+    vel->y -= (vel->y * scale);
+    vel->z -= (vel->z * scale);
```
engineer124, line 96:
```suggestion
    vel->y -= vel->y * scale;
    vel->z -= vel->z * scale;
```
Probably don't need these brackets

src/overlays/actors/ovl_Obj_Grass_Unit/z_obj_grass_unit.c
```diff
@@ -1,38 +1,168 @@
...
+
+static ObjGrassUnitPattern sGrassPatterns[2] = {
+    { ARRAY_COUNT(sGrassPatternCircle), sGrassPatternCircle },
+    { ARRAY_COUNT(sGrassPatternMixed), sGrassPatternMixed },
+};
+
+ObjGrass* sGrassManager = NULL;
+ObjGrassCarry* sGrassCarry0 = NULL;
+ObjGrassCarry* sGrassCarry1 = NULL;
+s32 sInitialized = 0;
```
engineer124, line 55:
```suggestion
s32 sInitialized = false;
```

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
```diff
@@ -56,35 +58,35 @@ static ColliderCylinderInit sCylinderInit = {
...
-Vec3f D_809AAB4C[] = {
-    { 0.0f, 0.707099974155f, 0.707099974155f },
-    { 0.707099974155f, 0.707099974155f, 0.0f },
-    { 0.0f, 0.707099974155f, -0.707099974155f },
-    { -0.707099974155f, 0.707099974155f, 0.0f },
+static Vec3f sUnitDirections[] = {
+    { 0.0f, 0.7071f, 0.7071f },
+    { 0.7071f, 0.7071f, 0.0f },
+    { 0.0f, 0.7071f, -0.7071f },
+    { -0.7071f, 0.7071f, 0.0f },
```
hensldm, line 65: Out of curiosity does `M_SQRT1_2` work for the `0.7071`s?

hensldm (approved):
1 minor suggestion.

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
mzxrules: Probably because I made a mistake when analyzing the variable.  assumed that dI assumed that `grassElem->dropTable` stored a single value ranging from 0x00-0x1F, but it seems to consistently test `(dropTable & 0x10) == 0` before passing the value to Item_DropCollectibleRandom,

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
mzxrules: nope, you need way too many digits of precision I'd assume

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.c
engineer124: Since there's multiple arrays that need to be synced to this size, I think using a define would be fine?

src/overlays/actors/ovl_Obj_Grass/z_obj_grass.h
```diff
@@ -7,40 +7,50 @@
...
+} ObjGrassElement; // size = 0x14
 
 typedef struct {
-    /* 0x00 */ Vec3f unk_00;
-    /* 0x0C */ ObjGrassStruct1_1 unk_0C[12];
-    /* 0xFC */ s16 unk_FC;
-    /* 0xFE */ u8 unk_FE;
-} ObjGrassStruct1; // size = 0x100
+    /* 0x00 */ Vec3f homePos;
+    /* 0x0C */ ObjGrassElement elements[12];
```
engineer124, line 30:
```suggestion
    /* 0x0C */ ObjGrassElement elements[OBJ_GRASS_GROUP_ELEM_COUNT_MAX];
```

engineer124 (approved):

AngheloAlf (approved):

PR #1176, opened 2023-02-24
Update `m2ctx.py`

AngheloAlf:
I cleaned up the script and modified it so macros are preserved. Useful for cases when you want to manually create a decomp.me scratch because gfx macros are preserved in context

### DIFF
--- a/tools/m2ctx.py
+++ b/tools/m2ctx.py
@@ -25,10 +25,8 @@ CPP_FLAGS = [
     "-D_Static_assert(x, y)=",
     "-D__attribute__(x)="
     "-D_MIPS_SZLONG=32",
-    "-D__attribute__(x)=",
     "-ffreestanding",
     "-DM2CTX",
-    "-DNON_MATCHING",
 
     "-std=gnu89",
 ]
--- a/tools/m2ctx.py
+++ b/tools/m2ctx.py
@@ -1,14 +1,43 @@
 #!/usr/bin/env python3
 
-import argparse, os, subprocess, sys
+import argparse
+import os
+import sys
+import subprocess
+import tempfile
 from pathlib import Path
 
-script_dir = os.path.dirname(os.path.realpath(__file__))
-root_dir = script_dir + "/../"
-src_dir = root_dir + "src/"
+script_dir = Path(os.path.dirname(os.path.realpath(__file__)))
+root_dir = script_dir / ".."
+src_dir = root_dir / "src"
+
+# Project-specific
+CPP_FLAGS = [
+    "-Iinclude",
+    "-Iassets",
+    "-Isrc",
+
+    "-D__sgi",
+    "-D_LANGUAGE_C",
+    "-DNON_MATCHING",
+    "-D_Static_assert(x, y)=",
+    "-D__attribute__(x)="
+    "-DNDEBUG",
+    "-D_FINALROM",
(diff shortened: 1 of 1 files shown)


hensldm (approved):
Works fine for me :)

tools/m2ctx.py
```diff
@@ -1,14 +1,43 @@
...
-src_dir = root_dir + "src/"
+script_dir = Path(os.path.dirname(os.path.realpath(__file__)))
+root_dir = script_dir / ".."
+src_dir = root_dir / "src"
+
+# Project-specific
+CPP_FLAGS = [
+    "-Iinclude",
+    "-Iassets",
+    "-Isrc",
```
hensldm, line 18: can you add `Ibuild` and `I.`, as well switch `Iassets` and `Isrc` to fully match with the Makefile?
`-Iinclude -Isrc -Iassets -Ibuild -I.`

hensldm (changes requested):

hensldm (approved):

isghj5 (approved):

tools/m2ctx.py
```diff
@@ -1,14 +1,45 @@
...
+    "-D_LANGUAGE_C",
+    "-DNON_MATCHING",
+    "-D_Static_assert(x, y)=",
+    "-D__attribute__(x)="
+    "-DNDEBUG",
+    "-D_FINALROM",
+    "-D_MIPS_SZLONG=32",
+    "-DSCRIPT(x)="
+    "-D__attribute__(x)=",
+    "-D__asm__(x)=",
```
EllipticEllipsis, line 32: This one is completely useless since we'll never use __asm__ when matching.

tools/m2ctx.py
```diff
@@ -1,14 +1,45 @@
...
+    "-Ibuild",
+    "-I.",
+
+    "-D__sgi",
+    "-D_LANGUAGE_C",
+    "-DNON_MATCHING",
+    "-D_Static_assert(x, y)=",
+    "-D__attribute__(x)="
+    "-DNDEBUG",
+    "-D_FINALROM",
```
EllipticEllipsis, line 28: Maybe, maybe not.

tools/m2ctx.py
```diff
@@ -1,14 +1,45 @@
...
+    "-D_MIPS_SZLONG=32",
+    "-DSCRIPT(x)="
```
EllipticEllipsis, line 30: ?

tools/m2ctx.py
```diff
@@ -48,63 +78,55 @@ def custom_replacements(output):
...
+def import_c_file(in_file) -> str:
+    in_file = os.path.relpath(in_file, root_dir)
 
-def get_c_file(directory):
-    for root, dirs, files in os.walk(directory):
-        for file in files:
-            if file.endswith(".c") and "data" not in file:
-                return file
+    cpp_command = ["gcc", "-E", "-P", "-undef", "-dM", *CPP_FLAGS, in_file]
+    cpp_command2 = ["gcc", "-E", "-P", "-undef", *CPP_FLAGS, in_file]
```
EllipticEllipsis, line 85: Why are there two of these?

AngheloAlf: The first one has `-dM` which is the thing that does the macro preservation

EllipticEllipsis (approved):

PR #1189, opened 2023-03-01
Misc cleanup 13

hensldm:
Does 4 things:

- Syncs up the FaultAddrConv functions in graph, actor overlay table, and kaleido manager together
- Adjusts Invisible Ruppe macros to actually take `thisx`
- Merges the `rot`s, `prim`s, and `env` into larger structs
- Change all z64 header defines to be `Z64MY_HEADER_H` format


src/code/z_actor_dlftbls.c
```diff
@@ -36,15 +36,15 @@ ActorOverlay gActorOverlayTable[] = {
...
-FaultAddrConvClient D_801ED940;
+FaultClient sActorOverlayTableFaultClient;
+FaultAddrConvClient sActorOverlayTableFaultAddrConvClient;
 
-void ActorOverlayTable_FaultPrint(void* arg0, void* arg1) {
+void ActorOverlayTable_FaultClient(void* arg0, void* arg1) {
     ActorOverlay* overlayEntry;
     u32 overlaySize;
-    ActorId i;
+    s32 i;
```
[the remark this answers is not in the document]
hensldm: Yeah looking at it again I agree keeping ActorId as the type is better. I've also renamed the variables to `actorId`.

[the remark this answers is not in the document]
fig02: only reason it would be preferred is sync with oot, since we dont use enums as types over there. 
idk if we'll ever come to an agreement on this topic, but the main argument for not doing so is that you cant consistently use enums as types everywhere. only where they happened to use s32 in their original code

AngheloAlf (approved):

include/z64animation.h
```diff
@@ -1,5 +1,5 @@
-#ifndef Z64_ANIMATION_H
-#define Z64_ANIMATION_H
+#ifndef Z64ANIMATION_H
+#define Z64ANIMATION_H
```
engineer124, line 2: Personally, I prefer as it was, with the underscore after the `Z64_`. I find it more readable that way. Is there an underlying reason for one way or the other?

src/code/z_actor_dlftbls.c
```diff
@@ -38,45 +38,44 @@ ActorOverlay gActorOverlayTable[] = {
...
+void* ActorOverlayTable_FaultAddrConv(void* address, void* param) {
+    uintptr_t addr = address;
+    ActorOverlay* actorOvl = &gActorOverlayTable[0];
+    uintptr_t ramConv;
+    void* ramStart;
+    uintptr_t diff;
+    ActorId actorId;
+
+    for (actorId = 0; actorId < gMaxActorId; actorId++, actorOvl++) {
+        diff = VRAM_PTR_SIZE(actorOvl);
```
engineer124, line 72: What's the reason for changing this from `size_t` to `uintptr_t`?

src/code/z_actor_dlftbls.c
```diff
@@ -38,45 +38,44 @@ ActorOverlay gActorOverlayTable[] = {
...
+        ramStart = actorOvl->loadedRamAddr;
+        ramConv = (uintptr_t)actorOvl->vramStart - (uintptr_t)ramStart;
+
         if (ramStart != NULL) {
-            if (ptr >= ramStart && ptr < ramEnd) {
-                return ptr + offset;
+            if (addr >= (uintptr_t)ramStart && addr < (uintptr_t)ramStart + diff) {
```
engineer124, line 77:
```suggestion
            if ((addr >= (uintptr_t)ramStart) && (addr < (uintptr_t)ramStart + diff)) {
```

src/code/z_vr_box.c
```diff
@@ -210,17 +210,17 @@ void Skybox_Setup(GameState* gameState, SkyboxContext* skyboxCtx, s16 skyboxId)
...
-            skyboxCtx->envG = 0;
-            skyboxCtx->envB = 40;
+            skyboxCtx->env.r = 40;
+            skyboxCtx->env.g = 0;
+            skyboxCtx->env.b = 40;
 
             // Inverted Stone Tower Temple and Inverted Stone Tower
             if ((play->sceneId == SCENE_F41) || (play->sceneId == SCENE_INISIE_R)) {
-                skyboxCtx->rotZ = 3.15f;
+                skyboxCtx->rot.z = 3.15f;
```
engineer124, line 223: Not even 3.14

include/z64animation.h
AngheloAlf: If I had to guess it would be because I have made a bunch of PRs without that underscore. I don't have a hard preference about this so I could change my PRs to have the underscore if that's preferred

src/code/z_vr_box.c
AngheloAlf: Reminded me of this image lmao
image

src/code/z_actor_dlftbls.c
hensldm: That is just what was there from the graph function, which is what I copied to the others. I do agree `size_t` fits better so will change. I also think `ramConv` also makes more sense as a `size_t` so will change that as well.

include/z64animation.h
hensldm: 2 reasons:
1. There seemed to be more already in this format.
2. The bigger reason is it matches the file name better which has that format.

I'm not opposed to changing it (I agree it does look a little better), but probably should change the file names as well imo.

engineer124: I suppose it's fine as you've changed it then

engineer124 (approved):

isghj5 (approved):

PR #1199, opened 2023-03-06
Port OoT's docs for `fault.c` and `fault_drawer.c`

AngheloAlf:
Most of the hard work were made by @Thar0 in https://github.com/zeldaret/oot/pull/1106


include/fault.h
```diff
@@ -0,0 +1,102 @@
...
+    /* 0x7CC */ u8 exitDebugger;
+    /* 0x7CD */ u8 msgId; // 1 - CPU Break; 2 - Fault; 3 - Unknown
+    /* 0x7CE */ u8 faultHandlerEnabled;
+    /* 0x7CF */ u8 autoScroll;
+    /* 0x7D0 */ OSThread* faultedThread;
+    /* 0x7D4 */ FaultPadCallback padCallback;
+    /* 0x7D8 */ FaultClient* clients;
+    /* 0x7DC */ FaultAddrConvClient* addrConvClients;
+    /* 0x7E0 */ UNK_TYPE1 pad7E0[0x4];
+    /* 0x7E4 */ Input padInput[4];
```
hensldm, line 95:
```suggestion
    /* 0x7E4 */ Input inputs[4];
```
Matches OoT

src/boot_O2_g3/fault.c
```diff
@@ -376,19 +462,20 @@ void Fault_PrintThreadContext(OSThread* t) {
     FaultDrawer_Printf("\n");
     FaultDrawer_SetCharPad(0, 0);
 
-    if (D_8009BE54 != 0) {
-        FaultDrawer_DrawText(0xA0, 0xD8, "%5.2f sec\n", D_8009BE54);
+    if (D_8009BE54 != 0.0f) {
+        FaultDrawer_DrawText(160, 216, "%5.2f sec\n", D_8009BE54);
```
hensldm, line 466: Hmm, I wonder if we can name `D_8009BE54` based on this. It seems like a timer/time taken value (think OoT speed_meter uses `TimeTotal`) of some kind, but doesn't seem to get set anywhere so always seems to be 0?

src/boot_O2_g3/fault.c
```diff
@@ -407,149 +494,193 @@ void osSyncPrintfThreadContext(OSThread* t) {
...
+    if (alignedAddr < K0BASE) {
+        alignedAddr = K0BASE;
     }
-    if (alignedAddr > (u32*)0x807FFF00) {
-        alignedAddr = (u32*)0x807FFF00;
+    // 8MB RAM, leave room to display 0x100 bytes on the final page
+    //! @bug The loop below draws 22 * 4 * 4 = 0x160 bytes per page. Due to this, by scrolling further than
+    //! 0x807FFEA0 some invalid bytes are read from outside of 8MB RDRAM space. This does not cause a crash,
+    //! however the values it displays are meaningless. On N64 hardware these invalid addresses are read as 0.
+    if (alignedAddr > K0BASE + 0x800000 - 0x100) {
```
hensldm, line 586:
```suggestion
    if (alignedAddr > (K0BASE + 0x800000 - 0x100)) {
```

src/boot_O2_g3/fault.c
```diff
@@ -407,149 +494,193 @@ void osSyncPrintfThreadContext(OSThread* t) {
...
-            addr = 0x80000000;
+        scrollCountdown = 0;
+        // Ensure address is within the bounds of RDRAM
+        if (addr < K0BASE) {
+            addr = K0BASE;
         }
-        if (addr > 0x807FFF00) {
-            addr = 0x807FFF00;
+        // 8MB RAM, leave room to display 0x100 bytes on the final page
+        if (addr > K0BASE + 0x800000 - 0x100) {
```
hensldm, line 640:
```suggestion
        if (addr > (K0BASE + 0x800000 - 0x100)) {
```

src/boot_O2_g3/fault.c
```diff
@@ -564,354 +695,440 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
 void Fault_DrawStackTrace(OSThread* t, u32 flags) {
-    s32 y;
-    u32 sp;
-    u32 ra;
-    u32 pc;
+    s32 line;
+    uintptr_t sp = t->context.sp;
+    uintptr_t ra = t->context.ra;
+    uintptr_t pc = t->context.pc;
     u32 pad;
```
hensldm, line 818:
```suggestion
    s32 pad;
```
Also can it go below `addr`?

src/boot_O2_g3/fault.c
```diff
@@ -564,354 +695,440 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+done:
     *spPtr = sp;
     *pcPtr = pc;
     *raPtr = ra;
 }
 
+/**
+ * Draws the stack trace page contents for the specified thread
+ */
 void Fault_DrawStackTrace(OSThread* t, u32 flags) {
```
hensldm, line 813:
```suggestion
void Fault_DrawStackTrace(OSThread* thread, u32 flags) {
```

src/boot_O2_g3/fault.c
```diff
@@ -564,354 +695,440 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
             osSyncPrintf(" -> ????????");
         }
         osSyncPrintf("\n");
 
-        Fault_FindNextStackCall(&sp, &pc, &ra);
+        Fault_WalkStack(&sp, &pc, &ra);
     }
 }
 
 void Fault_ResumeThread(OSThread* t) {
```
hensldm, line 868:
```suggestion
void Fault_ResumeThread(OSThread* thread) {
```
Might be good to double check I didn't miss any `OSThread* t -> OSThread* thread`

src/boot_O2_g3/fault_drawer.c
```diff
@@ -1,28 +1,64 @@
...
+    /* 0x35 */ u8 osSyncPrintfEnabled;
+    /* 0x38 */ FaultDrawerCallback inputCallback;
+} FaultDrawer; // size = 0x3C
+
 extern const u32 sFaultDrawerFont[];
 
-FaultDrawer sFaultDrawerStruct;
+FaultDrawer sFaultDrawer;
+
+FaultDrawer* sFaultDrawerRef = &sFaultDrawer;
```
hensldm, line 40:
```suggestion
FaultDrawer sFaultDrawerInstance;

FaultDrawer* sFaultDrawer = &sFaultDrawerInstance;
```
Mirrors what happens with `faultMgr` which I think is better, though it is worth noting that this does differ from `OoT` since they just use the instance directly, and don't have the reference variable.

include/fault.h
```diff
@@ -0,0 +1,102 @@
...
+void FaultDrawer_SetCharPad(s8 padW, s8 padH);
+void FaultDrawer_SetCursor(s32 x, s32 y);
+s32 FaultDrawer_VPrintf(const char* fmt, va_list ap);
+s32 FaultDrawer_Printf(const char* fmt, ...);
+void FaultDrawer_DrawText(s32 x, s32 y, const char* fmt, ...);
+
+
+typedef struct FaultMgr {
+    /* 0x000 */ OSThread thread;
+    /* 0x1B0 */ u8 stack[0x600]; // Seems leftover from an earlier version. The thread actually uses a stack of this size at 0x8009BE60
```
hensldm, line 83: Out of curiosity if this seems like a stack variable, would it use the `STACK` macro?

src/boot_O2_g3/fault.c
AngheloAlf: It isn't set anywhere so it is always zero.
Maybe we could name it something like `sFaultTimer`?

hensldm: tbh, I'd probably go with `TimeTotal` just because it would seem more like it keeps track of how long something takes, though am fine with `timer`. I do think it should have an unused comment as well.

include/fault.h
AngheloAlf: I was thinking in using the `STACK` macro, but I would need #1193 to be merged first. If this PR goes in first then I could change this in that PR

src/boot_O2_g3/fault.c
AngheloAlf: Annoyingly not even the debug rom sets this variable

src/libultra/os/createthread.c
```diff
@@ -1,6 +1,6 @@
 #include "global.h"
 
-void osCreateThread(OSThread* t, OSId id, void* entry, void* arg, void* sp, OSPri p) {
+void osCreateThread(OSThread* thread, OSId id, void* entry, void* arg, void* sp, OSPri p) {
```
hensldm, line 3: Don't forget to update the usage of the arg.

hensldm (approved):

include/fault.h
```diff
@@ -0,0 +1,103 @@
...
+#define FAULT_COLOR_STRINGIFY(s) #s
+#define FAULT_COLOR_EXPAND_AND_STRINGIFY(s) FAULT_COLOR_STRINGIFY(s)
+
+#define FAULT_ESC '\x1A'
+#define FAULT_COLOR(n) "\x1A" FAULT_COLOR_EXPAND_AND_STRINGIFY(FAULT_COLOR_ ## n)
+
+
+#define NORMAL_RDRAM_END 0x80400000
+// Address at the end of normal RDRAM after which is room for a screen buffer
+#define FAULT_FB_ADDRESS (NORMAL_RDRAM_END - SCREEN_HEIGHT * SCREEN_WIDTH * sizeof(u16))
```
mzxrules, line 32: Couple things here.

Considering that Fault_DisplayFrameBuffer implements it's framebuffer pointer as 
```c
fb = (void*)(PHYS_TO_K0(osMemSize) - SCREEN_HEIGHT * SCREEN_WIDTH * sizeof(u16));```

should we not express the address this way?

```c
#define FAULT_FB_ADDRESS (PHYS_TO_K0(0x400000) - SCREEN_HEIGHT * SCREEN_WIDTH * sizeof(u16))```


Second, the wording of the comment confused me for a bit, since the way it reads can make it seem like the framebuffer comes after the end of the hardware's 4MB RDRAM. Imo, the key thing to express here is that the framebuffer is reserved at the end of the 4MB already installed on the N64, since MM is expected to handle booting without an expansion pack installed.


Lastly, I truly despise the `NORMAL` in `NORMAL_RDRAM_END`. The word normal should only ever be used in the context of math and game options; for everything else you should be using something more descriptive.

Comment (AngheloAlf) [reply include/fault.h:32]: This is an already existing comment and define which I moved from another file to this header file. I don't really like it but it is used in sys_initial_check and I didn't want to mess what's going on there.
I guess I can yeet `NORMAL_RDRAM_END` and use the `PHYS_TO_K0(0x400000)` instead

File: src/boot_O2_g3/fault_drawer.c (hunk 8 above)
Comment (hensldm) [reply src/boot_O2_g3/fault_drawer.c]: Oh I actually swapped these from what `Fault` does. The `Instance` suffix should be on the pointer.

File: src/boot_O2_g3/fault.c [hunk 12]
```diff
@@ -572,354 +696,440 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+    Input* input3 = &sFaultInstance->inputs[3];
     u32 pad;
-    u32 graphPC;
-    u32 graphRA;
-    u32 graphSP;
+    uintptr_t pc;
+    uintptr_t ra;
+    uintptr_t sp;
 
     if (CHECK_BTN_ALL(input3->press.button, 0x80)) {
```
hensldm, line 933: https://discord.com/channels/688807550715560050/688851976746041391/1084273960721592360
Could probably use a comment that this is a neutral reset.

src/boot_O2_g3/fault_drawer.c
```diff
@@ -249,17 +294,17 @@ void FaultDrawer_DrawText(s32 x, s32 y, const char* fmt, ...) {
 }
 
 void FaultDrawer_SetDrawerFB(void* fb, u16 w, u16 h) {
```
engineer124, line 296: Since `Fault_SetFB` -> `Fault_SetFrameBuffer` and `Fault_CommitFB` -> `Fault_DisplayFrameBuffer`, would `FaultDrawer_SetDrawerFB` -> `FaultDrawer_SetDrawerFrameBuffer` make sense?

Similarly, would `void* fb` -> `void* frameBuffer` make sense?

AngheloAlf: I kept this name because that's how it is named in OoT, but you are right I'll rename it

src/boot_O2_g3/fault.c
```diff
@@ -572,354 +696,441 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
 
-void Fault_AddHungupAndCrashImpl(const char* arg0, char* arg1) {
+/**
+ * Immediately crashes the current thread, for cases where an irrecoverable
+ * error occurs. The parameters specify two messages detailing the error, one
+ * or both may be NULL.
+ */
+void Fault_AddHungupAndCrashImpl(const char* exp1, const char* exp2) {
     FaultClient client;
     u32 pad;
```
hensldm, line 1121:
```suggestion
    s32 pad;
```

src/boot_O2_g3/fault.c
```diff
@@ -572,354 +696,441 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+                   Z_THREAD_PRI_FAULT);
+    osStartThread(&sFaultInstance->thread);
 }
 
-void Fault_HangupFaultClient(const char* arg0, char* arg1) {
+/**
+ * Fault page for Hungup crashes. Displays the thread id and two messages
+ * specified in arguments to `Fault_AddHungupAndCrashImpl`.
+ */
+void Fault_HangupFaultClient(const char* exp1, char* exp2) {
```
hensldm, line 1105:
```suggestion
void Fault_HangupFaultClient(const char* exp1, const char* exp2) {
```

src/boot_O2_g3/fault.c
```diff
@@ -572,354 +696,441 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+            FaultDrawer_DrawText(64, 90, "All Pages are displayed.");
+            FaultDrawer_DrawText(64, 100, "       THANK YOU!       ");
+            FaultDrawer_DrawText(64, 110, " You are great debugger!");
             Fault_WaitForInput();
+        } while (!sFaultInstance->exit);
 
-        } while (!sFaultContext->exitDebugger);
-
-        while (!sFaultContext->exitDebugger) {
+        while (!sFaultInstance->exit) {
```
hensldm, line 1068:
```suggestion
        while (!sFaultInstance->exit) {}
```
As well as remove the `;` and `}` below.

src/boot_O2_g3/fault.c
```diff
@@ -37,34 +81,44 @@ const char* sFpuExceptions[] = {
...
 
-void Fault_AddClient(FaultClient* client, void* callback, void* param0, void* param1) {
+/**
+ * Registers a fault client.
+ *
+ * Clients contribute at least one page to the crash screen, drawn by `callback`.
+ * Arguments are passed on to the callback through `arg0` and `arg1`.
+ *
+ * The callback is intended to be
+ * `void (*callback)(void* arg0, void* arg1)`
```
hensldm, line 96: Probably an @Thar0 question, but any reason not to make a `FaultClientCallback` typedef and use it here and in the struct, instead of `void*`. Same question for the `AddrConvClients`.

src/boot_O2_g3/fault.c
```diff
@@ -143,25 +208,21 @@ void Fault_AddAddrConvClient(FaultAddrConvClient* client, void* callback, void*
...
                 lastIter->next = client->next;
             } else {
-                sFaultContext->addrConvClients = client;
-                if (sFaultContext->addrConvClients) {
-                    sFaultContext->addrConvClients = client->next;
+                sFaultInstance->addrConvClients = client;
+                if (sFaultInstance->addrConvClients) {
+                    sFaultInstance->addrConvClients = client->next;
                 } else {
                     listIsEmpty = 1;
```
hensldm, line 227:
```suggestion
                    listIsEmpty = true;
```

src/boot_O2_g3/fault.c
```diff
@@ -112,27 +165,39 @@ void Fault_RemoveClient(FaultClient* client) {
...
 
-void Fault_AddAddrConvClient(FaultAddrConvClient* client, void* callback, void* param) {
+/**
+ * Registers an address converter client. This enables the crash screen to look up virtual
+ * addresses of overlays relocated during runtime. Address conversion is carried out by
+ * `callback`, which either returns a virtual address or NULL if the address could not
+ * be converted.
+ *
+ * The callback is intended to be
+ * `uintptr_t (*callback)(uintptr_t addr, void* arg)`
```
hensldm, line 175: All of the current clients currently return `void*` so should probably update them to return `uintptr_t`s . Also I'm kind of thinking arg0 to the callback should remain an void*, since each client seems to do a cast to `uintptr_t` as one of the first things it does.

src/boot_O2_g3/fault.c
```diff
@@ -112,27 +165,39 @@ void Fault_RemoveClient(FaultClient* client) {
...
+ * The callback may return 0 if it could not convert the address
+ * The callback may return -1 to be unregistered
```
hensldm, line 177: I don't think this is true for MM's version. `Fault_ConvertAddress` doesn't have a `-1` check.

hensldm (changes requested):

src/boot_O2_g3/fault.c
AngheloAlf: I guess it is because some clients don't pass a function with the proper signature. For example 
```c

void CIC6105_PrintRomInfo(void) {
    FaultDrawer_DrawText(80, 200, "SP_STATUS %08x", HW_REG(SP_STATUS_REG, u32));
    FaultDrawer_DrawText(40, 184, "ROM_F [Creator:%s]", gBuildTeam);
    FaultDrawer_DrawText(56, 192, "[Date:%s]", gBuildDate);
}

void CIC6105_AddRomInfoFaultPage(void) {
    Fault_AddClient(&romInfoFaultClient, CIC6105_PrintRomInfo, NULL, NULL);
}
```

src/boot_O2_g3/fault.c
```diff
@@ -112,27 +162,38 @@ void Fault_RemoveClient(FaultClient* client) {
...
+ * be converted.
+ *
+ * The callback is intended to be
+ * `uintptr_t (*callback)(uintptr_t addr, void* arg)`
+ * The callback may return 0 if it could not convert the address
+ */
+void Fault_AddAddrConvClient(FaultAddrConvClient* client, FaultAddrConvClientCallback callback, void* arg) {
     OSIntMask mask;
-    u32 alreadyExists = 0;
+    bool alreadyExists = false;
```
hensldm, line 177:
```suggestion
    s32 alreadyExists = false;
```

src/boot_O2_g3/fault.c
hensldm: Tbh, I think return `uintptr_t` is still better for these addrconv clients, since that is how Fault uses them always. 

Also looks like this comment still needs to be updated to match the typedef.

hensldm: Bump on the second comment.

hensldm (commented):
Just want to bump on a comment that I made a few weeks ago.

hensldm (approved):
LGTM, though Jenkins is upset (my guess is bss)

src/boot_O2_g3/fault.c
```diff
@@ -182,98 +239,111 @@ void Fault_RemoveAddrConvClient(FaultAddrConvClient* client) {
...
-        if (sFaultContext->faultActive) {
+        if (sFaultInstance->autoScroll) {
             if (count-- < 1) {
-                return 0;
+                return false;
             }
         } else {
-            if (kDown == BTN_A || kDown == BTN_DRIGHT) {
-                return 0;
+            if (pressedBtn == BTN_A || pressedBtn == BTN_DRIGHT) {
```
engineer124, line 304:
```suggestion
            if ((pressedBtn == BTN_A) || (pressedBtn == BTN_DRIGHT)) {
```

src/boot_O2_g3/fault.c
```diff
@@ -283,11 +353,12 @@ void Fault_PrintFReg(s32 idx, f32* value) {
     if ((v0 >= -0x7E && v0 < 0x80) || raw == 0) {
         FaultDrawer_Printf("F%02d:%14.7e ", idx, *value);
     } else {
+        // Print subnormal floats as their ieee-754 hex representation
```
engineer124, line 356:
```suggestion
        // Print subnormal floats as their IEEE-754 hex representation
```

src/boot_O2_g3/fault.c
```diff
@@ -182,98 +239,111 @@ void Fault_RemoveAddrConvClient(FaultAddrConvClient* client) {
...
-    FaultDrawer_SetBackColor(0);
+    FaultDrawer_SetBackColor(GPACK_RGBA5551(0, 0, 0, 0));
 }
 
 void Fault_DrawCornerRec(u16 color) {
-    Fault_DrawRec(0x16, 0x10, 8, 1, color);
+    Fault_DrawRec(22, 16, 8, 1, color);
 }
 
 void Fault_PrintFReg(s32 idx, f32* value) {
```
engineer124, line 349:
```suggestion
void Fault_PrintFReg(s32 index, f32* value) {
```
I think we've been trying to purge `idx` -> `index`?

src/boot_O2_g3/fault.c
```diff
@@ -327,22 +401,23 @@ void osSyncPrintfFPCR(u32 value) {
     }
 }
 
-void Fault_PrintThreadContext(OSThread* t) {
+void Fault_PrintThreadContext(OSThread* thread) {
     __OSThreadContext* ctx;
```
engineer124, line 405:
```suggestion
    __OSThreadContext* threadCtx;
```
Maybe this was fine as is, but with so many ctx, it's pretty comment to use the prefix

src/boot_O2_g3/fault.c
```diff
@@ -416,149 +492,193 @@ void osSyncPrintfThreadContext(OSThread* t) {
...
-    OSTime start;
-    start = osGetTime();
+    s32 pad;
+    OSTime start = osGetTime();
+
     do {
-        Fault_Sleep(0x10);
-    } while ((osGetTime() - start) <= OS_USEC_TO_CYCLES(5000000));
+        Fault_Sleep(1000 / 60);
+    } while ((osGetTime() - start) < OS_SEC_TO_CYCLES(5) + 1);
```
engineer124, line 546: Why the `+ 1` instead of `<=`?

src/boot_O2_g3/fault.c
```diff
@@ -182,98 +239,111 @@ void Fault_RemoveAddrConvClient(FaultAddrConvClient* client) {
...
+ * Awaits user input
+ *
+ * L toggles auto-scroll
+ * DPad-Up enables osSyncPrintf output
+ * DPad-Down disables osSyncPrintf output
+ * A and DPad-Right continues and returns true
+ * DPad-Left continues and returns false
+ */
+u32 Fault_WaitForInputImpl(void) {
+    Input* input = &sFaultInstance->inputs[0];
```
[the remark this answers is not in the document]
hensldm: Personally I would say no, since the macro is meant for game states which sFaultInstance is not.

include/fault.h
```diff
@@ -0,0 +1,108 @@
...
+
+#define FAULT_COLOR_STRINGIFY(s) #s
+#define FAULT_COLOR_EXPAND_AND_STRINGIFY(s) FAULT_COLOR_STRINGIFY(s)
+
+#define FAULT_ESC '\x1A'
+#define FAULT_COLOR(n) "\x1A" FAULT_COLOR_EXPAND_AND_STRINGIFY(FAULT_COLOR_ ## n)
+
+
+// Address at the end of Jumper Pak
+#define FAULT_FB_ADDRESS (void*)((PHYS_TO_K0(0x400000) - SCREEN_HEIGHT * SCREEN_WIDTH * sizeof(u16)))
```
Thar0, line 32: "Jumper Pak" is not the right term for this, I would probably just call this "non-expansion" memory or similar

Thar0 (changes requested):

src/boot_O2_g3/fault.c
```diff
@@ -1,15 +1,59 @@
...
 
-extern FaultThreadStruct* sFaultContext;
-extern f32 D_8009BE54;
-extern u32 faultCustomOptions;
-extern u32 faultCopyToLog;
+// bss
+extern FaultMgr* sFaultInstance;
+extern f32 sFaultTimeTotal; // read but not set anywhere
+// extern u32 faultCustomOptions;
+// extern u32 faultCopyToLog;
```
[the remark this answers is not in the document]
AngheloAlf: The original reason was it was supposed to help remembering the original bss order, but I gave a try to import bss and got it matching without doing anything crazy.

https://github.com/zeldaret/mm/pull/1199/commits/b6f558d14e2ed917bb6bdf3b49b91084d64eb7a8

src/boot_O2_g3/fault.c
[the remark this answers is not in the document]
AngheloAlf: I think we should try to only use that macro for gamestates yeah

src/boot_O2_g3/fault_drawer.c
```diff
@@ -113,37 +149,39 @@ void FaultDrawer_DrawChar(char c) {
...
+void FaultDrawer_UpdatePrintColor(void) {
     s32 idx;
 
-    if (sFaultDrawContext->osSyncPrintfEnabled) {
+    if (sFaultDrawerInstance->osSyncPrintfEnabled) {
         osSyncPrintf(VT_RST);
-        idx = FaultDrawer_ColorToPrintColor(sFaultDrawContext->foreColor);
+
+        idx = FaultDrawer_ColorToPrintColor(sFaultDrawerInstance->foreColor);
         if (idx >= 0 && idx < 8) {
```
engineer124, line 167:
```suggestion
        if ((index >= 0) && (index < 8)) {
```
Just a small cleanup

engineer124 (approved):

src/boot_O2_g3/fault.c
```diff
@@ -573,354 +690,437 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+    for (line = 1; (line < 22) && (((ra != 0) || (sp != 0)) && (pc != (uintptr_t)__osCleanupThread)); line++) {
+        FaultDrawer_DrawText(0x24, line * 8 + 24, "%08x %08x", sp, pc);
 
         if (flags & 1) {
-            convertedPc = (u32)Fault_ConvertAddress((void*)pc);
-            if (convertedPc != 0) {
-                FaultDrawer_Printf(" -> %08x", convertedPc);
+            addr = Fault_ConvertAddress(pc);
+            if (addr != 0) {
+                FaultDrawer_Printf(" -> %08x", addr);
```
isghj5, line 826: I want some light documentation in this function explaining the purpose of this conversion from RAM back to VRAM for overlays, can just be a single comment line in the function header or in the block after the flag check.

src/boot_O2_g3/fault.c
```diff
@@ -573,354 +690,437 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+            Fault_DisplayFrameBuffer();
         }
-        iter = iter->next;
+        client = client->next;
     }
 }
 
-#ifdef NON_MATCHING
-// needs in-function static bss
 void Fault_SetOptionsFromController3(void) {
```
isghj5, line 917: What got it not matching? was it just using uintptr_t?

src/boot_O2_g3/fault_drawer.c
```diff
@@ -227,17 +266,21 @@ void* FaultDrawer_FormatStringFunc(void* arg, const char* str, size_t count) {
...
     va_list args;
+
     va_start(args, fmt);
 
-    FaultDrawer_VPrintf(fmt, args);
+    ret = FaultDrawer_VPrintf(fmt, args);
 
     va_end(args);
+
+    return ret;
```
isghj5, line 283: ... Really, IDO?

src/boot_O2_g3/fault.c
AngheloAlf: Nah, just importing bss and it worked

src/boot_O2_g3/fault.c
```diff
@@ -573,354 +690,439 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+    for (line = 1; (line < 22) && (((ra != 0) || (sp != 0)) && (pc != (uintptr_t)__osCleanupThread)); line++) {
+        FaultDrawer_DrawText(0x24, line * 8 + 24, "%08x %08x", sp, pc);
 
         if (flags & 1) {
-            convertedPc = (u32)Fault_ConvertAddress((void*)pc);
-            if (convertedPc != 0) {
-                FaultDrawer_Printf(" -> %08x", convertedPc);
+            // Try to convert the relocated program counter to the corresponding unrelocated virtual address
+            addr = Fault_ConvertAddress(pc);
+            if (addr != 0) {
```
isghj5, line 826:
```suggestion
            if (addr != NULL) {
```

src/boot_O2_g3/fault.c
```diff
@@ -573,354 +690,439 @@ void Fault_DrawMemDump(u32 pc, u32 sp, u32 unk0, u32 unk1) {
...
+    for (line = 1; (line < 22) && (((ra != 0) || (sp != 0)) && (pc != (uintptr_t)__osCleanupThread)); line++) {
         osSyncPrintf("%08x %08x", sp, pc);
 
         if (flags & 1) {
-            convertedPc = (u32)Fault_ConvertAddress((void*)pc);
-            if (convertedPc != 0) {
-                osSyncPrintf(" -> %08x", convertedPc);
+            // Try to convert the relocated program counter to the corresponding unrelocated virtual address
+            addr = Fault_ConvertAddress(pc);
+            if (addr != 0) {
```
isghj5, line 853:
```suggestion
            if (addr != NULL) {
```

src/boot_O2_g3/fault.c
AngheloAlf: No, this shouldn't use `NULL` since it isn't a pointer, it is a `uintptr_t`.
We may want to consider a `NULL` equivalent for `uintptr_t` tho

isghj5: ... but uintptr_t is a pointer type...?

And the data is a location in memory, (not actual RAM but VRAM) where zero is the error case and fits with C style null error conditions?

AngheloAlf: `uintptr_t` is a numeric type wide enough to hold a pointer, but it is not a pointer itself. Since `NULL` is a pointer type we would need to cast it like this
```c
if (addr != (uintptr_t)NULL) {
```
The point of `uintptr_t` is to be able to hold a pointer and do arithmetic on it

isghj5: I don't agree with this, I view it as a technicality, but it is not enough to stop me approving.

isghj5 (approved):

PR #1217, opened 2023-03-20
Add  `--convert-statics=global-with-filename` flag to asm-processor

AngheloAlf:
This flag makes the map file to actually have the static symbols and its corresponding addresses.

For example, having a `static` symbol like this:
```c
static struct_801F58B0 D_801F58B0[3][3];
```
Produces this in the map file
```
                0x00000000801f58b0                build/src/code/z_player_lib.o:D_801F58B0
```
Which should allow us to have less problems with static bss symbols.

It works by prefixing the filename to the symbol name.

It also works with in-function-static symbols

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ OBJCOPY    := $(MIPS_BINUTILS_PREFIX)objcopy
 OBJDUMP    := $(MIPS_BINUTILS_PREFIX)objdump
 ASM_PROC   := python3 tools/asm-processor/build.py
 
-ASM_PROC_FLAGS := --input-enc=utf-8 --output-enc=euc-jp
+ASM_PROC_FLAGS := --input-enc=utf-8 --output-enc=euc-jp --convert-statics=global-with-filename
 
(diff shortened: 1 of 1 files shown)


EllipticEllipsis (approved):
Worth noting that this does cause problems if there's more than one static of the same name in a file, which will probably cause problems in collision_check and sys_math3d.

AngheloAlf:
I think not being able to have static symbols with the same name in different functions of the same file is a worth tradeoff for having static symbols in the map file

EllipticEllipsis:
Yes, but this is something that could be fixed without too much trouble, I'd have thought, via line number or offset or something?

hensldm (approved):
Seems fine to me, though I do want to note that this does make us more reliant on asm-processor and so would be unable to remove it as easily in the future if we want. I don't think this is necessarily a bad thing, just want to mention it.

AngheloAlf:
I don't think it would make us more reliant on asm-processor, it would only help us identify static bss shifts. The only thing this change do is to make the `static` symbols to appear in the map file, nothing else

PR #1222, opened 2023-03-30
Load Docs

hensldm:
Add some more documentation for overlay loading (Thank you @Thar0 :) ). Also cleans up some macros.


AngheloAlf (approved):
Good job :D

isghj5 (approved):

tools/disasm/functions.txt
```diff
@@ -112,9 +112,9 @@
     0x800849A0:("Load_Relocate",),
     0x80084C0C:("Load_LoadOverlay",),
     0x80084CD0:("Load_AllocateAndLoad",),
-    0x80084DB0:("Load2_Relocate",),
-    0x8008501C:("Load2_LoadOverlay",),
-    0x800850C8:("Load2_AllocateAndLoad",),
+    0x80084DB0:("Overlay_Relocate",),
+    0x8008501C:("Overlay_Load",),
+    0x800850C8:("Overlay_AllocateAndLoad",),
```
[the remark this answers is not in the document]
hensldm: While I can agree on it being weird having completely different prefixes (though since one is completely unused I don't see it as important), I definitely would prefer consistency with OoT who use these function names. Though happy to take naming suggestions for the unused ones, maybe `UnusedOverlay_`?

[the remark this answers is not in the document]
AngheloAlf: We have been avoiding to name anything "unused", since the fact the system is unused is not a fundamental part of the system.
Since the system is unused, I guess we could have a name a bit longer too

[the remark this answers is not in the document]
EllipticEllipsis: > I don't see how `Fragment` is more descriptive than `Load` or `Overlay`, I think the rename is fine

Because it's what the chunks that are being loaded are actually called (for some reason). "Load" could load anything, "Overlay" isn't really correct either since `code` is just as much an overlay in that sense.

Since this system (and variants thereof) is used in other games, it makes sense to use the common terminology.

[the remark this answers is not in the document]
hensldm: So to be clear would you suggest going with `Fragment` and `Fragment2` for the system names, i.e.,
`Fragment_Relocate/Fragment2_Relocate`?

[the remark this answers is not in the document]
hensldm: So with that I don't mind using Fragment for the unused system (loadfragment.c), though I would still prefer `Overlay` for the used ones again for consistency with OoT.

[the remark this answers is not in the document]
EllipticEllipsis: Then at least add a note on what the known original name of the system is in the file header, and note in the file comment that it's specifically for the `fragment` type of overlays that have the .ovl section attached to the end, rather than any old overlay.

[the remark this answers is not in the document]
hensldm: So I added a few comments to help clarify. Let me know if that isn't enough.

src/boot_O2/loadfragment2.c
```diff
@@ -1,37 +1,78 @@
...
+#define MIPS_REG_RT(insn) (((insn) >> 0x10) & 0x1F)
+
+// Extract MIPS jump target from an instruction word
+#define MIPS_JUMP_TARGET(insn) (((insn)&0x03FFFFFF) << 2)
+
+/**
+ * Performs runtime relocation of overlay files, loadable code segments.
+ *
+ * Overlays are expected to be loadable anywhere in direct-mapped cached (KSEG0) memory, with some appropriate
+ * alignment requirements; memory addresses in such code must be updated once loaded in order to execute properly.
```
engineer124, line 28:
```suggestion
 * alignment requirements; memory addresses in such code must be updated once loaded to execute properly.
```
I find `in order to` can always be replaced by `to`

src/boot_O2/loadfragment.c
```diff
@@ -1,42 +1,81 @@
...
+#define MIPS_REG_RT(insn) (((insn) >> 0x10) & 0x1F)
+
+// Extract MIPS jump target from an instruction word
+#define MIPS_JUMP_TARGET(insn) (((insn)&0x03FFFFFF) << 2)
+
+/**
+ * Performs runtime relocation of overlay files, loadable code segments.
+ *
+ * Overlays are expected to be loadable anywhere in direct-mapped cached (KSEG0) memory, with some appropriate
+ * alignment requirements; memory addresses in such code must be updated once loaded in order to execute properly.
```
engineer124, line 31:
```suggestion
 * alignment requirements; memory addresses in such code must be updated once loaded to execute properly.
```

src/boot_O2/loadfragment2.c
```diff
@@ -1,37 +1,78 @@
...
+ * When compiled, overlays are given 'fake' KSEG0 RAM addresses larger than the total possible available main memory
+ * (>= 0x80800000), such addresses are referred to as Virtual RAM (VRAM) to distinguish them. When loading the overlay
+ * the relocation table produced at compile time is consulted to determine where and how to update these VRAM addresses
```
engineer124, line 31:
```suggestion
 * (>= 0x80800000), such addresses are referred to as Virtual RAM (VRAM) to distinguish them. When loading the overlay,
 * the relocation table produced at compile time is consulted to determine where and how to update these VRAM addresses
```
And in loadfragment.c

src/boot_O2/loadfragment2.c
```diff
@@ -51,10 +92,11 @@ void Load2_Relocate(void* allocatedVRamAddr, OverlayRelocationSection* ovl, uint
...
 
-                *relocDataP =
-                    (*relocDataP & 0xFC000000) |
-                    ((RELOCATE_ADDR(PHYS_TO_K0((*relocDataP & 0x03FFFFFF) << 2), vRamStart, allocu32) & 0x0FFFFFFF) >>
-                     2);
+                if (1) {
+                    *relocDataP =
+                        (*relocDataP & 0xFC000000) |
+                        (((PHYS_TO_K0(MIPS_JUMP_TARGET(*relocDataP)) - vramStart + allocu32) & 0x0FFFFFFF) >> 2);
+                }
```
engineer124, line 99: Why the extra `if (1) {}`? From debug? or needed with the change in macros?

hensldm: This is actually a rather interesting question. Elliptic and I actually went on adventure trying to match as many different versions of this particular function (spanning from Yoshi Story, OoT, MM, and Animal Forest). MM's debug actually has a slightly refactored version (which is why these are not just copy paste from OoT, since the GC rom would have used the new refactored version), but it did indeed require an `if(1)` (our theory is it was meant to be something like the other two cases that actually relocate and test against `0x0F00000` but was typo'd to always evaluate to true).  On the other hand, Yoshi Story JP (which seems to be the very first iteration, and is also slightly different, but more similar to what MM has) also requires an `if(1)` around the same area, so also suggests the same idea.

include/loadfragment.h
```diff
@@ -0,0 +1,42 @@
...
+    /* 4 */ RELOC_SECTION_MAX
+} RelocSectionId;
+
+typedef struct OverlayRelocationSection {
+    /* 0x00 */ u32 textSize;
+    /* 0x04 */ u32 dataSize;
+    /* 0x08 */ u32 rodataSize;
+    /* 0x0C */ u32 bssSize;
+    /* 0x10 */ u32 nRelocations;
+    /* 0x14 */ u32 relocations[1]; // size is nRelocations
```
engineer124, line 35:
```suggestion
    /* 0x10 */ u32 numRelocations;
    /* 0x14 */ u32 relocations[1]; // array count is numRelocations
```
Perhaps?

include/loadfragment.h
```diff
@@ -0,0 +1,42 @@
...
+    /* 2 */ RELOC_SECTION_DATA,
+    /* 3 */ RELOC_SECTION_RODATA,
+    /* 4 */ RELOC_SECTION_MAX
+} RelocSectionId;
+
+typedef struct OverlayRelocationSection {
+    /* 0x00 */ u32 textSize;
+    /* 0x04 */ u32 dataSize;
+    /* 0x08 */ u32 rodataSize;
+    /* 0x0C */ u32 bssSize;
```
engineer124, line 33: `size_t`?

include/loadfragment.h
```diff
@@ -0,0 +1,42 @@
+#ifndef LOADFRAGMENT_H
+#define LOADFRAGMENT_H
+
+#include "PR/ultratypes.h"
+#include "libc/stdint.h"
+#include "libc/stddef.h"
+
+extern s32 gOverlayLogSeverity;
```
engineer124, line 8: Just the smallest nit. I think we've been putting `externs` at the bottom of the file? I'm not sure if it's been consistent though so 🤷

engineer124 (approved):

include/loadfragment.h
hensldm: Yeah looks like we are pretty inconsistent. I would prefer leaving this here as is for now, and we can decide and clean them up later if that is okay.

PR #1231, opened 2023-04-22
Add a Couple Disasm Hacks (Fix NON_MATCHING Builds)

hensldm:
This adds a couple of disasm hacks to 1 fix up the collision check bss for non-matchings, and 2 remove a fake symbol from play.

Main motivation was to fix up the collision check bss, as when building with NON_MATCHING it would error out the first time due to duplicate bss symbols from the in static function ones. Since I needed to add a hack for that, I thought I would just add one to remove the fake symbol from play, rather than wait for the function to be matched.


AngheloAlf (approved):
LGTM

jpburnett (approved):

engineer124 (approved):

PR #1205, opened 2023-03-11
Some pathing docs

hensldm:
1. Names the two unk values in the `Path` struct
    - I'm not sure on `customValue` but can't really think of anything better. It is used as dayFlags in subs, pathtype for shooting gallery actors, and also seems to get compared to the current waypoint in `EnPr2`.
    - Documents a little for the shooting gallery actors that use `path.customValue`

2. Adds defines for each actor to indicate their NONE_PATH_INDEX (basically -1 to some bit width based on their own params), as well as two defines 1 for the general NONE_PATH_INDEX of just a regular -1, and one for the additional path index from scenes which is `-1` with 8 bits.

PR #1234, opened 2023-04-24
Document EnAob01 (doggy racetrack owner) and use macros for weekEventRegs in all dog racing files

tom-overton:
As part of this, I also applied some consistency to the horse racing state stuff and the talking to NPC weekEventReg flags. This was one of the most challenging actors I've ever had to document.

I'm not sure if others will really like the `BeforeRace`, `Race`, and `AfterRace` categories I've put on certain function names. This actor only sort of separates these various states, with some functions for some states scattered in random parts of the file. Let me know if you'd like me to remove this.


src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -91,105 +108,137 @@ static EnAobStruct D_809C384C[] = {
...
+    if (pathIndex != 0x3F) {
         do {
-            this->unk_1D8[i] = SubS_GetPathByIndex(play, temp_s0, 0x3F);
-            temp_s0 = this->unk_1D8[i]->unk1;
+            this->dogPaths[i] = SubS_GetPathByIndex(play, pathIndex, 0x3F);
+            pathIndex = this->dogPaths[i]->unk1;
             i++;
-        } while (temp_s0 != 0xFF);
+        } while (pathIndex != 0xFF);
     }
```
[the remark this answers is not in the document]
tom-overton: Which path macros? The ones in PR here? https://github.com/zeldaret/mm/pull/1205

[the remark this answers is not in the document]
tom-overton: As for the other part of your message, a plain-while adds this extra `beq` instruction:
image

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.h
```diff
@@ -8,38 +8,51 @@ struct EnAob01;
...
     /* 0x2DA */ Vec3s headRot;
     /* 0x2E0 */ Vec3s torsoRot;
-    /* 0x2E6 */ Vec3s unk_2E6;
-    /* 0x2EC */ Vec3s unk_2EC;
-    /* 0x2F2 */ Vec3s unk_2F2;
-    /* 0x2F8 */ s16 unk_2F8[MAMAMU_YAN_LIMB_MAX];
-    /* 0x318 */ s16 unk_318[MAMAMU_YAN_LIMB_MAX];
+    /* 0x2E6 */ Vec3s tempTrackTarget;
+    /* 0x2EC */ Vec3s tempHeadRot;
+    /* 0x2F2 */ Vec3s tempTorsoRot;
```
hensldm, line 42: I think we've generally used `prev` as the prefix for these kinds of variables. (e.g. EnBaba01)

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -488,124 +553,152 @@ void func_809C1EC8(EnAob01* this, PlayState* play) {
...
-    if ((this->unk_2D2 & 8) && !func_809C2EC4(this, play)) {
-        if ((this->unk_210 != 0) && (this->unk_210 != 0x3535) && (this->unk_210 != 0x3524) &&
-            (this->unk_210 != 0x3548) && (this->unk_210 != 0x3549) && (this->unk_210 != 0x354A)) {
-            this->unk_210 = 0x3523;
+    //! @bug: This block of code is *supposed* to act as a failsafe for when the player triggered this conversation by
+    //! bumping into the racetrack owner while holding the dog and, at the same time, dropped the dog frame-perfectly.
+    //! Without this code, doing those same actions will cause the ractrack owner be stuck in her "talk" action
+    //! permanently, so the developers probably put this code in to prevent it. However, this block of code, as-is,
+    //! causes a much worse problem; for some unknown reason, this code softlocks the game if the player manages to
+    //! perform the simultaneous actions described above.
```
hensldm, line 612: Perhaps it would be a good idea to mention what Zoey mentioned in discord.
https://discord.com/channels/688807550715560050/688851976746041391/1098248443291521044
TLDR. The code seems to work fine when you throw the dog, but if you shield drop it that is when there is an issue.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -488,124 +553,152 @@ void func_809C1EC8(EnAob01* this, PlayState* play) {
         Math_SmoothStepToS(&this->torsoRot.x, 0, 4, 0x3E8, 1);
         Math_SmoothStepToS(&this->torsoRot.y, 0, 4, 0x3E8, 1);
     }
-    func_809C10B0(this, 3);
-    SubS_FillLimbRotTables(play, this->unk_2F8, this->unk_318, ARRAY_COUNT(this->unk_2F8));
-    func_809C165C(this, play);
+
+    EnAob01_Blink(this, EN_AOB01_EYE_CLOSED + 1);
```
hensldm, line 557: Personally I think it would just be better to add a MAX value and use that.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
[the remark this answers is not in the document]
hensldm: It is probably because of the different comparisons being made. The first path index is coming from her individual paras, so is only 0x3F (5 bit -1) for the NONE_PATH_INDEX. On the other hand the additional path indices from the scene have a NONE_PATH_INDEX of 0xFF (8 bit -1). Technically the fact they use `0x3F` in the function call is wrong after the first iteration, since a none would be 0xFF at that point.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -76,12 +89,16 @@ static ColliderCylinderInit sCylinderInit = {
...
-    Vec3f unk_00;
-    s16 unk_04;
-    s16 unk_06;
-} EnAobStruct;
-
-static EnAobStruct D_809C384C[] = {
+    /* 0x0 */ Vec3f pos;
+    /* 0xC */ s16 rotY;
+    /* 0xE */ s16 pathIndex;
+} DogInfo;
```
hensldm, line 95:
```suggestion
} DogInfo; // size = 0x10
```

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
tom-overton: Whoops, I was going to fix up this comment, but I forgot to actually do that before pushing. I'll update it, thanks for the reminder.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -91,105 +109,137 @@ static EnAobStruct D_809C384C[] = {
     { { -4827.0f, 168.0f, 1328.0f }, 115, 4 }, { { -4130.0f, 150.0f, 1367.0f }, 112, 0 },
 };
 
-void func_809C10B0(EnAob01* this, s32 arg1) {
+void EnAob01_Blink(EnAob01* this, s32 indexToRestartBlink) {
```
hensldm, line 112: Perhaps `indexToRestartBlnk` -> `maxEyeIndex`? Up to you though

hensldm (approved):

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
tom-overton: Very good point. I didn't know what to name this, but now that we made `EN_AOB01_EYE_MAX`, it's pretty obvious.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -629,156 +728,205 @@ s32 func_809C2594(EnAob01* this, PlayState* play) {
...
     while (npc != NULL) {
         if ((npc->id == ACTOR_EN_RACEDOG) && (((EnRacedog*)npc)->raceStatus == RACEDOG_RACE_STATUS_FINISHED)) {
-            count++;
+            finishedCount++;
         }
         npc = npc->next;
     }
 
-    if (count >= 14) {
+    if (finishedCount >= 14) {
```
engineer124, line 748:
```suggestion
    if (finishedCount >= ARRAY_COUNT(sDogInfo)) {
```
If you were to add a dog, this number should be updated right?

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.h
```diff
@@ -8,38 +8,51 @@ struct EnAob01;
...
+    /* 0x2F2 */ Vec3s prevTorsoRot;
+    /* 0x2F8 */ s16 limbRotTableY[MAMAMU_YAN_LIMB_MAX];
+    /* 0x318 */ s16 limbRotTableZ[MAMAMU_YAN_LIMB_MAX];
     /* 0x338 */ UNK_TYPE1 unk338[0xB6];
     /* 0x3EE */ s16 eyeIndex;
     /* 0x3F0 */ s16 blinkTimer;
-    /* 0x3F4 */ Actor* unk_3F4;
-    /* 0x3F8 */ Actor* unk_3F8[14];
+    /* 0x3F4 */ Actor* dogToFollow;
+    /* 0x3F8 */ Actor* dogs[14];
```
engineer124, line 49: In fact, instead of `ARRAY_COUNT(sDogInfo)`, should you just have a `#define RACEDOG_COUNT 14`, 
```suggestion
    /* 0x3F8 */ Actor* dogs[RACEDOG_COUNT];
```
Then you can do: 
`static DogInfo sDogInfo[RACEDOG_COUNT] = {`
`ARRAY_COUNT(sDogInfo)` -> `RACEDOG_COUNT`
`ARRAY_COUNT(this->dogs)` -> `RACEDOG_COUNT`
More related to this on my next comment

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -885,40 +1050,46 @@ void func_809C2F34(EnAob01* this, PlayState* play) {
...
-    u8 index;
-    u8 orig;
-    u8 orig2;
-    u8 sp44[7];
-    u8 sp34[] = {
+    u8 dogTextOffsetsTemp;
+    u8 packedDogTextOffsets[7];
+    u8 dogTextOffsets[] = {
         0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13,
     };
```
engineer124, line 1066:
```suggestion
    u8 packedDogTextOffsets[RACEDOG_COUNT / 2];
    u8 dogTextOffsets[RACEDOG_COUNT] = {
        0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13,
    };
```
Then you could do something like this.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -885,40 +1050,46 @@ void func_809C2F34(EnAob01* this, PlayState* play) {
...
 
-    for (i = 0; i < ARRAY_COUNT(sp34); i++) {
+    // Swap around the dog text offsets randomly 14 times.
+    for (i = 0; i < ARRAY_COUNT(dogTextOffsets); i++) {
         rand = Rand_ZeroFloat(14.0f);
```
engineer124, line 1070:
```suggestion
    for (i = 0; i < RACEDOG_COUNT; i++) {
        rand = Rand_ZeroFloat(RACEDOG_COUNT);
```
And even this

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -885,40 +1050,46 @@ void func_809C2F34(EnAob01* this, PlayState* play) {
...
     }
 
-    for (i = 0; i < ARRAY_COUNT(sp44); i++) {
-        gSaveContext.save.saveInfo.weekEventReg[42 + i] = 0;
-        sp44[i] = 0;
+    // Clear out all weekEventRegs related to the dog text offsets.
+    for (i = 0; i < ARRAY_COUNT(packedDogTextOffsets); i++) {
+        CLEAR_WEEKEVENTREG_DOG_RACE_TEXT_OFFSETS(i);
+        packedDogTextOffsets[i] = 0;
     }
```
engineer124, line 1081: And would need to add a comment to the WEEKEVENTREG's (and the define of `RACEDOG_COUNT`) saying the number of WEEKEVENTREG's needs to be in sync with `RACEDOG_COUNT`? (sorry this is really just 1 long comment spread out over many comments)

engineer124 (commented):
Nice docs!

include/z64save.h
```diff
@@ -1561,6 +1577,25 @@ typedef enum {
 #define CLEAR_EVENTINF(flag) (gSaveContext.eventInf[(flag) >> 4] &= (u8)~(1 << ((flag) & 0xF)))
 #define CLEAR_EVENTINF_ALT(flag) (gSaveContext.eventInf[(flag) >> 4] &= ~(1 << ((flag) & 0xF)))
 
+typedef enum {
+    /* 0 */ EVENTINF_DOG_RACE_STATE_NOT_STARTED,
+    /* 2 */ EVENTINF_DOG_RACE_STATE_STARTED = 2,
+    /* 3 */ EVENTINF_DOG_RACE_STATE_ENDED,
+    /* 7 */ EVENTINF_DOG_RACE_STATE_MAX = 7
+} EventInfDogRaceState;
```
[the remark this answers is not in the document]
tom-overton: It should just be `EVENTINF_00` I think? But this is actually used in `z_game_over.c`:
```c
            gSaveContext.unk_3DC0 = 2000;
            gSaveContext.save.saveInfo.playerData.tatlTimer = 0;
            gSaveContext.seqId = (u8)NA_BGM_DISABLED;
            gSaveContext.ambienceId = AMBIENCE_ID_DISABLED;
            gSaveContext.eventInf[0] = 0;
            gSaveContext.eventInf[1] = 0;
            gSaveContext.eventInf[2] = 0;
            gSaveContext.eventInf[3] = 0;
```
So it shouldn't be commented out. I'll use this constant in the right places.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
tom-overton: This is clever, but unfortunately doesn't function properly in practice. The code makes the assumption that the number of dogs is even. It is particularly this code that is problematic:
```c
        if (i % 2) {
            packedDogTextOffsets[j] |= dogTextOffsetsTemp << 4;
            SET_WEEKEVENTREG_DOG_RACE_TEXT_OFFSETS(j, packedDogTextOffsets[j]);
        } else {
            packedDogTextOffsets[j] |= dogTextOffsetsTemp;
        }
```
Imagine, for example, that you bumped the number of dogs from 14 to 15. This means the indices will be 0, 1, 2, 3, ..., 12, 13, and 14. What happens when the index is 14? It will go to the second part of that if-else block and set the value in `packedDogTextOffsets`, but since there is no index 15 to follow, it will *never* actually write this to the weekEventRegs. I don't think this is fixable without breaking matching.

Given this, do you still want to use `RACEDOG_COUNT` here? Maybe with a comment that this function as-is only works when it's even?

src/code/flg_set.c
```diff
@@ -114,14 +114,14 @@ static FlagSetEntry sFlagEntries[] = {
...
-    { &gSaveContext.eventInf[6], "event_inf[6]" },
-    { &gSaveContext.eventInf[7], "event_inf[7]" },
+    { &gSaveContext.eventInf[EVENTINF_00], "event_inf[0]" },
+    { &gSaveContext.eventInf[EVENTINF_01], "event_inf[1]" },
+    { &gSaveContext.eventInf[EVENTINF_02], "event_inf[2]" },
+    { &gSaveContext.eventInf[EVENTINF_03], "event_inf[3]" },
+    { &gSaveContext.eventInf[EVENTINF_04], "event_inf[4]" },
+    { &gSaveContext.eventInf[EVENTINF_05], "event_inf[5]" },
+    { &gSaveContext.eventInf[EVENTINF_06], "event_inf[6]" },
+    { &gSaveContext.eventInf[EVENTINF_07], "event_inf[7]" },
```
AngheloAlf, line 124: The EVENTINF macros are bitpacked macros like the weekEventReg ones, so just using those 8 macros would be wrong.
See `CHECK_EVENTINF` and similar macros

src/overlays/actors/ovl_En_Go/z_en_go.c
```diff
@@ -1278,7 +1278,7 @@ void func_80A14018(EnGo* this, PlayState* play) {
     s32 phi_v0 = ENGO_GET_70(&this->actor) % 6;
 
     if (phi_v0 < 4) {
-        phi_v0 = ((gSaveContext.eventInf[2] & 0xF) + phi_v0) % 4;
+        phi_v0 = ((gSaveContext.eventInf[EVENTINF_02] & 0xF) + phi_v0) % 4;
```
AngheloAlf, line 1281: Same here.

src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
```diff
@@ -440,7 +440,8 @@ void EnMttag_PotentiallyRestartRace(EnMttag* this, PlayState* play) {
             CLEAR_EVENTINF(EVENTINF_11);
             CLEAR_EVENTINF(EVENTINF_12);
             CLEAR_EVENTINF(EVENTINF_13);
-            gSaveContext.eventInf[2] = ((gSaveContext.eventInf[2] & 0xF) + 1) | (gSaveContext.eventInf[2] & 0xF0);
+            gSaveContext.eventInf[EVENTINF_02] =
+                ((gSaveContext.eventInf[EVENTINF_02] & 0xF) + 1) | (gSaveContext.eventInf[EVENTINF_02] & 0xF0);
```
AngheloAlf, line 444: and here

include/z64save.h
```diff
@@ -1584,18 +1528,23 @@ typedef enum {
...
+         (gSaveContext.eventInf[EVENTINF_00] & EVENTINF_DOG_RACE_STATE_MAX) | (selectedDogIndex << 3))
 
 // This is exactly the same as the selected dog index; they never need both of them at the same time.
-#define GET_EVENTINF_DOG_RACE_RACE_STANDING ((gSaveContext.eventInf[0] & (u8)~EVENTINF_DOG_RACE_STATE_MAX) >> 3)
+#define GET_EVENTINF_DOG_RACE_RACE_STANDING \
+    ((gSaveContext.eventInf[EVENTINF_00] & (u8)~EVENTINF_DOG_RACE_STATE_MAX) >> 3)
 #define SET_EVENTINF_DOG_RACE_RACE_STANDING(raceStanding) \
-    (gSaveContext.eventInf[0] = (gSaveContext.eventInf[0] & EVENTINF_DOG_RACE_STATE_MAX) | (raceStanding << 3))
+    (gSaveContext.eventInf[EVENTINF_00] =                 \
+         (gSaveContext.eventInf[EVENTINF_00] & EVENTINF_DOG_RACE_STATE_MAX) | (raceStanding << 3))
```
AngheloAlf: Same comment as before about the EVENTINF macros.

src/code/flg_set.c
tom-overton: No, I think in this case, it's correct. It's *not* looking at the bitpacked values or anything here. The `CHECK_EVENTINF` macro looks like this:
```c
#define CHECK_EVENTINF(flag) (gSaveContext.eventInf[(flag) >> 4] & (1 << ((flag) & 0xF)))
```

Which is way more than what this is doing.

src/overlays/actors/ovl_En_Go/z_en_go.c
tom-overton: I'm fairly sure that this isn't bitpacked either. There's no shifting going on here.

include/z64save.h
tom-overton: And yeah, same thing here. The macros won't work for this one either.

src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
tom-overton: Well, this tells us that this *isn't* packed in the same way as other eventInf things. It's showing that it's packing two four-bit values into a single byte, similar to the dog text ID offset stuff.

src/code/flg_set.c
hensldm: Feel like we should copy
```c
#define WEEKEVENTREG(index) (gSaveContext.save.saveInfo.weekEventReg[(index)])
#define GET_WEEKEVENTREG(index) ((void)0, WEEKEVENTREG(index))
```
but for event infos, for these instances where they do not seem to be bitpacked like normal. Mainly because EVENTINF_02 doesn't mean eventInfo[2] and so would be named differently.

src/overlays/actors/ovl_En_Go/z_en_go.c
hensldm: This looks to be using `EVENTINF_20` through `EVENTINF_23`.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
engineer124: Oh that's so silly, so yeah it has to be even. I think I'd still use `RACEDOG_COUNT` with a note like this if I understand correctly? (can also write a better comment too)
```c
        if (i % 2) {
            packedDogTextOffsets[j] |= dogTextOffsetsTemp << 4;
            //! @note sets text offsets for both even and odd entry.
            //! Requires `RACEDOG_COUNT` to be even to work properly
            SET_WEEKEVENTREG_DOG_RACE_TEXT_OFFSETS(j, packedDogTextOffsets[j]);
        } else {
            packedDogTextOffsets[j] |= dogTextOffsetsTemp;
        }
```

src/code/flg_set.c
AngheloAlf: > No, I think in this case, it's correct. It's _not_ looking at the bitpacked values or anything here. The `CHECK_EVENTINF` macro looks like this:
> 
> ```c
> #define CHECK_EVENTINF(flag) (gSaveContext.eventInf[(flag) >> 4] & (1 << ((flag) & 0xF)))
> ```
> 
> Which is way more than what this is doing.

No, it isn't correct. Again, the macros are bitpacked. Look at it this way:
- Each macro consists on a 1 byte number (2 [nibbles](https://en.wikipedia.org/wiki/Nibble)) which can be seen as `0xXY`
- The `X` part of the macro is meant to be used to index the `eventInf` array
- The `Y` part is used _as a shift_ value to mask the the actual value contained on each entry of the array.

Knowing this, let's take a look at the macros you are using:
```c
#define EVENTINF_00 0x00
#define EVENTINF_01 0x01
#define EVENTINF_02 0x02
#define EVENTINF_03 0x03
#define EVENTINF_04 0x04
#define EVENTINF_05 0x05
#define EVENTINF_06 0x06
#define EVENTINF_07 0x07
```
All of them have an index value (the `X` part) as zero, so all of them are meant to be used to access the entry 0 of the array. But you are using the shift part of the macro (the `Y` part) to index the array, which is not how those macros are meant to be used.

Is it clear now?

src/overlays/actors/ovl_En_Go/z_en_go.c
AngheloAlf: To be clear, I'm not saying the actual access to the eventInf array here is bitpacked, but the `EVENTINF_02` isn't meant to be used in this way.

This specific line of code is accessing the values covered by the following macros
```c
#define EVENTINF_20 0x20
#define EVENTINF_21 0x21
#define EVENTINF_22 0x22
#define EVENTINF_23 0x23
```

It is using a range of the eventInf flags to store data that are not flag, like the Doggy racetrack does to store the textId in the weekEventReg, or the horse race flags

src/code/flg_set.c
tom-overton: Yeah, this was a misunderstanding on my part. I thought each of these was used to represent one *byte* in the `eventInf` array, but that's not how it works. I'll fix it up.

tom-overton: I ended up just reverting all this stuff back to raw integers. This PR is already pretty complicated, so I didn't see a lot of value in making it even more complicated. For now, all I did is comment out `EVENTINF_00` to `EVENTINF_07`, with a comment above saying the whole byte is used for dog racetrack info.

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -488,124 +554,152 @@ void func_809C1EC8(EnAob01* this, PlayState* play) {
...
 
-    if ((this->unk_2D2 & 8) && !func_809C2EC4(this, play)) {
-        if ((this->unk_210 != 0) && (this->unk_210 != 0x3535) && (this->unk_210 != 0x3524) &&
-            (this->unk_210 != 0x3548) && (this->unk_210 != 0x3549) && (this->unk_210 != 0x354A)) {
-            this->unk_210 = 0x3523;
+    //! @bug: This block of code acts as a failsafe for when the player triggered this conversation by
+    //! bumping into the racetrack owner while holding the dog and, at the same time, threw or dropped
+    //! the dog frame-perfectly by pressing A. In this case, the code successfully returns the racetrack
+    //! owner to her idle state. Unfortunately, there is another way to let go of the dog besides pressing
+    //! A to drop or throw it; the player can also drop the dog by shielding. If this shield drop is done
```
engineer124, line 612:
```suggestion
    //! owner to her idle state. Unfortunately, the player can also drop the dog by shielding. If this shield drop is done
```
I think part of this is fluff, and this still gets the point across

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -629,156 +728,205 @@ s32 func_809C2594(EnAob01* this, PlayState* play) {
     return false;
 }
 
-s32 func_809C25E4(EnAob01* this, PlayState* play) {
+/**
+ * Returns true if the race should be ended, which can happen in either of the following circumstances:
+ * - All 14 dogs have finished the race.
```
engineer124, line 733:
```suggestion
 * - All `RACEDOG_COUNT` dogs have finished the race.
```

src/overlays/actors/ovl_En_Aob_01/z_en_aob_01.c
```diff
@@ -885,40 +1050,48 @@ void func_809C2F34(EnAob01* this, PlayState* play) {
...
+        u8 dogTextOffsetsTemp = dogTextOffsets[i];
+        u8 j = i / 2;
 
         if (i % 2) {
-            sp44[index] |= orig2 << 0x4;
-            gSaveContext.save.saveInfo.weekEventReg[42 + index] =
-                ((void)0, gSaveContext.save.saveInfo.weekEventReg[42 + index]) | sp44[index];
+            //! @note sets text offsets for both even and odd entry.
+            //! Requires `RACEDOG_COUNT` to be even to work properly.
+            packedDogTextOffsets[j] |= dogTextOffsetsTemp << 4;
```
engineer124, line 1091: I know some comments would be duplicated, but I think it's worth mentioning this comment at the definition of `RACEDOG_COUNT` (Requires `RACEDOG_COUNT` to be even to work properly.`)

engineer124 (commented):
Final nitpick comments

engineer124 (approved):

PR #1235, opened 2023-04-30
Play_Hireso (Bombers Notebook) OK w/ Docs

hensldm:
Thanks Petrie for helping get it over the finish line.

Something interesting is there are quite a few arrays that need to be kept in sync if the entries are messed with. These would be `sBombersNotebookEventIcons`, `sBombersNotebookEventColorWeekEventFlags`, `sBombersNotebookTextIds`, and `gBombersNotebookWeekEventFlags` (which is in message). The last 2 in particular need to be updated as well if any new rows are added as those two hold both people and event information.

A lot of the names are rather long due to `BombersNotebook` being such a long prefix but don't have any idea to help with that.

 A little something that may seem odd is that for the entry macros I just use the `sBombersNotebookEntries` array directly rather than pass an entry as an argument. This significantly decreases the length of code on a single line, but perhaps we should just pass the entry anyways. This was all done in the (https://github.com/zeldaret/mm/pull/1235/commits/680b4028a1c81e64089838d307f8a8a84d208da8) commit, so could look at the diff to see how it looked like before.


include/PR/gbi.h
```diff
@@ -4699,7 +4699,7 @@ _DW({									\
 
 /* like gSPTextureRectangle but accepts negative position arguments */
 #define gSPScisTextureRectangle(pkt, xl, yl, xh, yh, tile, s, t, dsdx, dtdy) \
-{                                                                            \
+_DW({                                                                        \
```
AngheloAlf, line 4702: despair

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1450 @@
...
+};
+
+void BombersNotebook_DrawHeaders(Gfx** gfxP) {
+    Gfx* gfx = *gfxP;
+    s32 i;
+    s32 rectLeft;
+
+    gDPLoadTextureBlock(gfx++, gBombersNotebookHeaderBoxTex, G_IM_FMT_I, G_IM_SIZ_8b, 8, 24, 0,
+                        G_TX_NOMIRROR | G_TX_WRAP, G_TX_NOMIRROR | G_TX_WRAP, 3, G_TX_NOMASK, G_TX_NOLOD, G_TX_NOLOD);
+    for (i = 0, rectLeft = 120; i < 3; i++, rectLeft += 150) {
```
AngheloAlf, line 384:
```suggestion
    for (i = 0, rectLeft = 120; i < ARRAY_COUNT(sBombersNotebookHeaderColors); i++, rectLeft += 150) {
```

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1450 @@
...
+        gDPSetPrimColor(gfx++, 0, 0, sBombersNotebookHeaderColors[i][0], sBombersNotebookHeaderColors[i][1],
+                        sBombersNotebookHeaderColors[i][2], 192);
+        BombersNotebook_TextureRectangle(&gfx, rectLeft * 4, 74 * 4, (rectLeft + 143) * 4, 98 * 4, 0, 0, 0, 0x400,
+                                         0x400);
+    }
+
+    gDPSetPrimColor(gfx++, 0, 0, 255, 255, 255, 255);
+    for (i = 0, rectLeft = 124; i < 3; i++, rectLeft += 150) {
```
AngheloAlf, line 392:
```suggestion
    for (i = 0, rectLeft = 124; i < ARRAY_COUNT(sBombersNotebookDayTextures); i++, rectLeft += 150) {
```

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1450 @@
...
+    if (CURRENT_DAY == 0) {
+        onesDigit = 6;
+    }
+    hours = onesDigit;
+    do {
+        if (onesDigit >= 10) {
+            tensDigit++;
+            onesDigit -= 10;
+        }
+    } while (onesDigit >= 10);
```
AngheloAlf, line 852: err, what?

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1450 @@
...
+
+void BombersNotebook_Destroy(BombersNotebook* this) {
+    if (this->loadState == BOMBERS_NOTEBOOK_LOAD_STATE_STARTED) {
+        BombersNotebook_LoadFiles(this, OS_MESG_BLOCK);
+    }
+    if (this->scheduleDmaSegment != NULL) {
+        ZeldaArena_Free(this->scheduleDmaSegment);
+        this->scheduleDmaSegment = NULL;
+    }
+    //! @bug: Does not free malloc'd memory for schedule segment
```
AngheloAlf, line 1449: I wonder if opening and closing the bombers notebook many times could crash the game because of a memory leak

AngheloAlf:
> Something interesting is there are quite a few arrays that need to be kept in sync if the entries are messed with. These would be sBombersNotebookEventIcons, sBombersNotebookEventColorWeekEventFlags, sBombersNotebookTextIds, and gBombersNotebookWeekEventFlags (which is in message). The last 2 in particular need to be updated as well if any new rows are added as those two hold both people and event information.

something something bombers notebook table something

assets/xml/interface/schedule_static.xml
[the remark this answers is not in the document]
hensldm: Yeah it seems fine. Interface total stays the same, and this adds to the debinarized and described stats.

include/PR/gbi.h
hensldm: Fwiw this is the current only use in MM, and it doesn't get used at all in OoT.

src/code/z_play_hireso.c
hensldm: This is equivalent to 
```c
tensDigit = onesDigit / 10;
onesDigit = onesDigit % 10;
```
It is just getting the digit for each place to grab the correct digit texture. Though not sure why the if is inside the do while, rather than just a normal while loop.

src/code/z_play_hireso.c
hensldm: Destroy actually doesn't get called when you just close the notebook. It gets called in play destroy. So you have to open the notebook, close, and than force play to reload.

hensldm:
> > Something interesting is there are quite a few arrays that need to be kept in sync if the entries are messed with. These would be sBombersNotebookEventIcons, sBombersNotebookEventColorWeekEventFlags, sBombersNotebookTextIds, and gBombersNotebookWeekEventFlags (which is in message). The last 2 in particular need to be updated as well if any new rows are added as those two hold both people and event information.
> 
> something something bombers notebook table something

I added a couple tables let me know what you think. (when this gets in I could also update the message branch to use the tables). Also thanks Zoey for the bombers notebook docs in the message branch, of which I have taken quite a bit and added them here.

src/code/z_play_hireso.c
mzxrules: I don't think a memory leak is possible because Play_Init calls ZeldaArena_Init which I believe should free the whole thing for reuse.

include/tables/bombers_notebook/event_table.h
```diff
@@ -0,0 +1,46 @@
+/**
+ * Bombers Notebook Event Table
+ *
+ * DEFINE_EVENT should be used for events in the notebook
+ *    - Argument 0: Enum value for the event (without the pfefix)
```
engineer124, line 5:
```suggestion
 *    - Argument 0: Enum value for the event (without the prefix)
```
imo without the prefix makes searching a bit more annoying, because for `BOMBERS_NOTEBOOK_EVENT_RECEIVED_ROOM_KEY`, you wouldn't find the source immediately, then you'd have to search the substring `EVENT_RECEIVED_ROOM_KEY` which picks up a few other defines as well as this one.

I can see why it's done for the `DEFINE_PERSON` though since you're trying to define two enums with 1 name. But then should `BombersNotebookPerson` be a different enum to `BombersNotebookEvent`? (I haven't looked through the system yet so maybe it will answer itself).

edit: coming back to this later, you could also drop the `WEEKEVENTREG_` from the `WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_RECEIVED_ROOM_KEY` for both `event_table.h` and `person_table.h` and reuse that define for both `WEEKEVENTREG` and the enum introduced in the message branch `BombersNotebookEvent` (see https://github.com/Decompollaborate/mm/commit/c76ad75e42ee4da5f0dc4b6fcc390593563365b8)
For that reason, should the contents of Zoey's PR be included in this PR since they interlink? (i.e. add in that enum, name `D_801C6AB8`, and the docs for the two `BombersNotebookQueueEvent` functions)?

include/tables/bombers_notebook/event_table.h
```diff
@@ -0,0 +1,46 @@
...
+ *    - Argument 1: Icon to use (see BombersNotebookEventIcon)
+ *    - Argument 2: WeekEventFlag that updates the icons color
+ *    - Argument 3: Message id for the description of this event
+ *    - Argument 4: Message id for the completion of this event
+ *    - Argument 5: WeekEventFlag for the completion of this event
+ */
+/* 0x00 */ DEFINE_EVENT(RECEIVED_ROOM_KEY, BOMBERS_NOTEBOOK_EVENT_ICON_EXCLAMATION_POINT, WEEKEVENTREG_75_10, 0x2198, 0x2152, WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_RECEIVED_ROOM_KEY)
```
engineer124, line 12: Could these weekEventFlags/arg 5 be named here based on the event? Or would you rather do that separately? (i.e. `WEEKEVENTREG_75_10` and others)

include/z64bombers_notebook.h
```diff
@@ -0,0 +1,58 @@
...
+
+#undef DEFINE_PERSON
+#undef DEFINE_EVENT
+
+typedef struct {
+    /* 0x00 */ u8 loadState;
+    /* 0x01 */ UNK_TYPE1 pad01[0x3F];
+    /* 0x40 */ void* scheduleDmaSegment;
+    /* 0x44 */ u32 scheduleDmaSegmentStart;
+    /* 0x48 */ u32 scheduleDmaSegmentSize;
```
engineer124, line 38:
```suggestion
    /* 0x44 */ uintptr_t scheduleDmaSegmentStart;
    /* 0x48 */ size_t scheduleDmaSegmentSize;
```

src/code/z_play.c
```diff
@@ -1052,14 +1053,14 @@ void Play_Update(PlayState* this) {
     if (!sBombersNotebookOpen) {
         if (this->pauseCtx.bombersNotebookOpen) {
             sBombersNotebookOpen = true;
-            sBombersNotebook.unk_00 = 0;
+            sBombersNotebook.loadState = 0;
```
engineer124, line 1056:
```suggestion
            sBombersNotebook.loadState = BOMBERS_NOTEBOOK_LOAD_STATE_NONE;
```
Same below in z_play

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+        }
+        BombersNotebook_DrawArrows(this, &gfx);
+    }
+    POLY_OPA_DISP = gfx;
+
+    CLOSE_DISPS(gfxCtx);
+}
+
+void BombersNotebook_LoadFiles(BombersNotebook* this, s32 flag) {
+    switch (this->loadState) {
```
engineer124, line 1114: default case?

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
 #include "global.h"
+#include "z64bombers_notebook.h"
+#include "interface/schedule_static/schedule_static.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/func_8016AC10.s")
+// TODO: Create xml when build supports yaz0 compression/decompression
```
engineer124, line 5: Should make identical comments for other places that is waiting for the `yaz0` scripts, not a bit deal tho

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+    /* 1 */ BOMBERS_NOTEBOOK_EVENT_ICON_MASK,
+    /* 2 */ BOMBERS_NOTEBOOK_EVENT_ICON_RIBBON
+} BombersNotebookEventIcon;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/func_8016CD4C.s")
+#define BOMBERS_NOTEBOOK_EVENT_COLOR_WEEKEVENTREG_NONE 0xFFF0
+#define BOMBERS_NOTEBOOK_EVENT_COLOR_WEEKEVENTREG_DELIVERED_LETTER 0xFFF1
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_play_hireso/func_8016E40C.s")
+u16 sBombersNotebookEntries[BOMBERS_NOTEBOOK_PERSON_MAX][10 * BOMBERS_NOTEBOOK_ENTRY_SIZE] = {
```
engineer124, line 53: Could make a define for 10, like `BOMBERS_NOTEBOOK_ENTRY_MAX` or something?

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+
+    gDPPipeSync(gfx++);
+    *gfxP = gfx;
+}
+
+s16 sBombersNotebookColumnColors[][2][3] = {
+    { { 165, 183, 195 }, { 140, 158, 170 } },
+    { { 197, 195, 172 }, { 172, 170, 147 } },
+    { { 223, 190, 200 }, { 190, 165, 175 } },
+};
```
engineer124, line 360: The third dimension here feels unnecessary? Could you post a scratch of `BombersNotebook_DrawColumns`, is it really that bad?

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+            yOffset = 16;
+        }
+
+        startTime = BOMBERS_NOTEBOOK_ENTRY_GET_START_TIME(row, j) - CLOCK_TIME(6, 0);
+        endTime = BOMBERS_NOTEBOOK_ENTRY_GET_END_TIME(row, j) - CLOCK_TIME(6, 0);
+        entryRectLeft = sBombersNotebookDayRectRectLeft[BOMBERS_NOTEBOOK_ENTRY_GET_DAY(row, j)] + (startTime / 455);
+        entryRectRight = sBombersNotebookDayRectRectLeft[BOMBERS_NOTEBOOK_ENTRY_GET_DAY(row, j)] + (endTime / 455);
+        entryWidth = entryRectRight - entryRectLeft - 8;
+        if ((entryRectRight - entryRectLeft) < 8) {
+            entryRectLeft = ((entryRectLeft + entryRectRight) - entryRectLeft) - 8;
```
engineer124, line 456: Would this line not just be `entryRectRight - 8`?

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+        } else if (sBombersNotebookEntries[row][j] & BOMBERS_NOTEBOOK_ENTRY_POS_BELOW) {
+            yOffset = 24;
+        } else { // BOMBERS_NOTEBOOK_ENTRY_POS_CENTER
+            yOffset = 16;
+        }
+
+        startTime = BOMBERS_NOTEBOOK_ENTRY_GET_START_TIME(row, j) - CLOCK_TIME(6, 0);
+        endTime = BOMBERS_NOTEBOOK_ENTRY_GET_END_TIME(row, j) - CLOCK_TIME(6, 0);
+        entryRectLeft = sBombersNotebookDayRectRectLeft[BOMBERS_NOTEBOOK_ENTRY_GET_DAY(row, j)] + (startTime / 455);
+        entryRectRight = sBombersNotebookDayRectRectLeft[BOMBERS_NOTEBOOK_ENTRY_GET_DAY(row, j)] + (endTime / 455);
```
engineer124, line 453:
```suggestion
        entryRectLeft = sBombersNotebookDayRectRectLeft[BOMBERS_NOTEBOOK_ENTRY_GET_DAY(row, j)] + (startTime / CLOCK_TIME(0, 10));
        entryRectRight = sBombersNotebookDayRectRectLeft[BOMBERS_NOTEBOOK_ENTRY_GET_DAY(row, j)] + (endTime / CLOCK_TIME(0, 10));
```
I think this seems reasonable? There's a few `/ 455` that this could apply to

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+                gDPPipeSync(gfx++);
+                gDPSetCombineLERP(gfx++, PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT, TEXEL0, 0, PRIMITIVE, 0,
+                                  PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT, TEXEL0, 0, PRIMITIVE, 0);
+                gDPSetPrimColor(gfx++, 0, 0, 0, 0, 0, 255);
+                gDPSetEnvColor(gfx++, 0, 0, 0, 255);
+                gDPLoadTextureBlock(gfx++, gBombersNotebookStampTex, G_IM_FMT_IA, G_IM_SIZ_8b, 32, 48, 0,
+                                    G_TX_MIRROR | G_TX_WRAP, G_TX_MIRROR | G_TX_WRAP, G_TX_NOMASK, G_TX_NOMASK,
+                                    G_TX_NOLOD, G_TX_NOLOD);
+                BombersNotebook_TextureRectangle(&gfx, 567 * 4, (rectTop + 1) * 4, 599 * 4, (rectTop + 49) * 4, 0, 0, 0,
+                                                 0x400, 0x400);
```
engineer124, line 429: `0x400` can be `1 << 10` in many places. Also, I think `567 * 4` to `567 << 2` should be fine for constants, and leave `* 4` only when necessary for matching, which I acknowledge is in quite a few places. I think that's what @Thar0 prefers after dealing with this a lot in z_parameter

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+    } else {
+        sBarColorG += colorStep;
+    }
+
+    colorStep = ABS_ALT(sBarColorB - sBarColorTargetsB[sBarColorIndex]) / sBarColorTimer;
+    if (sBarColorB >= sBarColorTargetsB[sBarColorIndex]) {
+        sBarColorB -= colorStep;
+    } else {
+        sBarColorB += colorStep;
+    }
```
engineer124, line 616: I can see this was probably the interface dev then. A landmark styling of theirs

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+
+TexturePtr sBombersNotebookDigitTextures[] = {
+    gBombersNotebookDigit0Tex, gBombersNotebookDigit1Tex, gBombersNotebookDigit2Tex, gBombersNotebookDigit3Tex,
+    gBombersNotebookDigit4Tex, gBombersNotebookDigit5Tex, gBombersNotebookDigit6Tex, gBombersNotebookDigit7Tex,
+    gBombersNotebookDigit8Tex, gBombersNotebookDigit9Tex,
+};
+u8 sBombersNotebookDigitRectLeftOffsets[] = { 13, 8, 12, 13, 14, 13, 12, 12, 13, 12 };
+s16 sBombersNotebookDayTexRectLeftOffsets[] = { 16, 12, 9 };
+
+#define CURRENT_DAY_MIN_1 ((CURRENT_DAY == 0) ? 1 : CURRENT_DAY)
```
engineer124, line 701: I would have expected `CURRENT_DAY_MIN_1` to be `(CURRENT_DAY - 1)` (current day minus 1), what's the naming intention here?

src/code/z_play_hireso.c
engineer124: This is a classic interface dev loop, it's always how they process their digits. I think it always matches by doing:
```c
                    while (counterDigits[3] >= 10) {
                        counterDigits[2]++;
                        counterDigits[3] -= 10;
                    }
```
to make it a while-loop instead of a do-while loop. Surprised the digits aren't in an array though, that's normally how they do it

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+        case BOMBERS_NOTEBOOK_LOAD_STATE_NONE:
+            if (this->scheduleDmaSegment == NULL) {
+                break;
+            }
+            CmpDma_LoadAllFiles(this->scheduleDmaSegmentStart, this->scheduleDmaSegment, this->scheduleDmaSegmentSize);
+            osCreateMesgQueue(&this->loadQueue, this->loadMsg, ARRAY_COUNT(this->loadMsg));
+            DmaMgr_SendRequestImpl(&this->dmaRequest, this->scheduleSegment, this->scheduleSegmentStart,
+                                   this->scheduleSegmentSize, 0, &this->loadQueue, NULL);
+            this->loadState = BOMBERS_NOTEBOOK_LOAD_STATE_STARTED;
+            // fall-through
```
engineer124, line 1124:
```suggestion
            // fallthrough
```
Normally doesn't have the `-`

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1349 @@
...
+
+    if (this->loadState == BOMBERS_NOTEBOOK_LOAD_STATE_DONE) {
+        if (stickAdjX < -30) {
+            if (sStickXRepeatState == -1) {
+                sStickXRepeatTimer--;
+                if (sStickXRepeatTimer < 0) {
+                    sStickXRepeatTimer = 2;
+                } else {
+                    stickAdjX = 0;
+                }
```
engineer124, line 1175: Nice copy of kaleido input code

engineer124 (commented):
Great work! Some initial comments

src/code/z_play_hireso.c
hensldm: Unfortunately this doesn't match with an array, or with a while loop.

include/tables/bombers_notebook/event_table.h
hensldm: This would probably work better in Discord for more visibility, so will respond there.
(Small note I did have a version that used the full name in this commit https://github.com/zeldaret/mm/pull/1235/commits/13b1f9c4364d468ad13c4860b0aaa7a98024c7f8, but it seemed really long so tried token pasting in https://github.com/zeldaret/mm/pull/1235/commits/13b1f9c4364d468ad13c4860b0aaa7a98024c7f8 and liked it better)

include/tables/bombers_notebook/event_table.h
hensldm: I'm actually not quite sure what to name them. They are meant to be for the event, but on a temporary basis per cycle. That way it changes color when you have done it for that cycle.
Perhaps just use `BOMBERS_NOTEBOOK_EVENT_COLOR_`, but not 100% sure on that. If we can figure out a name I wouldn't mind doing it though.

src/code/z_play_hireso.c
hensldm: Where else do we need yaz0 scripts? I know map has some stuff, but that is still in a draft state #1032.

src/code/z_play_hireso.c
hensldm: https://decomp.me/scratch/kZ41w
The function more or less treats it as a single 1D array of s16s (notice the iterator is just a pointer to an s16, and it has to be updated by 3 in the loop. I made it 3D to be clearer on how it is used. The first dimension is which column (day) it is. The second corresponds to the subcolumns (if you look at the notebook each day's background is subdivided into 4, and uses two different colors). The last dimension is the usual RGB colors.

src/code/z_play_hireso.c
hensldm: AFAIK, yes but it doesn't match like that.

src/code/z_play_hireso.c
hensldm: I don't mind doing `1 << 10` for the `dsdx/dtdy` terms, but since `*4` is used basically everywhere I would personally prefer just to use `*4` for those constants as well.

src/code/z_play_hireso.c
hensldm: I was thinking `MIN` for `MINIMUM`. TBH, I had a difficult time trying to name it and couldn't think of anything better.

include/tables/bombers_notebook/event_table.h
hensldm: Ended up going with just the full enum names again, as well as bring over Zoey's work from the message branch.

include/tables/bombers_notebook/event_table.h
hensldm: Decided to go with actually naming them the same as the permanent Event flags, but without the BOMBERS_NOTEBOOK_EVENT prefix protion, since these are more general and used elsewhere.

AngheloAlf (approved):
Very good work!

src/code/z_play_hireso.c
engineer124: `gItemIcons` is a big one, that gets used in parameter and kaleido. 

In Kaleido Collect:
```c
extern TexturePtr D_08061000; // gBombersNotebookIconTex
extern TexturePtr D_08062000; // gSongNoteIconTex
```
In Kaleido Map:
```c
extern TexturePtr D_09007500; // gPlayerFaceIcon
```
In Parameter
```c
extern TexturePtr D_08095AC0; // gMagicArrowEquipEffectTex
```

Really just items and pause stuff, so maybe it felt like more just because that's what I work on. I can always just add comments later to these above, not a big deal here

src/code/z_play_hireso.c
engineer124: Oh I think I see. Yeah that makes sense, could you possibly explain that above the data?

src/code/z_play_hireso.c
hensldm: Sounds good, but I'm going to change it to just `// TODO: Needs YAZ0`. That way it would be easy to search for all of them if they are all the same.

include/z64.h
```diff
@@ -32,6 +32,7 @@
 #include "z64animation.h"
 #include "z64animation_legacy.h"
 #include "z64audio.h"
+#include "z64bombers_notebook.h"
 #include "z64bgcheck.h"
 #include "z64camera.h"
```
engineer124, line 37:
```suggestion
#include "z64audio.h"
#include "z64bgcheck.h"
#include "z64bombers_notebook.h"
#include "z64camera.h"
```
g before o

src/code/z_message.c
```diff
@@ -5,6 +5,27 @@
 #include "overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope.h"
 
 #if 0
+#define DEFINE_PERSON(_enum, _photo, _description, _metEnum, metMessage, _metFlag) metMessage,
+#define DEFINE_EVENT(_enum, _icon, _colorFlag, _description, completedMessage, _completedFlag) completedMessage,
+
+u16 sBombersNotebookEventMessages[] = {
```
engineer124, line 11:
```suggestion
u16 sBombersNotebookEventMessages[BOMBERS_NOTEBOOK_EVENT_MAX] = {
```
I think this makes sense?

src/code/z_message.c
```diff
@@ -20,8 +41,8 @@ s32 D_801D02F8[15] = { 0,1,2,3,4,0,1,0,0,0,0,0,1,1,0 };
 
 #endif
 
-extern u16 D_801C6AB8[];
-extern u16 D_801C6B28[];
+extern u16 sBombersNotebookEventMessages[];
+extern u16 gBombersNotebookWeekEventFlags[];
```
engineer124, line 45: These can be externed in z64message.h (and also removed from variables.h, and given `BOMBERS_NOTEBOOK_EVENT_MAX` as the previous comment). Don't worry about the other message variables though if you don't want, just these two

src/code/z_message.c
```diff
@@ -365,19 +386,19 @@ void func_80151A68(PlayState* play, u16 textId) {
...
+            msgCtx->bombersNotebookNewEventQueueSize++;
         }
-    } else if (arg1 >= 20) {
-        if (!CHECK_WEEKEVENTREG(D_801C6B28[arg1])) {
-            msgCtx->unk120B2[msgCtx->unk120B1] = temp;
-            msgCtx->unk120B1++;
+    } else if (event >= BOMBERS_NOTEBOOK_PERSON_MAX) {
+        // Non MET events are processed even if the player does not have the notebook yet
+        if (!CHECK_WEEKEVENTREG(gBombersNotebookWeekEventFlags[event])) {
+            msgCtx->bombersNotebookNewEventQueue[msgCtx->bombersNotebookNewEventQueueSize] = event;
```
engineer124, line 400:
```suggestion
            msgCtx->bombersNotebookEventQueue[msgCtx->bombersNotebookEventQueueSize] = event;
```
Does `New` need to be in the name?

src/code/z_message.c
```diff
@@ -365,19 +386,19 @@ void func_80151A68(PlayState* play, u16 textId) {
...
+            msgCtx->bombersNotebookNewEventQueueSize++;
```
engineer124, line 401:
```suggestion
            msgCtx->bombersNotebookNewEventQueueCount++;
```
It's more of a count instead of a size? Noting our style usually refers to size in bytes (`size_t`)

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1360 @@
...
+
+    gSPScisTextureRectangle(gfx++, xl, yl, xh, yh, tile, s, t, dsdx, dtdy);
+
+    *gfxP = gfx;
+}
+
+s16 sBombersNotebookHeaderColors[][3] = {
+    /* Column 1 */ { 116, 134, 146 },
+    /* Column 2 */ { 158, 156, 131 },
+    /* Column 3 */ { 174, 141, 151 },
```
engineer124, line 319:
```suggestion
    { 116, 134, 146 }, // Column 1
    { 158, 156, 131 }, // Column 2
    { 174, 141, 151 }, // Column 3
```
Any reason for having them on the left? (Same with `sBombersNotebookColumnColors`)

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1360 @@
...
+
+#define DEFINE_EVENT(_enum, _icon, colorFlag, _description, _completedMessage, _completedFlag) colorFlag,
+
+u16 sBombersNotebookEventColorWeekEventFlags[] = {
+#include "tables/bombers_notebook/event_table.h"
+};
+
+#undef DEFINE_EVENT
+
+void BombersNotebook_TextureRectangle(Gfx** gfxP, s32 rxl, s32 ryl, s32 rxh, s32 ryh, s32 tile, s32 s, s32 t, s32 dsdx,
```
engineer124, line 303:
```suggestion
void BombersNotebook_DrawTexRectScis(Gfx** gfxP, s32 rxl, s32 ryl, s32 rxh, s32 ryh, s32 tile, s32 s, s32 t, s32 dsdx,
```
At least for all the functions in parameter/kaleido/file-select that draw TextureRectangles, this is the style Tharo and I have been following. Not sure where to put the `Scis` in there, but the `DrawTexRect` is usually the base. Any thoughts?

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1360 @@
...
+        BombersNotebook_TextureRectangle(&gfx, rectLeft * 4, 86 * 4, (rectLeft + 48) * 4, 97 * 4, 0, 0, 0, 0x400,
+                                         0x400);
+        rectLeft += 150;
+    }
+
+    gDPPipeSync(gfx++);
+    *gfxP = gfx;
+}
+
+/* Each column/day gets subdivided into 4 sub columns, the first 2 sub cloumns use 1 color, the last 2 another */
```
engineer124, line 358:
```suggestion
// Each column/day gets subdivided into 4 sub columns, the first 2 sub columns use 1 color, the last 2 another
```

src/code/z_play_hireso.c
engineer124: I'm not too picky about it. But I think the `1 << 10` should be applied at least

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1360 @@
...
+
+void BombersNotebook_DrawRows(BombersNotebook* this, Gfx** gfxP) {
+    static s16 sBarColorR = 0;
+    static s16 sBarColorG = 0;
+    static s16 sBarColorB = 0;
+    static s16 sBarColorTimer = 10;
+    static s16 sBarColorIndex = 0;
+    static s16 sBarColorTargetsR[2] = { 100, 0 };
+    static s16 sBarColorTargetsG[2] = { 80, 0 };
+    static s16 sBarColorTargetsB[2] = { 255, 0 };
```
engineer124, line 592:
```suggestion
    static s16 sBarColorTargetsR[] = { 100, 0 };
    static s16 sBarColorTargetsG[] = { 80, 0 };
    static s16 sBarColorTargetsB[] = { 255, 0 };
```

src/code/z_play_hireso.c
engineer124: Yeah this is tricky. Maybe `CURRENT_DAY_CLAMP_MIN_1`? But it's not a true clamp.

src/code/z_play_hireso.c
engineer124: Yeah that's surprising. These are normally s16 arrays with the interface dev, but they made it s32. And yeah doesn't match. Interesting

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1360 @@
...
+    static s16 sLowerArrowColorG = 100;
+    static s16 sLowerArrowColorB = 255;
+    static s16 sUpperArrowColorIndex = 0;
+    static s16 sLowerArrowColorIndex = 0;
+    static s16 sUpperArrowColorTimer = 6;
+    static s16 sLowerArrowColorTimer = 6;
+    static s16 sUpperArrowOffsetY = 0;
+    static s16 sLowerArrowOffsetY = 0;
+    static s16 sArrowColorTargetsG[2] = { 175, 100 };
+    static s16 sArrowColorTargetsB[2] = { 205, 255 };
```
engineer124, line 994:
```suggestion
    static s16 sArrowColorTargetsG[] = { 175, 100 };
    static s16 sArrowColorTargetsB[] = { 205, 255 };
```

src/overlays/actors/ovl_Obj_Um/z_obj_um.c
```diff
@@ -14,15 +14,15 @@
 
 /**
  * weekEventReg flags checked by this actor:
- * - WEEKEVENTREG_22_01: Aliens defeated
+ * - WEEKEVENTREG_DEFENDED_AGAINST_THEM: Aliens defeated
```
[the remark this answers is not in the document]
AngheloAlf: Also `THEM` is even harder because you lose the capitalization.
From a quick search, seems like "Aliens" is more common in the codebase. "Them" is only used in 2 functions of EnMa4 (Romani)

src/code/z_play_hireso.c
hensldm: Seems fine. Though decided to go with `DrawScisTexRect` since that matches the macro better.

src/code/z_play_hireso.c
hensldm: Personally, I like `CURRENT_DAY_CLAMP_MIN_1` the best. I agree the intention is to avoid UB reads (though it is used directly without the -1 a couple of times), but since it only rounds up 0 I think the name should reflect that. (It is also curious that there is a `CURRENT_DAY_CLAMP_MIN_1 >= 3` check instead of just a `== 3`.)

src/code/z_play_hireso.c
```diff
@@ -1,27 +1,1363 @@
...
+
+    gDPPipeSync(gfx++);
+    *gfxP = gfx;
+}
+
+// Each column/day gets subdivided into 4 sub columns, the first 2 sub columns use 1 color, the last 2 another
+s16 sBombersNotebookColumnColors[][2][3] = {
+    { { 165, 183, 195 }, { 140, 158, 170 } }, // Column 2
+    { { 197, 195, 172 }, { 172, 170, 147 } }, // Column 1
+    { { 223, 190, 200 }, { 190, 165, 175 } }, // Column 3
```
engineer124, line 362:
```suggestion
    { { 165, 183, 195 }, { 140, 158, 170 } }, // Column 1
    { { 197, 195, 172 }, { 172, 170, 147 } }, // Column 2
    { { 223, 190, 200 }, { 190, 165, 175 } }, // Column 3
```

engineer124 (approved):

include/tables/bombers_notebook/event_table.h
```diff
@@ -0,0 +1,48 @@
...
+ *    - Argument 4: Message id for the completion of this event
+ *    - Argument 5: WeekEventFlag for the completion of this event
+ *
+ * Note: These events are after the MET events, so will start indexing at BOMBERS_NOTEBOOK_PERSON_MAX
+ */
+/* 0x14 */ DEFINE_EVENT(BOMBERS_NOTEBOOK_EVENT_RECEIVED_ROOM_KEY, BOMBERS_NOTEBOOK_EVENT_ICON_EXCLAMATION_POINT, WEEKEVENTREG_RECEIVED_ROOM_KEY, 0x2198, 0x2152, WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_RECEIVED_ROOM_KEY)
+/* 0x15 */ DEFINE_EVENT(BOMBERS_NOTEBOOK_EVENT_PROMISED_MIDNIGHT_MEETING, BOMBERS_NOTEBOOK_EVENT_ICON_EXCLAMATION_POINT, WEEKEVENTREG_PROMISED_MIDNIGHT_MEETING, 0x2199, 0x2153, WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_PROMISED_MIDNIGHT_MEETING)
+/* 0x16 */ DEFINE_EVENT(BOMBERS_NOTEBOOK_EVENT_PROMISED_TO_MEET_KAFEI, BOMBERS_NOTEBOOK_EVENT_ICON_EXCLAMATION_POINT, WEEKEVENTREG_HAD_MIDNIGHT_MEETING, 0x219A, 0x2154, WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_PROMISED_TO_MEET_KAFEI)
+/* 0x17 */ DEFINE_EVENT(BOMBERS_NOTEBOOK_EVENT_RECEIVED_LETTER_TO_KAFEI, BOMBERS_NOTEBOOK_EVENT_ICON_EXCLAMATION_POINT, WEEKEVENTREG_HAD_MIDNIGHT_MEETING, 0x219B, 0, WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_RECEIVED_LETTER_TO_KAFEI)
```
jpburnett, line 17:
```suggestion
/* 0x17 */ DEFINE_EVENT(BOMBERS_NOTEBOOK_EVENT_RECEIVED_LETTER_TO_KAFEI, BOMBERS_NOTEBOOK_EVENT_ICON_EXCLAMATION_POINT, WEEKEVENTREG_HAD_MIDNIGHT_MEETING, 0x219B, 0x0, WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_RECEIVED_LETTER_TO_KAFEI)
```
Negligible nit of maintaining hex format, 0 is zero is 0x0 :)

jpburnett (approved):
Nice work, negligible nit on my part.

include/tables/bombers_notebook/event_table.h
hensldm: Went with `0x0000` so it matches the same width as the other non zero text ids.

PR #1244, opened 2023-05-21
`z_fbdemo_wipe1` OK and assorted asset cleanup

tom-overton:
This file is seemingly identical to OoT, so I mostly ported the docs over as-is except for a few things. As part of working on this, I noticed that the assets in `z_fbdemo_triforce` were in an XML file called `object_fbdemo_triforce` in the `objects` folder. This is definitely *not* an object, so I moved it to the `overlays` folder and renamed it. While I was doing this, I noticed that `ovl_Arrow_Fire`'s asset XML didn't have the proper capitalization, so I fixed that too.


AngheloAlf (approved):
LGTM

engineer124 (approved):
Nice work

hensldm (approved):

PR #1251, opened 2023-05-26
sys_flashrom matched

louist103:
Also a little light documentation.


include/z64.h
```diff
@@ -723,6 +725,12 @@ typedef struct {
 #define FRAM_STATUS_REGISTER FRAM_BASE_ADDRESS // FRAM Base Address in Cart Memory
 #define FRAM_COMMAND_REGISTER 0x10000          // Located at 0x08010000 on the Cart
 
+#define FLASH_VERSION_MX_PROTO_A 0x00c20000
+#define FLASH_VERSION_MX_A       0x00c20001
+#define FLASH_VERSION_MX_C       0x00c2001e
+#define FLASH_VERSION_MX_B_AND_D 0x00c2001d
+#define FLASH_VERSION_MEI        0x003200f1
```
AngheloAlf, line 732: Could you use uppercase for those hex values?

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/SysFlashrom_ThreadEntry.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185DDC.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185EC4.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185F04.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185F64.s")
+//#include "prevent_bss_reordering.h"
```
AngheloAlf, line 6: I think we can remove this for now

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+        }
+    }
+    return -1;
+}
+
+s32 SysFlashrom_InitFlash(void) {
+    osCreateMesgQueue(&sFlashromMesgQueue, sFlashromMesg, ARRAY_COUNT(sFlashromMesg));
+    osFlashInit();
+    sFlashromFlashIsGood = true;
+    if (SysFlashrom_CheckFlashType()) {
```
AngheloAlf, line 65: It doesn't return a bool type
```suggestion
    if (SysFlashrom_CheckFlashType() != 0) {
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+
+s32 SysFlashrom_WriteData(u8* addr, u32 pageNum, u32 pageCount) {
+    OSIoMesg msg;
+    s32 result;
+    u32 i;
+
+    if (!SysFlashrom_FlashIsGood()) {
+        return -1;
+    }
+    if (pageNum & 0x7F) {
```
AngheloAlf, line 99: Maybe this expects a multiple of 0x80 / 128 ?
```suggestion
    if ((pageNum % 128) != 0) {
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+        return result;
+    }
+    result = SysFlashrom_WriteData(addr, pageNum, pageCount);
+    if (result != 0) {
+        if (i < 3) {
+            i++;
+            goto failRetry;
+        }
+        return result;
+    }
```
AngheloAlf, line 139: This feels like it could be a `while (i < 3)` loop with `continue`s

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+        case 2:
+            req->response = SysFlashrom_ReadData(req->addr, req->pageNum, req->pageCount);
+            osSendMesg(&req->messageQueue, req->response, OS_MESG_BLOCK);
+            break;
+    }
+}
+
+void SysFlashrom_CreateRequest(u8* addr, u32 pageNum, u32 pageCount) {
+    FlashromRequest* req = &sFlashromRequest;
+    if (SysFlashrom_FlashIsGood() != 0) {
```
AngheloAlf, line 208: This does return a boolean
```suggestion
    if (SysFlashrom_FlashIsGood()) {
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+        if (i < 3) {
+            i++;
+            goto failRetry;
+        }
+        return result;
+    }
+    return 0;
+}
+
+s32 func_80185BE4(void* data, void* addr, u32 pageCount) {
```
Thar0, line 143: My guess for why this function exists is to try and avoid erasures if they aren't necessary, but they gave up halfway. Flash write operations can only flip bits from 1 -> 0, and to flip from 0 -> 1 you need to run an erasure.
The check `((*(s32*)data & *(s32*)addr) != *(s32*)addr)` is saying
`(current_flash_data) & (incoming_data) != (incoming_data)`

On the level of individual bits:
```
Flash & Data  = Data
0       0       0
1       0       0
0       1       0
1       1       1
```
The third case `0 & 1 = 0` is the one disallowed by flash, so if this case is detected the check passes and 0 (false) is returned to indicate an erasure is necessary.

However given that, it seems they just ran the erasure either way. Not sure why, maybe some issues elsewhere?

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+
+    for (i = 0; i < pageCount * 128; i += 4) {
+        if ((*(s32*)data & *(s32*)addr) != *(s32*)addr) {
+            return 0;
+        }
+    }
+    return 1;
+}
+
+s32 SysFlashrom_WriteData2(void* addr, u32 pageNum, u32 pageCount) {
```
Thar0, line 155: Since I think the current `SysFlashrom_WriteData` can be renamed, this one could become `SysFlashrom_WriteData`

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+    void* data;
+    size_t size;
+    s32 ret;
+
+    if (!SysFlashrom_FlashIsGood()) {
+        return -1;
+    }
+    size = pageCount * 128;
+    data = SystemArena_Malloc(size);
+    if (data == 0) {
```
Thar0, line 165:
```suggestion
    if (data == NULL) {
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+        ret = SysFlashrom_AttemptWrite(addr, pageNum, pageCount);
+    } else {
+        SysFlashrom_ReadData(data, pageNum, pageCount);
+        if (bcmp(data, addr, size) == 0) {
+            ret = 0;
+        } else {
+            //! @bug Will attempt to write regardless of the result of the function call.
```
Thar0, line 172: Given the prior explanation of what `func_80185BE4` tries to do, it's not really a bug to run a write either way, I would say if anything the real bug is that it runs an erasure either way. I'm not entirely sure given how it doesn't really cause any bad side-effects whether it's worthy of "bug", but it is worth noting in some way for sure.

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+}
+
+void SysFlashrom_ThreadEntry(void* arg) {
+    FlashromRequest* req = (FlashromRequest*)arg;
+    switch (req->requestType) {
+        case 1:
+            req->response = SysFlashrom_WriteData2(req->addr, req->pageNum, req->pageCount);
+            osSendMesg(&req->messageQueue, req->response, OS_MESG_BLOCK);
+            break;
+        case 2:
```
Thar0, line 199: Defines for the request type seem nice
```suggestion
        case FLASHROM_REQUEST_WRITE:
            req->response = SysFlashrom_WriteData2(req->addr, req->pageNum, req->pageCount);
            osSendMesg(&req->messageQueue, req->response, OS_MESG_BLOCK);
            break;
        case FLASHROM_REQUEST_READ:
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+        osCreateMesgQueue(&req->messageQueue, sSysFlashromMsgBuf, ARRAY_COUNT(sSysFlashromMsgBuf));
+        StackCheck_Init(&sSysFlashromStackInfo, sSysFlashromStack, STACK_TOP(sSysFlashromStack), 0,
+                        0x100, "sys_flashrom");
+        osCreateThread(&sSysFlashromThread, Z_THREAD_ID_FLASHROM, SysFlashrom_ThreadEntry, req,
+                       STACK_TOP(sSysFlashromStack), Z_PRIORITY_FLASHROM);
+        osStartThread(&sSysFlashromThread);
+    }
+}
+
+s32 SysFlashrom_IsQueueFull(void) {
```
Thar0, line 222:
```suggestion
s32 SysFlashrom_IsBusy(void) {
```
The queue is one message large so more generally this is just a check for whether the flash thread is currently working or not

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+    OSMesgQueue* queue = &sFlashromRequest.messageQueue;
+
+    if (!SysFlashrom_FlashIsGood()) {
+        return -1;
+    }
+    return MQ_IS_FULL(queue); //queue->msgCount <= queue->validCount;
```
Thar0:
```suggestion
    return MQ_IS_FULL(queue);
```

src/code/sys_flashrom.c
louist103: It generates an extra branch and set. 
image

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+}
+
+s32 SysFlashrom_DestroyThread(void) {
```
Thar0, line 231: I think from an external perspective the important part of this function is that it cleans up the previous request and returns the result. I guess something like `SysFlashrom_AwaitResult`?

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+            osSendMesg(&req->messageQueue, req->response, OS_MESG_BLOCK);
+            break;
+        case 2:
+            req->response = SysFlashrom_ReadData(req->addr, req->pageNum, req->pageCount);
+            osSendMesg(&req->messageQueue, req->response, OS_MESG_BLOCK);
+            break;
+    }
+}
+
+void SysFlashrom_CreateRequest(u8* addr, u32 pageNum, u32 pageCount) {
```
Thar0, line 206:
```suggestion
void SysFlashrom_WriteDataAsync(u8* addr, u32 pageNum, u32 pageCount) {
```
Since this function only issues a write request.
It's maybe worth noting that async read requests are unimplemented despite the thread function supporting them, I suppose since flash reads are fast enough that they don't need to run on a separate thread

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,242 @@
...
+    osRecvMesg(&sFlashromRequest.messageQueue, NULL, OS_MESG_BLOCK);
+    osDestroyThread(&sSysFlashromThread);
+    StackCheck_Cleanup(&sSysFlashromStackInfo);
+    return sFlashromRequest.response;
+}
+
+void func_80185F64(void* addr, u32 pageNum, u32 pageCount) {
+    SysFlashrom_CreateRequest(addr, pageNum, pageCount);
+    SysFlashrom_DestroyThread();
+}
```
Thar0, line 244:
```suggestion
s32 SysFlashrom_WriteDataSync(void* addr, u32 pageNum, u32 pageCount) {
    SysFlashrom_CreateRequest(addr, pageNum, pageCount);
    return SysFlashrom_DestroyThread();
}
```
Since it issues a write request then immediately waits for the result, which blocks the calling thread while just running the first function does not block the calling thread.

Thar0 (commented):
Nice job on this file, finally bridging the gap between z_sram and the low level flash routines 🎉 

Lots of comments about various things (Yes I am a flash enthusiast, how could you tell?)

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,243 @@
...
+    data = SystemArena_Malloc(size);
+    if (data == NULL) {
+        ret = SysFlashrom_AttemptWrite(addr, pageNum, pageCount);
+    } else {
+        SysFlashrom_ReadData(data, pageNum, pageCount);
+        if (bcmp(data, addr, size) == 0) {
+            ret = 0;
+        } else {
+            // Will always erase the sector even if it wouldn't normally need to.
+            if (SysFlashrom_NeedsToErase(data, addr, pageCount) != 0) {
```
hensldm, line 174:
```suggestion
            if (SysFlashrom_NeedsToErase(data, addr, pageCount)) {
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,243 @@
...
+
+// Flash bits can only by flipped with a write from 1 -> 0. Going from 0 -> 1 requires an erasure.
+// This function will try to determine if any sectors need to be erased before writing saving erase cycles.
+s32 SysFlashrom_NeedsToErase(void* data, void* addr, u32 pageCount) {
+    u32 size;
+    u32 i;
+
+    for (i = 0; i < pageCount * FLASH_PAGE_SIZE; i += 4) {
+        if ((*(s32*)data & *(s32*)addr) != *(s32*)addr) {
+            return 0;
```
hensldm, line 150:
```suggestion
            return false;
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,243 @@
...
+        }
+    }
+    return 1;
```
hensldm, line 153:
```suggestion
    return true;
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,243 @@
...
+}
+
+void SysFlashrom_ThreadEntry(void* arg) {
+    FlashromRequest* req = (FlashromRequest*)arg;
+    switch (req->requestType) {
+        case FLASHROM_REQUEST_WRITE:
+            req->response = SysFlashrom_WriteData(req->addr, req->pageNum, req->pageCount);
+            osSendMesg(&req->messageQueue, req->response, OS_MESG_BLOCK);
+            break;
+        case FLASHROM_REQUEST_READ:
```
hensldm, line 200:
```suggestion
            break;

        case FLASHROM_REQUEST_READ:
```

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,243 @@
...
+            req->response = SysFlashrom_ReadData(req->addr, req->pageNum, req->pageCount);
+            osSendMesg(&req->messageQueue, req->response, OS_MESG_BLOCK);
+            break;
+    }
+}
+
+void SysFlashrom_WriteDataAsync(u8* addr, u32 pageNum, u32 pageCount) {
+    FlashromRequest* req = &sFlashromRequest;
+    if (SysFlashrom_IsInit()) {
```
hensldm, line 209:
```suggestion
    FlashromRequest* req = &sFlashromRequest;

    if (SysFlashrom_IsInit()) {
```

hensldm (commented):
Just a few small things.

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,243 @@
...
+s32 SysFlashrom_CheckFlashType(void) {
+    u32 flashType;
+    u32 flashVendor;
+
+    if (!SysFlashrom_IsInit()) {
+        return -1;
+    }
+    osFlashReadId(&flashType, &flashVendor);
+    sFlashromVendor = flashVendor;
+    if (flashType == 0x11118001) {
```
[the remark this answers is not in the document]
louist103: Not a clue. Maybe @Thar0 has some idea.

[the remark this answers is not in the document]
Thar0: This is as magic a number as it gets as far as I know. It's "just" the type value that all the flash chips respond with, whether the bits have a decipherable meaning is unknown as far as I'm aware. Might be worth moving to a define like
`FLASH_TYPE` or something, but there's not a lot more that can be done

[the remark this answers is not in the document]
louist103: Maybe `FLASH_MAGIC` would be better?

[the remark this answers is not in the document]
hensldm: I'd say `FLASH_TYPE_MAGIC`. `FLASH_MAGIC` by itself reminds me of things like `NODE_MAGIC` where it is used as a canary value to determine corruption, which this is not.

hensldm:
I think your merge did not get noticed as an actual merge. Review shows differences for the two merged in commits. Probably should rebase or reset back a few commits, remerge, and fix bss/rename the define.

Thar0 (approved):

include/functions.h
```diff
@@ -2224,22 +2224,15 @@ void func_8018450C(PlayState* play, SkeletonInfo* skeleton, Mtx* mtx, OverrideKe
 // void func_801850A0(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE4 param_5, UNK_TYPE4 param_6, UNK_TYPE4 param_7);
 // void func_801853C8(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE4 param_5, UNK_TYPE4 param_6);
 // void func_80185460(void);
-
 // void func_801857C0(void);
```
engineer124, line 2227: We've been leaving space between files when it comes to `functions.h`. The functions above belong to c_keyframe.c, while the functions below belong to sys_flashrom.c, so could you bring back this space?

include/functions.h
```diff
@@ -2224,22 +2224,15 @@ void func_8018450C(PlayState* play, SkeletonInfo* skeleton, Mtx* mtx, OverrideKe
...
-// void func_80185A2C(void);
-// void func_80185B1C(void);
-// void func_80185BE4(void);
-// void func_80185C24(void);
-void SysFlashrom_ThreadEntry(s80185D40* param_1);
-void func_80185DDC(u8* arg0, u32 arg1, u32 arg2);
-s32 func_80185EC4(void);
-s32 func_80185F04(void);
-void func_80185F64(void* arg0, UNK_TYPE arg1, UNK_TYPE arg2);
+// void SysFlashrom_CheckFlashType(void);
```
engineer124, line 2229: If a function isn't used externally, could you remove it from `functions.h`?

include/variables.h
```diff
@@ -2520,7 +2520,6 @@ extern Vec3f D_801FBDA4;
 extern Vec3f D_801FBDE8;
 extern MtxF* sMatrixStack;
 extern MtxF* sCurrentMatrix;
-
 extern s32 D_801FD120;
```
engineer124, line 2523: Same here, this space separates out audio bss

src/code/sys_flashrom.c
```diff
@@ -3,42 +3,252 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/SysFlashrom_ThreadEntry.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185DDC.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185EC4.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185F04.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/code/sys_flashrom/func_80185F64.s")
+#include "prevent_bss_reordering.h"
```
engineer124, line 6: We try to put this above all other `#include`'s unless it's required below to match

engineer124 (commented):
Nice work. Just some nitpicks. Also, there will be conflicts with https://github.com/zeldaret/mm/pull/1158 as that PR names the data used by this file, but it shouldn't be too bad to resolve whichever one goes in second

hensldm (approved):

include/functions.h
```diff
@@ -2224,22 +2224,16 @@ void func_8018450C(PlayState* play, SkeletonInfo* skeleton, Mtx* mtx, OverrideKe
 // void func_801850A0(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE4 param_5, UNK_TYPE4 param_6, UNK_TYPE4 param_7);
 // void func_801853C8(UNK_TYPE1 param_1, UNK_TYPE1 param_2, UNK_TYPE1 param_3, UNK_TYPE1 param_4, UNK_TYPE4 param_5, UNK_TYPE4 param_6);
 // void func_80185460(void);
-
 // void func_801857C0(void);
 // char* func_801857D0(void);
```
engineer124, line 2228: Could you remove these two `SysFlashrom` functions too?

engineer124 (approved):

AngheloAlf (approved):

PR #1266, opened 2023-06-01
Enable `-Wint-conversion` warning

AngheloAlf:
Turning this warning on makes CC_CHECK to warn about implicit int/pointer conversions, to avoid it we just need to make a explicit cast.

- For Physical/rom addresses I try to us `uintptr_t` as much as I can (as we mostly do already)
- Because of the previous point, I added a `uintptr_t` cast in the `SEGMENT_ROM_START`/`SEGMENT_ROM_END` macros
- I also had to make a `SEGMENT_ROM_START_OFFSET` macro because some places in the codebase do arithmetic with  a `u8*` pointer and doesn't like the address to be casted to `uintptr_t` before the arithmetic. Same reason for `SEGMENT_ROM_SIZE2`.


include/segment_symbols.h
```diff
@@ -24,9 +24,11 @@
...
 #define SEGMENT_SIZE(segment) ((uintptr_t)SEGMENT_END(segment) - (uintptr_t)SEGMENT_START(segment))
 
-#define SEGMENT_ROM_START(segment) (_ ## segment ## SegmentRomStart)
-#define SEGMENT_ROM_END(segment) (_ ## segment ## SegmentRomEnd)
-#define SEGMENT_ROM_SIZE(segment) ((uintptr_t)SEGMENT_ROM_END(segment) - (uintptr_t)SEGMENT_ROM_START(segment))
+#define SEGMENT_ROM_START(segment) ((uintptr_t) _ ## segment ## SegmentRomStart)
+#define SEGMENT_ROM_START_OFFSET(segment, offset) ((uintptr_t) (( _ ## segment ## SegmentRomStart ) + (offset)))
+#define SEGMENT_ROM_END(segment) ((uintptr_t) _ ## segment ## SegmentRomEnd)
+#define SEGMENT_ROM_SIZE(segment) (SEGMENT_ROM_END(segment) - SEGMENT_ROM_START(segment))
+#define SEGMENT_ROM_SIZE2(segment) ((size_t)( _ ## segment ## SegmentRomEnd - _ ## segment ## SegmentRomStart ))
```
EllipticEllipsis, line 31: I'm not really sure what to do here, and to a large extent it depends on what we decide to do with types for rom addresses and segmented addresses. I don't like the `SEGMENT_ROM_START_OFFSET` macro, though, it seems unnecessary, it looks too much like getting the offset of something in a struct, when really it's just the most basic pointer arithmetic operation.

src/audio/lib/heap.c
```diff
@@ -1438,7 +1438,7 @@ void AudioHeap_ApplySampleBankCacheInternal(s32 apply, s32 sampleBankId) {
 
     sampleBankTable = gAudioCtx.sampleBankTable;
     numFonts = gAudioCtx.soundFontTable->numEntries;
-    change.oldAddr = AudioHeap_SearchCaches(SAMPLE_TABLE, CACHE_EITHER, sampleBankId);
+    change.oldAddr = (uintptr_t)AudioHeap_SearchCaches(SAMPLE_TABLE, CACHE_EITHER, sampleBankId);
```
EllipticEllipsis, line 1441: I defer to @engineer124 on all the audio stuff in here, most of these structs are quite deep.

src/code/TwoHeadArena.c
```diff
@@ -103,7 +103,7 @@ void* THA_AllocTailAlign(TwoHeadArena* tha, size_t size, uintptr_t mask) {
  * @return Remaining size. A negative number indicates an overflow.
  */
 s32 THA_GetRemaining(TwoHeadArena* tha) {
```
EllipticEllipsis, line 105: Do we have ptrdiff_t? If not this ought to use intptr_t, probably.

src/code/TwoHeadArena.c
```diff
@@ -32,7 +32,7 @@ void* THA_GetTail(TwoHeadArena* tha) {
 void* THA_AllocHead(TwoHeadArena* tha, size_t size) {
     void* start = tha->head;
 
-    tha->head = (u8*)tha->head + size;
+    tha->head = (void*)((uintptr_t)tha->head + size);
```
EllipticEllipsis, line 35: I don't really see the point of casting out of and back into a pointer here when it's not necessary, (u8*) just seems better for this sort of arithmetic where possible.

src/code/TwoHeadArena.c
```diff
@@ -103,7 +103,7 @@ void* THA_AllocTailAlign(TwoHeadArena* tha, size_t size, uintptr_t mask) {
  * @return Remaining size. A negative number indicates an overflow.
  */
 s32 THA_GetRemaining(TwoHeadArena* tha) {
-    return (s32)((u8*)tha->tail - (u8*)tha->head);
+    return (uintptr_t)tha->tail - (uintptr_t)tha->head;
```
EllipticEllipsis, line 106: Actually won't this output an unsigned thing that then gets cast? I think keeping it as u8* is a much better option.

src/code/z_kanfont.c
```diff
@@ -39,8 +39,7 @@ void Font_LoadOrderedFont(Font* font) {
             loadOffset = 0;
         }
 
-        DmaMgr_SendRequest0(writeLocation, (uintptr_t)SEGMENT_ROM_START(nes_font_static) + loadOffset,
-                            FONT_CHAR_TEX_SIZE);
+        DmaMgr_SendRequest0(writeLocation, SEGMENT_ROM_START(nes_font_static) + loadOffset, FONT_CHAR_TEX_SIZE);
```
EllipticEllipsis, line 42: So does SEGMENT_ROM_START_OFFSET work here or not? Either way I think it confuses the operation here.

EllipticEllipsis (commented):
I like the idea of this in principle but we need to decide what to do with rom addresses and segmented addresses first, I think. I have a few comments on the current state, mostly where one or the other cast seems better.

Makefile
```diff
@@ -102,8 +102,8 @@ endif
 
 # Check code syntax with host compiler
 ifneq ($(RUN_CC_CHECK),0)
-  CHECK_WARNINGS := -Wall -Wextra -Wno-format-security -Wno-unknown-pragmas -Wno-unused-parameter -Wno-unused-variable -Wno-missing-braces -Wno-int-conversion -Wno-unused-but-set-variable -Wno-unused-label -Wno-sign-compare -Wno-tautological-compare
```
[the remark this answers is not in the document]
AngheloAlf: It does nothing currently. https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat-security

It warns you for cases when you misuse printf/scanf functions, like passing a float when you have a `%u`

We could annotate some variadic functions that work similar to printf (with `__attribute__ ((format (printf, 2, 3)))`) but that's a out of scope for this PR

Makefile
```diff
@@ -102,8 +102,8 @@ endif
...
-  CC_CHECK   := gcc -fno-builtin -fsyntax-only -funsigned-char -fdiagnostics-color -std=gnu89 -D _LANGUAGE_C -D NON_MATCHING $(IINC) -nostdinc $(CHECK_WARNINGS)
```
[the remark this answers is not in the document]
AngheloAlf: At some point I had different flags for the audio directory, but ended up scrapping it.
Also I think it is a bit cleaner to have this separated. We have the compiler invocation separated from the flags passed

src/boot_O2/system_heap.c
```diff
@@ -8,9 +8,9 @@
 #include "global.h"
 #include "system_malloc.h"
 
-typedef void (*BlockFunc)(void*);
-typedef void (*BlockFunc1)(void*, u32);
-typedef void (*BlockFunc8)(void*, u32, u32, u32, u32, u32, u32, u32, u32);
+typedef void (*BlockFunc)(uintptr_t);
+typedef void (*BlockFunc1)(uintptr_t, u32);
+typedef void (*BlockFunc8)(uintptr_t, u32, u32, u32, u32, u32, u32, u32, u32);
```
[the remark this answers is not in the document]
AngheloAlf: I don't know what I'm supposed to look at

src/code/code_8012EC80.c
```diff
@@ -97,137 +97,137 @@ u32 gGsFlagsShift[] = {
...
-    0x09006C00,        // ITEM_MAGIC_LARGE
-    0x08062180,        // ITEM_HEART_PIECE_2
-    0x08062A80,        // ITEM_INVALID_1
-    0x08063380,        // ITEM_INVALID_2
-    gOcarinaCUpTex,    // ITEM_INVALID_3
-    gOcarinaCDownTex,  // ITEM_INVALID_4
-    gOcarinaCLeftTex,  // ITEM_INVALID_5
-    gOcarinaCRightTex, // ITEM_INVALID_6
-    gOcarinaATex,      // ITEM_INVALID_7
+    (TexturePtr)0x08000000, // ITEM_OCARINA
```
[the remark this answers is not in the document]
AngheloAlf: What do you mean? This cast is necessary if this gets merged before the other PR

src/code/sys_cmpdma.c
```diff
@@ -43,7 +43,7 @@ void func_80178AC0(u16* src, void* dst, size_t size) {
     }
 }
 
-void CmpDma_GetFileInfo(u8* segmentRom, s32 id, uintptr_t* outFileRom, size_t* size, s32* flag) {
+void CmpDma_GetFileInfo(uintptr_t segmentRom, s32 id, uintptr_t* outFileRom, size_t* size, s32* flag) {
     u32 dataStart;
```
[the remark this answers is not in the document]
AngheloAlf: do you mean the temps?

src/code/z_DLF.c
```diff
@@ -17,23 +17,23 @@ void Overlay_LoadGameState(GameStateOverlay* overlayEntry) {
                                                         (uintptr_t)vramStart, (uintptr_t)overlayEntry->vramEnd);
     if (overlayEntry->loadedRamAddr != NULL) {
 
-        overlayEntry->unk_14 = (uintptr_t)(
+        overlayEntry->unk_14 = (void*)(uintptr_t)(
```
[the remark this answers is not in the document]
AngheloAlf: yeah, it is sadly required

include/z64.h
```diff
@@ -121,7 +121,7 @@ typedef struct {
 
 typedef struct {
     /* 0x00 */ s32 requestType;
-    /* 0x04 */ OSMesg response;
+    /* 0x04 */ s32 response;
```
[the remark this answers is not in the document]
AngheloAlf: In sys_flashrom itself, it doesn't make sense to have this as a type

src/code/z_lib.c
```diff
@@ -702,7 +702,7 @@ void* Lib_SegmentedToVirtual(void* ptr) {
 }
 
 void* Lib_SegmentedToVirtualNull(void* ptr) {
-    if (((uintptr_t)ptr >> 28) == 0) {
+    if ((((uintptr_t)ptr) >> 28) == 0) {
```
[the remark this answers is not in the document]
AngheloAlf: No, it didn't complain, but I think this is better this way

src/libultra/os/createthread.c
```diff
@@ -9,9 +9,9 @@ void osCreateThread(OSThread* t, OSId id, void* entry, void* arg, void* sp, OSPr
     t->next = NULL;
     t->queue = NULL;
     t->context.pc = (u32)entry;
-    t->context.a0 = arg;
+    t->context.a0 = (u64)(intptr_t)arg;
     t->context.sp = (u64)(s32)sp - 16;
-    t->context.ra = __osCleanupThread;
+    t->context.ra = (u64)(intptr_t)__osCleanupThread;
```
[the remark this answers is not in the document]
AngheloAlf: Yes, I did check, that's actually where I got the double cast idea from.

But actually, just using the `intptr_t` cast matches now and doesn't give a warning...

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -19241,7 +19241,7 @@ void func_80859C60(PlayState* play, Player* this, void* arg2) {
     PlayerAnimation_Update(play, &this->skelAnime);
 }
 
-void func_80859CA0(PlayState* play, Player* this, s32 arg2) {
+void func_80859CA0(PlayState* play, Player* this, UNK_TYPE arg2) {
```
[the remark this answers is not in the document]
AngheloAlf: The forward declaration already uses an `UNK_TYPE` so I just made it consistent with that.
It will conflict with that PR either way so I don't mind much about this

src/overlays/effects/ovl_Effect_Ss_Bubble/z_eff_ss_bubble.c
```diff
@@ -34,7 +34,7 @@ u32 EffectSsBubble_Init(PlayState* play, u32 index, EffectSs* this, void* initPa
     {
         TexturePtr tex = (Rand_ZeroOne() < 0.5f) ? gEffBubble1Tex : gEffBubble2Tex;
 
-        this->gfx = VIRTUAL_TO_PHYSICAL(SEGMENTED_TO_VIRTUAL(tex));
+        this->gfx = (void*)VIRTUAL_TO_PHYSICAL(SEGMENTED_TO_VIRTUAL(tex));
```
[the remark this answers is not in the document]
AngheloAlf: Depends on how we want to interpret physical addresses, which we have never able to agree

Makefile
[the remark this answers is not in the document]
EllipticEllipsis: That's a confusing attribute. Yeah, don't think we need to worry about that.

Makefile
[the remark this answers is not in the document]
EllipticEllipsis: I don't see that here, most of the flags are still in CC_CHECK itself?

include/segment_symbols.h
AngheloAlf: `SEGMENT_ROM_START_OFFSET` is necessary because of some places in the codebase requiring some arithmetic before the `uintptr_t` cast or it wouldn't match.

I added the cast as part of the macros to avoid having to use the cast explicitly on every use of `SEGMENT_ROM_START` and `SEGMENT_ROM_END`

src/boot_O2/system_heap.c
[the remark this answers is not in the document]
EllipticEllipsis: The point is that the casting will surely follow the same pattern, so if there's discrepancy it's probably wrong.

src/code/sys_cmpdma.c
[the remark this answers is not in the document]
EllipticEllipsis: Yes.

src/code/z_lib.c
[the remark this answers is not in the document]
EllipticEllipsis: I don't see why, we don't usually put () round any other cast (unless it's in a macro, probably).

src/code/TwoHeadArena.c
AngheloAlf: Personally I see doing pointer arithmetic as something bad and which should be avoided. That's why we have types to do proper pointer arithmetic

src/libultra/os/createthread.c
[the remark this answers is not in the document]
EllipticEllipsis: If there's a way to match debug and avoid this warning that would be my preference. But I suspect we'll turn this off for libultra when it's a subrepo anyway.

src/code/TwoHeadArena.c
AngheloAlf: We can't keep this as `u8*` without casting somewhere, that's the point of `-Wint-conversion`

src/overlays/effects/ovl_Effect_Ss_Bubble/z_eff_ss_bubble.c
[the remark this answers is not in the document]
EllipticEllipsis: I suggest we open an issue for it to summarise where we are and where we need to go, then.

include/segment_symbols.h
EllipticEllipsis: It's not necessary in that it wasn't required before, though, it's only become necessary because of the way you currently have the macros arranged, surely?

src/code/z_kanfont.c
AngheloAlf: Weirdly no, this place requires the cast to happen before the arithmetic and not before. The devs were definitely not consistent with this

Makefile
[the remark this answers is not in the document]
AngheloAlf: Forgot to link the docs about it
https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes
I think it could actually be useful in the future

Makefile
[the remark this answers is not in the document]
AngheloAlf: Yes, I didn't want to split those yet

src/code/TwoHeadArena.c
EllipticEllipsis: But this is standard pointer arithmetic, not alignment, it's not an unusual operation. Arithmetic is one of the two main properties of pointers, after all.

src/code/TwoHeadArena.c
EllipticEllipsis: So cast to the return type afterwards.

src/boot_O2/system_heap.c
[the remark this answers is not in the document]
AngheloAlf: Or maybe they got reworked like a lot of stuff between OoT and MM? :man_shrugging:

Makefile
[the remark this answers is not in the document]
EllipticEllipsis: I don't really see why you'd want the compiler invocation separate from the flags at the point of use anyway, though, when it just means duplicating more. It's not really being used as a compiler anyway here, just a single program invocation.

src/boot_O2/system_heap.c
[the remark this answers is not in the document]
EllipticEllipsis: The only reworking I see there is the interrupt handling is removed for MM and OoT GC.

src/code/z_kaleido_manager.c
```diff
@@ -39,7 +39,8 @@ void* KaleidoManager_FaultAddrConv(void* address, void* param) {
 
 void KaleidoManager_LoadOvl(KaleidoMgrOverlay* ovl) {
     ovl->loadedRamAddr = sKaleidoAreaPtr;
-    Load2_LoadOverlay(ovl->vromStart, ovl->vromEnd, ovl->vramStart, ovl->vramEnd, ovl->loadedRamAddr);
+    Load2_LoadOverlay(ovl->vromStart, ovl->vromEnd, (uintptr_t)ovl->vramStart, (uintptr_t)ovl->vramEnd,
```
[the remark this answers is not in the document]
AngheloAlf: We usually try to use `void*` to hold actual vram pointers. I guess I could change the struct members to be `uintptr_t`

include/segment_symbols.h
AngheloAlf: It was an issue tho, there were some awful casting in place already
image

src/code/z_kaleido_manager.c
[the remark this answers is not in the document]
AngheloAlf: Actually no, I think i prefer the members to be `void*`. Also I managed to get this function to match with the `void*` parameters instead of `uintptr_t`, so I'll keep this members as `void*`

src/overlays/effects/ovl_Effect_Ss_Bubble/z_eff_ss_bubble.c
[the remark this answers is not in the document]
AngheloAlf: https://github.com/zeldaret/mm/issues/1274

src/code/TwoHeadArena.c
AngheloAlf: Casting to `uintptr_t` to perform pointer arithmetic is way more common in the codebase than using `u8*`

src/audio/lib/heap.c
engineer124: Admittedly, low level audio does a lot of weird pointer manipulation with offsets and indexing when both transferring audio assets from rom to ram (sequences/soundfonts/samples), and even more weird pointer/offset math when extracting meaningful information from it (especially soundfonts, that's really messy when it comes to pointers).

Given that, I think I'd need to sit down and summarize all the pointer math involved in audio and share that to figure out what kind of types we need. That would take time though, so for now I think I'm fine with just making sure it's consistent i.e. warning free?

src/audio/lib/synthesis.c
```diff
@@ -1110,7 +1110,8 @@ Acmd* AudioSynth_ProcessSample(s32 noteIndex, NoteSampleState* sampleState, Note
                     case CODEC_REVERB:
                         reverbAddrSrc = (void*)0xFFFFFFFF;
                         if (gAudioCustomReverbFunction != NULL) {
-                            reverbAddrSrc = gAudioCustomReverbFunction(sample, numSamplesToLoadAdj, flags, noteIndex);
+                            reverbAddrSrc =
+                                (void*)gAudioCustomReverbFunction(sample, numSamplesToLoadAdj, flags, noteIndex);
```
engineer124, line 1114:
```suggestion
                            reverbAddrSrc = gAudioCustomReverbFunction(sample, numSamplesToLoadAdj, flags, noteIndex);
```
Instead, change 
`typedef s32 (*AudioCustomReverbFunction)(Sample*, s32, s8, s32);`
to
`typedef void* (*AudioCustomReverbFunction)(Sample*, s32, s8, s32);`

src/audio/lib/thread.c
```diff
@@ -95,7 +95,7 @@ AudioTask* AudioThread_UpdateImpl(void) {
     if (gAudioCtx.resetStatus != 0) {
         if (AudioHeap_ResetStep() == 0) {
             if (gAudioCtx.resetStatus == 0) {
-                osSendMesg(gAudioCtx.audioResetQueueP, gAudioCtx.specId, OS_MESG_NOBLOCK);
+                osSendMesg(gAudioCtx.audioResetQueueP, (OSMesg)(uintptr_t)gAudioCtx.specId, OS_MESG_NOBLOCK);
```
engineer124, line 98: Is the double cast needed?

src/audio/lib/thread.c
```diff
@@ -790,13 +790,13 @@ void AudioThread_ProcessChannelCmd(SequenceChannel* channel, AudioCmd* cmd) {
             break;
 
         case AUDIOCMD_OP_CHANNEL_SET_SFX_STATE:
-            channel->sfxState = cmd->asUInt;
+            channel->sfxState = cmd->data;
```
engineer124, line 793: I think `cmd->data` is suppose to be the internal variable not accessed by the commands. The type should be paired with whatever function it's tied to:
`AudioThread_QueueCmdF32`
`AudioThread_QueueCmdS32`
`AudioThread_QueueCmdS8`
`AudioThread_QueueCmdU16`
(although S32 covers both s32/u32, U16 covers both s16/u16, and S8 covers both s8/u8, names should reflect that).

What would the casting look like without using `data` for everything?

src/audio/lib/thread.c
AngheloAlf: Annoyingly casting a non 32-bit wide type to a pointer type produces this warning
```
src/audio/lib/thread.c: In function 'AudioThread_UpdateImpl':
src/audio/lib/thread.c:98:56: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
   98 |                 osSendMesg(gAudioCtx.audioResetQueueP, (OSMesg)gAudioCtx.specId, OS_MESG_NOBLOCK);
      |                                                        ^
```

src/audio/lib/thread.c
AngheloAlf: It looks like this https://github.com/zeldaret/mm/pull/1266/commits/af9e263c96d9fdfab772c8327bdb2638ef9b7292
I'm not a huge fan of stuff like this
```c
gAudioCustomUpdateFunction = (AudioCustomUpdateFunction)cmd->asUInt;
```
casting a number to a function seems fishy

include/segment_symbols.h
hensldm, line 31:
```suggestion
#define SEGMENT_ROM_SIZE_ALT(segment) ((size_t)( _ ## segment ## SegmentRomEnd - _ ## segment ## SegmentRomStart ))
```
Feel like `Alt` is common for these kind of things.

src/code/z_skelanime.c
```diff
@@ -1007,7 +1007,7 @@ void AnimationContext_SetLoadFrame(PlayState* play, PlayerAnimationHeader* anima
 
     if (entry != NULL) {
         PlayerAnimationHeader* playerAnimHeader = Lib_SegmentedToVirtual(animation);
-        void* ram = (void*)frameTable;
```
hensldm, line 1010: Hmm replacing this temp with a pad and just using frameTable in the Dma function seems to work. Not 100% sure it is better, though I do think we should rename the temp variable if we do.

AngheloAlf: Sounds better, I'll go with it

src/audio/lib/thread.c
engineer124: Yeah, that is kinda ugly. I wonder if it's worth adding one more member `cmd->asPtr` and define `asPtr` as `void*`?

AngheloAlf: yeah, that sounds like a better solution, I'll go with it

src/audio/lib/thread.c
```diff
@@ -296,24 +296,24 @@ void AudioThread_ProcessGlobalCmd(AudioCmd* cmd) {
...
+                gAudioCtx.customSeqFunctions[cmd->arg2] = cmd->asPtr;
             }
             break;
 
         case AUDIOCMD_OP_GLOBAL_SET_DRUM_FONT:
         case AUDIOCMD_OP_GLOBAL_SET_SFX_FONT:
         case AUDIOCMD_OP_GLOBAL_SET_INSTRUMENT_FONT:
             if (AudioPlayback_SetFontInstrument(cmd->op - AUDIOCMD_OP_GLOBAL_SET_DRUM_FONT, cmd->arg1, cmd->arg2,
-                                                cmd->asInt)) {}
+                                                (void*)cmd->asUInt)) {}
```
hensldm, line 316: Should this use `asPtr`?

src/code/sched.c
```diff
@@ -550,14 +552,14 @@ void Sched_FaultClient(void* param1, void* param2) {
  * threads or the OS.
  */
 void Sched_ThreadEntry(void* arg) {
-    OSMesg msg = NULL;
+    s32 msg = 0;
     SchedContext* sched = (SchedContext*)arg;
 
     while (true) {
-        osRecvMesg(&sched->interruptQ, &msg, OS_MESG_BLOCK);
+        osRecvMesg(&sched->interruptQ, (OSMesg)&msg, OS_MESG_BLOCK);
```
hensldm, line 559:
```suggestion
        osRecvMesg(&sched->interruptQ, (OSMesg*)&msg, OS_MESG_BLOCK);
```

src/audio/lib/thread.c
engineer124: Yeah, this can be `asPtr` assuming it doesn't cause any issues

src/code/z_camera.c
```diff
@@ -476,7 +476,7 @@ PlayerMeleeWeaponState func_800CBAAC(Camera* camera) {
     }
 }
 
-s32 Camera_GetFocalActorPos(Vec3f* dst, Camera* camera) {
+Vec3f* Camera_GetFocalActorPos(Vec3f* dst, Camera* camera) {
     PosRot focalPosRot;
     Actor* focalActor = camera->focalActor;
```
engineer124, line 481: Amazed I never caught that

src/code/z_play.c
```diff
@@ -2316,7 +2316,8 @@ void Play_Init(GameState* thisx) {
     THA_GetRemaining(&this->state.heap);
     zAllocSize = THA_GetRemaining(&this->state.heap);
     zAlloc = (uintptr_t)THA_AllocTailAlign16(&this->state.heap, zAllocSize);
-    ZeldaArena_Init(((zAlloc + 8) & ~0xF), (zAllocSize - ((zAlloc + 8) & ~0xF)) + zAlloc); //! @bug: Incorrect ALIGN16s
+    ZeldaArena_Init((void*)((zAlloc + 8) & ~0xF),
+                    (zAllocSize - ((zAlloc + 8) & ~0xF)) + zAlloc); //! @bug: Incorrect ALIGN16s
```
engineer124, line 2320: Would the bug table above help the formatter?

src/overlays/actors/ovl_En_Go/z_en_go.c
```diff
@@ -462,7 +462,7 @@ void EnGo_InitDust(EnGoEffect effect[ENGO_OTHER_EFFECT_COUNT], Vec3f pos, Vec3f
  *
  * @param effect The EnGoEffect table
  */
-void EnGo_DrawDust(EnGoEffect effect[ENGO_EFFECT_COUNT], PlayState* play2) {
+void EnGo_DrawDust(EnGoEffect effect[], PlayState* play2) {
```
engineer124, line 465: I thought we wanted these counts in?

src/code/z_camera.c
AngheloAlf: Stuff like this is so easy to miss, this is why I want this warning to be enabled

src/overlays/actors/ovl_En_Go/z_en_go.c
AngheloAlf: I removed them before the other PR that actually fixes them, I'll add them again

include/z64.h
```diff
@@ -80,7 +80,7 @@ typedef struct {
 // End of RDRAM with the Expansion Pak installed
 #define EXPANDED_RDRAM_END 0x80800000
 // Address at the end of normal RDRAM after which is room for a screen buffer
-#define FAULT_FB_ADDRESS (NORMAL_RDRAM_END - sizeof(u16[SCREEN_HEIGHT][SCREEN_WIDTH]))
+#define FAULT_FB_ADDRESS ((void*)(NORMAL_RDRAM_END - sizeof(u16[SCREEN_HEIGHT][SCREEN_WIDTH])))
```
hensldm, line 83:
```suggestion
#define FAULT_FB_ADDRESS (void*)(NORMAL_RDRAM_END - sizeof(u16[SCREEN_HEIGHT][SCREEN_WIDTH]))
```
Very nitpicky, but don't feel the extra parens are needed.

src/overlays/actors/ovl_En_Fish2/z_en_fish2.c
```diff
@@ -1091,15 +1091,15 @@ void func_80B2ADB0(EnFish2* this, Vec3f* vec, s16 arg2) {
...
             if (Rand_ZeroOne() < 0.5f) {
-                phi_v0 = gEffBubble2Tex;
+                texture = gEffBubble2Tex;
             } else {
-                phi_v0 = gEffBubble1Tex;
+                texture = gEffBubble1Tex;
             }
 
-            ptr->unk_20 = VIRTUAL_TO_PHYSICAL(SEGMENTED_TO_VIRTUAL(phi_v0));
+            ptr->unk_20 = (void*)VIRTUAL_TO_PHYSICAL(SEGMENTED_TO_VIRTUAL(texture));
```
hensldm, line 1102:
```suggestion
            ptr->unk_20 = (TexturePtr)VIRTUAL_TO_PHYSICAL(SEGMENTED_TO_VIRTUAL(texture));
```
Based on this feel like `unk_20` should actually be a `TexturePtr`.

hensldm (approved):
A couple more nits though overall LGTM. Does seem good to enable this warning.

engineer124 (approved):

tom-overton (approved):
Admittedly, I don't know if all of Elliptic's concerns were fully addressed, but this has been open for months. I think it's better for us to get this in and fix things later.

src/overlays/effects/ovl_Effect_Ss_Extra/z_eff_ss_extra.c
```diff
@@ -34,8 +34,8 @@ u32 EffectSsExtra_Init(PlayState* play, u32 index, EffectSs* this, void* initPar
     s32 objectSlot;
 
     objectSlot = Object_GetSlot(&play->objectCtx, OBJECT_YABUSAME_POINT);
-    if ((objectSlot > OBJECT_SLOT_NONE) && (Object_IsLoaded(&play->objectCtx, objectSlot))) {
-        void* segBackup = gSegments[6];
+    if ((objectSlot >= 0) && (Object_IsLoaded(&play->objectCtx, objectSlot))) {
```
engineer124, line 37: Just a bad merge. Could you bring back `OBJECT_SLOT_NONE` and remove the extra brackets around `Object_IsLoaded`

PR #1283, opened 2023-06-11
EnSw OK

StickyThwomp:
Decompiled the last remaining function.

Planning on doing docs next, just as a separate PR.

### DIFF
--- a/src/overlays/actors/ovl_En_Sw/z_en_sw.c
+++ b/src/overlays/actors/ovl_En_Sw/z_en_sw.c
@@ -358,7 +358,6 @@ s32 func_808D9440(PlayState* play, Vec3f* posA, Vec3f* posB, Vec3f* posResult, C
     return ret;
 }
 
-#ifdef NON_MATCHING
 void func_808D94D0(EnSw* this, PlayState* play, s32 arg2, s32 arg3, s16 arg4) {
     CollisionPoly* spA4;
     CollisionPoly* spA0;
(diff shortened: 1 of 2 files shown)


AngheloAlf (approved):

src/overlays/actors/ovl_En_Sw/z_en_sw.c
```diff
@@ -433,6 +431,8 @@ void func_808D94D0(EnSw* this, PlayState* play, s32 arg2, s32 arg3, s16 arg4) {
                 break;
             }
         }
+
+        if (i == 3) {}
```
isghj5, line 435: lol, does C even keep the reference to variables out of scope like this?

I know you want to document in the next build but can we add a tag to this for now?
```suggestion
        //! FAKE
        if (i == 3) {}
```

StickyThwomp: I was very surpised that it worked, but the Bomchu code gave me the hint... In that instance, if ever it satisfies this condition, the bomchu explodes. I'm not sure how picky to get over the usage of "FAKE", but I'd assume this is here for something like a debug assert/message? If that still tracks with what we wanto to use FAKE for, gimme a thumbs up and I'll do that.

isghj5: No its just a comment for us to search for later, it means we think its fake and there is a better match to be found later, a bunch of recent `if (1) {}` fake matches were recently discovered to be early returns by Engineer124, but this looks different to me, and we might figure it out later, this comment just helps us find it easier by searching a large project.

isghj5 (approved):

engineer124 (approved):
Great find!

PR #1287, opened 2023-06-13
Document EnSyatekiOkuta (Shooting Gallery Octorok)

tom-overton:
This is the last undocumented shooting gallery actor. After this is merged, I'll do one final cleanup of the shooting gallery stuff, because I know that this process has introduced a bit of inconsistency.


src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.c
```diff
@@ -76,15 +85,15 @@ static InitChainEntry sInitChain[] = {
     ICHAIN_F32(targetArrowOffset, 6500, ICHAIN_STOP),
 };
 
-Color_RGBA8 D_80A37B90 = { 255, 255, 255, 255 };
+Color_RGBA8 sDustPrimColor = { 255, 255, 255, 255 };
```
hensldm, line 88: If you are going to use pretty generic names, should probably make them static now.

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.h
```diff
@@ -15,13 +47,13 @@ typedef struct EnSyatekiOkuta {
...
+    /* 0x1DC */ Vec3f headScale;
     /* 0x1E4 */ Vec3s jointTable[OCTOROK_LIMB_MAX];
     /* 0x244 */ Vec3s morphTable[OCTOROK_LIMB_MAX];
-    /* 0x2A4 */ s16 unk_2A4;
-    /* 0x2A6 */ s16 unk_2A6;
+    /* 0x2A4 */ s16 deathTimer;
+    /* 0x2A6 */ s16 type;
     /* 0x2A8 */ UNK_TYPE1 unk_2A8[0x2];
-    /* 0x2AA */ s16 unk_2AA;
+    /* 0x2AA */ s16 circleOrCrossAlpha;
```
hensldm, line 56: I remember this got mentioned, and `hitmark` came up as a name to describe the circle/cross (would be nice to think of a name as well for the torch jumping minigame). I saw one opposition, but hitmark seems fine to me. Or maybe something like `minigameHitmark`?

hensldm (commented):
A couple thoughts.

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.h
tom-overton: I didn't like it in this case because, when you hit them, an actual hitmark shows up too.
GLideN64_ZELDA_MAJORA'S_MASK_009
I didn't want to cause too much confusion between the *actual* hitmark and the circle/cross that appears. That's why I went with the current name - it's as unambiguous as possible. I also think that, if the goal is synchronize this name with the jumping minigame one, then putting "hit" in the name is the opposite of what we want. In the jumping minigame, you don't really "hit" anything.

What do you think? I definitely think this name is a little unwieldy, but I tried to avoid using "hitmark" for a reason. Do you think I should go with "hitmark" anyways?

hensldm: Oh true hitmark doesn't really work for jumping minigame. A few other brainstorming ideas are `minigameMark` (just remove hit) or perhaps `scoreMark`, maybe mix the two `minigameScoreMark`

tom-overton: Hmm, the thing with `scoreMark` is that only the circle actually corresponds to the score. When the circle appears, the score increases by one. When the cross appears, the timer goes down, and the score remains the same. I forgot what happens in the jumping minigame, but I think it's similar in the sense that the cross doesn't actually change the score.

For `minigameMark`, I think it's a little better, but it's also more ambiguous than what currently exists. `circleOrCrossAlpha` makes it pretty clear what it controls, but `minigameMarkAlpha` is a bit more vague. If you don't already know the specifics as to how the minigame works, it might be confusing what this actually changes.

hensldm (approved):

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.h
tom-overton: And just to add even more confusion to this, in MM3D, in North America and Europe, they changed the circle to a checkmark:
image

So I think there *is* a better name for this...but I can't quite think of it. Because if this was ported to MM3D, we'd have to change the name of this to `circleCheckOrCrossAlpha`, which is a pretty crazy name.

I don't know if this helps, or if it adds even more confusion, but this is currently where I'm at. I think there's a name better than what I have, but I don't know what that name might be.

engineer124: What about `hitResultMark`? or just `hitResult`? `hitResultAlpha` as well

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.c
```diff
@@ -121,187 +120,241 @@ void EnSyatekiOkuta_Destroy(Actor* thisx, PlayState* play) {
...
         }
 
         this->actor.scale.y = this->actor.scale.x;
         this->actor.scale.z = this->actor.scale.x;
     }
 }
 
-void func_80A368E0(EnSyatekiOkuta* this, PlayState* play) {
+/**
+ * Adjust's the collider's dimensions and position based on a few different factors, like the Octorok's
```
engineer124, line 348:
```suggestion
 * Adjusts the collider's dimensions and position based on a few different factors, like the Octorok's
```

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.h
engineer124: That's true. I suppose just by nature of being different minigames, I'd be okay with `hitResult` here and `jumpResult` for the jumping game, but I can see why consistency would be nice. What about `eventResult` or just `resultMark`?

tom-overton: I think `hitResult` is the best thing proposed so far. The thing with `eventResult` is that "event" in this codebase usually means something specific, and in this case it means something else, so I can see how it might cause confusion. For `resultMark`, it's a bit ambiguous. The result of what? It's the result of getting hit, hence why I think `hitResult` is the best thing so far.

However, I took a look at the jumping game stuff, because I was curious if *the developers* treated them as the same. And they totally did. The names are `jgame_maru_model` and `jgame_peke_model`, just like how the Octoroks have it. They even swapped the circle for a green checkmark in the US and EU releases of MM3D! 

image

Now, we're not beholden to their conventions; we can do whatever we want. But at the same time, these are clearly the same thing, so it feels a bit odd to deviate. I'd be fine with either the current `circleOrCrossAlpha` or `hitResultAlpha` (and presumably `jumpResultAlpha` for the jumping game), but after all this discussion, I'm actually kind of leaning to the former, even if I wasn't that big of a fan to begin with. What do you think?

tom-overton: Sure, let's go with that then.

src/overlays/actors/ovl_En_Syateki_Man/z_en_syateki_man.c
```diff
@@ -1323,27 +1324,27 @@ void EnSyatekiMan_Town_RunGame(EnSyatekiMan* this, PlayState* play) {
...
-        if (this->perGameVar2.octorokHitType != SG_OCTO_HIT_TYPE_NONE) {
-            if (this->perGameVar2.octorokHitType == SG_OCTO_HIT_TYPE_BLUE) {
+        if (this->perGameVar2.octorokHitType != SG_OCTO_TYPE_NONE) {
+            if (this->perGameVar2.octorokHitType == SG_OCTO_TYPE_BLUE) {
                 gSaveContext.timerTimeLimits[TIMER_ID_MINIGAME_1] -= SECONDS_TO_TIMER_PRECISE(2, 50);
                 sModFromLosingTime = (sModFromLosingTime + 25) % 50;
             }
 
-            this->perGameVar2.octorokHitType = SG_OCTO_HIT_TYPE_NONE;
+            this->perGameVar2.octorokHitType = SG_OCTO_TYPE_NONE;
```
engineer124, line 1333:
```suggestion
            this->perGameVar2.octorokType = SG_OCTO_TYPE_NONE;
```
Should this reflect dropping `HIT` in the enum as well?

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.c
```diff
@@ -121,187 +120,241 @@ void EnSyatekiOkuta_Destroy(Actor* thisx, PlayState* play) {
...
 
     while (actorIt != NULL) {
         if (actorIt->id == ACTOR_EN_SYATEKI_MAN) {
             this->actor.parent = actorIt;
-            func_80A362F8(this);
+            EnSyatekiOkuta_SetupDoNothing(this);
             break;
         } else {
             actorIt = actorIt->next;
         }
```
engineer124, line 193:
```suggestion
            EnSyatekiOkuta_SetupDoNothing(this);
            break;
        }
        actorIt = actorIt->next;
```
Just personal preference

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.c
```diff
@@ -121,187 +120,241 @@ void EnSyatekiOkuta_Destroy(Actor* thisx, PlayState* play) {
...
 }
 
-void func_80A363B4(EnSyatekiOkuta* this, PlayState* play) {
-    if (Animation_OnFrame(&this->skelAnime, 2.0f) || Animation_OnFrame(&this->skelAnime, 15.0f)) {
-        if (func_80A361F4(this)) {
+/**
+ * Jumps out of the water and starts floating.
+ */
+void EnSyatekiOkuta_Appear(EnSyatekiOkuta* this, PlayState* play) {
+    if (Animation_OnFrame(&this->skelAnime, 2.0f) || (Animation_OnFrame(&this->skelAnime, 15.0f))) {
```
engineer124, line 223:
```suggestion
    if (Animation_OnFrame(&this->skelAnime, 2.0f) || Animation_OnFrame(&this->skelAnime, 15.0f)) {
```

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.c
```diff
@@ -121,187 +120,241 @@ void EnSyatekiOkuta_Destroy(Actor* thisx, PlayState* play) {
...
 
-void func_80A365EC(EnSyatekiOkuta* this, PlayState* play) {
-    Vec3f sp84;
-    Vec3f sp78;
+/**
+ * Plays the death animation and slowly shrinks the Octorok. Also spawns various bubble and dust
+ * effects as it dies. Once the Octorok is finished with its death animation, this function will
+ * make it do nothing until the Shooting Gallery Man tells it to appear again.
+ */
+void EnSyatekiOkuta_Die(EnSyatekiOkuta* this, PlayState* play) {
```
engineer124, line 289: So the `Octorok` doesn't die in the sense that it's despawned, but more goes into hiding until it can reappear for the next round it's involved in?

src/overlays/actors/ovl_En_Syateki_Man/z_en_syateki_man.c
tom-overton: No, I don't think we want to do this, though I think this name can definitely be improved. What this variable represents is the type of the last Octorok that was hit by the player. This block of code runs every single frame the minigame is running:
```c
        if (this->perGameVar2.octorokHitType != SG_OCTO_TYPE_NONE) {
            if (this->perGameVar2.octorokHitType == SG_OCTO_TYPE_BLUE) {
                gSaveContext.timerTimeLimits[TIMER_ID_MINIGAME_1] -= SECONDS_TO_TIMER_PRECISE(2, 50);
                sModFromLosingTime = (sModFromLosingTime + 25) % 50;
            }

            this->perGameVar2.octorokHitType = SG_OCTO_TYPE_NONE;
        }
```

And every time an Octorok *gets* hit, it does this:
```c
    if (EnSyatekiOkuta_CheckCollision(this, play)) {
        syatekiMan = (EnSyatekiMan*)this->actor.parent;
        if (this->type == SG_OCTO_TYPE_RED) {
            Actor_PlaySfx(&this->actor, NA_SE_SY_TRE_BOX_APPEAR);
            play->interfaceCtx.minigamePoints++;
            syatekiMan->score++;
            syatekiMan->perGameVar2.octorokHitType = SG_OCTO_TYPE_RED;
        } else {
            Actor_PlaySfx(&this->actor, NA_SE_SY_ERROR);
            syatekiMan->perGameVar2.octorokHitType = SG_OCTO_TYPE_BLUE;
        }

        EnSyatekiOkuta_SetupDie(this);
    } 
```

In other words, every time the player hits an Octorok, it updates a variable on the Shooting Gallery Man. He then checks to see if it was a blue Octorok, and if so, reduces the timer a bit. No matter what, he clears the variable so that the next time an Octorok gets hit, he'll be aware of it.

Maybe something like `lastHitOctorokType` would work here? But we definitely don't want to just do `octorokType`, since that's too vague as to what the purpose of this variable is. Hopefully, that helps explain it!

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.c
tom-overton: Yeah, that's correct. You'll notice that there's only a single `Actor_Kill` in here, and it's in the Init function. In other words, once the Octorok is spawned, it *stays* spawned forever; "killing" it just makes it hide until it's ready again.

src/overlays/actors/ovl_En_Syateki_Man/z_en_syateki_man.c
engineer124: As I was reading that, `lastHitOctorokType` was the name that came to mind before I got to the end, so that's a good sign. Works for me

src/overlays/actors/ovl_En_Syateki_Okuta/z_en_syateki_okuta.c
```diff
@@ -60,6 +60,15 @@ static ColliderCylinderInit sCylinderInit = {
...
+    /* 0 */ SG_OCTO_ANIM_SHOOT, // unused
+    /* 1 */ SG_OCTO_ANIM_DIE,
+    /* 2 */ SG_OCTO_ANIM_HIDE,
+    /* 3 */ SG_OCTO_ANIM_FLOAT,
+    /* 4 */ SG_OCTO_ANIM_APPEAR,
+    /* 5 */ SG_OCTO_ANIM_HIT // unused
+} ShootingGalleryOctorokAnimation;
+
 static AnimationInfo sAnimationInfo[] = {
     { &gOctorokShootAnim, 1.0f, 0.0f, 0.0f, ANIMMODE_ONCE, -1.0f },
```
engineer124, line 73:
```suggestion
    /* 5 */ SG_OCTO_ANIM_HIT, // unused
    /* 6 */ SG_OCTO_ANIM_MAX
} ShootingGalleryOctorokAnimation;

static AnimationInfo sAnimationInfo[SG_OCTO_ANIM_MAX] = {
    { &gOctorokShootAnim, 1.0f, 0.0f, 0.0f, ANIMMODE_ONCE, -1.0f }, // SG_OCTO_ANIM_SHOOT
```

I've been adding `MAX` to the animations to make sure any added animations have their corresponding enum remain in sync.

Also possibly hasn't been finalised, but I like having enum values commented along with the animations (i.e. the `// SG_OCTO_ANIM_SHOOT`), but this does feel a little redundant once the values are named. It is a quick way to go from the enum to the Animation though. It becomes annoying though when the animation name is long enough to force a new line from the formatter.

Long story short, you don't need the comments by the side until we decide if we want them in my animation cleanup

engineer124 (approved):

AngheloAlf (approved):

PR #1299, opened 2023-06-19
Create an enum for the various `face_reaction` reaction sets

tom-overton:
Was looking into this while cleaning up Syateki-related actors, and I figured it was so easy I should just knock it out. The only thing I couldn't figure out is what header to put this into; I know many people have been trying to un-bloat the headers, and I'm worried I may be bloating them by doing this. For now, I put this in `z64message.h`, since it's somewhat message-related.


include/z64message.h
```diff
@@ -12,6 +12,17 @@ struct OcarinaStaff;
...
+    /* 0x11 */ FACE_REACTION_SET_JIM = 0x11,
+    /* 0x12 */ FACE_REACTION_SET_BOMBERS,
+    /* 0x15 */ FACE_REACTION_SET_BOMBERS_HIDEOUT_GUARD = 0x15,
+    /* 0x2D */ FACE_REACTION_SET_TREASURE_CHEST_SHOP_GAL = 0x2D,
+    /* 0x2F */ FACE_REACTION_SET_CURIOSITY_SHOP_MAN = 0x2F,
+    /* 0x30 */ FACE_REACTION_SET_TOWN_SHOOTING_GALLERY_MAN,
+    /* 0x31 */ FACE_REACTION_SET_SWAMP_SHOOTING_GALLERY_MAN,
+    /* 0x36 */ FACE_REACTION_SET_PART_TIMER = 0x36
+} FaceReactionSet;
+
```
AngheloAlf, line 25: I'm not sure if this is the better header for this stuff, but I think it is fine for now, we can always move it after.

Btw, can you move the function declaration from `functions.h` to this header too?

AngheloAlf: Oh, and can you check if the function would still match using the enum as a type too?
```c
u16 Text_GetFaceReaction(PlayState* play, FaceReactionSet reactionSet);
```

src/overlays/actors/ovl_En_Bombers/z_en_bombers.c
```diff
@@ -295,11 +295,11 @@ void func_80C03AF4(EnBombers* this, PlayState* play) {
             break;
     }
 
-    if (!Text_GetFaceReaction(play, 0x12)) {
+    if (!Text_GetFaceReaction(play, FACE_REACTION_SET_BOMBERS)) {
```
hensldm, line 298:
```suggestion
    if (Text_GetFaceReaction(play, FACE_REACTION_SET_BOMBERS) != 0) {
```

hensldm (commented):
1 small nit

include/z64message.h
tom-overton: >Btw, can you move the function declaration from functions.h to this header too?

No, I can't, at least not without further changes. This function declaration relies on `PlayState` being defined, but `PlayState` is defined in z64.h, which includes `z64message.h`. So as-is, this creates a circular dependency.

src/overlays/actors/ovl_En_Bombers/z_en_bombers.c
tom-overton: Did you mean `== 0`? Because doing `!= 0` here isn't equivalent (and doesn't match).

hensldm: Ah yeah :), that should be a `==0`

include/z64message.h
```diff
@@ -8,10 +8,23 @@
...
+u16 Text_GetFaceReaction(struct PlayState* play, FaceReactionSet reactionSet);
 typedef enum TextState {
```
AngheloAlf, line 28: Sorry for being annoying, but can you add an empty line between those?

tom-overton: You're not being annoying! I just missed this (or rather, my IDE was *already* showing a blank line here when none actually existed in practice).

AngheloAlf (approved):

hensldm (approved):

PR #1302, opened 2023-06-20
Cleanup of Shooting Gallery actors

tom-overton:
This PR has a few main objectives:
- Some actors used `EN_SYATEKI_*` for things like macros, enums, etc., while others used `SG_*`. This PR standardizes on the much shorter `SG_*`.
- While working on the individual Syateki actors (e.g., Crow, Dekunuts, Wf, and Okuta), I deliberately tried to avoid using the word "spawn", because that means something specific in this codebase. I applied that same deliberation to EnSyatekiMan; it was the first actor I documented of this set, so I just wasn't thinking about it at the time.
- I tried to make defines for things like scores and waves to make it a little more "shiftable," for lack of a better term. The code as-is relied on a lot of magic numbers that made modifying the game logic pretty tricky. For proof that it works, here's a video of a very simple mod that adds a fifth wave; it's quite tricky to finish it in time!

https://github.com/zeldaret/mm/assets/12245827/9c724ed7-e4e7-4f21-8895-d6a8945480a8

There's some other stuff in here like code cleanup, additional comments, defining `WEEKEVENTREGS` etc., but these are pretty simple and fairly uncontroversial. The main three changes I listed above are probably the ones that are going to cause contention.


src/overlays/actors/ovl_En_Shn/z_en_shn.c
```diff
@@ -72,8 +72,8 @@ s32 func_80AE615C(EnShn* this, s32 animIndex) {
     static AnimationInfoS sAnimationInfo[] = {
         { &gBurlyGuyHandsOnTableAnim, 1.0f, 0, -1, 0, 0 },
         { &gBurlyGuyHandsOnTableAnim, 1.0f, 0, -1, 0, -4 },
-        { &gSwampGuideChinScratchAnim, 1.0f, 0, -1, 0, 0 },
-        { &gSwampGuideChinScratchAnim, 1.0f, 0, -1, 0, -4 },
+        { &gBurlyGuyChinScratchAnim, 1.0f, 0, -1, 0, 0 },
+        { &gBurlyGuyChinScratchAnim, 1.0f, 0, -1, 0, -4 },
```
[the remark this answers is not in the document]
tom-overton: Very strange that this wasn't already the case; I thought anon or someone else made a big PR to update all of these before? Whatever the case, I updated them.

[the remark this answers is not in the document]
engineer124: There's actually a few scattered throughout the repo that are missing these animmode that I plan to catch in my anim cleanup sweep

src/overlays/actors/ovl_En_Syateki_Crow/z_en_syateki_crow.c
```diff
@@ -206,21 +206,21 @@ void EnSyatekiCrow_Fly(EnSyatekiCrow* this, PlayState* play) {
...
         this->isActive = false;
-        syatekiMan->guayFlags &= ~(1 << EN_SYATEKI_CROW_GET_INDEX(&this->actor));
+        syatekiMan->guayFlags &= ~(1 << SG_GUAY_GET_INDEX(&this->actor));
         EnSyatekiCrow_SetupWaitForSignal(this);
     }
 
     SkelAnime_Update(&this->skelAnime);
     this->actor.shape.yOffset = (fabsf(this->skelAnime.curFrame - 3.0f) * 150.0f) + 1700.0f;
-    if ((syatekiMan->perGameVar1.guaySpawnTimer % 90) == 0) {
+    if ((syatekiMan->guayAppearTimer % 90) == 0) {
```
engineer124, line 215: Outside the scope of this PR and definitely don't change here, but just writing down that I sort of wish we had something more to indicate that a variable is a union/shared between others. Even a convention like `syatekiMan->u.guayAppearTimer` idk.

engineer124 (approved):
lgtm

src/overlays/actors/ovl_En_Syateki_Crow/z_en_syateki_crow.c
tom-overton: Whether unions should be named or unnamed, or whether they were even *allowed* to be unnamed at all, seemed to change wildly depending on whatever the whims of any given lead were at any given point. I would absolutely *love* some degree of consistency or guidance here, because I've been told to both add a name to a union and remove it at different points in time. In particular, I always hated the `perGameVarX` thing here, and I only did it to get it merged; I removed it here because I saw that you in particular were using unnamed unions in EnTest6, and that this might be my chance to get rid of something I always disliked.

engineer124: Yeah, that's fair. It's definitely not consistent and people have different preferences. OoT also tends to avoid unions much more than MM for reused variables, so it'll be an interesting discussion moving forward.

tom-overton: A couple of actors in OoT do a thing where they use unions in all but name. For example, `z_en_tite` has struct variables with names like `actionVar1`, then those are redefined to be more specific names like so:
```c
// EnTite_Attack (vQueuedJumps also used by EnTite_MoveTowardPlayer)
#define vAttackState actionVar1
#define vQueuedJumps actionVar2

//  EnTite_FlipOnBack
#define vOnBackTimer actionVar1
#define vLegTwitchTimer actionVar2
```

I actually started with this approach with ObjTokeidai (an actor with a huge number of reused variables), and I was told to use unions instead. So that's what I've been doing with all my actors, and in doing so, I've repeatedly run into mixed messaging about *how* I should be using unions.

Like you said, this isn't really in scope for this PR. I would love to see and participate in discussions about how we should be handling these sort of reused variables. But I would *also* love to get rid of this thing I never truly liked in the first place, hence why I snuck it into this PR. I can revert it if desired, though as you might guess, I would really prefer it if I didn't.

engineer124: I'm fine with leaving it to personal preference until a consensus is reached.

Tbh that's the first time I've seen that trick used in OoT. I knew the defines with the `v` prefix was used for locally scoped temps, but not for struct members (other than overlay effects which uses that trick for `this->regs[X]`). Not sure what that offers over unions except for dispersing the information to more places.

I used to like the `v` prefix defines for multi-use temps but am not as fond of it anymore.

jpburnett (approved):
A pretty straight shot if you ask me! (Left a comment in discord about a simple question, but the comments and documentation was very helpful!)

engineer124 (approved):

PR #1312, opened 2023-06-30
Talk Exchange System Docs

hensldm:
Documents the actor/subs functions revolving around the exchange system. It seems very similar to the GetItem system in that the actor offers the ability, and it is up to player to implement and do it. With this in mind I tried to mirror that system in naming and comments. In MM, this actually doesn't get used for the main purpose of exchanging items with NPCs without first talking to them. Instead it is generally used as a way to toggle the A button, and if passed a -1 ~~(PLAYER_IA_HELD)~~ will auto continue the conversation with the player (for example after closing the textbox after receiving an item). While documenting the SubS exchange functions, added an enum for the item mode for `SubS_OfferItem`, which is generally the 3 least significant bits in the actors stateflags for the actors that use it. Again this is just used to offer the ability to speak with the player, and doesn't actually offer items.

~~Last I renamed PLAYER_IA_MINUS1->PLAYER_IA_HELD mainly due `func_80123C58` which seems to set the itemaction to the current held item action. This actually doesn't seem to be used much, and the main use of PLAYER_IA_HELD is the exchange system function of continuing conversations.~~


src/overlays/actors/ovl_En_Rg/z_en_rg.c
```diff
@@ -750,7 +750,7 @@ void EnRg_Init(Actor* thisx, PlayState* play) {
         this->actor.flags &= ~ACTOR_FLAG_1;
         this->unk_310 = 8;
         this->actor.gravity = -1.0f;
-        SubS_UpdateFlags(&this->unk_310, 3, 7);
+        SubS_UpdateFlags(&this->unk_310, SUBS_OFFER_MODE_ONSCREEN, SUBS_OFFER_MODE_MAX);
```
isghj5, line 753: Er, it doesn't look like this actor used one of the documented Subs functions, and the UpdateFlags function doesn't seem to be specific to any system, are you sure these are the right enums for this actor's stateFlag variable?

isghj5 (commented):
Ah good maybe this documentation will solve some of my dialogue bugs

src/overlays/actors/ovl_En_Rg/z_en_rg.c
hensldm: Ah your right, I'll change it back. Though it is weird since I actually don't see those 3 bits getting used in `EnRg` (or `EnMtTag`) though maybe some other actor uses it 🤷.

src/code/z_actor.c
```diff
@@ -1873,10 +1873,32 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
...
+ * - `PLAYER_IA_NONE`:
+ *    Allows the player to speak to the actor (by pressing A).
+
+ *
+ * @return true If the player actor is capable of accepting the offer.
+ *
+ * Note: There is only one instance of using this for actually exchanging an item with the player
+ * without speaking first. Every other instance is to either offer to speak, or continue the current conversation.
+ */
+s32 Actor_OfferExchangeItem(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, PlayerItemAction exchangeItemId) {
```
engineer124, line 1901: Thanks for these docs! I've actually been interested in the exchange system since it was interlinked with bottles which I was looking at in detail, but then the exchange system is also interlinked with the talk system.

I think anytime you want to talk to an actor, you need to go through this function. An actor needs to indicate that it can be talked to through an offer, and I think this function name should capture that. Of course, they mainly do that through the wrapper of the function `Actor_OfferSpeak`, but this function should capture that you can talk with the choice of exchange.

My rough first idea for these function names are:
`Actor_OfferToTalkWithExchange`
`Actor_OfferToTalkWithExchangeRadius`
`Actor_OfferToTalk`
`Actor_OfferToTalkNearby`

I also think `Talk` should be used with `Speak` because this would apply to both `Speak` and `Check` (for example, signs and posters in Clock Town).

It's definitely open to discussion though, thoughts?

src/code/z_actor.c
```diff
@@ -1873,10 +1873,32 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
...
+ *
+ * This function carries an exchange item request to the player actor if context allows it (e.g. the player is in range
+ * and not busy with certain things). The player actor performs the requested action itself.
+ *
+ * The following description of what the `ItemAction` values can do is provided here for completeness, but these
+ * behaviors are entirely out of the scope of this function. All behavior is defined by the player actor.
+ *
+ * - Positive values in range (`PLAYER_IA_NONE < ItemAction < PLAYER_IA_MAX`):
+ *    Offers the ability to exhange an item with the player.
+ *    Not all postivie values are implemented properly for this to work.
```
engineer124, line 1888:
```suggestion
 *    Not all positive values are implemented properly for this to work.
```

src/code/z_player_lib.c
```diff
@@ -621,7 +621,7 @@ PlayerItemAction func_80123810(PlayState* play) {
             play->interfaceCtx.unk_222 = 0;
             play->interfaceCtx.unk_224 = 0;
             Interface_SetHudVisibility(play->msgCtx.unk_120BC);
-            return PLAYER_IA_MINUS1;
+            return PLAYER_IA_HELD;
```
engineer124, line 624: You might also be interested in `func_80123810` just to help evaluate the `PLAYER_IA_HELD` name. This also ties into `play->unk_18794` which processes the exchange request, Currently:
```c
/**
 * Gives the player a request to use a C-Button to present an item to exchange
 * Return `exchangeItemAction` if an item is presented
 * Return `PLAYER_IA_NONE` if no option is offered yet
 * Return `PLAYER_IA_HELD` if the offer is declined or invalid
 */
func_80123810

```
Here's `func_80123810`
https://github.com/engineer124/mm/blob/b400822da32565533ff6d7a940a01c0aa373288a/src/code/z_player_lib.c#L616-L622
And `play->unk_18794`
https://github.com/engineer124/mm/blob/b400822da32565533ff6d7a940a01c0aa373288a/src/overlays/actors/ovl_player_actor/z_player.c#L10636
https://github.com/engineer124/mm/blob/b400822da32565533ff6d7a940a01c0aa373288a/src/overlays/actors/ovl_player_actor/z_player.c#L20923

And here's the main action for processing the exchange item:
https://github.com/engineer124/mm/blob/b400822da32565533ff6d7a940a01c0aa373288a/src/overlays/actors/ovl_player_actor/z_player.c#L17677

And some exchange item logic:
https://github.com/engineer124/mm/blob/b400822da32565533ff6d7a940a01c0aa373288a/src/overlays/actors/ovl_player_actor/z_player.c#L7412-L7441

src/overlays/actors/ovl_En_Nb/z_en_nb.c
```diff
@@ -587,9 +583,9 @@ s32 func_80BC0B98(EnNb* this, PlayState* play, ScheduleOutput* scheduleOutput) {
 
 s32 func_80BC0C0C(EnNb* this, PlayState* play, ScheduleOutput* scheduleOutput) {
     if (!CHECK_EVENTINF(EVENTINF_43)) {
-        SubS_UpdateFlags(&this->stateFlags, EN_NB_FLAG_1 | EN_NB_FLAG_2, EN_NB_FLAG_1 | EN_NB_FLAG_2 | EN_NB_FLAG_4);
+        SubS_UpdateFlags(&this->stateFlags, SUBS_OFFER_MODE_ONSCREEN, SUBS_OFFER_MODE_MAX);
     } else {
-        SubS_UpdateFlags(&this->stateFlags, EN_NB_FLAG_4, EN_NB_FLAG_1 | EN_NB_FLAG_2 | EN_NB_FLAG_4);
+        SubS_UpdateFlags(&this->stateFlags, SUBS_OFFER_MODE_AUTO, SUBS_OFFER_MODE_MAX);
```
engineer124, line 588:
```suggestion
        SubS_UpdateOfferFlags(&this->stateFlags, SUBS_OFFER_MODE_AUTO, SUBS_OFFER_MODE_MAX);
```
Are these always used with the `SUBS_OFFER_XXX` flags?

src/code/z_actor.c
hensldm: I had actually been considering something like that. Though I would probably suggest dropping the `To` and `With` as I'm not sure they add much. Something like
* `Actor_OfferTalkExchange`
* `Actor_OfferTalkExchangeRadius`
* `Actor_OfferTalk`
* `Actor_OfferTalkNearby`
And for completeness the Subs Functions
* `SubS_OfferItem` -> `SubS_OfferTalkExchange` (while getItem can be used here, the focus seemed to be on the exchange system so I think this is fine)
* `SubS_OfferExchangeItemFacing` -> `SubS_OfferTalkExchangeFacing`

src/overlays/actors/ovl_En_Nb/z_en_nb.c
hensldm: There is currently 1 use that is possibly not using the `OfferFlags` (it still uses 7 as the unset bits, but the actor doesn't actually use the function, and those bits don't seem to be used anywhere https://github.com/zeldaret/mm/pull/1312#discussion_r1247461152).
The function is general enough it doesn't have to be used with just the `OfferFlags` but that does seem to be how it was used, so I'm actually not really sure what is better.

include/z64subs.h
```diff
@@ -10,16 +10,28 @@ extern Vec3f gOneVec3f;
...
+typedef enum {
+    /* 0 */ SUBS_OFFER_MODE_NONE,
+    /* 1 */ SUBS_OFFER_MODE_GET_ITEM,
+    // Furthur modes all deal with offering exhange items
+    /* 2 */ SUBS_OFFER_MODE_NEARBY,
+    /* 3 */ SUBS_OFFER_MODE_ONSCREEN,
+    /* 4 */ SUBS_OFFER_MODE_AUTO,
+    /* 5 */ SUBS_OFFER_MODE_AUTO_TARGETED, // Also checks nearby and on screen
+    /* 6 */ SUBS_OFFER_MODE_AUTO_NEARBY_ONSCREEN,
+    /* 7 */ SUBS_OFFER_MODE_MAX
```
AngheloAlf, line 23: This MAX value is a bit weird, because it is used for and'ing bits, it is more all the bits that can be used by this system instead of just the next value.
For example if we added an extra value to this enum (having the value 7), then MAX should have the value 7 too and not 8, because those are the bits that we want to mask.

So, what about renaming this to `SUBS_OFFER_MODE_MASK` (or similar) and defining it as all the values of this enum or'd?

AngheloAlf (commented):
Could you also add the stuff you renamed to namefixer.py?

include/z64subs.h
```diff
@@ -10,16 +10,28 @@ extern Vec3f gOneVec3f;
 
 #define SUBS_TIME_PATHING_ORDER 3
 
+// NOTE: all on screen checks are bugged, and always evaluate to true, even if off screen
+typedef enum {
+    /* 0 */ SUBS_OFFER_MODE_NONE,
+    /* 1 */ SUBS_OFFER_MODE_GET_ITEM,
+    // Furthur modes all deal with offering exhange items
```
engineer124, line 17:
```suggestion
    // Further modes all deal with offering exchange items
```

include/z64subs.h
```diff
@@ -10,16 +10,28 @@ extern Vec3f gOneVec3f;
...
+
 typedef enum {
     /* 0 */ SUBS_CUTSCENE_WITH_PLAYER,
     /* 1 */ SUBS_CUTSCENE_NORMAL,
     /* 2 */ SUBS_CUTSCENE_WITH_PLAYER_SET_FLAG
 } SubSCutsceneType;
 
-//! TODO: rename based on func_8013E748 and func_800B8500
-typedef s32 (*func_8013E748_VerifyFunc)(struct PlayState*, Actor*, void*);
+typedef s32 (*VerifyExchangeItemActor)(struct PlayState*, Actor*, void*);
```
engineer124, line 32:
```suggestion
typedef s32 (*VerifyExchangeActor)(struct PlayState*, Actor*, void*);
```
Might be fine?

src/code/z_actor.c
```diff
@@ -1873,14 +1873,36 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
...
+ *
+ * The following description of what the `ItemAction` values can do is provided here for completeness, but these
+ * behaviors are entirely out of the scope of this function. All behavior is defined by the player actor.
+ *
+ * - Positive values in range (`PLAYER_IA_NONE < ItemAction < PLAYER_IA_MAX`):
+ *    Offers the ability to initate the conversation with an item from the player.
+ *    Not all positive values are implemented properly for this to work.
+ *    Working ones seem to be PLAYER_IA_PICTO_BOX and PLAYER_IA_BOTTLE_MIN <= ItemAction < PLAYER_IA_MASK_MIN
+ * - `PLAYER_IA_NONE`:
+ *    Allows the player to speak to the actor (by pressing A).
```
engineer124, line 1891:
```suggestion
 *    Allows the player to speak to or check the actor (by pressing A).
```

src/code/z_actor.c
```diff
@@ -1873,14 +1873,36 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
...
+ * This function carries an exchange item request to the player actor if context allows it (e.g. the player is in range
+ * and not busy with certain things). The player actor performs the requested action itself.
+ *
+ * The following description of what the `ItemAction` values can do is provided here for completeness, but these
+ * behaviors are entirely out of the scope of this function. All behavior is defined by the player actor.
+ *
+ * - Positive values in range (`PLAYER_IA_NONE < ItemAction < PLAYER_IA_MAX`):
+ *    Offers the ability to initate the conversation with an item from the player.
+ *    Not all positive values are implemented properly for this to work.
+ *    Working ones seem to be PLAYER_IA_PICTO_BOX and PLAYER_IA_BOTTLE_MIN <= ItemAction < PLAYER_IA_MASK_MIN
```
engineer124, line 1889: Yeah, this logic seems consistent with:
https://github.com/engineer124/mm/blob/c04dc9d02e0f63cae7a5c5c6e836e8e81bca306f/src/overlays/actors/ovl_player_actor/z_player.c#L7371-L7396

The only notable thing missed is that they also use the exchange system to plant a magic bean, but there is a special case and a different function used for magic beans within the exchange system. You can the code below used it used in z_obj_bean.c (Floating Bean Plant / Soft Soil)
`Actor_OfferTalkExchangeRadius(&this->dyna.actor, play, 28.0f, PLAYER_IA_MAGIC_BEANS);`

src/overlays/actors/ovl_En_Rg/z_en_rg.c
engineer124: Because bits 1,2,4 are never used, that seems a bit too suspicious that those flags were still tied to the SUBS_OFFER modes and that it's simply doesn't use that feature. I would still tie this to the SUBS_OFFER since this seems like the only exception where `SubS_UpdateFlags` seems designed for the SUBS_OFFER? And then just add a comment noting these flags are unused. I would even still name `SubS_UpdateFlags` -> `SubS_UpdateOfferFlags` too.

Overall thoughts?

include/z64subs.h
hensldm: Hmm, we've already kind of done this with the race flags (EventInfDogRaceState and WeekEventRegHorseRaceState) so still kind of like using `MAX` as the name. Though I think it would help if I do hard code it to 7, that way it is more noticeable.

include/z64subs.h
hensldm: Still think it should involve `TalkExchange` since that is the function it uses.

src/code/z_actor.c
hensldm: Removed the `seem to be` and added a note
`Note: While PLAYER_IA_BEANS works, it is special cased to just plant the bean with no talking.`
Is that good?

src/overlays/actors/ovl_En_Nb/z_en_nb.c
```diff
@@ -587,9 +583,9 @@ s32 func_80BC0B98(EnNb* this, PlayState* play, ScheduleOutput* scheduleOutput) {
 
 s32 func_80BC0C0C(EnNb* this, PlayState* play, ScheduleOutput* scheduleOutput) {
     if (!CHECK_EVENTINF(EVENTINF_43)) {
-        SubS_UpdateFlags(&this->stateFlags, EN_NB_FLAG_1 | EN_NB_FLAG_2, EN_NB_FLAG_1 | EN_NB_FLAG_2 | EN_NB_FLAG_4);
```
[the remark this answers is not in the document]
hensldm: They were never declared in the header?

src/code/z_player_lib.c
hensldm: Discussed on discord,
decided to go with two enum values both equaling `-1` `PLAYER_IA_EXCH_CONTINUE` and `PLAYER_IA_HELD`.

include/z64subs.h
EllipticEllipsis: Agreed, the enum should not be used for masking, it's both not semantically correct, and liable to cause errors. The other question is whether it makes sense to use the MAX as a function argument, and I agree with Anghelo that it doesn't seem to make sense for that either, an ALL define makes more sense for both.

hensldm: Ok, I'll take it out of the enum and make a `SUBS_OFFER_MODE_MASK` define (I'll do the same for the race flags)

src/overlays/actors/ovl_En_Rg/z_en_rg.c
isghj5: Yeah I would be fine with it so long as there is documentation supporting the suspicion.

hensldm: Added a comment above, let me know if it isn't enough.

include/z64save.h
```diff
@@ -1442,16 +1442,17 @@ typedef enum {
     /* 1 */ WEEKEVENTREG_HORSE_RACE_STATE_START,
     /* 2 */ WEEKEVENTREG_HORSE_RACE_STATE_2,
     /* 3 */ WEEKEVENTREG_HORSE_RACE_STATE_3,
-    /* 4 */ WEEKEVENTREG_HORSE_RACE_STATE_4,
-    /* 7 */ WEEKEVENTREG_HORSE_RACE_STATE_MAX = 7
+    /* 4 */ WEEKEVENTREG_HORSE_RACE_STATE_4
 } WeekEventRegHorseRaceState;
 
-#define GET_WEEKEVENTREG_HORSE_RACE_STATE (WEEKEVENTREG(92) & WEEKEVENTREG_HORSE_RACE_STATE_MAX)
+#define WEEKEVENTREG_HORSE_RACE_STATE_MASK 7
```
engineer124, line 1448: Should you define the mask here in a similar manner to `SUBS_OFFER_MODE_MASK`, or'ing all the states? (and `EVENTINF_DOG_RACE_STATE_MASK `)

src/code/z_sub_s.c
```diff
@@ -1513,15 +1521,18 @@ Actor* SubS_FindActorCustom(PlayState* play, Actor* actor, Actor* actorListStart
...
+ * Will extend a TalkExchange offer to the player if the actor is verified with a custom callback.
+ * The callback should return `true` when the actor is succesfully verified.
+ */
+s32 SubS_OfferTalkExchangeCustom(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, s32 exchangeItemAction,
+                                 void* data, VerifyTalkExchangeActor verifyActor) {
+    s32 canAccept = false;
 
-    if ((verifyFunc == NULL) || ((verifyFunc != NULL) && verifyFunc(play, actor, data))) {
-        ret = func_800B8500(actor, play, xzRange, yRange, exchangeItemId);
+    if ((verifyActor == NULL) || ((verifyActor != NULL) && verifyActor(play, actor, data))) {
```
engineer124, line 1532:
```suggestion
    if ((verifyActorFunc == NULL) || ((verifyActorFunc != NULL) && verifyActorFunc(play, actor, data))) {
```
I think `Func` should still be part of the name? Without it, it feels it should be of type `Actor*`

src/code/z_actor.c
```diff
@@ -1873,14 +1873,37 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
...
+ * Passing an exchangeItemAction (see `ItemAction`) allows the player to also use the item to initiate the conversation.
+ *
+ * This function carries talk exchange request to the player actor if context allows it (e.g. the player is in range
+ * and not busy with certain things). The player actor performs the requested action itself.
+ *
+ * The following description of what the `ItemAction` values can do is provided here for completeness, but these
+ * behaviors are entirely out of the scope of this function. All behavior is defined by the player actor.
+ *
+ * - Positive values in range (`PLAYER_IA_NONE < ItemAction < PLAYER_IA_MAX`):
+ *    Offers the ability to initate the conversation with an item from the player.
```
engineer124, line 1887:
```suggestion
 *    Offers the ability to initiate the conversation with an item from the player.
```

src/code/z_sub_s.c
```diff
@@ -1513,15 +1521,18 @@ Actor* SubS_FindActorCustom(PlayState* play, Actor* actor, Actor* actorListStart
     return actorIter;
 }
 
-//! TODO: Needs docs with func_800B8500
-s32 func_8013E748(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, s32 exchangeItemId, void* data,
-                  func_8013E748_VerifyFunc verifyFunc) {
-    s32 ret = false;
+/**
+ * Will extend a TalkExchange offer to the player if the actor is verified with a custom callback.
+ * The callback should return `true` when the actor is succesfully verified.
```
engineer124, line 1526:
```suggestion
 * The callback should return `true` when the actor is successfully verified.
```

src/code/z_sub_s.c
```diff
@@ -1543,14 +1554,14 @@ s32 SubS_ActorAndPlayerFaceEachOther(PlayState* play, Actor* actor, void* data)
     return areFacing;
 }
 
-//! TODO: Needs docs with func_800B8500
-s32 func_8013E8F8(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, s32 exhangeItemId, s16 playerYawTol,
-                  s16 actorYawTol) {
+s32 SubS_OfferTalkExchangeFacing(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, s32 exchangeItemAction,
+                                 s16 playerYawTol, s16 actorYawTol) {
     Vec3s yawTols;
```
engineer124, line 1559: What about `YawRange` instead of `YawTol`?

src/code/z_actor.c
```diff
@@ -1873,14 +1873,37 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
...
-// Actor_OfferTalk / Actor_OfferGetItemExchange? Seems to be called with PLAYER_IA_MINUS1 if the same actor used
-// Actor_OfferGetItem.
-// This function is also used to toggle the "Speak" action on the A button
-s32 func_800B8500(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, PlayerItemAction exchangeItemId) {
+/**
+ * This function covers offering the ability to `Talk` with the player.
+ * Passing an exchangeItemAction (see `ItemAction`) allows the player to also use the item to initiate the conversation.
+ *
+ * This function carries talk exchange request to the player actor if context allows it (e.g. the player is in range
+ * and not busy with certain things). The player actor performs the requested action itself.
```
fig02, line 1881: `This function carries talk exchange requests to the player actor` or `This function carries the talk exchange request to the player actor`, both sound fine

include/z64player.h
```diff
@@ -52,7 +52,7 @@ typedef enum {
 } PlayerEnvHazard;
 
 typedef enum PlayerItemAction {
-    /*   -1 */ PLAYER_IA_MINUS1 = -1,
+    /*   -1 */ PLAYER_IA_CONTINUE = -1, // TODO: determine usages with more player docs, possibly split into seperate values
```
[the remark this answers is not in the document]
engineer124: The name `CONTINUE` definitely wouldn't make sense in OoT since MM decided to merge the itemAction and ExchangeItem systems, and this applies more for the exchangeItemAction usages of `-1`. There are other uses of `-1` like it being connected to where it remembers what the previous item was when you're holding shield (with R) eg. if you have hookshot in hand and you shield and let go, you still have hookshot in hand. So it's a value that has different meanings in different contexts.

We recently had 2 defines for -1, but decided to keep it to 1 until more is understood, see the thread convo here:
https://discord.com/channels/688807550715560050/1124540483591491684/1125074160889114644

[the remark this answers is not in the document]
engineer124: That said, I think I would prefer to keep it generally named instead of naming it after only 1 specific purpose when it has other purposes too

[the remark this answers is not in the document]
fig02: Yeah the usage with shield and returning to previous item is what I know it to do from oot. but yeah if its different here, then fair enough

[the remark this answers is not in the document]
engineer124: The -1 itemAction still have that same shielding/return to previous item functionality in MM it seems, they just made it multipurpose

src/code/z_actor.c
```diff
@@ -1889,24 +1912,24 @@ s32 func_800B8500(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, Player
...
+    player->exchangeItemAction = exchangeItemAction;
 
     CutsceneManager_Queue(CS_ID_GLOBAL_TALK);
     return true;
 }
 
-s32 func_800B85E0(Actor* actor, PlayState* play, f32 radius, PlayerItemAction exchangeItemId) {
-    return func_800B8500(actor, play, radius, radius, exchangeItemId);
+s32 Actor_OfferTalkExchangeRadius(Actor* actor, PlayState* play, f32 radius, PlayerItemAction exchangeItemAction) {
+    return Actor_OfferTalkExchange(actor, play, radius, radius, exchangeItemAction);
```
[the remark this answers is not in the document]
engineer124: Technically, the radius here is the radius defined on an "Equilateral Cylinder" (the blue shape below) since `xz` and `y` are measured separately. I think `radius` is the right term for that (you won't find a better term I don't think), but it might be worth a comment or an `@radius` that this is the radius of an equilateral cylinder?
https://en.wikipedia.org/wiki/Right_circular_cylinder#:~:text=.-,Equilateral%20cylinder,-%5Bedit%5D

image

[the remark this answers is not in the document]
AngheloAlf: `Actor_OfferTalkExchangeRadius` sounds fine to me, we could explain the cylinder stuff in a comment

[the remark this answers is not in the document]
fig02: what makes this radius special that it deserves to be in the name vs `Actor_OfferTalkExchange` which also has a radius? This is my main point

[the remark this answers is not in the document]
fig02: Another way of phrasing, the radius isnt what makes it special, its the fact that the value gets used for both the radius and height of the cylinder (vs `Actor_OfferTalkExchange` which has separately configurable radius and height). Im hoping to find a name that better reflects this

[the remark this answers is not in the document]
hensldm: Tbh, I actually think `Actor_OfferTalkExchangeEquiCylinder` isn't too bad. A bit lengthy, but does seem better than `Radius` to get why this one is different to the original `Actor_OfferTalkExchange`.

include/z64player.h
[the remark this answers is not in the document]
hensldm: > That said, I think I would prefer to keep it generally named instead of naming it after only 1 specific purpose when it has other purposes too

I'll revert it back to `PLAYER_IA_MINUS1` for the time being. I also plan on adding a comment just above the enum (it is rather too big to put it directly next to the value) explain the 3 current known usages of it.

include/z64save.h
hensldm: I could here. However, `EVENTINF_DOG_RACE_STATE_MASK` actually wouldn't work since the enum values only go up to 3, which means the `Mask` would only go up to 3. I'd rather both be just 7 since that is the case.

src/code/z_actor.c
```diff
@@ -1889,24 +1912,24 @@ s32 func_800B8500(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, Player
...
+s32 Actor_OfferTalk(Actor* actor, PlayState* play, f32 radius) {
+    return Actor_OfferTalkExchangeRadius(actor, play, radius, PLAYER_IA_NONE);
 }
 
-s32 func_800B863C(Actor* actor, PlayState* play) {
+s32 Actor_OfferTalkNearby(Actor* actor, PlayState* play) {
     f32 cylRadius = actor->colChkInfo.cylRadius + 50.0f;
 
-    return func_800B8614(actor, play, cylRadius);
+    return Actor_OfferTalk(actor, play, cylRadius);
```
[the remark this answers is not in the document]
fig02: The comment added is good, but I do still kinda think that "nearby" doesnt describe it all to well. Its only nearby if the cylinder radius is small. Are you assuming the colcheckinfo cylinder radius is kept small/as default? cause it is configurable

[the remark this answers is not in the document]
hensldm: At least for me, even if the collider is huge I would still consider the Player nearby the actor if they are just outside the collider reach (I kind of consider the collider as part of the actor, even if the part drawn is much smaller than the collider). I also like `Nearby` as it mirrors the GetItem function in purpose, though granted the implementation details are different.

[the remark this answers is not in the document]
fig02: This isnt the actors collider, to be clear. its an unrelated cylinder that only gets used for a few niche things. mainly talking (as seen here) and some actors use it to avoid projectiles. It can be configured to be much larger than the actual collider cause its seperate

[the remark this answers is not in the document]
hensldm: Ah, that was my disconnect. Hmm, I can possibly think of a better name. Do you have any suggestions though?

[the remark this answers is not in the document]
fig02: I dont, other than just explicitly saying ColChkInfoCyl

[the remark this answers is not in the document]
fig02: If we come up with a better name for this cylinder it can be named after it. But the issue is the cylinder doesnt have an exact purpose. Its just a misc cylinder radius and height that lives in the actor struct and the actor system will sometimes use it if it feels like. Its pretty half baked overall

src/code/z_actor.c
```diff
@@ -1889,24 +1912,31 @@ s32 func_800B8500(Actor* actor, PlayState* play, f32 xzRange, f32 yRange, Player
...
+ * Offers a talk exchange request within an equilateral cylinder with the radius specified.
+ */
+s32 Actor_OfferTalkExchangeEquiCylinder(Actor* actor, PlayState* play, f32 radius,
+                                        PlayerItemAction exchangeItemAction) {
+    return Actor_OfferTalkExchange(actor, play, radius, radius, exchangeItemAction);
 }
 
-s32 func_800B8614(Actor* actor, PlayState* play, f32 radius) {
-    return func_800B85E0(actor, play, radius, PLAYER_IA_NONE);
+s32 Actor_OfferTalk(Actor* actor, PlayState* play, f32 radius) {
```
engineer124, line 1929:
```suggestion
/**
 * Offers a talk request within an equilateral cylinder with the radius specified.
 */
s32 Actor_OfferTalk(Actor* actor, PlayState* play, f32 radius) {
```
Since the radius here also represents the equilateral cylinder. Basically the identical comment above but `talk request` instead of `talk exchange request`

hensldm: Forgot to mention, I also updated the `Actor_OfferTalkNearColChkInfoCylinder` comment to something similar.

src/code/z_actor.c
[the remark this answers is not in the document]
engineer124: Does the full name of the collider `ColChkInfoCyl` need to be in the name or could `Actor_OfferTalkNearColChkInfoCylinder` be simplified to `Actor_OfferTalkNearCol` and just leave it to the summary above the function to specify which of the actor colliders it uses?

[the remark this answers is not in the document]
hensldm: Hmm, not sure we would want `NearCol` cause the first thing I think of is `NearCollider` which from what I understand above isn't really that. Perhaps `Actor_OfferTalkNearCylinder` with the comment explaining the cylinder it is referring to?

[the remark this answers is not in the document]
fig02: I feel like one could easily assume they know it means the collider cylinder and then end up being wrong. Being fully explicit here seems like the best option imo, and maybe one day we'll have a better overall name for this cylinder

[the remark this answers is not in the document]
fig02: this conversation needs to happen outside of this PR, but I could see something like "interact cylinder" being used to describe it (think of its main uses, talking and projectile detection)
but have not put any thought into it and double checked all uses. just giving an example of the direction it could go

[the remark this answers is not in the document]
hensldm: That does seem appropriate, but yeah outside the scope of this PR ( side note `Actor_OfferTalkNearInteractCylinder` does seem nice :) ).

src/code/z_actor.c
```diff
@@ -1878,14 +1878,37 @@ s32 Actor_ProcessTalkRequest(Actor* actor, GameState* gameState) {
...
+ * This function carries a talk exchange request to the player actor if context allows it (e.g. the player is in range
+ * and not busy with certain things). The player actor performs the requested action itself.
+ *
+ * The following description of what the `ItemAction` values can do is provided here for completeness, but these
+ * behaviors are entirely out of the scope of this function. All behavior is defined by the player actor.
+ *
+ * - Positive values in range (`PLAYER_IA_NONE < ItemAction < PLAYER_IA_MAX`):
+ *    Offers the ability to initiate the conversation with an item from the player.
+ *    Not all positive values are implemented properly for this to work.
+ *    Working ones are PLAYER_IA_PICTO_BOX and PLAYER_IA_BOTTLE_MIN <= ItemAction < PLAYER_IA_MASK_MIN
```
engineer124, line 1894:
```suggestion
 * The following description of what the `exchangeItemAction` values can do is provided here for completeness, but these
 * behaviors are entirely out of the scope of this function. All behavior is defined by the player actor.
 *
 * - Positive values in range (`PLAYER_IA_NONE < exchangeItemAction < PLAYER_IA_MAX`):
 *    Offers the ability to initiate the conversation with an item from the player.
 *    Not all positive values are implemented properly for this to work.
 *    Working ones are PLAYER_IA_PICTO_BOX and PLAYER_IA_BOTTLE_MIN <= exchangeItemAction < PLAYER_IA_MASK_MIN
```
Should these be updated? Or does it intentionally use the enum name?

src/code/z_actor.c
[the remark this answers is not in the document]
engineer124: Alright, that's fair. We can revisit this along side this collider docs

src/code/z_actor.c
hensldm: Good catch.

engineer124 (approved):

AngheloAlf (approved):

isghj5 (approved):

PR #1318, opened 2023-07-07
Deku Playground Docs

hensldm:
Wanted to document the `Actor` functions that used `ActorContext_unk_20C`, so just decided to document the full deku playground minigame (EnLiftNuts, EnGamelupy, and the object object_dnt) since they are the only ones that make use of the system.
Basically it is a way for actors to be able to access a shared memory location which they can than use to communicate with each other by just reading/writing to it (other actors usually just use health for example). In this case EnLiftNuts shares a static variable in its overlay so that EnGamelupy can write to it when it has been collected. The system is also general enough that it can even malloc the memory location, and you don't even have to declare a variable in any overlay.


src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -112,109 +134,137 @@ static ColliderCylinderInit sCylinderInit = {
...
-    if ((this->actionFunc != func_80AEB934) && (this->actionFunc != func_80AEB8A4) &&
-        (this->actionFunc != func_80AEACF8) && (this->actionFunc != func_80AEAC64) &&
-        (this->actionFunc != func_80AEA044) && (this->actionFunc != func_80AEB598) &&
-        (this->actionFunc != func_80AEB698) && !func_80AE9B4C(0, 3) && (func_80AE9B8C() == 3) &&
-        (GET_PLAYER_FORM == PLAYER_FORM_DEKU) && (this->actor.xzDistToPlayer < 150.0f)) {
-        func_80AEB828(this);
+/**
+ * Will check if the actor should hide, and if so starts hiding.
+ */
+void EnLiftNuts_CheckHide(EnLiftNuts* this, PlayState* play) {
```
engineer124, line 247: This happens a lot in Player where a set of conditions are checked against, and if certain conditions are met, change the state of the actor to something different, and if they are not met, do nothing.

For these types of functions, I like adapting the word `Try` in the function. So `EnLiftNuts_TryHiding`. Something like `EnLiftNuts_CheckHide` makes me think it should return true if hiding and false if not, and it doesn't indicate that there would be a change in state if true. Does that make sense?

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.h
```diff
@@ -8,25 +8,25 @@ struct EnLiftNuts;
...
+    /* 0x1F0 */ Vec3s jointTable[BUSINESS_SCRUB_LIMB_MAX];
+    /* 0x298 */ Vec3s morphTable[BUSINESS_SCRUB_LIMB_MAX];
     /* 0x340 */ UNK_TYPE1 unk_340[0xC];
     /* 0x34C */ s16 textId;
-    /* 0x34E */ s16 unk_34E;
+    /* 0x34E */ s16 autotalk;
     /* 0x350 */ UNK_TYPE1 unk_350[0x4];
-    /* 0x354 */ s16 unk_354;
-    /* 0x356 */ s16 unk_356;
+    /* 0x354 */ s16 frames; // Frame counter used for various different things
```
engineer124, line 28: Any reason for `frames` over `timer`?

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -904,64 +968,63 @@ void func_80AEB698(EnLiftNuts* this, PlayState* play) {
...
 
-void func_80AEB974(EnLiftNuts* this) {
-    s16 textId = this->textId;
-
-    switch (textId) {
-        default:
-            this->unk_1E4 = 0;
-            break;
-
+void EnLiftNuts_SetEyeTexIndex(EnLiftNuts* this) {
```
engineer124, line 1001: Currently, functions like this are split between `XXX_Blink` and `XXX_UpdateEyes`. These should probably be consolidated into 1, but for now I'd name this function with one of these two options

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -61,32 +61,54 @@ ActorInit En_Lift_Nuts_InitVars = {
...
+    /* 12 */ ENLIFTNUTS_ANIM_SHOCKED_POUND,
+    /* 13 */ ENLIFTNUTS_ANIM_SHOCKED_END,
+    /* 14 */ ENLIFTNUTS_ANIM_THINK,
+    /* 15 */ ENLIFTNUTS_ANIM_BOB,
+    /* 16 */ ENLIFTNUTS_ANIM_JUMP,
+    /* 17 */ ENLIFTNUTS_ANIM_BURROW_HALF,
+    /* 18 */ ENLIFTNUTS_ANIM_BURROW_SHORT
+} EnLiftNutsAnimation;
+
 static AnimationInfo sAnimations[] = {
```
engineer124, line 86:
```suggestion
static AnimationInfo sAnimationInfo[ENLIFTNUTS_ANIM_MAX] = {
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -665,174 +719,179 @@ void func_80AEACF8(EnLiftNuts* this, PlayState* play) {
...
-        temp_fv0 = 40.0f;
+        magnitude = 40.0f;
     } else {
-        temp_fv0 = 80.0f;
+        magnitude = 80.0f;
     }
 
     play->actorCtx.unk268 = 1;
-    func_800B6F20(play, &play->actorCtx.unk_26C, temp_fv0, yaw);
+    func_800B6F20(play, &play->actorCtx.unk_26C, magnitude, playerYaw);
```
engineer124, line 800: `actorCtx.unk268` and `actorCtx.unk_26C` are interesting. Together, they seem to overwrite controller input for player for a frame. `unk268` is a boolean to signal to player to read the input from `unk_26C` once, and `unk268` is turned off every frame after player input is processed.

Could at least make `unk268 = true;`

engineer124 (commented):
Nice work! Some initial thoughts

assets/xml/objects/object_dnt.xml
```diff
@@ -1,91 +1,106 @@
...
+
+        <!-- Animations -->
+        <Animation Name="gBusinessScrubFlyAnim" Offset="0x1BC8" /> <!-- Original name is "dnt_fly" -->
+        <Animation Name="gBusinessScrubShockedShakeHeadAnim" Offset="0x1E2C" /> <!-- Original name is "dnt_iyaiya" -->
+        <Animation Name="gBusinessScrubShockedEndAnim" Offset="0x2268" /> <!-- Original name is "dnt_iyaiyaTOmuun" -->
+        <Animation Name="gBusinessScrubFlyEndAnim" Offset="0x2670" /> <!-- Original name is "dnt_kouka" -->
+        <Animation Name="gBusinessScrubBurrowAnim" Offset="0x29E8" /> <!-- Original name is "dnt_moguru" -->
+        <Animation Name="gBusinessScrubThinkAnim" Offset="0x2F08" /> <!-- Original name is "dnt_muun" -->
+        <Animation Name="gBusinessScrubShockedStartAnim" Offset="0x3438" /> <!-- Original name is "dnt_nooo" -->
+        <Animation Name="gBusinessScrubExcitedAnim" Offset="0x38CC" /> <!-- Original name is "dnt_onedari" -->
```
tom-overton, line 20: You have a `gBusinessScrubExcitedStartAnim` and a `gBusinessScrubExcitedEndAnim`, so shouldn't this be `gBusinessScrubExcitedLoopAnim`? That tends to be how we name things, though perhaps we want to change that.

assets/xml/objects/object_dnt.xml
```diff
@@ -1,91 +1,106 @@
...
+        <!-- Animations -->
+        <Animation Name="gBusinessScrubShockedPoundAnim" Offset="0x994" /> <!-- Original name is "dnt_banban" -->
+        <Animation Name="gBusinessScrubTakeOffHatAnim" Offset="0x12F4" /> <!-- Original name is "dnt_boyon" -->
+
+        <!-- Additional Eyes DLs -->
+        <DList Name="gBusinessScrubEyesWideDL" Offset="0x1350" />
+        <DList Name="gBusinessScrubEyesSquintDL" Offset="0x1420" />
+
+        <!-- Animations -->
+        <Animation Name="gBusinessScrubFlyAnim" Offset="0x1BC8" /> <!-- Original name is "dnt_fly" -->
```
tom-overton, line 13: Same thing with this one regarding the lack of `Loop`.

assets/xml/objects/object_dnt.xml
```diff
@@ -1,91 +1,106 @@
...
+        <Animation Name="gBusinessScrubThinkAnim" Offset="0x2F08" /> <!-- Original name is "dnt_muun" -->
+        <Animation Name="gBusinessScrubShockedStartAnim" Offset="0x3438" /> <!-- Original name is "dnt_nooo" -->
+        <Animation Name="gBusinessScrubExcitedAnim" Offset="0x38CC" /> <!-- Original name is "dnt_onedari" -->
+        <Animation Name="gBusinessScrubExcitedEndAnim" Offset="0x3CC0" /> <!-- Original name is "dnt_onedariTOwait" -->
+        <TextureAnimation Name="gBusinessScrubEmptyTexAnim" Offset="0x3CD0" />
+        <Animation Name="gBusinessScrubFlyStartAnim" Offset="0x4700" /> <!-- Original name is "dnt_spoon" -->
+        <Animation Name="gBusinessScrubRiseUpAnim" Offset="0x4AA0" /> <!-- Original name is "dnt_tobidasu01" -->
+        <Animation Name="gBusinessScrubJumpAnim" Offset="0x4E38" /> <!-- Original name is "dnt_tobidasu02" -->
+        <Animation Name="gBusinessScrubIdleStandAnim" Offset="0x5488" /> <!-- Original name is "dnt_wait" -->
+        <Animation Name="gBusinessScrubBobAnim" Offset="0x577C" /> <!-- Original name is "dnt_wait02" --> <!-- used as the idle anim when half burrowed into a deku flower-->
```
tom-overton, line 27: For these two, it feels like you could drop "Stand" from the first one and just call it `gBusinessScrubIdleAnim`. If the second one was called `gBusinessScrubIdleBobAnim`, I could see why you would want to make the distinction, but at the moment there's only one idle animation, so you don't need to further disambiguate it.

Alternatively, you could just drop the word "idle" entirely and just call it `gBusinessScrubStandAnim`. Or you could put "idle" in both of them and call the second one `gBusinessScrubIdleBobAnim`. Up to you which you prefer; my key point here is that I don't think you need "idle" *and* "stand" in the first one unless "idle" is also in the second one.

assets/xml/objects/object_dnt.xml
```diff
@@ -1,91 +1,106 @@
...
+        <!-- Skeleton -->
+        <Limb Name="gBusinessScrubRootLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_ROOT" Offset="0xAAC0" />
+        <Limb Name="gBusinessScrubBodyLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_BODY" Offset="0xAACC" />
+        <Limb Name="gBusinessScrubRightThighLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_RIGHT_THIGH" Offset="0xAAD8" />
+        <Limb Name="gBusinessScrubRightShinLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_RIGHT_SHIN" Offset="0xAAE4" />
+        <Limb Name="gBusinessScrubRightFootLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_RIGHT_FOOT" Offset="0xAAF0" />
+        <Limb Name="gBusinessScrubLeftThighLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_LEFT_THIGH" Offset="0xAAFC" />
+        <Limb Name="gBusinessScrubLeftShinLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_LEFT_SHIN" Offset="0xAB08" />
+        <Limb Name="gBusinessScrubLeftFootLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_LEFT_FOOT" Offset="0xAB14" />
+        <Limb Name="gBusinessScrubFaceLimb" Type="Standard" EnumName="BUSINESS_SCRUB_LIMB_FACE" Offset="0xAB20" />
```
tom-overton, line 80: I know this guy has some weird anatomy, but the use of "face" here is weird when it wraps around his whole body:
image

Maybe just call this one "head"? Or, if you don't want to use that term, call them "upper body" and "lower body"?

tom-overton (commented):
Looks great, thanks for doing this! Didn't have time tonight to go over it all, so I left some comments on the object.

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
hensldm: Yes that is much better. I did not like `Check` but for the life of me couldn't figure out anything better. `Try` is a good way to describe it though.

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.h
hensldm: Mainly just because it counts up instead of down, which I thought we wanted to distinguish with a different name. I don't really have an opinion, but I'll change it to timer.

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -112,109 +134,137 @@ static ColliderCylinderInit sCylinderInit = {
...
+void EnLiftNuts_CheckHide(EnLiftNuts* this, PlayState* play) {
+    if ((this->actionFunc != EnLiftNuts_Hide) && (this->actionFunc != EnLiftNuts_StartHiding) &&
+        (this->actionFunc != EnLiftNuts_HandleConversation) && (this->actionFunc != EnLiftNuts_StartConversation) &&
+        (this->actionFunc != EnLiftNuts_Burrow) && (this->actionFunc != EnLiftNuts_GiveReward) &&
+        (this->actionFunc != EnLiftNuts_ResumeConversation) &&
+        !EnLiftNuts_MinigameState(ENLIFTNUTS_MINIGAME_STATE_MODE_CHECK, ENLIFTNUTS_MINIGAME_STATE_AFTER) &&
+        (EnLiftNuts_GetNumDaysWon() == 3) && (GET_PLAYER_FORM == PLAYER_FORM_DEKU) &&
+        (this->actor.xzDistToPlayer < 150.0f)) {
+        EnLiftNuts_SetupStartHiding(this);
     }
```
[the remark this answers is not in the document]
hensldm: Decided to break the action function portions into there own, as well as the minigame state function. Let me know what you think. Does seem easier to see what is going on.

assets/xml/objects/object_dnt.xml
hensldm: Removing `Idle` seems fine. Though I think I prefer `Standing` over anim. I know there has been some movement to remove `ing` from animations, but since this isn't an anim of standing up, but the anim done while standing I think it makes sense to keep the `ing`.

AngheloAlf (approved):

assets/xml/objects/object_dnt.xml
tom-overton: Yeah, in this case, I think it's fine. The "movement" you've talked about is mostly me getting mad that sometimes we use `ing` and sometimes we don't; I don't think it's been widely discussed or agreed upon, so you shouldn't feel like you *have* to dogmatically conform to it.

hensldm: Fwiw I agree with said movement to an extent. In particular leave `ing` if it is more about the animation during that state (I have seen a couple like that, like the above mentioned one), and no `ing` if the verb describes the animation itself (for example if this was about the business scrub standing up I would use `StandAnim`)

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.h
engineer124: There was a discussion about this at one point, mainly for using `timer` for counting down and `counter` for counting up, but then there was some weird cases where it could count up and down, so we leaned towards using `timer` for counting both up and down

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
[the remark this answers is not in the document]
engineer124: Over time, I've become more attached to early returns, especially for large if-blocks like this to help readability. I like the way it was changed

src/overlays/actors/ovl_En_Gamelupy/z_en_gamelupy.c
```diff
@@ -91,77 +91,77 @@ void EnGamelupy_Destroy(Actor* thisx, PlayState* play) {
...
+    this->actionFunc = EnGamelupy_Collected;
 }
 
-void func_80AF6A78(EnGamelupy* this, PlayState* play) {
+void EnGamelupy_Collected(EnGamelupy* this, PlayState* play) {
     f32 scale;
     Player* player = GET_PLAYER(play);
 
-    if (this->unk_19C > 30) {
+    if (this->collectedFrames > 30) {
```
engineer124, line 151:
```suggestion
    if (this->collectedTimer > 30) {
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -228,109 +285,111 @@ void EnLiftNuts_Init(Actor* thisx, PlayState* play) {
...
     this->actor.targetMode = 0;
-    this->unk_354 = 0;
-    this->unk_34E = 0;
-    this->unk_356 = 0;
-    this->unk_1E4 = 0;
-    this->unk_1E8 = 0;
+    this->timer = 0;
+    this->autotalk = 0;
+    this->isFirstTimeHiding = 0;
```
engineer124, line 290:
```suggestion
    this->isFirstTimeHiding = false;
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -228,109 +285,111 @@ void EnLiftNuts_Init(Actor* thisx, PlayState* play) {
         }
     }
     this->actor.targetMode = 0;
-    this->unk_354 = 0;
-    this->unk_34E = 0;
-    this->unk_356 = 0;
-    this->unk_1E4 = 0;
-    this->unk_1E8 = 0;
+    this->timer = 0;
+    this->autotalk = 0;
```
engineer124, line 289:
```suggestion
    this->autotalk = false;
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -228,109 +285,111 @@ void EnLiftNuts_Init(Actor* thisx, PlayState* play) {
...
+        EnLiftNuts_SpawnDust(this, play);
     }
 }
 
-void func_80AEA1A0(EnLiftNuts* this, PlayState* play) {
-    if ((func_80AE9B4C(0, 3) || func_80AE9B4C(0, 1)) && (this->unk_34E == 1)) {
+void EnLiftNuts_Idle(EnLiftNuts* this, PlayState* play) {
+    if ((EnLiftNuts_MinigameState(ENLIFTNUTS_MINIGAME_STATE_MODE_CHECK, ENLIFTNUTS_MINIGAME_STATE_AFTER) ||
+         EnLiftNuts_MinigameState(ENLIFTNUTS_MINIGAME_STATE_MODE_CHECK, ENLIFTNUTS_MINIGAME_STATE_STARTING)) &&
+        (this->autotalk == 1)) {
```
engineer124, line 378:
```suggestion
        (this->autotalk == true)) {
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -61,32 +61,55 @@ ActorInit En_Lift_Nuts_InitVars = {
...
-Gfx* D_80AEBF18[] = {
-    object_dnt_DL_008290,
-    object_dnt_DL_001350,
-    object_dnt_DL_001420,
+Gfx* sEyeTextures[] = {
+    gBusinessScrubEyesDL,
+    gBusinessScrubEyesWideDL,
+    gBusinessScrubEyesSquintDL,
     NULL,
 };
```
engineer124, line 114:
```suggestion
Gfx* sEyeTextures[] = {
    gBusinessScrubEyesDL,
    gBusinessScrubEyesWideDL,
    gBusinessScrubEyesSquintDL,
};

static s32 sPad = 0;
```
I think this would be more appropriate?

engineer124 (approved):

src/overlays/actors/ovl_En_Gamelupy/z_en_gamelupy.h
```diff
@@ -5,17 +5,25 @@
...
     /* 0x000 */ Actor actor;
     /* 0x144 */ ColliderCylinder collider;
     /* 0x190 */ EnGamelupyActionFunc actionFunc;
-    /* 0x194 */ s32 rupeeIndex;
-    /* 0x198 */ s16* unk_198;
-    /* 0x19C */ s16 unk_19C;
-    /* 0x19E */ s16 unk_19E;
-    /* 0x1A0 */ s16 unk_1A0;
+    /* 0x194 */ s32 type;
+    /* 0x198 */ s16* minigameScore; // Shared memory location with actor EnLiftNuts
```
tom-overton, line 23: Just a minor nit to make this super clear. Feel free to push back; I just felt it was nice considering this concept literally does not appear in any other actor I think?
```suggestion
    /* 0x198 */ s16* minigameScore; // Pointer to shared memory location with actor EnLiftNuts
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.h
```diff
@@ -8,25 +8,25 @@ struct EnLiftNuts;
...
-    /* 0x1D8 */ Vec3f vec_1D8;
-    /* 0x1E4 */ s32 unk_1E4;
-    /* 0x1E8 */ s32 unk_1E8;
-    /* 0x1EC */ s16* ptr_1EC;
-    /* 0x1F0 */ Vec3s jointTable[OBJECT_DNT_LIMB_MAX];
-    /* 0x298 */ Vec3s morphTable[OBJECT_DNT_LIMB_MAX];
+    /* 0x1D8 */ Vec3f waypointPos;
+    /* 0x1E4 */ s32 eyeTexIndex;
+    /* 0x1E8 */ s32 unk1E8; // Set but never used
+    /* 0x1EC */ s16* minigameScore; // Shared memory location with Actor EnGamelupy
```
tom-overton, line 21: Same thing:
```suggestion
    /* 0x1EC */ s16* minigameScore; // Pointer to shared memory location with Actor EnGamelupy
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -112,109 +132,143 @@ static ColliderCylinderInit sCylinderInit = {
...
+    /* 0 */ ENLIFTNUTS_AUTOTALK_MODE_CHECK_OFF,
+    /* 1 */ ENLIFTNUTS_AUTOTALK_MODE_SET_ON,
+    /* 2 */ ENLIFTNUTS_AUTOTALK_MODE_SET_OFF
+} EnLiftNutsAutoTalkMode;
 
-    switch (arg1) {
-        case 0:
-            if (D_80AEBF6C == 0) {
+s32 EnLiftNuts_AutoTalk(EnLiftNuts* this, EnLiftNutsAutoTalkMode mode) {
+    static s32 sIsAutotalkOn = false;
```
tom-overton, line 162: Since the function is called `AutoTalk`:
```suggestion
    static s32 sIsAutoTalkOn = false;
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -112,109 +132,143 @@ static ColliderCylinderInit sCylinderInit = {
...
             }
             break;
 
-        case 1:
-            if (D_80AEBF6C == 0) {
-                this->unk_34E = 1;
-                D_80AEBF6C = 1;
+        case ENLIFTNUTS_AUTOTALK_MODE_SET_ON:
+            if (!sIsAutotalkOn) {
+                this->autotalk = true;
```
tom-overton, line 173: Maybe do the same as above? I don't know if you think it looks weird, I'm sort of 50/50 on this one. Feel free to say no:
```suggestion
                this->autoTalk = true;
```

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
```diff
@@ -112,109 +132,143 @@ static ColliderCylinderInit sCylinderInit = {
...
+    /* 1 */ ENLIFTNUTS_MINIGAME_STATE_STARTING,
+    /* 2 */ ENLIFTNUTS_MINIGAME_STATE_RUNNING,
+    /* 3 */ ENLIFTNUTS_MINIGAME_STATE_AFTER,
+    /* 4 */ ENLIFTNUTS_MINIGAME_STATE_HIDING
+} EnLiftNutsMiniGameState;
+
+/**
+ * Will either check or set the current minigame state based on the given mode
+ */
+s32 EnLiftNuts_MinigameState(EnLiftNutsMiniGameStateMode mode, EnLiftNutsMiniGameState state) {
```
tom-overton, line 210: This returns a boolean, but the name of the function doesn't immediately indicate what the boolean means in this context. Can you add to the comment what true/false from this function means?

src/code/z_actor.c
```diff
@@ -955,33 +955,37 @@ f32 Actor_GetPlayerImpact(PlayState* play, f32 range, Vec3f* pos, PlayerImpactTy
...
+ * Initializes an element of the `play->actorCtx.actorSharedMemory` array to the `ptr` pointer, or allocates one using
+ * the `size` argument in case `ptr` is NULL. This element is associated to an `id`.
  *
- * In success returns the allocated pointer if `arg2` was NULL or the `arg2` pointer otherwise
- * In failure (There's no space left in `play->actorCtx.unk_20C` or an allocation error happened) returns NULL
+ * This allows allows different actors the ability to access the varible, and thus communicate with each other by
+ * reading/setting the value.
+ *
+ * In success returns the allocated pointer if `ptr` was NULL or the `ptr` pointer otherwise.
+ * In failure (There's no space left in `play->actorCtx.actorSharedMemory` or an allocation error happened) returns
```
tom-overton, line 965: The tiniest of nits:
```suggestion
 * In success: returns the allocated pointer if `ptr` was NULL or the `ptr` pointer otherwise.
 * In failure (There's no space left in `play->actorCtx.actorSharedMemory` or an allocation error happened): returns
```

tom-overton (commented):
Excellent work! Just some very small comments

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
hensldm: How about the other way around, name the function `Autotalk`? (the main reason is I had already kind of made it one word in Baba, so seems better to keep it consistent).

src/overlays/actors/ovl_En_Lift_Nuts/z_en_lift_nuts.c
hensldm: Added, let me know what you think.

tom-overton: Looks good to me, thanks!

tom-overton (approved):

PR #1072, opened 2022-09-10
EnGe3 (Aveil) OK and documented, object_geldb documented

EllipticEllipsis:
Even simpler than `EnGe2`, though a few interesting things going on in here. This was how I figured out that week_event_reg. `object_geldb` is rather different from OoT: the model is the same, a couple of animations come from OoT, most of the rest is gone (EnKaizoku uses a completely different object). Retains the classic colour fix hack from OoT np1 and later, they really should have just put the colours in the displaylists this time. 

En_Ge3 gv

PR #1323, opened 2023-07-11
Add original asset names to all `object_g*` files

tom-overton:
I was waiting for #1072 to be merged before starting this, but at this point it has a lot of conflicts, so what's one more?

This adds the original names from MM3D for objects that start with `object_g*`. Do note that I skipped all the GI objects, since I firmly believe that those are original Grezzo names, so I don't think they're very useful. I also skipped `object_gm` (because it's empty) and `object_gs` (because, like the GI objects, it seems to have been completely redone by Grezzo and named with their own naming scheme). Thus, this is a pretty small list of changes.

### DIFF
--- a/assets/xml/objects/object_ghaka.xml
+++ b/assets/xml/objects/object_ghaka.xml
@@ -1,12 +1,12 @@
 ﻿<Root>
     <File Name="object_ghaka" Segment="6">
-        <Collision Name="object_ghaka_Colheader_0000E8" Offset="0xE8" />
+        <Collision Name="object_ghaka_Colheader_0000E8" Offset="0xE8" /> <!-- Original name is "goron_hakaisi_bgdatainfo" -->
         <DList Name="object_ghaka_DL_001980" Offset="0x1980" />
-        <DList Name="object_ghaka_DL_001A20" Offset="0x1A20" />
+        <DList Name="object_ghaka_DL_001A20" Offset="0x1A20" /> <!-- Original name is "z2_goron_hakaisi_model" -->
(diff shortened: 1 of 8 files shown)


hensldm (approved):

assets/xml/objects/object_ghaka.xml
```diff
@@ -1,12 +1,12 @@
 ﻿<Root>
     <File Name="object_ghaka" Segment="6">
-        <Collision Name="object_ghaka_Colheader_0000E8" Offset="0xE8" />
+        <Collision Name="object_ghaka_Colheader_0000E8" Offset="0xE8" /> <!-- Original name is "goron_hakaisi_bgdatainfo" -->
```
AngheloAlf, line 3: No `z2_` prefix?

tom-overton: Nope, not in this case:
image

I compared the .zsis from MM3D to the extracted collision data from MM64 to make sure I named them right. The `Vec3s` arrays from MM64 can actually be found verbatim in MM3D (assuming the collision vertices didn't change, which they rarely do), so it's pretty easy to line them up.

tom-overton: I should also note the one without the `z2_` prefix is unused; only the latter one is used in ObjGhaka.

tom-overton: One last thing: since we have a collision viewer in Z64Utils (I forgot about this, despite literally being the person to finish Tharo's implementation). I took a look at what these two pieces of collision look like:
z2_goron_hakaisi_bgdatainfo (used):
image

goron_hakaisi_bgdatainfo (unused):
image

So the unused one is smaller and less detailed.

AngheloAlf (approved):

engineer124 (approved):

PR #1337, opened 2023-07-25
Added some missing includes in ZAPD

Vertmo:
I have added some includes of the `stdint` header in three files from ZAPD. Without these includes, ZAPD did not compile on my machine, showing the error message "unknown type name`uint32_t`".
I hope this can be useful to other contributors :)

### DIFF
--- a/tools/ZAPD/ZAPD/CrashHandler.cpp
+++ b/tools/ZAPD/ZAPD/CrashHandler.cpp
@@ -7,6 +7,7 @@
 #define HAS_POSIX 0
 #endif
 
+#include <stdint.h>
 #include <array>
 #include <cstdio>
 #include <cstdlib>
--- a/tools/ZAPD/ZAPD/Declaration.h
+++ b/tools/ZAPD/ZAPD/Declaration.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <stdint.h>
 #include <string>
 #include <vector>
 
--- a/tools/ZAPD/ZAPD/OutputFormatter.h
+++ b/tools/ZAPD/ZAPD/OutputFormatter.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <stdint.h>
 #include <map>
 #include <string>
 #include <vector>


tom-overton:
Hi! Thanks for the contribution. ZAPD is a subrepo that we pull from: https://github.com/NEstelami/ZAPD
While I've never made changes to ZAPD before, I believe the standard procedure is to PR them to that repo, then update the subrepo on OoT and MM's side of things.

Vertmo:
> Hi! Thanks for the contribution. ZAPD is a subrepo that we pull from: https://github.com/NEstelami/ZAPD While I've never made changes to ZAPD before, I believe the standard procedure is to PR them to that repo, then update the subrepo on OoT and MM's side of things.

Oh OK ! Sorry that was not completely clear to me. It looks like the issue has already been fixed in ZAPD.
In that case, may I ask why the project does not use the `submodule` feature of Git for external repositories ? I think that it may be more explicit and clear than the `subrepo` command that you use (which I admit I dont know, so I may not see the whole picture)

tom-overton:
> In that case, may I ask why the project does not use the `submodule` feature of Git for external repositories ? I think that it may be more explicit and clear than the `subrepo` command that you use (which I admit I dont know, so I may not see the whole picture)

ZAPD *was* a submodule in the past, and it was an absolutely miserable user experience. It puts the onus on users to make sure their local state is correct (having to run the proper commands, debugging issues with the submodule not updating when changing branches, etc.). We switched to subrepos because it puts the burden of updating these dependencies on the repo maintainers, not the consumers of the repo.

In this case, though, it's a bit of a downside, because we haven't updated the subrepo in a while; no one expressed a need for it. Sounds like we need to do that, though.

Vertmo:
Ok, that makes sense, I usually use submodules on small projects (3 to 5 people who all know each other), so I never had any usability issue, but I understand how this could become difficult on a larger scale.
Thank you for your answer ! Yes if you could update ZAPD it would be perfect. I will close this PR.

PR #1342, opened 2023-07-28
Fix first_diff and sym_info for older and new ld

hensldm:
Changes `first_diff.py` and `sym_info.py` to use mapfile_parser which works with older and newer ld. Changed `assist.py` to make it easier to work with new ld (change gAddressWidth to 10), but didn't feel it worthwhile at the moment to change it work with mapfile_parser.


AngheloAlf (approved):

first_diff.py
```diff
@@ -1,255 +1,50 @@
...
-        diff_args = "-" + diff_args
-    check_call(
-        [
-            "python3",
-            "diff.py",
-            diff_args,
-            diff_sym,
-        ]
-    )
+import mapfile_parser
```
Dragorn421, line 6: one request:
```suggestion
try:
    import mapfile_parser
except ImportError:
    print("Missing dependency mapfile_parser, install it with `python3 -m pip install mapfile-parser`")
    exit(1)
```
(written in github, didn't test, there may be an obvious mistake)

engineer124 (approved):

jpburnett (approved):

PR #1364, opened 2023-08-30
document object_kanban (cuttable sign)

MegaIDK:
funy multi part objects


AngheloAlf (approved):

assets/xml/objects/object_kanban.xml
```diff
@@ -1,18 +1,19 @@
...
-        <DList Name="object_kanban_DL_000FF8" Offset="0xFF8" />
-        <DList Name="object_kanban_DL_0010B8" Offset="0x10B8" />
-        <DList Name="object_kanban_DL_0011C0" Offset="0x11C0" />
-        <DList Name="object_kanban_DL_0012C8" Offset="0x12C8" />
-        <DList Name="object_kanban_DL_0013D0" Offset="0x13D0" />
-        <DList Name="object_kanban_DL_001488" Offset="0x1488" />
-        <DList Name="object_kanban_DL_001540" Offset="0x1540" />
-        <DList Name="object_kanban_DL_001630" Offset="0x1630" />
-        <Texture Name="object_kanban_Tex_0016B0" OutName="tex_0016B0" Format="i8" Width="16" Height="16" Offset="0x16B0" />
+        <DList Name="gSignSetupDL" Offset="0xC30" />
```
hensldm, line 4:
```suggestion
        <DList Name="gSignMaterialDL" Offset="0xC30" />
```
Generally the two dl sets are named `Material` (which sets everything up but does not have any vertex information) and `Model` (which does hold the vertex information). I'd probably say add `Model` to the other dls (excpet `SignParticle`) as well.
Something like `gSignUpperLeftModelDL`.

MegaIDK: oh yea i was wondering whether that one was a material or something but usually (as in the ones ive seen lol) its just one material and on model so i got a bit confused but yea on it

hensldm (approved):

tom-overton (commented):
If I'm understanding the naming right, the sign looks like this, correct?
image

I found the names pretty confusing, so I had to make a diagram to help me understand. I think what's tripping me up is that the sign is divided into 8 segments, but there are only 4 words used to describe each piece, and instead the word *order* is being used to distinguish things; "left lower" and "lower left" use the same words, but the order is what's supposed to tell you which is which.

What do you think about something like this?
image
Feel free to customize it as you like as well. The point I'm trying to make with this diagram is that labeling them in a manner similar to this would make things a little simpler to understand, and would make it so word order isn't a distinguishing factor. For example, if you rewrote "upper left middle" to "left upper middle", it would still be distinct from any other piece.

MegaIDK:
aye it is quite confusing for sure, took a while for me to understand too. i took the names from the enum in the actor's header and the parts flags array so it wouldnt confuse people when they had to compare them, so maybe if we do change the DLs' names wed have to change all that other stuff as well lol. also yeah i think the last example is best to avoid confusion since the sign is kinda divided in a way that makes you think about the corners instead of sides, might go with that

tom-overton:
> i took the names from the enum in the actor's header and the parts flags array so it wouldnt confuse people when they had to compare them, so maybe if we do change the DLs' names wed have to change all that other stuff as well lol

Ahhhh, I missed this. Yeah, in that case, if you update the DL names, it would be good to update the defines in the header and actor as well. I do think it's worth it, though.

MegaIDK:
another suggestion could be to specify that the first word refers to the side of the idk rectangle
264594649-767c4ccf-a0d1-43a6-ac18-907acb1cdb03
to maybe minimize how much has to be changed and keep names a bit shorter idk could also still be somewhat confusing

tom-overton:
I don't think minimizing how much needs to be changed is really important; if the names are better, then that's good enough; it's okay if the names have to change pretty dramatically to accomplish that. However, you bring up a good point regarding the length of the names; I hadn't considered that.

tom-overton (approved):
I'm still not 100% in love with these names, but they're way better than what came before, and I don't see a reason to hold up this PR any longer. Thanks for changing them!

MegaIDK:
yeah they do be kinda funky, if yall prefer one of the others its fine i can change 👍

hensldm:
Just want to note the conflicts due to #1388 getting merged.

assets/xml/objects/object_kanban.xml
```diff
@@ -1,18 +1,19 @@
 ﻿<Root>
+    <!-- Assets for cuttable signs-->
```
engineer124, line 2:
```suggestion
    <!-- Assets for cuttable signs -->
```
Just the tiniest nitpick

engineer124 (approved):
Apologies for the delay on this. I've really been thinking about a better way to name a rectangle split into 8 pieces like this, but I haven't come up with anything satisfactory. So it's fine as is. Just the tiniest nitpick

PR #1418, opened 2023-10-06
object_f40_switch

hack-nuss:
Object for Stone Tower Switch


engineer124 (approved):

AngheloAlf (approved):

isghj5:
Oh it's the foot switch, calling it the "Stone Tower Switch" made me think it was the textures for the light switch that flips the temple. Can we change to "Foot switch" or "Weight switch" or something that specifies? I'm not actually sure what the current foot switch terminology we use is.

tom-overton:
The terminology we use presently is "floor switch". You can see that in `gameplay_dangeon_keep`.

isghj5 (approved):

PR #1423, opened 2023-10-06
obj_b_heart

Wimmerrad:
Unused object for an Heart Piece


hack-nuss:
i think you accidentally also commit Obj_visiblock because you have an PR for it

PR #1438, opened 2023-10-11
`SCHEDULE_TIME`

AngheloAlf:
Introduce a `SCHEDULE_CALC_TIME_ALT` macro to compare a time value that uses the Schedule format, for example comparing to `SCHEDULE_TIME_NOW`

Currently the only known cases that use this macro are in DmTag, Kankyo and EnAn, of which only the first one is in main. I decided to bring over the kankyo function too since it is small.

New script `schedule_timeconv.py`, which converts a raw value to this macro, similar to `timeconv.py`.

I also made a few semi related cleanups


engineer124 (approved):

include/z64schedule.h
```diff
@@ -28,10 +28,15 @@
  */
 
 // Macro to convert the time format used in the save struct into the format used in Schedule
-#define SCHEDULE_CONVERT_TIME(time) ((u16)((time) - 0x10000 / 360 * 90))
+#define SCHEDULE_CONVERT_TIME_ALT(time) ((time) - 0x10000 / 360 * 90)
+#define SCHEDULE_CONVERT_TIME(time) ((u16)(s32)SCHEDULE_CONVERT_TIME_ALT(time))
+
+#define SCHEDULE_CALC_TIME_ALT(hour, minute) SCHEDULE_CONVERT_TIME((((hour)*60.0f) + (minute)) * (0x10000 / 60 / 24.0f))
```
hensldm, line 34: Any reason not to just use `SCHEDULE_TIME` to match `CLOCK_TIME`?

include/z64schedule.h
```diff
@@ -28,10 +28,15 @@
  */
 
 // Macro to convert the time format used in the save struct into the format used in Schedule
-#define SCHEDULE_CONVERT_TIME(time) ((u16)((time) - 0x10000 / 360 * 90))
+#define SCHEDULE_CONVERT_TIME_ALT(time) ((time) - 0x10000 / 360 * 90)
+#define SCHEDULE_CONVERT_TIME(time) ((u16)(s32)SCHEDULE_CONVERT_TIME_ALT(time))
```
hensldm, line 32: I am not 100% sure on having the duplicate macro. We could just have the non cast version, and cast as needed. (I think that is how it was originally, but we never found a no cast instance until now).

src/overlays/actors/ovl_Dm_Tag/z_dm_tag.c
```diff
@@ -187,10 +187,12 @@ s32* func_80C22880(DmTag* this, PlayState* play) {
 
     switch (this->unk_18E) {
         case 1:
-            time = gSaveContext.save.time - 0x3FFC;
-            if ((time >= 0xA54B) && (time < 0xB54A) && (gSaveContext.save.day == 2)) {
-                this->msgEventCallback = func_80C227E8;
-                return D_80C22BF0;
+            time = SCHEDULE_TIME_NOW_ALT;
```
hensldm, line 190: This needs the no cast version?

hensldm (commented):
Some thoughts

include/z64schedule.h
AngheloAlf: Because I didn't thought on that. It's a great idea

src/overlays/actors/ovl_Dm_Tag/z_dm_tag.c
AngheloAlf: The other macro doesn't match sadly

include/z64schedule.h
AngheloAlf: I'll give this a try and see how it ends up looking

AngheloAlf: It was surprisingly less bad than I expected
https://github.com/zeldaret/mm/pull/1438/commits/1f3269580702cc434ca2e8eb069e256a02780804

hensldm (approved):

hensldm:
Not a big deal, but may want to change the title of the PR now.

tom-overton (approved):

PR #1449, opened 2023-10-21
Progress LD fix

hensldm:
fixes the progress script to work with both newer (>= 2.40) and older LD versions.


AngheloAlf (approved):
Still works on my machine (ld 2.38)

engineer124 (approved):
I'm also on ld 2.38 and it works just fine

tom-overton (approved):
Works on my machine (2.38) as well.

PR #1454, opened 2023-10-26
Fix ultratypes types

hensldm:
Following https://github.com/zeldaret/oot/pull/1546 from OoT. Note the same situation seems to occur with only boolean functions needing to be typed as `int` to match. I would argue for a bool typedef at this point for this situation.


include/PR/ultratypes.h
```diff
@@ -22,12 +22,6 @@ typedef volatile s64 vs64;
 typedef float  f32;
 typedef double f64;
 
-typedef u32 size_t;
-
-#ifndef NULL
-#define NULL (void*)0
-#endif
-
```
AngheloAlf, line 30: libultra defines size_t and NULL on ultratypes.h. Why did you remove them from here?

src/boot/O2/system_malloc.c
```diff
@@ -19,7 +19,7 @@ void SystemArena_Free(void* ptr) {
     __osFree(&gSystemArena, ptr);
 }
 
-void* SystemArena_Calloc(u32 elements, size_t size) {
+void* SystemArena_Calloc(u32 elements, u32 size) {
```
AngheloAlf, line 22: A shame

AngheloAlf (commented):
Personally I think a typedef for bool would be great. I wouldn't be opposed to include it on this PR

include/PR/ultratypes.h
hensldm: Oh, I was following what OoT has. Didn't think they were actually from `ultratypes`, will add them back.

AngheloAlf: It does some ifdef shenanigans to avoid redefining `size_t`
https://github.com/decompals/ultralib/blob/eddcfcd44e6bf181996c577dc2384e823e999fed/include/PR/ultratypes.h#L60-L69

Tharo did a similar ifdef check on his libc PR, but only for `stddef.h`
https://github.com/zeldaret/oot/pull/1568/files#diff-4e201337732f1fcf8edb60e2ec377650650121fb6078dba6cbc647134b487a4a

include/libc/stddef.h
```diff
@@ -3,18 +3,8 @@
...
-#endif
-
 typedef s32 ptrdiff_t;
 
-#if defined(_MIPS_SZLONG) && (_MIPS_SZLONG == 64)
-typedef unsigned long size_t;
-#else
-typedef unsigned int  size_t;
-#endif
-
```
AngheloAlf: I wouldn't be opposed to have this here too if we wanted to go with the same ifdef checks libultra does

AngheloAlf (approved):

src/boot/O2/system_malloc.c
Dragorn421: What am I missing here, `size_t` used to be `u32` aka `unsigned int`, and with the PR it becomes `unsigned int` so what changed?

AngheloAlf: Actually you are right, this should expand to `unsigned int` like it used to :thinking:

hensldm: Its the fact that `u32` changed to an `unsigned long`, so they now had different types. 
Fwiw it actually matches if you use `size_t` for the first arg and `u32` for the second, and since both instances go unused I just decided to swap the args so they still work.

tom-overton (approved):
Echoing that it would be nice to have a bool define at some point.

src/boot/O2/system_malloc.c
Dragorn421: Second argument should match as `size_t` afaict
Also both arguments should be `size_t`, and it would be better to keep the standard nelem,elemsize order
Cf e.g. man calloc https://www.man7.org/linux/man-pages/man3/calloc.3p.html

hensldm: Yeah both work as `size_t`.

hensldm:
> Echoing that it would be nice to have a bool define at some point

I'll work on another PR after this one to do that (probably going to be rather large)

hensldm:
Note I noticed we actually already had a bool (though it was a define and a `u32`). So I just converted it to an int typedef and used it where only where necessary (basically everywhere I needed to change to an `s32` to an `int`).

tom-overton (approved):
Thanks for doing the `bool` thing!

Thar0:
If you want to be compliant with stdbool.h `bool` should be a macro rather than a typedef
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdbool.h.html

I suggest this based on my approach in oot's libc cleanup
```c
#if (__STDC_VERSION__ >= 199901L)
#define bool    _Bool
#else
#define bool    int
#endif
```

AngheloAlf:
If we are going to use `__STDC_VERSION__` to check the standard in use then we should also check for C23 which adds `bool` as a proper type, so a header shouldn't define it

`bool` being a type in C23: https://en.cppreference.com/w/c/language/arithmetic_types
version check for C23: https://en.cppreference.com/w/c/preprocessor/replace

```c
#if (__STDC_VERSION__ < 202311L)
// bool is a type in C23, do not define it
#ifelif (__STDC_VERSION__ >= 199901L)
#define bool    _Bool
#else
// technically this could be a typedef, but for consistency a define is fine
#define bool    int
#endif
```

engineer124 (approved):
I'm still not a fan of having the `bool` variable introduced here, but it seems most MM contributors would like to see it, so I'll approve based on that👍

I'll still wait a day to merge in case anyone else wants to add anything to the conversation

PR #1456, opened 2023-10-29
Document `object_boss05` (Bio Deku Baba)

tom-overton:
Similar to what I did with the Garo object, I split this one out to make the eventual actor PR a little smaller. I do think this object may have a bit of controversy around the term "head", though I hope I explained my position well enough in the comments.

### DIFF
--- a/assets/xml/objects/object_boss05.xml
+++ b/assets/xml/objects/object_boss05.xml
@@ -1,81 +1,110 @@
 ﻿<Root>
+    <!-- Assets for Bio Deku Baba. The "Head" in this file refers to both the chomping head attached to the lily pad as well as -->
+    <!-- the entity with legs and eyes that spawns from the head after it is detached. The same skeleton and limbs are used for -->
+    <!-- both of these; when the head is attached to the lily pad, the Boss_05 actor is responsible for hiding most of the limbs. -->
     <File Name="object_boss05" Segment="6">
-        <Animation Name="object_boss05_Anim_0006A4" Offset="0x6A4" /> <!-- Original name is "wdb_atackH" -->
-        <Animation Name="object_boss05_Anim_000A5C" Offset="0xA5C" /> <!-- Original name is "wdb_atackwalkH" -->
-        <Animation Name="object_boss05_Anim_000ABC" Offset="0xABC" />
-        <DList Name="object_boss05_DL_0011C0" Offset="0x11C0" />
-        <DList Name="object_boss05_DL_001288" Offset="0x1288" />
-        <DList Name="object_boss05_DL_001320" Offset="0x1320" />
-        <DList Name="object_boss05_DL_0013D0" Offset="0x13D0" />
-        <DList Name="object_boss05_DL_0014A0" Offset="0x14A0" />
-        <DList Name="object_boss05_DL_001538" Offset="0x1538" />
-        <DList Name="object_boss05_DL_0015F0" Offset="0x15F0" />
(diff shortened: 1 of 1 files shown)


assets/xml/objects/object_boss05.xml
```diff
@@ -1,81 +1,110 @@
...
+        <Limb Name="gBioDekuBabaHeadRightLowerLegLimb" Type="Standard" EnumName="BIO_DEKU_BABA_HEAD_LIMB_RIGHT_LOWER_LEG" Offset="0x6308" />
+        <Limb Name="gBioDekuBabaHeadRightFootLimb" Type="Standard" EnumName="BIO_DEKU_BABA_HEAD_LIMB_RIGHT_FOOT" Offset="0x6314" />
+        <Limb Name="gBioDekuBabaHeadLeavesLimb" Type="Standard" EnumName="BIO_DEKU_BABA_HEAD_LIMB_LEAVES" Offset="0x6320" />
+
+        <!-- Bio Deku Baba Head Skeleton -->
+        <Skeleton Name="gBioDekuBabaHeadSkel" Type="Flex" LimbType="Standard" LimbNone="BIO_DEKU_BABA_HEAD_LIMB_NONE" LimbMax="BIO_DEKU_BABA_HEAD_LIMB_MAX" EnumName="BioDekuBabaHeadLimb" Offset="0x6378" />
+
+        <!-- Bio Deku Baba Head Animations -->
+        <Animation Name="gBioDekuBabaHeadChompAnim" Offset="0x6484" /> <!-- Original name is "wdb_pakupaku" ("repeatedly opening and closing (one's mouth)") -->
+        <Animation Name="gBioDekuBabaHeadTransformAnim" Offset="0x6E50" /> <!-- Original name is "wdb_startH" -->
```
AngheloAlf, line 106: Transform? BioDekuBaba transforms?

AngheloAlf (approved):

assets/xml/objects/object_boss05.xml
tom-overton: This is the animation:
transform

When you detach a Bio Baba from the lily pad, the head still has bits of the lily pad attached to it as it falls. When it hits the ground, the lily pad parts disappear and it plays this animation, For reasons I still haven't quite figured out, though, this animation doesn't really look quite right in-game, so the Bio Baba looks more like the limbs are "popping in" rather than unfolding like this animation.

Definitely a weird animation to describe, so if you have a better term for it, I'm all ears.

AngheloAlf: "Detach" maybe?
But yes, indeed it is hard to name

hensldm (approved):

assets/xml/objects/object_boss05.xml
tom-overton: I don't think "detach" is quite right. This video shows what happens when a Bio Baba gets detached. It spawns a Bio Baba with params of 3; once it hits the floor, the lily pad limbs shrink. Once they reach 0.0f scale, it spawns a new Bio Baba with params of 4, which grows the head limbs and plays this animation:

https://github.com/zeldaret/mm/assets/12245827/d8af50af-f357-4698-910f-59768957b53b

You can also see in this video the thing I talked about before; you don't really see much of this animation in-game besides the mouth opening. But in conclusion, this isn't the animation that plays when it gets detached. This is the animation that plays when the detached head *touches the ground*.

engineer124 (approved):

PR #1464, opened 2023-11-05
add D_801D7E04 and D_801D7EA4 as labels to fix 2 missed pointers

RevoSucks:
Added these 2 variable definitions so it will catch their missing pointers in code_8019AF00.data.s.


hensldm:
Since all the data from code_8019AF00.data.s has been imported, code_8019AF00.data.s shouldn't be used anymore. Were you having issues with these pointers missing from the data asm file?

RevoSucks:
I was trying to diagnose a crash I was getting where Darmani's Ghost going back to his cave (after getting the lens of truth) keeps causing an RCP crash from shifting certain symbols in z_actor. At first I saw those pointers and thought it was because of missed pointers but that didnt fix it.

Honestly I got no clue why you can get an RCP crash from shifting symbols, especially because I dont see any remaining pointers in this repo.

RevoSucks:
Closing, as the data file seemingly isnt needed anymore.

PR #1465, opened 2023-11-05
Match func_809F24C8 NON_MATCHING

sonicdcer:
No reviews on this one please, i'm not familiar with the function at all, i just fixed it by changing two types

### DIFF
--- a/src/overlays/actors/ovl_Boss_06/z_boss_06.c
+++ b/src/overlays/actors/ovl_Boss_06/z_boss_06.c
@@ -182,8 +182,6 @@ void func_809F24A8(Boss06* this) {
     this->unk_1AC = 0.0f;
 }
 
-#ifdef NON_MATCHING
-// The 1 constant from the switch branch is reused in case 2 for some reason.
 void func_809F24C8(Boss06* this, PlayState* play) {
     s16 sp4E = 0;
     Player* player = GET_PLAYER(play);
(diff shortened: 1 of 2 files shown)


AngheloAlf:
What do you mean by not being familiar with the function? It didn't match itself, didn't it? So you can't explain how you figured it this match then?

Also, asking for not reviews isn't reasonable. Why wouldn't you expect feedback when PRing a patch to a project?

sonicdcer:
> What do you mean by not being familiar with the function? It didn't match itself, didn't it? So you can't explain how you figured it this match then?
> 
> Also, asking for not reviews isn't reasonable. Why wouldn't you expect feedback when PRing a patch to a project?

What i meant is that i'm not the one who decompiled it in the first place so i don't want to be responsible for documenting it.

sonicdcer:
Regarding the matching, changing types is in my usual checklist to getting a match.

RevoSucks:
I agree that trying to avoid a review (unless it's like for an extremely urgent time sensitive hotfix or something, definitely doesn't apply here) is unreasonable. In any case, it's preferable to explain how this was a nonmatching in the first place, and what the fix was, which seems to have been explained now.

sonicdcer:
Maybe i choose poor words, i didn't mean i didn't want a review, i was referring to the documentation part.

AngheloAlf:
> What i meant is that i'm not the one who decompiled it in the first place so i don't want to be responsible for documenting it.

We usually don't expect for a PR like this to document the function, just to clean it up, and there isn't much to clean up in a PR like this.

I'm mainly complaining about how you were literally asking for no reviews at all, which is not how the PR process on this repo works. If you don't want reviews then you don't get a merge either

sonicdcer:
Sorry for the misunderstanding, i corrected the description to the appropriate message.

hensldm (commented):
Out of curiosity have you taken a look at `Environment_Init`? It has the exact some problem as here.

sonicdcer:
> Out of curiosity have you taken a look at `Environment_Init`? It has the exact some problem as here.

Just did, unfortunately the same fix doesn't apply here, that switch is a bit suspicious tho, because later the same var is being evaluated with an if statement.

engineer124 (approved):

hensldm (approved):

isghj5 (approved):

AngheloAlf (approved):

PR #1488, opened 2023-11-11
Move the system heap and the framebuffers to their own segments

AngheloAlf:
This PR is a bit too technical

The game expects the framebuffers (`gPictoPhotoI8`, `D_80784600` and `gFramebuffer0`) to be placed at a fixed location in VRAM. All the framebuffers should be placed at the end of the VRAM, 0x80800000 and backwards (0x80800000 is the end of the Expansion Pak address space).

On a modded rom, those addresses may shift due to the new code, making the framebuffers shorter, producing weird and unexpected behavior.

To keep those framebuffers on a fixed place in VRAM, I moved them to their own segment. Because the game wants a hardcoded address on `Main` instead of a symbol to be able to match, I added a define (`SYSTEM_HEAP_END_ADDR`) to be able to control this value. I defined it on the makefile because that the same define can be used for both C files and the spec file. I'm not completely satisfied with this define on the makefile and I can remove it if it is preferred.

Since I was already working on this I decided to work on the pre-boot buffers too. If we don't like how I made those, I can revert them


spec
```diff
@@ -621,7 +635,22 @@ beginseg
...
     include "build/src/buffers/gfxstack.o"
     include "build/src/buffers/gfxpools.o"
-    include "build/data/code/buffers.bss.o"
+    include "build/src/buffers/audio_heap.o"
+endseg
+
+beginseg
+    name "system_heap"
+    flags NOLOAD
+    // This segment is a just a dummy that is used to know where the other buffers (non framebuffers) end
```
jpburnett, line 644:
```suggestion
    // This segment is just a dummy that is used to know where the other buffers (non framebuffers) end
```
typo

spec
```diff
@@ -4,12 +4,26 @@
...
+    name "framebuffer_1"
+    address 0x80000500
+    flags NOLOAD
+    include "build/src/buffers/framebuffer_1.o"
+endseg
+
+beginseg
+    name "pre_boot_buffer"
+    flags NOLOAD
+    // This segment is a just a dummy so we can symbolize this address range.
```
jpburnett, line 23:
```suggestion
    // This segment is just a dummy so we can symbolize this address range.
```

jpburnett: Would "this segment is a placeholder so we can symbolize the address range" be any better/make sense?

jpburnett (approved):

Makefile
```diff
@@ -132,6 +132,22 @@ MIPS_VERSION := -mips2
 # we support Microsoft extensions such as anonymous structs, which the compiler does support but warns for their usage. Suppress the warnings with -woff.
 CFLAGS += -G 0 -non_shared -fullwarn -verbose -Xcpluscomm $(IINC) -nostdinc -Wab,-r4300_mul -woff 649,838,712,516
 
+# The system heap is an area of memory that covers all the VRAM that is not
+# used by compiled code/data.
```
Dragorn421, line 136:
```suggestion
# The system heap is an area of memory that covers all the RAM that is not
# taken by compiled code/data.
```
vram -> ram: because either "vram" as in the linker view of memory includes >8MiB, yet the heap is a very concrete area so may as well not confuse it with it,  or "vram" as in virtual refering to the cpu address space being irrelevant in the context

used -> taken: since the heap is "used" by code too, I think you meant the bytes that the various code and data sections are made of, saying "taken" for that seems less confusing

Makefile
```diff
@@ -132,6 +132,22 @@ MIPS_VERSION := -mips2
...
+# It starts just after the end of compiled code (that is not an overlay) and
```
Dragorn421, line 137: "compiled code" is not wrong but it may be misleading to someone as refering to text only

If it's correct maybe
```suggestion
# It starts just after the end of the code segment
```
Would be clearer?

Makefile
```diff
@@ -132,6 +132,22 @@ MIPS_VERSION := -mips2
...
+# ends at a fixed location in VRAM.
+# That fixed location in VRAM is used by the framebuffers of the game and have
+# a fixed size.
+# This address is calculated by doing `0x80800000 - (size of framebuffers)`,
+# where 0x80800000 is the end of the Expansion Pak address range
+# This is a define in the Makefile because references in code require a raw
+# value to match (opposed to a linker symbol). This way, the define can also be
+# used in the spec file.
+SYSTEM_HEAP_END_ADDR ?= 0x80780000
```
Dragorn421, line 146: A few comments in general

1) the comments are nice but don't really fit in the Makefile imo. Though since we still didn't solve the discussion on where to put extended documentation I guess it can do.

2) I'm not sure it's worth abstracting so much of the memory layout instead of just having a copypasted value in two spots, if MM is anything like OoT (and I hear it's one of its best romhacks) the memory layout/usage/numbers details are sprinkled all over and you can't easily change one thing anyway unless you do know what you're doing

src/buffers/pre_boot_buffer.c
```diff
@@ -0,0 +1,4 @@
+#include "buffers.h"
+
+// Dummy, the size of this symbol will adjust itself based on the spec
```
Dragorn421, line 3: Not a fan of the comment, the symbol itself will stay at a size of 1 element, maybe you could instead say something like "marks the start of xxx space whose actual size depends on the spec"

src/boot/idle.c
```diff
@@ -52,9 +53,9 @@ void Main_InitMemory(void) {
     void* memStart = (void*)0x80000400;
     void* memEnd = OS_PHYSICAL_TO_K0(osMemSize);
 
-    Main_ClearMemory(memStart, gFramebuffer1);
-    Main_ClearMemory(D_80025D00, bootproc);
-    Main_ClearMemory(gGfxSPTaskYieldBuffer, memEnd);
+    Main_ClearMemory(memStart, SEGMENT_START(framebuffer_1));
+    Main_ClearMemory(SEGMENT_START(pre_boot_buffer), SEGMENT_START(boot));
+    Main_ClearMemory(SEGMENT_END(code), memEnd);
```
Dragorn421, line 58: Kinda would have expected something like (pseudocode)

```suggestion
    Main_ClearMemory(memStart, SEGMENT_START(framebuffer_1));
    Main_ClearMemory(SEGMENT_START(pre_boot_buffer), SEGMENT_END(pre_boot_buffer));
    Main_ClearMemory(SEGMENT_START(heap), memEnd);
```
, doesn't match? Or is this not what this effectively does?

Dragorn421 (commented):
🧐

Makefile
hensldm: Think I would agree with both points. Instead of a Makefile define, I wonder if it would just be better to add a `ifdef NON_MATCHING/`AVOID_UB` that uses the linker symbol.

mzxrules: I also think it's very unusual to have an address hardcoded through the makefile.

>if MM is anything like OoT (and I hear it's one of its best romhacks)

MM is a way better "romhack" than Master Quest :)

src/boot/idle.c
mzxrules: also, shouldn't `Main_ClearMemory(memStart, SEGMENT_START(framebuffer_1));` not be hardcoded to end at the start of `framebuffer_1`?

sonicdcer (approved):
I've been trying to do some mods myself and the game was crashing hard on the sGraphStack on real hardware. This PR fixed my issue and it made the GCC build also work fine on real hardware which wasn't working before. This is a great contribution.

Makefile
AngheloAlf: I agree with both points. Tbh this started like a "cool idea" I had when rearranging the spec, but it wasn't that cool when I saw it implemented :sweat_smile: 

I guess we could just simplify it to only leave a note in the spec mentioning that the user should only change that value only if they know what they are doing and telling them to also update all the other references on the repo

Roman971: Is there a reason not to put the define in a header then include that header in the spec and where it's used in code?
The spec file should be getting preprocessed so it's always an option

src/boot/idle.c
Dragorn421: I'm not sure what the first clear does but maybe (again idk if matches)

Makefile
AngheloAlf: Yours isn't totally correct. The system heap doesn't start after the `code` segment, it starts after the `buffers` segment (which starts after the `code` segment).
I tried to avoid referring to specific segments, but I don't have a good reason to do so either. What do you think to change the comment to refer to the `buffers` segment instead?

src/boot/idle.c
AngheloAlf: I'll answer in the order of the function calls

As a reference, this is on main right now:
```c
    Main_ClearMemory(memStart, gFramebuffer1);
    Main_ClearMemory(D_80025D00, bootproc);
    Main_ClearMemory(gGfxSPTaskYieldBuffer, memEnd);
```
https://github.com/zeldaret/mm/blob/71d83a5e15c69e5f2e91f45f55448f49bb215897/src/boot/idle.c#L55C45-L57

- `Main_ClearMemory(memStart, SEGMENT_START(framebuffer_1));`

Is clearing the range `0x80000400 ~ 0x80000500`. I have no idea what's in that range or why it is clearing it up.
We don't have any symbol that points to that range either. `gFramebuffer1` happens to be at `0x80000500` so we were using that.
I thought it would be better to use a linker symbol for this so I changed it (I don't think it is correct either, but less wrong than what we had).

- `Main_ClearMemory(SEGMENT_START(pre_boot_buffer), SEGMENT_START(boot));`

Sadly this can't reference the end of `pre_boot_buffer` because the way I set it up.
That segment only has 1 symbol that has size 1, so the end of the segment is at `start(pre_boot_buffer) + 0x10`, which isn't the address that we want to point to, that's why I used `SEGMENT_START(boot)` instead.
If we really wanted to change this to refer to the end of this segment, then we could change the size of the only symbol it has to have size `0x5A360`, that should bump the end linker symbol up to were we need it.

- `Main_ClearMemory(SEGMENT_END(code), memEnd);`

This one clears all the ram after `code`, since all of that is assumed to usable by the game as buffers. This PR is in a weird position because now there are 3 segments after the `code` one, so referring to only one of them feels a bit weird.
From all the 3 I think this one is the only one that would match by simply changing the symbol to `SEGMENT_START(buffers)`

Makefile
AngheloAlf: I changed the spec to have a hardcoded value and moved the explanation to `buffers.h`. `Main` is still using the define
https://github.com/zeldaret/mm/pull/1488/commits/c1debe5d0ad2e389c107ccb28dbd002b55800f92

src/boot/idle.c
mzxrules: Just remembered when I worked on the z_syscfb pr, I came up with these notes on how the buffers are allocated: https://github.com/zeldaret/mm/pull/1159#discussion_r1118077166

pre_boot_buffer is phony, the "framebuffer_1" segment should actually be u16 [HIRES_BUFFER_HEIGHT][HIRES_BUFFER_WIDTH] (454 x 576), which will get you to 0x80080000, then the missing 0x60 bytes is the space reserved for the makerom entrypoint.

So the second Main_ClearMemory is wiping `framebuffer_1` and the makerom entrypoint, only it's skipping over the lores color framebuffer being stored there for w/e reason

include/buffers.h
```diff
@@ -5,13 +5,31 @@
...
 
+/**
+ * The system heap is an area of memory that covers all the RAM that is not
+ * taken by compiled code/data.
+ * It starts just after the end of compiled code/data (that is not an overlay)
+ * and ends at a fixed location in RAM.
+ * That fixed location in RAM is used by the framebuffers of the game and have
+ * a fixed size.
+ * This address is calculated by doing `0x80800000 - (size of framebuffers)`,
+ * where 0x80800000 is the end of the Expansion Pak address range.
```
hensldm, line 27: Could we possibly frame the define like this comment, something like
`#define SYSTEM_HEAP_END_ADDR 0x80800000 - (size of framebuffers)`, though probably want 0x80800000 like FAULT_FB_ADDRESS and do `PHYS_TO_K0(0x800000)` instead.

Also not sure but would it be better to have the define be something like `FB_ADDRESS` instead and phrase it in terms of the framebuffers, rather than the end of something.

src/boot/idle.c
AngheloAlf: This is interesting.
The second `Main_ClearMemory` only wipes `pre_boot_buffer` segment, it doesn't touch the `framebuffer_1` segment.
`gFramebufferHiRes1` (which is placed at the same address of `framebuffer_1`) is bigger than the `framebuffer_1` segment and it leaks into `pre_boot_buffer`, so this call technically wipes a chunk of `gFramebufferHiRes1`, but it doesn't touch `gFramebuffer1` at all

include/buffers.h
AngheloAlf: I named it as the end of the system_heap segment because that's how it is used on `Main`:
```
    sysHeap = (intptr_t)SEGMENT_START(system_heap);
    fb = SYSTEM_HEAP_END_ADDR;
    gSystemHeapSize = fb - sysHeap;
    SystemHeap_Init((void*)sysHeap, gSystemHeapSize);
```

hensldm: Yeah, though you could see it as the beginning of the framebuffers (which is why it is the end of the system heap), which is actually how it currently portrayed with the temp variable being named `fb`.

src/boot/idle.c
mzxrules: And now I understand why the lores buffer is avoided; the memory at 80000500 to 80025D00 has already initialized by `Main_InitScreen`.

hensldm (approved):

engineer124 (approved):
I'm not exactly following the technicals of this PR, but if others who do are happy with it, then my approval is based on that

mzxrules:
I'm still not happy with it in it's current state. I don't like the prebuffer stuff, I think it's way too much of a hack to make a single function call look cleaner, and the names for `framebuffers.c` `framebuffer_1.c` should be cleaned up before merging.

Maybe it should be looking more like this: https://github.com/AngheloAlf/mm/compare/buffers_heaps_shenanigans...mzxrules:mm:framebuffer?expand=1

hensldm:
> I'm still not happy with it in it's current state. I don't like the preboot stuff, I think it's way too much of a hack to make a single function call look cleaner, and the names for `framebuffers.c` `framebuffer_1.c` should be cleaned up before merging.
> 
> Maybe it should be looking more like this: https://github.com/AngheloAlf/mm/compare/buffers_heaps_shenanigans...mzxrules:mm:framebuffer?expand=1

Fwiw I also think these changes would be good to have. There are a few things I might change, but overall they look good.

AngheloAlf:
> I'm still not happy with it in it's current state. I don't like the preboot stuff, I think it's way too much of a hack to make a single function call look cleaner, and the names for `framebuffers.c` `framebuffer_1.c` should be cleaned up before merging.
> 
> Maybe it should be looking more like this: https://github.com/AngheloAlf/mm/compare/buffers_heaps_shenanigans...mzxrules:mm:framebuffer?expand=1

An issue I noticed with the `union` approach is that functions that reference both `gFramebuffer1` and `gFramebufferHiRes1` still need to reference two distinct symbols to match, you can't do `gLoBuffer.gFramebuffer1` and `gLoBuffer.gFramebufferHiRes1`

Example:
https://decomp.me/scratch/wmaw7

hensldm (approved):
Any reason to leave the needs-discussion tag?

AngheloAlf:
Because I forgot to remove it :sweat_smile:

PR #1499, opened 2023-11-17
Organize `libc` and `libm` files

AngheloAlf:
Not to be confused with `libc64` ( #1492  )

Similarly to the `libc64` PR, this PR aims to identify functions that are part of `libc` and `libm` (the C standard library).
I'm naming these functions as the libc functions mainly because they are at the bottom of the `boot` segment.

I also cleaned up the libc headers to prevent them to depend on other non-libc headers, like libultra's headers.


hensldm (approved):

engineer124 (approved):

src/libultra/gu/sins.c
```diff
@@ -1,5 +1,6 @@
 #include "ultra64.h"
 #include "libc/math.h"
+#include "ultra64.h"
```
hensldm, line 3: duplicate

tom-overton (approved):

PR #1500, opened 2023-11-18
Add Shift-JIS support for some voice and message files

hensldm:
~~Uses iconv to convert a few files to shift-jis before compiling~~ Uses `shiftjis_conv.py` to convert shift-jis char literals to their hex values so that way we can use shift-jis characters. Note this does break the permuter for those few files, but they are already matched so I don't think it is too big a deal.


hensldm:
Changed the implementation so that it uses a python script that comes from ultralib, that instead of using encodings just replaces the char's with the correct hex. This allows more flexibility and shouldn't interfere with euc-jp strings.

tools/shiftjis_conv.py
```diff
@@ -0,0 +1,80 @@
...
+    for i, char in enumerate(buf):
+        if skipTimer > 0:
+            skipTimer -= 1
+            continue
+        if char == "'" and buf[i+1] in sjis_table and buf[i+2] == "'":
+            if sjis_table[buf[i+1]] == 0:
+                print("Error: Please map %s in %s" % (buf[i+1], sys.argv[0]))
+                exit(1)
+            outfile.write("0x%X" % sjis_table[buf[i+1]])
+            skipTimer = 2
```
AngheloAlf, line 64: I was wondering if it could be better to use Python's built-in encoding functions instead of a manual look-up table.

For example, you can do something like this
```python
character = "ー".encode("shift-jis")
hex_value = struct.unpack_from(">H", character)[0]
outfile.write(f"0x{hex_value:04X}")
```

This snipped lacks error checking (in case the character is not a Japanese one), but something among those lines should work. What do you think?

hensldm: Hmm does seem better than having to manage a look up table. Do you have any thoughts on how check if the character is a Japanese one?

A simple approach would be just check if the value is more than 1 byte `ord(TEST[i+1]) > 255`, but that would catch more than just Japanese characters. Also note we will need some characters that are not considered Japanese characters, but are still a part of shift-jis (e.g. ０)

hensldm: Implemented. Also changed the script to allow for outputting to stdout as well if desired.

AngheloAlf (approved):

engineer124 (approved):

yemenalcyone230:
Listen, I got all 23 masks (not encluding fierce deity mask) so I was in the twinmold room (on the moon should've said that earlier lol) and I realized I was a mask short. This happened before too. So I restarted the game and double checked, I had every slot filled except for that one slot where the fierce deity mask was supposed to go. Now I'm confused af someone please help!!

yemenalcyone230:
Is it the fierce diety mask and giant mask??If it is can you tell me where can i find it?

hirottau302:
I'm trying to beat snowhead right now, and i want to know: are there more than 2 switches that change the height of the round platform in the middle of the center room? cuz i'm stuck and trying to get to the boss room.

hirottau302:
I was genuinely wondering if I need all 24 masks for the moon children

ginomeelectra202:
The room where you get the chest of bombchus after defeating the knight. Anyone know how to get past this room? I've been sending them at the crack in the wall but nothing happens.

ginomeelectra202:
?meht teg ot laiceps gnihtyna deen uoy od dna ,taht si nehW .elbaliava emoceb yeht ,doirep yad-eerht eht gnirud emit emos gniruD .aera looP yrdnuaL eht ni etag retaw eht dniheb neddih seepuR deR era erehT

ginomeelectra202:
?niaga gnirpS ti ekam nac I yaw yna ereht sI .ereht gorf eht teg nac I os niaga gnirpS ti ekam ot woh wonk t'nod I won tub ,ti tog dna emit ni kcab tnaw I .enog saw eh tub ,ksaM oreG noD eht teg ot deirt ,elpmeT daehwonS taeb I

PR #1530, opened 2023-12-25
Text Extraction + some Message Docs

hensldm:
MERRY CHRISTMAS :)

This adds text extraction for both `message_data` (NES) and `staff_message_data_static` (Credits). The text data gets extracted to "assets/text/message_data.h" and "assets/text/staff_message_data.h" respectively. This also adds the ability to add new text by just adding things to either `message_data.h` or `staff_message_data.h` (though do note the very last message will not work as it will get DMA errors, so suggest just adding above both the 0xFFFx texts).

To view the control cmds message allows for each type look at the "message_data_fmt_nes.h"/"message_data_fmt_staff.h" files.

This also adds quite a bit of general documentation around Message with the ability to view and even change different text. This started to get rather large, so while more documentation could be added, I wanted to open this to start getting eyes on this and add more documentation later.

This also extracts the other message assets from `message_static` and `message_texture_static`. This also changes the spec to use the extracted `nes_font_static` as well.


include/message_data_fmt_nes.h
```diff
@@ -0,0 +1,340 @@
...
+ */
+
+#ifndef GLUE
+#define GLUE(a, b) a##b
+#endif
+
+#define STRINGIFY(s) #s
+#define EXPAND_AND_STRINGIFY(s) STRINGIFY(s)
+
+#define HEX(N) GLUE(0x, N)
```
jpburnett, line 17: Glue?

jpburnett (approved):
The tooling part all looked fine to me. Just curious what GLUE means in the message headers.

include/message_data_fmt_nes.h
hensldm: It glues the two arguments together into 1 token. For example `HEX(10) = GLUE(0x, 10) = 0x10`.

extract_assets.py
```diff
@@ -111,23 +111,46 @@ def main():
     signal.signal(signal.SIGINT, SignalHandler)
 
     extractedAssetsTracker = manager.dict()
-    if os.path.exists(EXTRACTED_ASSETS_NAMEFILE) and not args.force:
+    if not args.force and os.path.exists(EXTRACTED_ASSETS_NAMEFILE):
         with open(EXTRACTED_ASSETS_NAMEFILE, encoding='utf-8') as f:
             extractedAssetsTracker.update(json.load(f, object_hook=manager.dict))
 
+    extract_text_path = "assets/text/message_data.h"
+    extract_staff_text_path = "assets/text/staff_message_data.h"
```
AngheloAlf, line 119: I would suggest using `pathlib.Path` here, but given this script doesn't use it at all then it doesn't matter

tools/msg/nes/msgdisNES.py
```diff
@@ -0,0 +1,300 @@
...
+                    if not prevText or prevNewline:
+                        self.decodedText += '"'
+
+                    self.decodedText += f'\\n"\n'
+
+                    prevText = False
+                    prevNewline = True
+                else: # Control Codes (see message_data_fmt_nes.h)
+                    if prevText and not prevNewline:
+                        self.decodedText += '" '
```
AngheloAlf, line 229: I noted there's a trailing space in some lines of the extracted text.
This line and the `macro` function seem to be the culprits.

Example:
image

Not sure if it is worth go over the trouble of fixing this since it is a so small issue

AngheloAlf (approved):
Awesome work!

tools/msg/nes/msgdisNES.py
hensldm: Not as simple as just the one line (had more to do with the assumption that each cmd always ended with a space). Though I think I prefer the new algorithm anyways so thanks for bringing it up.

extract_assets.py
hensldm: Yeah, there are other places that should use it (for example the ZAPD path). Feel it would be better in a different PR.

src/code/z_message.c
```diff
@@ -1061,9 +1078,9 @@ void Message_DrawTextDefault(PlayState* play, Gfx** gfxP) {
     Gfx* gfx = *gfxP;
     u16 character;
     s16 sp130;
-    s16 sp12E;
-    s16 sp12C;
-    s16 sp12A;
+    s16 prevR;
+    s16 prevG;
+    s16 prevC;
```
hensldm, line 1083: Reminder to self: `prevC` -> `prevB` (possibly use a `Color_RGB16`?)

tools/msg/staff/msgdisStaff.py
```diff
@@ -0,0 +1,231 @@
...
+                prevNewline = False
+                prevCmd = True
+
+def parseTable(start):
+    table = {}
+    with open("baserom/code","rb") as f:
+        f.seek(start)
+        buf = f.read(8)
+        textId, typePos, segment = struct.unpack(">HBxI", buf)
+        while segment != 0:
```
hensldm, line 179: Reminder to self: Actually check if textId is` 0xFFFF` instead of segment is 0 like `Message_FindMessageCredits`

tools/msg/nes/msgdisNES.py
```diff
@@ -0,0 +1,316 @@
...
+                prevNewline = False
+                prevCmd = True
+
+def parseTable(start):
+    table = {}
+    with open("baserom/code","rb") as f:
+        f.seek(start)
+        buf = f.read(8)
+        textId, typePos, segment = struct.unpack(">HBxI", buf)
+        while segment != 0:
```
hensldm, line 259: Reminder to self: Actually check if textId is` 0xFFFF` instead of segment is 0 like `Message_FindMessageNES`

src/code/z_message.c
```diff
@@ -1031,24 +1048,24 @@ Color_RGB16 D_801CFE74[] = {
     { 255, 120, 0 },  { 70, 255, 80 },   { 80, 110, 255 },  { 255, 255, 30 },
     { 90, 180, 255 }, { 210, 100, 255 }, { 170, 170, 170 }, { 255, 130, 30 },
 };
-Color_RGB16 D_801CFEA4[] = {
+Color_RGB16 sColorsNormalJPN[] = {
```
engineer124, line 1051: I'd say either make these static or add a `Message` prefix?

src/overlays/actors/ovl_En_Ginko_Man/z_en_ginko_man.c
```diff
@@ -508,7 +508,7 @@ void EnGinkoMan_Dialogue(EnGinkoMan* this, PlayState* play) {
             EnGinkoMan_WaitForDialogueInput(this, play);
             break;
 
-        case TEXT_STATE_5:
+        case TEXT_STATE_EVENT:
```
engineer124, line 511: Could you add these text-states to namefixer?

engineer124 (approved):

PR #1532, opened 2023-12-27
Allow the mips tool chain to be picked via command argument

louist103:
Bring over OOT's makefile section to allow changing the default mips tool chain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,8 @@ OBJDUMP_BUILD ?= 0
 ASM_PROC_FORCE ?= 0
 # Number of threads to disassmble, extract, and compress with
 N_THREADS ?= $(shell nproc)
-
+#MIPS toolchain
+MIPS_BINUTILS_PREFIX ?= mips-linux-gnu-
 #### Setup ####
 
 # Ensure the map file being created using English localization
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,10 +65,11 @@ endif
 
 #### Tools ####
 
-ifeq ($(shell type mips-linux-gnu-ld >/dev/null 2>/dev/null; echo $$?), 0)
-  MIPS_BINUTILS_PREFIX := mips-linux-gnu-
-else
-  $(error Please install or build mips-linux-gnu)
+MIPS_BINUTILS_PREFIX := mips-linux-gnu-
```
hensldm, line 68: Could you bring this to the top in the Defaults section with a comment? Also from what I remember I think we need to use `?=` to allow overriding in `.make_options`, though I could be wrong.

Dragorn421: Yeah `:=`/`=` will replace any previous value, `?=` doesn't

hensldm (approved):

Makefile
```diff
@@ -27,7 +27,8 @@ OBJDUMP_BUILD ?= 0
 ASM_PROC_FORCE ?= 0
 # Number of threads to disassmble, extract, and compress with
 N_THREADS ?= $(shell nproc)
-
+#MIPS toolchain
```
AngheloAlf, line 30:
```suggestion
# MIPS toolchain
```

AngheloAlf (approved):

engineer124 (approved):

PR #1536, opened 2024-01-14
Extract Kanji + some adjacent docs

hensldm:
Extracts everything from the kanji file. I added the codepoint index similar to nes font, ~~though did not actually fully name the textures just used their offset (would prefer someone who actually knows the japanese characters to do it)~~. Textures are named thanks to Emily.
Also doc'd as much as possible code_80185F90.text.s (kanread.s) which is related.


AngheloAlf:
> Also doc'd as much as possible code_80185F90.text.s (kanread.s) which is related.
I don't see the docs anywhere. In case you documented the actual assembly file remember that we don't commit the assembly files

hensldm:
> > > > Also doc'd as much as possible code_80185F90.text.s (kanread.s) which is related.

> > I don't see the docs anywhere. In case you documented the actual assembly file remember that we don't commit the assembly files

Yeah. By doc'd I mean named the file, function, and data. Don't think I can do anything else without commit the assembly.

AngheloAlf:
> > I don't see the docs anywhere. In case you documented the actual assembly file remember that we don't commit the assembly files
> 
> Yeah. By doc'd I mean named the file, function, and data. Don't think I can do anything else without commiting the assembly.

I guess we could have a header and comments explaining what does the function do, what takes as arguments and whats the return value. But it is fine as it currently is

hensldm:
> > > I don't see the docs anywhere. In case you documented the actual assembly file remember that we don't commit the assembly files
> > 
> > 
> > Yeah. By doc'd I mean named the file, function, and data. Don't think I can do anything else without commiting the assembly.
> 
> I guess we could have a header and comments explaining what does the function do, what takes as arguments and whats the return value. But it is fine as it currently is

Ah I really like that idea. Let me do that real quick.

assets/xml/interface/kanji.xml
```diff
@@ -0,0 +1,3989 @@
+<Root>
+    <File Name="kanji">
+        <!-- The four-digit number after the prefix is the character's codepoint: the hexadecimal short it corresponds to in the text data -->
+
+        <!-- NonKanji -->
+        <Texture Name="gMsgKanji8140_000000" OutName="msg_kanji_8140_000000" Format="i4" Width="16" Height="16" Offset="0x0"/>
+        <Texture Name="gMsgKanji8141_000080" OutName="msg_kanji_8141_000080" Format="i4" Width="16" Height="16" Offset="0x80"/>
+        <Texture Name="gMsgKanji8142_000100" OutName="msg_kanji_8142_000100" Format="i4" Width="16" Height="16" Offset="0x100"/>
```
AngheloAlf, line 8: It took me a bit to realize the second set of values are the offset in the kanji segment itself. What do you think about adding a note explaining it?

AngheloAlf (approved):

assets/xml/interface/kanji.xml
hensldm: Done, also added a todo to remove the comment when the textures are properly named without the offset.

hensldm:
Kanji textures are now documented. Thank you Emily :) .

tom-overton (approved):

AngheloAlf (approved):
I don't really have a way to verify if every Japanese character is correct, but I trust you two for getting them right

engineer124 (approved):

PR #1600, opened 2024-01-31
Build Sync

hensldm:
Work to sync up with OoT's build system (including some minor multiversion support) as mentioned here <https://discord.com/channels/688807550715560050/688851976746041391/1199852164256899082>. I tried to make the commits pretty standalone with simple messages describing them.

Note: I did do a little more than described in that discord message.
1. Ended up adding $(BUILD_DIR) to the spec and added minor multiversion support similar to what OoT has (though no `setup`/`disassembly` of other versions)
2. Renamed the spec `SYMS` flag to `UNSET` to match the name I gave the dma entry macro (also added it to a few more segments)
3. Moved the built roms and elf to the build dir, along with renaming the map and ldscript based on the version (see: https://github.com/zeldaret/oot/pull/1651).
4. Added a run target similar to what OoT has
5. Minor Makefile clean up

Note2: A lot of tools are now version aware, but most are still hardcoded to just use the US version.

Note3: Would recommend rereading the README to setup the baserom in the correct directory.


Makefile
```diff
@@ -5,8 +5,16 @@
...
+# Ensure the build fails if a piped command fails
+SHELL = /bin/bash
+.SHELLFLAGS = -o pipefail -c
+
+
 #### Defaults ####
 
+# Target game version. Currently only the following version is supported:
+#   us   N64 USA (default)
+VERSION ?= us
```
AngheloAlf, line 17: Since there are versions for other platforms we should at least have `n64` somewhere in the version name
```suggestion
# Target game version. Currently only the following version is supported:
#   us   N64 USA (default)
VERSION ?= n64-us
```

Makefile
```diff
@@ -29,8 +37,13 @@ ASM_PROC_FORCE ?= 0
 N_THREADS ?= $(shell nproc)
 # MIPS toolchain prefix
 MIPS_BINUTILS_PREFIX ?= mips-linux-gnu-
+# Python virtual environment
+VENV ?= .venv
 # Python interpreter
-PYTHON ?= python3
+PYTHON ?= $(VENV)/bin/python3
```
AngheloAlf, line 43: Something interesting that I learnt a few days ago when debugging a bug of a python library that only happened on Windows (it was a blast as you can imagine), is that the Python binary and activate script are not put under `venv/bin/`, but under the `venv/Scripts/` path instead. https://docs.python.org/3/library/venv.html#how-venvs-work

Sounds like it could be a problem for Windows users not using WSL

Makefile
```diff
@@ -297,120 +311,137 @@ $(O_FILES): | asset_files
...
 
+venv:
+	test -d $(VENV) || python3 -m venv $(VENV)
+	$(PYTHON) -m pip install -U pip
+	$(PYTHON) -m pip install -U -r requirements.txt
+
 ## Extraction step
-setup:
+setup: venv
 	$(MAKE) -C tools
```
AngheloAlf, line 339: As mentioned in discord, I think the `setup` step shouldn't setup the venv and update the dependencies each time it is run.

```suggestion
setup:
	$(MAKE) -C tools
```

Makefile
```diff
@@ -297,120 +311,137 @@ $(O_FILES): | asset_files
...
 	cp -r build expected/build
 
-init:
-	$(MAKE) distclean
+init: distclean
 	$(MAKE) setup
 	$(MAKE) assets
 	$(MAKE) disasm
 	$(MAKE) all
 	$(MAKE) diff-init
```
AngheloAlf, line 362: Following `setup` not implicitly using `venv`, that step would need to be added to the `init` target
```suggestion
init: distclean
	$(MAKE) venv
	$(MAKE) setup
	$(MAKE) assets
	$(MAKE) disasm
	$(MAKE) all
	$(MAKE) diff-init
```

docs/tutorial/vscode.md
```diff
@@ -39,18 +39,18 @@ You can create a `.vscode/c_cpp_properties.json` file with `C/C++: Edit Configur
 {
     "configurations": [
         {
-            "name": "Linux",
+            "name": "N64 mm-us",
```
AngheloAlf, line 42: The `mm` part feels kinda redundant given that mm is the only game that we will ever have on this repo
```suggestion
            "name": "n64 us",
```

tools/extract_baserom.py
```diff
@@ -1,1618 +1,1633 @@
...
+
+def main():
+    SEGMENTS_PATH.mkdir(parents=True, exist_ok=True)
+
+    # read baserom data
+    try:
+        rom_data = ROM_FILE_PATH.read_bytes()
+    except IOError:
+        print(f"failed to read file {ROM_FILE_PATH}")
+        exit(1)
```
AngheloAlf, line 1623: I think it would be better if we checked if the file exists instead with `.exists()`

docs/tutorial/vscode.md
fig02: you can have local configurations that you can pick from within vscode, since the .vscode folder isnt committed. it was dragorns reason for renaming it in the first place: https://github.com/zeldaret/oot/pull/1669#discussion_r1472072110

so including the game name does make sense

AngheloAlf: Yeah, but will ever have MM related configurations on your local `.vscode`. It is not like you would put an OoT configuration on `mm/.vscode`, so I don't understand the idea of putting the game name

README.md
```diff
@@ -91,55 +92,41 @@ This will copy the GitHub repository contents into a new folder in the current d
...
 ```
 
-#### 3. Install python dependencies
+#### 3. Prepare a base ROM
 
-The build process has a few python packages required that are located in `requirements.txt`.
+Place a copy of the US ROM inside the `baseroms/us/` folder.
 
-It is recommend to setup a virtual environment for python to localize all dependencies. To create a virtual environment:
+Rename the file to `baserom.z64`, `baserom.n64` or `baserom.v64`, depending on the original extension..
```
jpburnett, line 99:
```suggestion
Rename the file to `baserom.z64`, `baserom.n64` or `baserom.v64`, depending on the original extension.
```
Extra period

README.md
```diff
@@ -91,55 +92,41 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+#### 4. Make and Build the ROM
 
 To start the extraction/build process, run the following command:
 
 ```bash
 make init
 ```
 
-This will extract all the individual files in the ROM into a newly created baserom folder, as well as decompress the compressed files in a newly created decomp folder. This will create the build folders as well as a new folder with the ASM as well as containing the disassemblies of nearly all the files containing code.
+This downloads some dependencies (from pip), and compiles tools for the build process, extracts all the individual files in the ROM into a newly created baserom folder, as well as decompress the compressed files in a newly created decomp folder. This will create the build folders as well as a new folder with the ASM as well as containing the disassemblies of nearly all the files containing code.
```
jpburnett, line 109:
```suggestion
The build process:
1. Fetches dependencies: Downloads necessary tools from pip.
2. Prepares build environment: Compiles tools for the build process.
3. Extracts ROM contents:
    - Creates a baserom folder to house individual files extracted from the ROM.
    - Decompresses compressed files into a decomp folder.
4. Builds and disassembles code:
    - Creates folders for the build process.
    - Generates a folder containing ASM files and disassemblies of most code-containing files.
```
This gives a more clear outline of the process I think.

requirements.txt
```diff
@@ -1,12 +1,12 @@
-# Setup
+# setup
 colorama>=0.4.3
 crunch64>=0.3.1,<1.0.0
 ipl3checksum>=1.2.0,<2.0.0
 
 # disasm
 rabbitizer>=1.0.0,<2.0.0
```
jpburnett, line 7: This is just a comment in general (and perhaps solicits a wider discussion...), can we adopt pinning exact versions? Loose version constraints are okay, but pinning exact versions would offer more control over the environment (venvs in this case) and prevent unexpected changes due to package updates. It would also facilitate troubleshooting.

jpburnett (commented):
Thanks for bringing it over! This is what I had to comment right now, will finish later tonight.

requirements.txt
AngheloAlf: I'm not a huge fan of pinning exact versions. Or at least not for the libraries that we have control over.
We usually add new features and fix issues pretty quickly, it would be annoying to be out of date and need to do PRs that only do version bumps on the requirements.txt file

docs/tutorial/vscode.md
hensldm: Since this is a recommendation anyways, I feel like it doesn't really matter too much. With that in mind I just decided to go with `n64-us`.

README.md
hensldm: Adjusted wording slightly, let me know if it still works.

Makefile
hensldm: Latest commits should fix this. Note I moved the os detection before the defaults, so PYTHON would still be editable in .make_options.

requirements.txt
AngheloAlf: Yeah, our packages already adhere to semver so it shouldn't be an issue

README.md
```diff
@@ -91,55 +92,50 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+The extraction/build process:
+1. Fetches dependencies: Downloads necessary tools from pip.
+2. Prepares build environment: Compiles tools for the build process.
+3. Extracts ROM contents:
+    - Creates a baserom folder to house individual files extracted from the ROM.
+    - Decompresses compressed files into a decomp folder.
+4. Disassembles code:
+    - Creates folders for the build process.
+    - Generates a folder containing ASM files and disassemblies of most code-containing files.
+5. Builds the ROM
```
AngheloAlf, line 118: This is missing asset extraction, isn't it?

hensldm: Expanded on this. Let me know if I should add/change anything.

Makefile
```diff
@@ -29,6 +27,9 @@ else
 endif
 
 
+-include .make_options
```
AngheloAlf, line 30: Why this was moved down here?

hensldm: Hmm, at first I thought it seemed nicer, but on second thought it really should be above. I'll move it back.

AngheloAlf:
I think the change overall is fine.
I did a bit of local testing to check everything works fine and noticed that the compiled schedules are still being put under the `build/src` path instead of the `build/n64-us/src` path

hensldm:
> I did a bit of local testing to check everything works fine and noticed that the compiled schedules are still being put under the `build/src` path instead of the `build/n64-us/src` path

Should be fixed now. Note I had to change the includes and remove the `build/` so that the `Ibuild/$(VERSION)` works.

AngheloAlf (approved):

hensldm:
Note brought over the changes made in OoT's 1704 PR.

isghj5 (approved):

PR #1601, opened 2024-02-04
Bss script update

AngheloAlf:
- Update the bss script to use mapfile parser
- Add flag to specify sections other than `.bss`, allowing to check for reordering in other sections.
- Hardcode the path to the mapfile
  - This way it is more consistent with the other scripts.
- Some code cleanups

Misc changes:
- Update mapfile_parser version to 2.x.x
  - The most remarkable change is the actual mapfile parsing was rewritten in Rust, so it should be faster
- Changed the `python-Levenshtein` requirement to `Levenshtein`
  - According to its [PyPI page](https://pypi.org/project/python-Levenshtein/) the project was renamed, so I think it is better to use the new name instead.
-


tools/global_bss_check.py
```diff
@@ -142,63 +123,37 @@ def main():
...
-                print(f"bss reordering in {file}", file=os.sys.stderr)
-            print("", file=os.sys.stderr)
-            
-            if not args.no_fun_allowed:
-                print(colorama.Fore.LIGHTWHITE_EX +
-                "  BSS is REORDERED!!\n"
-                "  Oh! MY GOD!!" 
-                + colorama.Style.RESET_ALL, file=os.sys.stderr)
-                print("", file=os.sys.stderr)
+    mapfilePath = Path("build/mm.map")
```
hensldm, line 132: More so note to self. Uses mapfile so would need to be updated when the build sync PR gets in (or vice versa if this gets in first).

hensldm (commented):
Since this no longer just checks `bss` should we rename the script? Perhaps `check_reordering.py`?

Dragorn421 (commented):
Does it check specifically for reordering, ignoring shift?
e.g. it wouldn't yell if a,b,c at 0x10 become a,b,c at 0x14 ?

Would it be possible to not restrict by section if that's an operation that makes sense?
could be `--section='*'` or `--all-sections`

(also spdx reminder)

AngheloAlf:
If it sees a symbol is shifted then it will complain.
Example:
```c
f32 sKaleidoSetupRightPageEyeX[] = {
    PAUSE_EYE_DIST * -PAUSE_MAP_X,   // PAUSE_ITEM
    PAUSE_EYE_DIST * -PAUSE_QUEST_X, // PAUSE_MAP
    PAUSE_EYE_DIST * -PAUSE_MASK_X,  // PAUSE_QUEST
    PAUSE_EYE_DIST * -PAUSE_ITEM_X,  // PAUSE_MASK
    0 // Added this
};
f32 sKaleidoSetupRightPageEyeZ[] = {
    PAUSE_EYE_DIST * -PAUSE_MAP_Z,   // PAUSE_ITEM
    PAUSE_EYE_DIST * -PAUSE_QUEST_Z, // PAUSE_MAP
    PAUSE_EYE_DIST * -PAUSE_MASK_Z,  // PAUSE_QUEST
    PAUSE_EYE_DIST * -PAUSE_ITEM_Z,  // PAUSE_MASK
};

```
```
Build mapfile:    build/mm.map
Expected mapfile: expected/build/mm.map

Symbol Name,Build Address,Build File,Expected Address,Expected File,Difference,GOOD/BAD/MISSING
sKaleidoSetupRightPageEyeZ,801BDB1C,src/code/z_kaleido_setup.c,801BDB1C,src/code/z_kaleido_setup.c,4,BAD

  BAD
Symbol reordering in src/code/z_kaleido_setup.c

  Symbols are REORDERED!!
  Oh! MY GOD!!

```

Dragorn421:
is it intended both the build and expected addresses are `801BDB1C` in this example?

AngheloAlf:
No, that's a bug on the library

tools/check_reordering.py
```diff
@@ -0,0 +1,167 @@
...
+    description = "Check that globally visible symbols has not been reordered."
+    epilog = """\
+    N.B. Since this script reads the map files, it can only see globally visible symbols; in-function static symbols must be examined with other tools.
+    """
+
+    parser = argparse.ArgumentParser(description=description, epilog=epilog, formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument("-a", "--print-all", help="Print all symbols of the section, not just non-matching.", action="store_true")
+    parser.add_argument("-n", "--no-fun-allowed", help="Remove amusing messages.", action="store_true")
+    parser.add_argument("-r", "--no-reverse-check", help="Disable looking for symbols on the expected map that are missing on the built map file.", action="store_true")
+    parser.add_argument("-s", "--section", help="Specify which section should be checked for reordered symbols. Use * to check all sections. Defaults to .bss", default=".bss")
```
hensldm, line 134: I wonder if we should have some kind of check for section names? I could easily see someone pass `bss` and think `bss` is fine when they needed to pass `.bss` instead.

tools/check_reordering.py
```diff
@@ -0,0 +1,167 @@
...
+    if args.no_fun_allowed:
+        return 0
+
+    eprint("\n" + colorama.Fore.LIGHTWHITE_EX +
+    colorama.Back.RED + f"                                  " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"         CONGRATURATIONS!         " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"    All Global {section} is correct.    " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"             THANK YOU!           " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"      You are great decomper!     " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"                                  " + colorama.Style.RESET_ALL )
```
hensldm, line 161: Note that since the section name is variable length now, it does kind of mess with the red background where that one line is longer then the rest. I'm fine with leaving it as is, but just wanted to mention it in case it was something you wanted to look at.

docs/tutorial/vscode.md
```diff
@@ -76,6 +76,7 @@ Add the following to (or create) the `.vscode/settings.json` file for VSCode to
         "build/**": true,
         "expected/**": true,
         "nonmatchings/**": true,
+        ".venv": true,
```
hensldm, line 79:
```suggestion
        ".venv/**": true,
```
Don't think it makes any difference, but this matches the style of the others.

tools/check_reordering.py
hensldm, line 134: I wonder if the script should except `all` to check all sections, rather than "*" just to avoid `*` being the wildcard in bash as well.

tools/check_reordering.py
Dragorn421: That's absolutely notOK

You could use (untested)
```py
"{:^34}".format(f"All Global {section} is correct.")
```
to center the string in a 34-columns string, padding with spaces
( https://docs.python.org/3/library/string.html#grammar-token-format-spec-format_spec )

tools/check_reordering.py
Dragorn421: could move `mapfile_parser.MapFile()` and `readMapFile` out of `compareMapFiles` and into `main`,
then do something like
```py
all_section_types = {f.sectionType for seg in mapFile for f in seg}
if section not in all_section_types:
    approx_section_types = [section_type for section_type in all_section_types if section in section_type]
    if len(approx_section_types) == 1:
        print("Unknown section", section, "using", approx_section_types[0], "instead")
        section = approx_section_types[0]
    elif not approx_section_types:
        print("Unknown section", section)
        exit(1)
    else:
        print("Unknown section", section, "did you mean", approx_section_types)
        exit(1)
```

Dragorn421: it is a standard wildcard symbol though, including in linker scripts and binutils section manipulation
I think it's fine to expect the user knows to quote it like `--section='*'`
I'm not even sure `blabla=*` would cause the shell to expand `*` to anything tbh

AngheloAlf: Shell does expand the `*`
```
$ tools/check_reordering.py --section *
usage: check_reordering.py [-h] [-a] [-n] [-r] [-s SECTION]
check_reordering.py: error: unrecognized arguments: assets baserom baserom.mm.us.rev1.z64 baseroms baserom_uncompressed.z64 build ...
```
Doing `--section=*` and `-s=*` do not expand to everything on the root folder

Given that this is kinda bad I think I'll go with `all` instead

AngheloAlf: I went with a hardcoded choices list on the argument instead

tools/check_reordering.py
```diff
@@ -0,0 +1,167 @@
...
+    eprint("")
+    eprint(colorama.Fore.GREEN + "  GOOD" + colorama.Style.RESET_ALL)
+
+    if args.no_fun_allowed:
+        return 0
+
+    eprint("\n" + colorama.Fore.LIGHTWHITE_EX +
+    colorama.Back.RED + f"                                  " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"         CONGRATURATIONS!         " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "{:^34}".format(f"All Global {section} is correct.") + colorama.Back.RESET + "\n" +
```
hensldm, line 158:
```suggestion
    colorama.Back.RED + "{:^34}".format(f"All Global {section if section != "all" else "Symbols"} is correct.") + colorama.Back.RESET + "\n" +
```
Very minor (tbh I don't expect all to be used all that often).

hensldm (approved):

tools/global_bss_check.py
hensldm: Bump. Mapfile path will now need to be updated.

tools/check_reordering.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+
+    parser = argparse.ArgumentParser(description=description, epilog=epilog, formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument("-a", "--print-all", help="Print all symbols of the section, not just non-matching.", action="store_true")
+    parser.add_argument("-n", "--no-fun-allowed", help="Remove amusing messages.", action="store_true")
+    parser.add_argument("-r", "--no-reverse-check", help="Disable looking for symbols on the expected map that are missing on the built map file.", action="store_true")
+    parser.add_argument("-s", "--section", help="Specify which section should be checked for reordered symbols. Use `all` to check all sections. Defaults to .bss", default=".bss", choices=[".text", ".data", ".rodata", ".bss", "all"])
+    args = parser.parse_args()
+
+    mapfilePath = Path("build/mm.map")
```
hensldm, line 137: Mapfile path needs to be updated now.

tools/check_reordering.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if args.no_fun_allowed:
+        return 0
+
+    eprint("\n" + colorama.Fore.LIGHTWHITE_EX +
+    colorama.Back.RED + f"                                  " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"         CONGRATURATIONS!         " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + "{:^34}".format(f"All Global {section if section != 'all' else 'Symbols'} is correct.") + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"             THANK YOU!           " + colorama.Back.RESET + "\n" +
+    colorama.Back.RED + f"      You are great decomper!     " + colorama.Back.RESET + "\n" +
```
jpburnett, line 161: <3

jpburnett (approved):

PR #1603, opened 2024-02-07
`ovl_En_Door` docs pass

AngheloAlf:
- Named all the used `EnDoorType`
- What originally was `switchFlag` was really a multipurpose, where its meaning depended on the current `doorType`.
  - Because of this I changed it to be an `union` named `actionVar`. Each member of the `union` is named after its corresponding `doorType`.
- Introduced the `EnDoorScheduleType` and `EnDoorFramedType` enums. They should be used with their specific door types
- Now that we know that `SubS_FindDoor` is not a general "find door" rutine, but instead it is specific for schedule doors, so I renamed it to `SubS_FindDoorSchedule`.
  - Also made all the uses of this function consistent.
- Gave a documentation pass to the schedules themselves.
  - Would be nice to get a bit extra attention here since this is the first actor that has its schedule documented this way.
  - Due to how the schedule system is used by `EnDoor`, I opted to not introduce an `ScheduleResult` enum like we usually do, because I feel like the raw number holds more value than an enum.
- Renamed the `playOpenAnim` member of the `KnobDoorActor` struct to `requestOpen`.

Another important thing is I tried a new experiment with `static_assert`s on this branch. I'm using them in two ways:
- Ensure the size of an array matches what is expected (either checking against an enum or similar).
  - Instead of our usual way of putting the MAX value of an enum as the array size (ie `someArray[VALUE_MAX]`) I'm checking the size of the array is equal to the MAX value.
    ```c
    s32 someArray[] = { ... };
    static_assert(ARRAY_COUNT(someArray) == VALUE_MAX, "error message");
    ```
  - Our current approach has the issue that if the enum is extended or if an element of the array is removed then the array would silently grow with zeroes to fill the required size. This approach would make the compiler bark at the user if the array does not match exactly what is expected.
- Ensure two different enums don't go out of sync.
  - Some enums have subranges that should be synced with other enums, like the `EnDoorObjectInfoIndex` and `EnDoorScheduleType` enums, so I'm comparing the start and end of the ranges of those enums that should match between them, to make sure one is not getting extra entries on the middle without the other from being updated.

I'm open to discuss those two approaches, and I can remove them if we don't end up liking them.


src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -5,30 +5,127 @@
...
+ * | 0 0 0 0     | 0 0 0 
+ * | 4           | 3
+ * |
+ *
+ * Text offset          0000000001111000    Used as an offset relative to a base text id. The base text id is different depending on the door type.
+ * Half day bit index   0000000000000111    Index into the D_801AED48 array
+ *
+ *
+ * The rest of the door types either don't use the actionv var or don't bitpack extra parameters on it (using the full 7 bits)
+ */
```
AngheloAlf, line 37: Shamelessly stolen from OoT

src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -5,30 +5,127 @@
...
-    /* 0x1A5 */ u8 switchFlag;
-    /* 0x1A6 */ u8 unk_1A6;
-    /* 0x1A7 */ s8 openTimer; // For how long the door will be open. positive/negative means the opening direction
+    /* 0x1A5 */ union {
+                    u8 actionVar; // generic
+                    u8 halfDayCheck; // ENDOOR_TYPE_WHOLE_DAY, ENDOOR_TYPE_DAY and ENDOOR_TYPE_NIGHT
+                    u8 switchFlag; // ENDOOR_TYPE_LOCKED
+                    u8 schType; // ENDOOR_TYPE_SCHEDULE
+                    u8 frameType; // ENDOOR_TYPE_FRAMED
+                } actionVar; // TODO: think on a better name
```
AngheloAlf, line 126: I'm not completely sure if `actionVar` is a good name tbh. I'm open to suggestions

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -165,23 +302,28 @@ static u8 sAnimCloseFrames[2 * PLAYER_FORM_MAX] = {
...
+    { gPostOfficeTradingPostDoorDL, gPostOfficeTradingPostDoorDL }, // DOOR_DL_POST_OFFICE_TRAIDING_POST
+    { gInnSchoolDoorDL, gInnSchoolDoorDL },                         // DOOR_DL_INN_SCHOOL
+    { gMilkBarDoorDL, gMilkBarDoorDL },                             // DOOR_DL_MILK_BAR
+    { gMusicBoxHouseDoorDL, gMusicBoxHouseDoorDL },                 // DOOR_DL_MUSIC_BOX
+    { gPiratesFortressDoorDL, gPiratesFortressDoorDL },             // DOOR_DL_PIRATES_FORTESS
+    { gOceansideSpiderHouseDoorDL, gOceansideSpiderHouseDoorDL },   // DOOR_DL_OCEANSIDE_SPIDER_HOUSE
+    { gFieldWoodDoorLeftDL, gFieldWoodDoorRightDL },                // DOOR_DL_DEFAULT_FIELD_KEEP
 };
+static_assert(ARRAY_COUNT(sDoorDLists) == DOOR_DL_MAX,
+              "The entry count of `sDoorDLists` should match the `EnDoorDListIndex` enum");
```
AngheloAlf, line 326: Instead of using the `DOOR_DL_MAX` as the size of the array I'm using an `static_assert` to make sure the size matches what it should, preventing silent issues

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -57,63 +97,155 @@ ActorInit En_Door_InitVars = {
...
+    /* 49 */ DOOR_OBJINFO_MAX
+} EnDoorObjectInfoIndex;
+
+// These static asserts try to ensure the two enums don't get out of sync
+static_assert(ENDOOR_SCH_TYPE_SWORDSMANS_SCHOOL == DOOR_OBJINFO_17 - DOOR_OBJKIND_SCHEDULE,
+              "The enums values of `EnDoorScheduleType` and `EnDoorObjectInfoIndex` (from `DOOR_OBJKIND_SCHEDULE` "
+              "onwards) must be synced.");
+static_assert(ENDOOR_SCH_TYPE_MAX == DOOR_OBJINFO_MAX - DOOR_OBJKIND_SCHEDULE,
+              "The enums values of `EnDoorScheduleType` and `EnDoorObjectInfoIndex` (from `DOOR_OBJKIND_SCHEDULE` "
+              "onwards) must be synced.");
```
AngheloAlf, line 189: Here I check for the start of the schedule range of `EnDoorObjectInfoIndex` and the max of this enum to be synced with the `EnDoorScheduleType` enum.
So if one enum gets an extra entry without the other one being updated, the compiler will bark at you

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -57,63 +97,155 @@ ActorInit En_Door_InitVars = {
...
     /* 0x4 */ s16 objectId;
 } EnDoorInfo; // size = 0x6
 
+// TODO: This enum feels kinda overkill...
+typedef enum EnDoorObjectInfoIndex {
+    /*  0 */ DOOR_OBJKIND_DEFAULT,
+    /*  0 */ DOOR_OBJINFO_0 = DOOR_OBJKIND_DEFAULT,
+    /*  1 */ DOOR_OBJINFO_1,
+    /*  2 */ DOOR_OBJINFO_2,
+    /*  3 */ DOOR_OBJINFO_3,
```
AngheloAlf, line 130: I'm not completely sure about this enum.
I initially thought it would be like Anju's animation enum because it also has ranges where each range has its own meaning, but most values of this enum are not explicitly referenced in the code, so I'm not really sure if it is worth having.
At the same time, having it and keeping the `static_assert`s give a bit extra safety on the code's shiftability.

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -193,165 +335,241 @@ void EnDoor_Init(Actor* thisx, PlayState* play2) {
...
+               (Actor_WorldDistXZToActor(&this->knobDoor.dyna.actor, &GET_PLAYER(play)->actor) >
+                DOOR_AJAR_SLAM_RANGE)) {
+        this->actionFunc = EnDoor_AjarWait;
+        this->knobDoor.dyna.actor.world.rot.y = -0x1800;
     }
 }
 
-void func_80866B20(EnDoor* this, PlayState* play) {
-    static s32 D_80867BC0;
+s32 D_80867BC0;
```
hensldm, line 432: Perhaps `isMilkBarMember`? since that seems to be when it gets set to true.

src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -5,30 +5,128 @@
...
+    /* 1 */ ENDOOR_TYPE_LOCKED,
+    /* 2 */ ENDOOR_TYPE_DAY, // Allows to specify the door may be closed for specific dawns. Which dawns the door is closed is controlled by the "half day bit index" part of the parameter
+    /* 3 */ ENDOOR_TYPE_NIGHT, // Allows to specify the door may be closed for specific nights. Which nights the door is closed is controlled by the "half day bit index" part of the parameter
+    /* 4 */ ENDOOR_TYPE_AJAR, // unused
+    /* 5 */ ENDOOR_TYPE_SCHEDULE,
+    /* 6 */ ENDOOR_TYPE_6, // unreferenced
+    /* 7 */ ENDOOR_TYPE_FRAMED // unused
 } EnDoorType;
 
+// ENDOOR_TYPE_SCHEDULE
```
hensldm, line 54:
```suggestion
```
Probably not needed anymore?

src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -5,30 +5,128 @@
...
+    /* 26 */ ENDOOR_SCH_TYPE_ROMANI_RANCH_BEDROOM,
+    /* 27 */ ENDOOR_SCH_TYPE_IKANA_CANYON_MUSIC_BOX_HOUSE,
+    /* 28 */ ENDOOR_SCH_TYPE_DAMPES_HOUSE,
+    /* 29 */ ENDOOR_SCH_TYPE_MAGIC_HAGS_POTION_SHOP,
+    /* 30 */ ENDOOR_SCH_TYPE_30, // unused
+    /* 31 */ ENDOOR_SCH_TYPE_SWAMP_SHOOTING_GALLERY,
+    /* 32 */ ENDOOR_SCH_TYPE_MAX
+} EnDoorScheduleType;
+
+// ENDOOR_TYPE_FRAMED
```
hensldm, line 91:
```suggestion
```
Same here

src/overlays/actors/ovl_En_Door/z_en_door.h
hensldm: Possibly name after type `typeVar` (still don't like `var` but can't think of a good way to remove it). would also probably say change the first element to something like just `data` that way it is generic and different then the union name.

hensldm (commented):
Need a bit more time to think about the static asserts. I think I like them, but not sure.

src/overlays/actors/ovl_En_Pm/z_en_pm.c
```diff
@@ -1344,7 +1344,7 @@ s32 func_80AF9BF8(EnPm* this, PlayState* play, ScheduleOutput* scheduleOutput) {
 }
 
 s32 func_80AF9D04(EnPm* this, PlayState* play) {
-    EnDoor* door = (EnDoor*)func_80AF7D60(play, this->unk_258);
+    EnDoor* door = (EnDoor*)EnPm_FindDoorSchedule(play, this->unk_258);
```
hensldm, line 1347:
```suggestion
    EnDoor* door = EnPm_FindDoorSchedule(play, this->unk_258);
```
Cast doesn't seem necessary?

hensldm (commented):
Some more thoughts. Thinking I'm liking the static asserts, but still not completely sure.

include/z64subs.h
```diff
@@ -97,7 +97,7 @@ typedef struct ActorPathing {
     /* 0x68 */ ActorPathingUpdateFunc setNextPointFunc; // Return true if should compute and update again
 } ActorPathing; // size = 0x6C
 
-struct EnDoor* SubS_FindDoor(struct PlayState* play, s32 switchFlag);
+struct EnDoor* SubS_FindDoorSchedule(struct PlayState* play, s32 schType);
```
hensldm, line 100: I keep coming back to this, but I kind of want to to do `FindScheduleDoor` or maybe even `FindScheduledDoor`. Right now the name makes me think it is going to find the schedule for the door (i.e. gets the script).

hensldm (approved):
1 more thought but everything looks good :)

include/z64subs.h
AngheloAlf: `FindScheduleDoor` sounds fine, I'll change it

jpburnett (approved):

PR #1607, opened 2024-02-22
Mayor Dotour (ovl_en_dt) OK and documented 

GinNoOokami:
- This NPC is closely related to two additional NPCs: z_en_baisen (the guard) and z_en_muto (the carptener)
- Filled out some previously unknown fields in z_en_baisen
- Documented some event flags


src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+
+    /* 23 */ EN_DT_CS_FOCUS_TARGET_MAYOR /* [2AC5] "You take refuge, too." */
+};
+
+typedef struct {
+    s16 D_80BEB268[1];
+    s16 D_80BEB26A[25];
+} CutsceneIndexTable;
+
+static CutsceneIndexTable CsIndexTable = { { 0x2ABB },
```
GinNoOokami, line 138: This table is a bit strange to me. It seems like the developers offset the initial data pointer depending on whether they wanted the cutscene ID index or the string ID, and then incremented by 2. I couldn't find a more appropriate data representation that still resulting in a match.

Open to suggestions here.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+    if (this->animIndex == EN_DT_ANIMATION_2 || this->animIndex == EN_DT_ANIMATION_5) {
+        morphFrames = 0.0f;
+    }
+
+    this->animEndFrame = (f32)Animation_GetLastFrame(sAnimations[this->animIndex]);
+    Animation_Change(&this->skelAnime, sAnimations[this->animIndex], 1.0f, 0.0f, this->animEndFrame,
+                     sAnimationModes[this->animIndex], morphFrames);
+}
+
+void EnDt_UpdateVisualState(EnDt* this) {
```
GinNoOokami, line 248: I had a bit of trouble matching this method and had to rely on some permuter craziness to fix some final reg alloc mismatching. I also could not find a more appropriate representation for the data that still resulted in a match, so I'm open to ideas here as well.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -15,7 +19,70 @@ void EnDt_Destroy(Actor* thisx, PlayState* play);
...
+    /* 3 */ EN_DT_NPC_STATE_OFFERED_MEETING_REWARD,
+    /* 4 */ EN_DT_NPC_STATE_WAIT_FINAL_NIGHT_TALK,
+    /* 5 */ EN_DT_NPC_STATE_DONE_FINAL_NIGHT_TALK
+} EnDtNpcState;
+
+typedef enum {
+    /* 0 */ EN_DT_CS_STATE_NONE,
+    /* 1 */ EN_DT_CS_STATE_WAITING,
+    /* 2 */ EN_DT_CS_STATE_PLAYING
+} EnDtCutsceneState;
```
GinNoOokami, line 52: It seems like the Baisen/Muto npcs also share this enum, so it might be a good candidate to extract into a shared header.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+
+    if ((gSaveContext.save.day == 3) && (gSaveContext.save.isNight)) {
+        EnDt_SetupFinalNightState(this, play);
+    } else {
+        s32 csId = this->actor.csId;
+        s32 i = 0;
+
+        // clang-format off
+        while (csId != CS_ID_NONE) { this->csIds[i] = csId; csId = CutsceneManager_GetAdditionalCsId(csId); i++; }
+        // clang-format on
```
jpburnett, line 207: Any particular reason for wanting this while loop to be a single line? IMO this could be formatted just fine.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+    }
+    this->actionFunc = EnDt_SetupCutsceneNpcs;
+}
+
+void EnDt_Destroy(Actor* thisx, PlayState* play) {
+    EnDt* this = THIS;
+    Collider_DestroyCylinder(play, &this->collider);
```
jpburnett, line 214:
```suggestion
    EnDt* this = THIS;
    
    Collider_DestroyCylinder(play, &this->collider);
```
Space here

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+
+void EnDt_UpdateHeadRotate(EnDt* this) {
+    s32 deltaAngle = BINANG_SUB(this->actor.yawTowardsPlayer, this->actor.world.rot.y);
+    s32 absDeltaAngle = ABS_ALT(deltaAngle);
+
+    this->headRotTarget.y = 0;
+    if (absDeltaAngle < 20000) {
+        this->headRotTarget.y = BINANG_SUB(this->actor.yawTowardsPlayer, this->actor.world.rot.y);
+
+        if (this->headRotTarget.y > 10000) {
```
jpburnett, line 225: Not sure if it is standardized, but I think once upon a time it was said we wanted angles in hex and not decimal?

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+    f32 morphFrames;
+
+    this->animIndex = animIndex;
+
+    morphFrames = -4.0f;
+    if (this->animIndex == EN_DT_ANIMATION_2 || this->animIndex == EN_DT_ANIMATION_5) {
+        morphFrames = 0.0f;
+    }
+
+    this->animEndFrame = (f32)Animation_GetLastFrame(sAnimations[this->animIndex]);
```
jpburnett, line 243: Is the cast here needed for matching?

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+            case EN_DT_CS_FOCUS_TARGET_MAYOR:
+                npcEnMuto->targetActor = &this->actor;
+                npcEnBaisen->targetActor = &this->actor;
+                this->targetActor = &this->actor;
+                break;
+        }
+    }
+}
+
+void EnDt_SetupCutsceneNpcs(EnDt* this, PlayState* play) {
```
jpburnett, line 294: This function looks more like a `EnDt_FindMutoOrBaisen`

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+            Message_StartTextbox(play, this->actor.textId, &this->actor);
+            Player_SetCsActionWithHaltedActors(play, &this->actor, PLAYER_CSACTION_WAIT);
+
+            actionFn = EnDt_UpdateMeetingCutscene;
+            this->npcState = EN_DT_NPC_STATE_IDLE;
+            this->actionFunc = actionFn;
+            return;
+        }
+    } else if (CHECK_WEEKEVENTREG(WEEKEVENTREG_63_80)) {
+        this->textIdIndex = 23; // "You take refuge too"
```
jpburnett, line 331: Remove direct quotes to the dialog since that's an asset.

src/overlays/actors/ovl_En_Dt/z_en_dt.h
```diff
@@ -2,16 +2,42 @@
...
+    /* 0x24A */ s16 blinkTimer;
+    /* 0x24C */ s16 eyeTexIndex;
+    /* 0x24E */ // compiler padding
+    /* 0x250 */ f32 animEndFrame;
+    /* 0x254 */ s16 npcState;
+    /* 0x256 */ s16 textIdIndex;
+    /* 0x258 */ s16 csIds[EN_DT_CUTSCENE_COUNT];
+    /* 0x26E */ s16 csIdIndex;
+    /* 0x270 */ s16 cutsceneState;
+    /* 0x272 */ // compiler padding
```
jpburnett, line 32: you could add an `UNK_TYPE1 unk272` // struct padding comment?

src/overlays/actors/ovl_En_Dt/z_en_dt.c
jpburnett: Looking at this function, I think this is the function to handle blinking.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
-#endif
+static AnimationHeader* sAnimations[EN_DT_ANIMATION_MAX] = { &object_dt_Anim_00112C, &object_dt_Anim_0005A4,
+                                                             &object_dt_Anim_000854, &object_dt_Anim_000DA8,
+                                                             &object_dt_Anim_000BE0, &object_dt_Anim_00B500 };
+
+static u8 sAnimationModes[EN_DT_ANIMATION_MAX] = { ANIMMODE_LOOP, ANIMMODE_LOOP, ANIMMODE_ONCE,
+                                                   ANIMMODE_LOOP, ANIMMODE_LOOP, ANIMMODE_ONCE };
+
+// clang-format off
+static s32 sVisualStateTable[] = {
```
jpburnett, line 171: I think this table are the Eye Textures and needs to be adjusted to be the right format.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+static u16 sTextIds[] = { 0x2ABD, 0x2ABE, 0x2ABF, 0x2AC0, 0x2ABE, 0x2AC1, 0x2AC2, 0x2AC3, 0x2AC4, 0x2ABB,
+                          0x2ABC, 0x2AC6, 0x2AC7, 0x2AC8, 0x2AC9, 0x2ACA, 0x2ACB, 0x2ACC, 0x2ACD, 0x2ACE,
+                          0x2ACF, 0x2AD0, 0x2AD1, 0x2AC5, 0x2AD3, 0x2AD4, 0x2AD2, 0x0000 };
+
+// The code uses EnDt::textIdIndex as an index into this lookup table
+static s32 sCutsceneFocusTargetTable[] = {
+    /* 00 */ EN_DT_CS_FOCUS_TARGET_BAISEN, /* [2ABD] "Most of the townsfolk already sheltered" */
+    /* 01 */ EN_DT_CS_FOCUS_TARGET_MAYOR,  /* [2ABE] "Ahh...hmm...well..." */
+    /* 02 */ EN_DT_CS_FOCUS_TARGET_MUTO,   /* [2ABF] "You cowards! The moon won't fall" */
+    /* 03 */ EN_DT_CS_FOCUS_TARGET_MUTO,   /* [2AC0] "Isn't that right Mayor?" */
```
isghj5, line 107: These aren't exact quotes right? Current project policy is to avoid one-to-one text recreations in the code (you don't have to throw them away entirely we just don't want them to perfectly match)

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+            this->visualState = 5;
+            EnDt_UpdateVisualState(this);
+
+            this->actor.textId = sTextIds[this->textIdIndex];
+            Message_StartTextbox(play, this->actor.textId, &this->actor);
+            Player_SetCsActionWithHaltedActors(play, &this->actor, PLAYER_CSACTION_WAIT);
+
+            actionFn = EnDt_UpdateMeetingCutscene;
+            this->npcState = EN_DT_NPC_STATE_IDLE;
+            this->actionFunc = actionFn;
```
isghj5, line 327: Interesting, feels like this could be eliminated somehow such that actionFn can be removed, since it happens twice in this function, almost want to leave a comment about it

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+    s32 index;
+    s16* pTextData;
+
+    if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_63_80)) {
+        if (this->npcEnMuto != NULL) {
+            if (this->npcEnBaisen != NULL) {
+                npcMuto = (EnMuto*)this->npcEnMuto;
+                npcBaisen = (EnBaisen*)this->npcEnBaisen;
+            }
+        }
```
isghj5, line 417:
```suggestion
        if ((this->npcEnMuto != NULL) && (this->npcEnBaisen != NULL)){
            npcMuto = (EnMuto*)this->npcEnMuto;
            npcBaisen = (EnBaisen*)this->npcEnBaisen;
        }
```
Does it still match if you merge the two conditions?

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+    if (Message_GetState(&play->msgCtx) == TEXT_STATE_EVENT && Message_ShouldAdvance(play)) {
+        Message_CloseTextbox(play);
+        if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_60_40)) {
+            this->textIdIndex = 25; // "The carnival is... on"
+            Message_ContinueTextbox(play, sTextIds[this->textIdIndex]);
+            SET_WEEKEVENTREG(WEEKEVENTREG_60_40);
+            return;
+        }
+        EnDt_SetupFinalNightState(this, play);
+    }
```
isghj5, line 728:
```suggestion
        if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_60_40)) {
            this->textIdIndex = 25; // "The carnival is... on"
            Message_ContinueTextbox(play, sTextIds[this->textIdIndex]);
            SET_WEEKEVENTREG(WEEKEVENTREG_60_40);
        } else {
            EnDt_SetupFinalNightState(this, play);
        }
    }
```
Should still match.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: https://decomp.me/scratch/TuEEh

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: It does not match without the single line formatting. I puzzled over this one for a while until I found similar loop patterns in other actors, such as bombers or fish2.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: Not actually needed, leftover from m2c. Good catch.

src/overlays/actors/ovl_En_Dt/z_en_dt.h
GinNoOokami: Is that a standard? I couldn't find anything in the docs specifying one way or another, and if the goal is to match original source as closely as possible, it seems strange to introduce unused fields that are inserted automatically by the compiler for alignment.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: The shorter ones are exact, while most are paraphrased. I can remove them, but this was useful to me as I deciphered the various cutscene dialog and thought it might be helpful to others as well.

src/overlays/actors/ovl_En_Dt/z_en_dt.h
hensldm: Generally we don't even document it and leave it implicit. We just remove the entry along with the offset comment.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: I thought this was odd as well, and it may very well not match the original. The branching is a bit complex in this one though, and I couldn't find another arrangement of the conditionals to match (without using labels anyway). If you have any suggestions, I'm all ears!

src/overlays/actors/ovl_En_Dt/z_en_dt.h
GinNoOokami: Understood, that makes sense to me.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
hensldm:
```suggestion
            this->npcState = EN_DT_NPC_STATE_IDLE;
            this->actionFunc = EnDt_UpdateMeetingCutscene;
```
Seems to actually not be needed. You can just assign directly without the actionfunc temp.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+
+    this->npcState = EN_DT_NPC_STATE_VIEWING_MEETING;
+    this->actionFunc = EnDt_UpdateMeetingCutscene;
+}
+
+void EnDt_UpdateMeetingCutscene(EnDt* this, PlayState* play) {
+    EnMuto* muto = NULL;
+    EnBaisen* baisen = NULL;
+    s32 index = CsIndexTable.D_80BEB26A[this->csIdIndex];
```
hensldm:
```suggestion
    s32 index = sCsIndexTable[this->csIdIndex + 1];
```
The example of accessing D_80BEB26A I was talking about.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+
+            this->visualState++;
+            if (this->visualState > 5) {
+                this->visualState = 5;
+            }
+        }
+    } else if (this->animEndFrame <= currFrame) {
+        Camera* pCamera;
+        s32 index;
+        s32 currTextId = CsIndexTable.D_80BEB26A[this->csIdIndex];
```
hensldm:
```suggestion
        s32 csIdIndex = CsIndexTable.D_80BEB26A[this->csIdIndex];
```
Doesn't seem to be used as a textId, but like a `csIdIndex` in other functions.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+        Message_CloseTextbox(play);
+        this->actor.textId = sTextIds[this->textIdIndex];
+        Message_StartTextbox(play, this->actor.textId, &this->actor);
+
+        if (this->textIdIndex == 8) { // "Let's not bring my wife into this"
+            Message_BombersNotebookQueueEvent(play, BOMBERS_NOTEBOOK_PERSON_MAYOR_DOTOUR);
+        }
+
+        // Loop through the text id array and set the index if it matches the current text id
+        for (index = 0; index < 24; index++) {
```
hensldm, line 612: I'm wondering if this is worth a bug tag. index should be incrementing by 2 like the other functions. Pretty harmless though.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+                EnDt_UpdateVisualState(this);
+            }
+
+            this->visualState++;
+            if (this->visualState > 5) {
+                this->visualState = 5;
+            }
+        }
+    } else if (this->animEndFrame <= currFrame) {
+        Camera* pCamera;
```
hensldm, line 597:
```suggestion
        Camera* subCam;
```
This would be a pretty standard name for how it is used.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/func_80BE9C74.s")
+    if (this->npcState != 4 && this->npcState != 5 && gSaveContext.save.day == 3 && gSaveContext.save.isNight) {
+        EnDt_SetupFinalNightState(this, play);
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/func_80BE9CE8.s")
+    if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_63_80) &&
+        (gSaveContext.save.day != 3 || (gSaveContext.save.day == 3 && !gSaveContext.save.isNight))) {
+        Audio_PlaySequenceAtPos(SEQ_PLAYER_BGM_SUB, &gSfxDefaultPos, NA_BGM_MAYORS_OFFICE, 1000.0f);
+        Actor_PlaySfx(&this->actor, 0x205DU);
```
hensldm, line 737: If you run (note the trailing `.` is required)
```bash
./tools/sfxconvert.py src/overlays/actors/ovl_En_Dt/z_en_dt.c .
```
it should convert the 0x205D to the proper sfx.

src/overlays/actors/ovl_En_Dt/z_en_dt.h
```diff
@@ -2,16 +2,42 @@
...
+    /* 0x274 */ Actor* npcEnMuto;
+    /* 0x278 */ Actor* npcEnBaisen;
+    /* 0x27C */ Actor* targetActor;
+    /* 0x280 */ s32 visualState;
+    /* 0x284 */ Vec3s headRotValue;
+    /* 0x28A */ Vec3s headRotTarget;
+    /* 0x290 */ s32 meetingFinished;
+    /* 0x294 */ ColliderCylinder collider;
+} EnDt;                                             /* size = 0x2E0 */
```
hensldm, line 39:
```suggestion
} EnDt; // size = 0x2E0
```
Back to as it originally was

src/overlays/actors/ovl_En_Dt/z_en_dt.c
hensldm: With text now extracted and pretty easy to lookup in the repo, this might be something we want to discuss again. I'm not sure.
However please no exact, even for short ones.

hensldm (commented):
Really nice work, especially for a first actor :) .
Here are my initial thoughts

include/z64save.h
```diff
@@ -1108,10 +1108,10 @@ typedef enum {
 #define WEEKEVENTREG_60_01 PACK_WEEKEVENTREG_FLAG(60, 0x01)
 #define WEEKEVENTREG_60_02 PACK_WEEKEVENTREG_FLAG(60, 0x02)
 #define WEEKEVENTREG_60_04 PACK_WEEKEVENTREG_FLAG(60, 0x04)
-#define WEEKEVENTREG_60_08 PACK_WEEKEVENTREG_FLAG(60, 0x08)
-#define WEEKEVENTREG_60_10 PACK_WEEKEVENTREG_FLAG(60, 0x10)
+#define WEEKEVENTREG_60_08 PACK_WEEKEVENTREG_FLAG(60, 0x08) // Seen the mayor's meeting
+#define WEEKEVENTREG_60_10 PACK_WEEKEVENTREG_FLAG(60, 0x10) // Completed mayor meeting event
 #define WEEKEVENTREG_60_20 PACK_WEEKEVENTREG_FLAG(60, 0x20)
-#define WEEKEVENTREG_60_40 PACK_WEEKEVENTREG_FLAG(60, 0x40)
+#define WEEKEVENTREG_60_40 PACK_WEEKEVENTREG_FLAG(60, 0x40) // Talked to mayor on final night
```
AngheloAlf, line 1114: I think we could try naming those flags.
What do you think about this:
- `WEEKEVENTREG_60_08`: `WEEKEVENTREG_ATTENDED_MAYOR_DOTOUR_MEETING`
- `WEEKEVENTREG_60_10`: A bit tricky, maybe `WEEKEVENTREG_ENDED_MAYOR_DOTOUR_MEETING`?
- `WEEKEVENTREG_60_40`: `WEEKEVENTREG_TALKED_MAYOR_DOTOUR_NIGHT_3`

We could possibly drop either the `MAYOR` or `DOTOUR` part, since it is a bit redundant

include/z64save.h
```diff
@@ -1141,7 +1141,7 @@ typedef enum {
 #define WEEKEVENTREG_63_20 PACK_WEEKEVENTREG_FLAG(63, 0x20)
 #define WEEKEVENTREG_63_40 PACK_WEEKEVENTREG_FLAG(63, 0x40)
 
-// showed Couple's Mask to meeting
+// Showed Couple's Mask to meeting
 #define WEEKEVENTREG_63_80 PACK_WEEKEVENTREG_FLAG(63, 0x80)
```
AngheloAlf, line 1145: I guess we could also name this one. Maybe `WEEKEVENTREG_SHOWED_COUPLES_MASK_TO_MAYOR_DOTOUR_MEETING`. Kinda too verbose tho

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: I will change it if that's what's preferred, but in this case I would like to point out that the original devs almost certainly used integers to represent these angles given the nice round numbers. 10000 is a lot easier to grok at a glance than 0x2710. Perhaps it's something the community could reconsider at a later time.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,706 @@ ActorInit En_Dt_InitVars = {
...
+void EnDt_Update(Actor* thisx, PlayState* play) {
+    EnDt* this = THIS;
+    ColliderCylinder* collider;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/EnDt_Destroy.s")
+    SkelAnime_Update(&this->skelAnime);
+    Actor_SetScale(&this->actor, 0.01f);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/func_80BE9C74.s")
+    if (this->npcState != 4 && this->npcState != 5 && gSaveContext.save.day == 3 && gSaveContext.save.isNight) {
```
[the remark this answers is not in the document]
GinNoOokami: Oh, I thought I read somewhere in the style guide that redundant parentheses were discouraged? I can add them if that's not the case.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: Yep, makes perfect sense! I seriously over-complicated this because I didn't realize that indexing into a single dimension array with an offset of one like that would generate two distinct addresses in the final asm, so I thought they had to be separate symbols. Thanks for helping me understand that.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: You're absolutely right. Fixed, and thanks for the suggestion.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
     /**/ EnDt_Draw,
 };
 
-// static ColliderCylinderInit sCylinderInit = {
-static ColliderCylinderInit D_80BEB29C = {
-    { COLTYPE_NONE, AT_NONE, AC_NONE, OC1_ON | OC1_TYPE_ALL, OC2_TYPE_2, COLSHAPE_CYLINDER, },
-    { ELEMTYPE_UNK0, { 0x00000000, 0x00, 0x00 }, { 0xF7CFFFFF, 0x00, 0x00 }, TOUCH_NONE | TOUCH_SFX_NORMAL, BUMP_NONE, OCELEM_ON, },
+static u16 sTextIds[] = { 0x2ABD, 0x2ABE, 0x2ABF, 0x2AC0, 0x2ABE, 0x2AC1, 0x2AC2, 0x2AC3, 0x2AC4, 0x2ABB,
+                          0x2ABC, 0x2AC6, 0x2AC7, 0x2AC8, 0x2AC9, 0x2ACA, 0x2ACB, 0x2ACC, 0x2ACD, 0x2ACE,
+                          0x2ACF, 0x2AD0, 0x2AD1, 0x2AC5, 0x2AD3, 0x2AD4, 0x2AD2, 0x0000 };
```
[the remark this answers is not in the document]
GinNoOokami: I thought I saw that trailing commas would introduce warnings, but it appears I am mistaken.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+static AnimationHeader* sAnimations[EN_DT_ANIMATION_MAX] = { &object_dt_Anim_00112C, &object_dt_Anim_0005A4,
+                                                             &object_dt_Anim_000854, &object_dt_Anim_000DA8,
+                                                             &object_dt_Anim_000BE0, &object_dt_Anim_00B500 };
+
+static u8 sAnimationModes[EN_DT_ANIMATION_MAX] = { ANIMMODE_LOOP, ANIMMODE_LOOP, ANIMMODE_ONCE,
+                                                   ANIMMODE_LOOP, ANIMMODE_LOOP, ANIMMODE_ONCE };
+
+// clang-format off
+static s32 sVisualStateTable[] = {
+    0, EN_DT_ANIMATION_1, EN_DT_EYE_TEXTURE_LOOK_DOWN, true,
```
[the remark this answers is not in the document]
GinNoOokami: I took your initial suggestion of the empty comments for now, but let me mull over the enum idea for a bit.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,714 @@ ActorInit En_Dt_InitVars = {
...
+}
+
+void EnDt_Destroy(Actor* thisx, PlayState* play) {
+    EnDt* this = THIS;
+    Collider_DestroyCylinder(play, &this->collider);
+}
+
+void EnDt_UpdateHeadRotate(EnDt* this) {
+    s32 deltaAngle = BINANG_SUB(this->actor.yawTowardsPlayer, this->actor.world.rot.y);
+    s32 absDeltaAngle = ABS_ALT(deltaAngle);
```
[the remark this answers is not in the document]
GinNoOokami: I thought I had trouble matching without a second var, but I just tried again and it worked.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +95,706 @@ ActorInit En_Dt_InitVars = {
...
+
+typedef struct {
+    s16 D_80BEB268[1];
+    s16 D_80BEB26A[25];
+} CutsceneIndexTable;
+
+static CutsceneIndexTable sCsIndexTable = { { 0x2ABB },
+                                            { 0x0000, 0x2ABD, 0x0000, 0x2AC0, 0x0001, 0x2AC1, 0x0002, 0x2AC4, 0x0003,
+                                              0x2AC6, 0x0004, 0x2AC7, 0x0005, 0x2AC8, 0x0006, 0x2AC9, 0x0007, 0x2ACC,
+                                              0x0008, 0x2ACF, 0x0009, 0x2AD0, 0x000A, 0x0309, 0x0309 } };
```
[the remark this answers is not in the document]
GinNoOokami: Absolutely, thanks to you and hensldm for pointing this out. I was definitely overcomplicating it based on my own misunderstanding of how the data addresses were treated. I didn't realize that `D_80BEB268[index+1]` would translate to `D_80BEB26A` in the final assembly.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: Understood. I've removed the text and replaced them with markers where the individual cutscenes begin and end. Would that be acceptable?

include/z64save.h
GinNoOokami: Yep, I can definitely rename these. Originally I didn't because it looked like most were the default naming pattern, and I wasn't sure if that was preferred. I suppose that could just be because a lot of the code is still undocumented?

GinNoOokami:
> Really nice work, especially for a first actor :) . Here are my initial thoughts

Thanks, but I really feel like I'm standing on the shoulders of giants here. The project was well structured and documented, had great tooling, and had plenty of examples to draw on for inspiration in places where I got stuck or was unsure of naming/formatting. Great work by the community, and I'm excited to be a part of it!

src/overlays/actors/ovl_En_Dt/z_en_dt.c
hensldm: I would agree that this particular actor probably used decimal and round numbers, however they were no where near consistent in this aspect. Some places have nice round decimal numbers while others have nice round hex numbers. We decided to prefer consistency across the whole code base, and so decided to just use Hex for all of them especially with their connection to binary angles.

include/z64save.h
GinNoOokami: How about this? I analyzed their usage and realized that  `WEEKEVENTREG_60_10` is actually only used in one place internally in the actor to handle what I can only imagine is an edge case, so I ended up renaming that to `WEEKEVENTREG_RECEIVED_MAYOR_REWARD`. 

However, `WEEKEVENTREG_63_80` is more widely referenced by several actors to check whether the event has been completed, so I changed that to `WEEKEVENTREG_RESOLVED_MAYOR_MEETING` (and left the comment about the Couple's Mask trigger for clarity). Thoughts?

include/z64save.h
```diff
@@ -1108,10 +1108,10 @@ typedef enum {
 #define WEEKEVENTREG_60_01 PACK_WEEKEVENTREG_FLAG(60, 0x01)
 #define WEEKEVENTREG_60_02 PACK_WEEKEVENTREG_FLAG(60, 0x02)
 #define WEEKEVENTREG_60_04 PACK_WEEKEVENTREG_FLAG(60, 0x04)
-#define WEEKEVENTREG_60_08 PACK_WEEKEVENTREG_FLAG(60, 0x08)
-#define WEEKEVENTREG_60_10 PACK_WEEKEVENTREG_FLAG(60, 0x10)
+#define WEEKEVENTREG_ATTENDED_MAYOR_MEETING PACK_WEEKEVENTREG_FLAG(60, 0x08)
+#define WEEKEVENTREG_RECEIVED_MAYOR_REWARD PACK_WEEKEVENTREG_FLAG(60, 0x10)
 #define WEEKEVENTREG_60_20 PACK_WEEKEVENTREG_FLAG(60, 0x20)
-#define WEEKEVENTREG_60_40 PACK_WEEKEVENTREG_FLAG(60, 0x40)
+#define WEEKEVENTREG_SPOKE_WITH_MAYOR_ON_NIGHT_3 PACK_WEEKEVENTREG_FLAG(60, 0x40)
```
hensldm, line 1114:
```suggestion
#define WEEKEVENTREG_TALKED_MAYOR_NIGHT_3 PACK_WEEKEVENTREG_FLAG(60, 0x40)
```
We have standardized these kinds of week event regs.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+
+            this->visualState++;
+            if (this->visualState > 5) {
+                this->visualState = 5;
+            }
+        }
+    } else if (this->animEndFrame <= currFrame) {
+        Camera* subCam;
+        s32 index;
+        s32 currCsIndex = sStringIdCsIndexTable[this->csIdIndex + 1];
```
hensldm, line 620:
```suggestion
        s32 csIdIndex = sStringIdCsIndexTable[this->csIdIndex + 1];
```
Would prefer this named the same as other functions.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+    Math_SmoothStepToS(&this->headRotValue.y, this->headRotTarget.y, 1, 3000, 0);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/func_80BEA254.s")
+    Actor_SetFocus(&this->actor, 60.0f);
+    Actor_MoveWithGravity(&this->actor);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/func_80BEA394.s")
+    collider = &this->collider;
+    Collider_UpdateCylinder(&this->actor, collider);
+    CollisionCheck_SetOC(play, &play->colChkCtx, &collider->base);
```
hensldm, line 788:
```suggestion
    Collider_UpdateCylinder(&this->actor, &this->collider);
    CollisionCheck_SetOC(play, &play->colChkCtx, &this->collider.base);
```
Is this temp actually needed. If it causes the stack to be too small just add an `s32 pad`. Note this pad would actually be real, and if you curious ping me in discord (anon58) and I can explain it.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+    Actor_SetScale(&this->actor, 0.01f);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/func_80BE9C74.s")
+    if ((this->state != 4) && (this->state != 5) && (gSaveContext.save.day == 3) && gSaveContext.save.isNight) {
+        EnDt_SetupFinalNightState(this, play);
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/ovl_En_Dt/func_80BE9CE8.s")
+    if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_RESOLVED_MAYOR_MEETING) &&
+        (gSaveContext.save.day != 3 || (gSaveContext.save.day == 3 && !gSaveContext.save.isNight))) {
```
hensldm, line 757:
```suggestion
        ((gSaveContext.save.day != 3) || ((gSaveContext.save.day == 3) && !gSaveContext.save.isNight))) {
```

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+            case EN_DT_CS_FOCUS_TARGET_BAISEN:
+                npcEnMuto->targetActor = this->npcEnBaisen;
+                npcEnBaisen->targetActor = this->npcEnBaisen;
+                this->targetActor = this->npcEnBaisen;
+                break;
+            case EN_DT_CS_FOCUS_TARGET_MAYOR:
+                npcEnMuto->targetActor = &this->actor;
+                npcEnBaisen->targetActor = &this->actor;
+                this->targetActor = &this->actor;
+                break;
```
hensldm, line 304:
```suggestion
            case EN_DT_CS_FOCUS_TARGET_MUTO:
                npcEnMuto->targetActor = this->npcEnMuto;
                npcEnBaisen->targetActor = this->npcEnMuto;
                this->targetActor = this->npcEnMuto;
                break;

            case EN_DT_CS_FOCUS_TARGET_BAISEN:
                npcEnMuto->targetActor = this->npcEnBaisen;
                npcEnBaisen->targetActor = this->npcEnBaisen;
                this->targetActor = this->npcEnBaisen;
                break;

            case EN_DT_CS_FOCUS_TARGET_MAYOR:
                npcEnMuto->targetActor = &this->actor;
                npcEnBaisen->targetActor = &this->actor;
                this->targetActor = &this->actor;
                break;

            default:
                break;
```
Newlines please after breaks, makes it easier to tell different switch cases apart. If possible we like to add defaults as well if possible.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+    0, EN_DT_ANIMATION_3, EN_DT_EYE_TEXTURE_SHOCK,     false, //
+    0, EN_DT_ANIMATION_5, EN_DT_EYE_TEXTURE_SHOCK,     false, //
+    0, EN_DT_ANIMATION_4, EN_DT_EYE_TEXTURE_SHOCK,     false  //
+};
+
+static TexturePtr sEyeTextures[] = {
+    gDotourEyeShockTex, gDotourEyeOpenTex, gDotourEyeClosedTex, gDotourEyeLookDownTex, gDotourEyeSquintTex,
+};
+
+static TexturePtr sBrowTextures[] = {
```
hensldm, line 203:
```suggestion
static TexturePtr sEyebrowTextures[] = {
```
Possibly my own opinion but so it matches the names of the textures.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+    ActorShape_Init(&this->actor.shape, 0.0f, ActorShadow_DrawCircle, 19.0f);
+    SkelAnime_InitFlex(play, &this->skelAnime, &object_dt_Skel_00B0CC, &object_dt_Anim_00112C, this->jointTable,
+                       this->morphTable, OBJECT_DT_LIMB_MAX);
+
+    this->actor.targetMode = TARGET_MODE_6;
+    this->npcEnMuto = NULL;
+    this->npcEnBaisen = NULL;
+    Collider_InitAndSetCylinder(play, &this->collider, &this->actor, &sCylinderInit);
+
+    if ((gSaveContext.save.day == 3) && (gSaveContext.save.isNight)) {
```
hensldm, line 222:
```suggestion
    if ((gSaveContext.save.day == 3) && gSaveContext.save.isNight) {
```

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+    }
+
+    Actor_OfferTalk(&this->actor, play, 150.0f);
+}
+
+void EnDt_SetupMeetingCutscene(EnDt* this, PlayState* play) {
+    EnMuto* npcMuto;
+    EnBaisen* npcBaisen;
+    s32 index;
+    s16* pTextData;
```
hensldm, line 421:
```suggestion
```
Looks like this isn't used?

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,725 @@ ActorInit En_Dt_InitVars = {
...
+        }
+        npc = npc->next;
+    }
+
+    EnDt_SetupRegularState(this, play);
+}
+
+void EnDt_SetupRegularState(EnDt* this, PlayState* play) {
+    u32 textId;
+
```
hensldm, line 326:
```suggestion
```
Looks like this isn't used?

hensldm (commented):
2nd pass

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: Agreed. Meant to do this in an earlier pass, but overlooked it. Taken care of now though.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: You're right, it is not actually needed with the padding.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,726 @@ ActorInit En_Dt_InitVars = {
...
+static AnimationHeader* sAnimations[EN_DT_ANIMATION_MAX] = {
+    &object_dt_Anim_00112C, &object_dt_Anim_0005A4, &object_dt_Anim_000854,
+    &object_dt_Anim_000DA8, &object_dt_Anim_000BE0, &object_dt_Anim_00B500,
+};
+
+static u8 sAnimationModes[EN_DT_ANIMATION_MAX] = {
+    ANIMMODE_LOOP, ANIMMODE_LOOP, ANIMMODE_ONCE, ANIMMODE_LOOP, ANIMMODE_LOOP, ANIMMODE_ONCE,
+};
+
+static s32 sVisualStateTable[] = {
```
hensldm, line 190: Might need more opinions, but perhaps something like `sModelInfo`. Not like other ModelInfos I've seen, but it data used to deal with the model.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,726 @@ ActorInit En_Dt_InitVars = {
...
+    if ((this->animIndex == EN_DT_ANIMATION_2) || (this->animIndex == EN_DT_ANIMATION_5)) {
+        morphFrames = 0.0f;
+    }
+
+    this->animEndFrame = Animation_GetLastFrame(sAnimations[this->animIndex]);
+    Animation_Change(&this->skelAnime, sAnimations[this->animIndex], 1.0f, 0.0f, this->animEndFrame,
+                     sAnimationModes[this->animIndex], morphFrames);
+}
+
+void EnDt_UpdateVisualState(EnDt* this) {
```
hensldm, line 271: Similar here, I would recommend something like `EnDt_UpdateModel`

hensldm (approved):
A couple of related doc suggestions, but overall looks good to me :)

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: Admittedly I struggled a bit with this one, since I couldn't find another example of this type of state table in any other actor. As far as I can tell, it's unique to this one. That said, I'm not sure I agree `ModelInfo` is an appropriate name for this. If I were to glance at that, I would expect it to contain properties pertaining to the actor's model and/or mesh. Instead, what this actually controls is the actor's look and behavior at key points during the player's interaction with it, including the animation, eye texture, and blinking behavior. 

That said, I'm not committed to the name, and I realize I'm new to the project and don't have a lot of experience with the idioms in it. If people can agree on a better name, I will certainly change it.

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -28,59 +98,726 @@ ActorInit En_Dt_InitVars = {
...
+    s32 index = this->visualState * 4;
```
AngheloAlf, line 272: From what I see `visualState` could use an enum like the state does. Could you add one too?

src/overlays/actors/ovl_En_Dt/z_en_dt.c
GinNoOokami: I took another swing at these names, let me know if they make sense.

AngheloAlf (approved):
Great work on your first actor!

I have just one final nitpick

src/overlays/actors/ovl_En_Dt/z_en_dt.c
```diff
@@ -179,31 +189,48 @@ static ColliderCylinderInit sCylinderInit = {
...
 
 static AnimationHeader* sAnimations[EN_DT_ANIMATION_MAX] = {
-    &object_dt_Anim_00112C, &object_dt_Anim_0005A4, &object_dt_Anim_000854,
-    &object_dt_Anim_000DA8, &object_dt_Anim_000BE0, &object_dt_Anim_00B500,
+    &object_dt_Anim_00112C, // EN_DT_ANIMATION_WAIT
+    &object_dt_Anim_0005A4, // EN_DT_ANIMATION_DISTRESS
+    &object_dt_Anim_000854, // EN_DT_ANIMATION_SHOCK
+    &object_dt_Anim_000DA8, // EN_DT_ANIMATION_TAP_DESK
+    &object_dt_Anim_000BE0, // EN_DT_ANIMATION_UPRIGHT
+    &object_dt_Anim_00B500, // EN_DT_ANIMATION_SIT_UP
```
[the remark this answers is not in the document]
GinNoOokami: Took a bit to figure out how the asset extractor worked, but it was easier than I thought it would be.

GinNoOokami:
I don't have write access, so feel free to merge at your leisure.

jpburnett (approved):
nice work!

PR #1609, opened 2024-03-09
bg_breakwall: fix great bay storm dl reference

Archez:
This fixes an incorrect DL reference for the great bay storm effect. `func_808B7B54` renders two separate "storm" effects. One of the DL calls goes to `object_posthouse_obj_DL_000A50`, but that is the postmans backpack and hat in the post office. The correct DL for the storm should be `object_kumo30_DL_000A50`.

I presume since both `object_posthouse_obj_DL_000A50` and `object_kumo30_DL_000A50` have the same offset of `0xA50` and objects are loaded into segment 6, that that is why both are considered matching.


hensldm (approved):
Yeah this seems right. Looks like this function should be used with `OBJECT_KUMO30`.

AngheloAlf (approved):
Nice catch, thanks

jpburnett (approved):

PR #1610, opened 2024-03-09
Increase audio seq cmd list size

Archez:
I'm not sure why this list had a size of `0xB0`. `sAudioSeqCmds` is indexed by `sSeqCmdWritePos` and `sSeqCmdReadPos` which are both `u8` values. These values are only ever incremented with `++`, meaning that they eventually will rollover. Setting the size to `0x100` (256) ensures there is no OOB access with the read/write pointers.

This makes it a `sizeof 0x400` which lines up with the next symbol location
```
0x801FFD40 sAudioSeqCmds (+ 0x400)
0x80200140 gActiveSeqs
```
I believe I have updated `variables.txt` correctly. If not, please let me know.


AngheloAlf (approved):
The reasoning makes sense.
`sAudioSecCmds` probably was 0xB0 because some other fake symbols that eventually got cleaned up

Dragorn421 (commented):
can confirm this is technically correct (it's how it is in OoT too)

hensldm (approved):

PR #1615, opened 2024-04-21
Name a lot of the weekeventregs for when the player receives a Heart Piece

tom-overton:
I wanted to start chipping away at all the unnamed weekeventregs, so I figured I'd start by naming most of the ones that are currently being used to track if the player has received a given Heart Piece or not. I managed to get most of the actors that give Heart Pieces done, but there are a few that probably need some more documentation before I can understand which weekeventreg is being used for certain purposes.


isghj5 (approved):
Seeing them this long kinda makes me want to switch the naming convention to `*_HP` but looks good to me.

jpburnett (approved):

AngheloAlf (approved):

PR #1619, opened 2024-04-27
Add original asset names to all `object_s*` files

tom-overton:
A couple of notes:
- MM3D removed `object_sth`, `object_stream`, and all of the `object_sekihi*` objects. The first was just `object_boj` with the animations removed, and the second was the object used for an unused actor. However, while most of the `sekihi` objects were unused grave variations, `object_sekihiz` *was* used for the Song of Soaring pedestal in MM64. How did they handle that pedestal in MM3D? They simply baked it into the Southern Swamp scene.
- The original skulltula token assets from MM64 are actually in MM3D's version of `object_st`. This isn't exactly noteworthy by itself, but it explicitly confirms that the `_model` DList comes before the `_modelT` DList, which is extremely rare.
- Speaking of Skulltulas, the unused animation in `object_sth` is clearly intended to be a death animation, as it looks exactly like the animation of the same name used in `object_st`.
- It's probably possible to name more assets in `object_syoten` once we have some `c_keyframe` docs and some docs for the actor. I only managed to name one thing because it was the only asset that was obvious just by looking at it.

### DIFF
--- a/assets/xml/objects/object_st.xml
+++ b/assets/xml/objects/object_st.xml
@@ -1,6 +1,6 @@
 ﻿<Root>
     <File Name="object_st" Segment="6">
-        <Animation Name="object_st_Anim_000304" Offset="0x304" />
+        <Animation Name="object_st_Anim_000304" Offset="0x304" /> <!-- Original name is "st_matsu" -->
         <Texture Name="object_st_Tex_000320" OutName="tex_000320" Format="i4" Width="16" Height="8" Offset="0x320" />
         <Texture Name="object_st_Tex_000360" OutName="tex_000360" Format="rgba16" Width="16" Height="8" Offset="0x360" />
         <Texture Name="object_st_Tex_000460" OutName="tex_000460" Format="rgba16" Width="8" Height="8" Offset="0x460" />
@@ -33,8 +33,8 @@
         <DList Name="object_st_DL_0048B8" Offset="0x48B8" />
         <DList Name="object_st_DL_004950" Offset="0x4950" />
         <DList Name="object_st_DL_0049E8" Offset="0x49E8" />
-        <DList Name="gSkulltulaTokenDL" Offset="0x4DB0" />
-        <DList Name="gSkulltulaTokenFlameDL" Offset="0x4EB8" />
+        <DList Name="gSkulltulaTokenDL" Offset="0x4DB0" /> <!-- Original name is "gi_sutaru_model" -->
+        <DList Name="gSkulltulaTokenFlameDL" Offset="0x4EB8" /> <!-- Original name is "gi_sutaru_modelT" -->
         <Limb Name="object_st_Standardlimb_0050C8" Type="Standard" EnumName="OBJECT_ST_LIMB_01" Offset="0x50C8" />
         <Limb Name="object_st_Standardlimb_0050D4" Type="Standard" EnumName="OBJECT_ST_LIMB_02" Offset="0x50D4" />
         <Limb Name="object_st_Standardlimb_0050E0" Type="Standard" EnumName="OBJECT_ST_LIMB_03" Offset="0x50E0" />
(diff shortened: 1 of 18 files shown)


AngheloAlf (approved):

assets/xml/objects/object_st.xml
```diff
@@ -33,8 +33,8 @@
         <DList Name="object_st_DL_0048B8" Offset="0x48B8" />
         <DList Name="object_st_DL_004950" Offset="0x4950" />
         <DList Name="object_st_DL_0049E8" Offset="0x49E8" />
-        <DList Name="gSkulltulaTokenDL" Offset="0x4DB0" />
-        <DList Name="gSkulltulaTokenFlameDL" Offset="0x4EB8" />
+        <DList Name="gSkulltulaTokenDL" Offset="0x4DB0" /> <!-- Original name is "gi_sutaru_model" -->
+        <DList Name="gSkulltulaTokenFlameDL" Offset="0x4EB8" /> <!-- Original name is "gi_sutaru_modelT" -->
```
isghj5, line 37: Out of curiosity, Why do only some of these display lists have names? Are those other ones no longer in 3ds?

tom-overton: In this case, it's because this object has a lot of limb displaylists; for something like 99.9% of limb displaylists, we currently have no means of getting the names from MM3D.

To explain a little further, for skeletal meshes, what we tend to get is a single CMB file with all the limbs like so:
image

These CMB files tend to have everything the model would ever need all bundled together; they don't tend to do the thing that Zelda 64 does where certain parts of the model are replaced at runtime with other models. For an extreme example of this, see:
image

For these CMB models, we don't get names of the individual limbs. And since the CMB tends to have everything the model needs, that means things that were individual DLs in MM64 often don't get names either. However, we have pretty strong proof that they *were* named originally. The way we know this is because Grezzo accidentally included some XML data used to author their assets in MM3D. Specifically, there's a file in `zelda2_kaizoku_obj.gar.lzs` (the MM3D equivalent of `object_kaizoku_obj`), they included a file called `Y2_boat_model.cmata`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<NintendoWareIntermediateFile>
	<GraphicsContentCtr Version="1.3.0" Namespace="">
		<EditData>
			<MetaData>
				<Key>MetaData</Key>
				<Create Date="2013-06-25T19:48:56">
					<ToolDescriptions Name="NintendoWare Creative Studio." Version="1.4.5.39788" />
				</Create>
				<Modify Date="2013-06-25T19:54:41">
					<ToolDescriptions Name="NintendoWare Creative Studio." Version="1.4.5.39788" />
				</Modify>
			</MetaData>
		</EditData>
		<Animations>
			<MaterialAnimationData Name="" FrameSize="30" LoopMode="Loop" IsAdditiveAnimation="false">
				<MemberAnimationDataSet>
					<MemberAnimationData>
						<Path>Materials["boat_03"].TextureCoordinators[0].Rotate</Path>
						<FloatAnimation>
							<SegmentsFloatCurve PreRepeatMethod="None" PostRepeatMethod="None" StartFrame="0" EndFrame="30">
								<Segments>
									<LinearFloatSegment>
										<Keys>
											<LinearFloatKey Frame="0" Value="0" />
											<LinearFloatKey Frame="30" Value="-6.283185" />
										</Keys>
									</LinearFloatSegment>
								</Segments>
							</SegmentsFloatCurve>
						</FloatAnimation>
					</MemberAnimationData>
				</MemberAnimationDataSet>
			</MaterialAnimationData>
		</Animations>
	</GraphicsContentCtr>
</NintendoWareIntermediateFile>
```

There's a reference to `boat_03` in here, which is probably the name of one of the bones of the boat. Another thing to note is that MM3D (and *only* MM3D, not OoT3D) stores a 32-bit hash derived from the mesh's name. You can see us discuss this within the Discord here: https://discord.com/channels/688807550715560050/873211000122384415/1148065617958682656

However, you can also see it fizzles out pretty quickly. We simply don't have a lot of information available to try and crack this hash. And even if we did, we have no guarantees that what we get back is even relevant for MM64. We are really confident that for skeletal meshes, Grezzo named the CMBs themselves; we can deduce this from the fact that Grezzo is much better at English than Nintendo was in the late '90s and from the fact that these skeletal meshes tend to be named after the character's name in the final game rather than their codename used during development. Since they renamed the whole mesh, they might have renamed the individual bones too. In the `boat_03` example, this boat wasn't even a skeleton in MM64, so we know that the name `boat_03` had to have been created by Grezzo.

For *non*-skeletal meshes (like the Skulltula Token that started this whole conversation), Grezzo often (but not always) just left the names of those alone. These names follow fairly consistent patterns, and they have leftovers that indicate their MM64 heritage (like how many of them use the same names as actors, or how many vestigial `_modelT` models there are; this is a remnant of how MM64 uses different models for the OPA and XLU buffers). In this case, the original N64 asset is actually in the MM3D object, so we can be pretty sure that this was the original name.

I know that was long, but I hope that answers your question. I went into a lot of detail because I wanted to "demystify" what I'm doing in these sorts of PRs. I don't have any magical or special knowledge; I'm just looking at the data present in the MM3D assets and doing a bit of pattern matching.

isghj5 (approved):

engineer124 (approved):

PR #1620, opened 2024-05-01
Document missing gDPSetEnvColor

garrettjoecox:
Examples of what the ammo count looks like with each of the wallet tiers:
Screenshot 2024-05-01 at 10 07 26 AM
Screenshot 2024-05-01 at 10 07 59 AM
Screenshot 2024-05-01 at 10 08 11 AM

### DIFF
--- a/src/code/z_parameter.c
+++ b/src/code/z_parameter.c
@@ -4055,6 +4055,11 @@ void Interface_DrawAmmoCount(PlayState* play, s16 button, s16 alpha) {
         }
 
         gDPPipeSync(OVERLAY_DISP++);
+        // @bug Missing a gDPSetEnvColor here, which means the ammo count will be drawn with the last env color set.
+        // Once you have the magic meter, this becomes a non issue, as the magic meter will set the color to black,
+        // but prior to that, when certain conditions are met, the color will have last been set by the wallet icon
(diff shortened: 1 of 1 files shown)


src/code/z_parameter.c
```diff
@@ -4055,6 +4055,11 @@ void Interface_DrawAmmoCount(PlayState* play, s16 button, s16 alpha) {
         }
 
         gDPPipeSync(OVERLAY_DISP++);
+        // @bug Missing a gDPSetEnvColor here, which means the ammo count will be drawn with the last env color set.
+        // Once you have the magic meter, this becomes a non issue, as the magic meter will set the color to black,
+        // but prior to that, when certain conditions are met, the color will have last been set by the wallet icon
+        // causing the ammo count to be drawn incorrectly. This is most obvious when you get deku nuts early on, and
+        // the ammo count is drawn with a shade of green.
```
AngheloAlf, line 4062: I had no idea about this bug. Nice find.

Just a question, why document the bug here and not somewhere else?

garrettjoecox: My guess was if placed correctly, it would have been placed anywhere from lines 4057-4085, so I just opted for the top. Anywhere is fine with me if you have a better idea

hensldm (approved):

AngheloAlf (approved):

PR #1621, opened 2024-05-02
CLOSE_DISPS fixes and avoid early return

Archez:
After seeing the bug about early returns in a OPEN/CLOSE_DISPS pairs, I did some poking around and found some areas that I think can be improved.

* Happy Mask Salesman draw had a misplaced `CLOSE_DISPS`. Originally this matched for retail, but I imagine would fail for debug rom matching (when CLOSE_DISPS becomes more than a closing bracket)
* The Monkey face draw had some returns in a switch, but reworking this to move the one `gSPSegment` up into the `default` case means we can swap the `return` for `break` in the other cases and still be matching
* The last one I found was in `eff_blure` where there are returns for null mtx or no elements to draw. ~~Here I opted for `goto close_disps` which sorta matches other places that this is done (`z_eff_spark`, `z_skin`)~~ Instead, per feedback about this being like this even in debug rom, I've added `@bug` tags indicating that the CLOSE_DISPS is skipped.

### DIFF
--- a/src/code/z_eff_blure.c
+++ b/src/code/z_eff_blure.c
@@ -658,7 +658,8 @@ void EffectBlure_DrawSmooth(EffectBlure* this2, GraphicsContext* gfxCtx) {
     OPEN_DISPS(gfxCtx);
 
     if (this->numElements < 2) {
-        goto close_disps;
+        //! @bug Skips CLOSE_DISPS
+        return;
     }
 
     this->elements[0].flags &= ~(EFFECT_BLURE_ELEMENT_FLAG_1 | EFFECT_BLURE_ELEMENT_FLAG_2);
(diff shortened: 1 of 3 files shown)


src/code/z_eff_blure.c
```diff
@@ -705,6 +705,7 @@ void EffectBlure_DrawSmooth(EffectBlure* this2, GraphicsContext* gfxCtx) {
         }
     }
 
+close_disps:
```
hensldm, line 708: Based on OoT debug rom and also MM debug rom, these actually were early returns since this Graph_CloseDisps function doesn't get called in those control flows. This only works since this is retail and doesn't use that function.

hensldm (commented):
Not sure about the eff_blure changes but the other 2 look good.

src/code/z_eff_blure.c
Archez: Ah, this one I wasn't sure about. I should have checked OoT decomp for that 😅 
I can revert that change if that makes more sense.

Would this be a candidate for `// @bug Skips CLOSE_DISPS` like the other places, or just leave as is?

hensldm: I would say add bug tags to both

Archez: Updated

hensldm (approved):

louist103 (approved):

AngheloAlf (approved):

PR #1622, opened 2024-05-03
Document Bug in EnMttag_GetCurrentCheckpoint and Cleanup

inspectredc:
Not sure what the extent of the bug causes but surprised it doesn't cause crashes since from my understanding, the upper limit of the loop should be 3268542464 for sceneExitIndex 18.

### DIFF
--- a/src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
+++ b/src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
@@ -153,7 +153,8 @@ s32 EnMttag_GetCurrentCheckpoint(Actor* actor, PlayState* play, s32* upcomingChe
     checkpointIterator = sStartingCheckpointPerSceneExitIndex[sceneExitIndex];
 
     // Iterates through all possible checkpoints that are associated with this sceneExitIndex.
-    //! @bug - sStartingCheckpointPerSceneExitIndex is indexed out of bounds when sceneExitIndex is 18, due to the `sceneExitIndex + 1` access.
+    //! @bug - sStartingCheckpointPerSceneExitIndex is indexed out of bounds when sceneExitIndex is 18, due to the
+    //! `sceneExitIndex + 1` access.
     do {
         if (Math3D_PointDistToLine2D(
                 actor->world.pos.x, actor->world.pos.z, sCheckpointPositions[checkpointIterator - 1].x,
--- a/src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
+++ b/src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
@@ -155,7 +155,7 @@ s32 EnMttag_GetCurrentCheckpoint(Actor* actor, PlayState* play, s32* upcomingChe
     // Iterates through all possible checkpoints that are associated with this sceneExitIndex.
     //! @bug - sStartingCheckpointPerSceneExitIndex is indexed out of bounds when sceneExitIndex is 18
     do {
-        if (Math3D_PointDistSqToLine2DImpl(
+        if (Math3D_PointDistToLine2D(
                 actor->world.pos.x, actor->world.pos.z, sCheckpointPositions[checkpointIterator - 1].x,
                 sCheckpointPositions[checkpointIterator - 1].z, sCheckpointPositions[checkpointIterator + 1].x,
                 sCheckpointPositions[checkpointIterator + 1].z, &perpendicularPointX, &perpendicularPointZ,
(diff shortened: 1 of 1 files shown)


src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
```diff
@@ -153,25 +153,26 @@ s32 EnMttag_GetCurrentCheckpoint(Actor* actor, PlayState* play, s32* upcomingChe
     checkpointIterator = sStartingCheckpointPerSceneExitIndex[sceneExitIndex];
 
     // Iterates through all possible checkpoints that are associated with this sceneExitIndex.
+    //! @bug - sStartingCheckpointPerSceneExitIndex is indexed out of bounds when sceneExitIndex is 18
```
AngheloAlf, line 156: I think it would be more clear to specify that the OoB is produced because of the `checkpointIterator + 1` accesses

AngheloAlf:
Something to take in consideration that we are trying to get everybody who has contributed / wants to contribute to MM decomp to agree to licensing the repo before merging PRs from people who has not yet agreed to it. It is better explained here: https://github.com/louist103/mm-licensing/issues/1.

src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
inspectredc: There are 2 out of bounds accesses, first sStartingCheckpointPerSceneExitIndex[sceneExitIndex + 1] which causes sCheckpointPositions[checkpointIterator + 1] (and possibly sCheckpointPositions[checkpointIterator - 1]), just wasn't sure the best way to phrase it.

How would this be?
```
    //! @bug - sStartingCheckpointPerSceneExitIndex is indexed out of bounds when sceneExitIndex is 18, causing the loops upper bound to exceed what is expected.
    //         In turn, out of bounds accesses of sCheckpointPositions will occur as checkpointIterator increases past 45.
```

AngheloAlf: Maybe keep it more simple and only talk about `sStartingCheckpointPerSceneExitIndex`, since that's the bigger offender here
```
    //! @bug - sStartingCheckpointPerSceneExitIndex is indexed out of bounds when sceneExitIndex is 18, due to the `sceneExitIndex+1` access.
```

src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
```diff
@@ -153,25 +153,27 @@ s32 EnMttag_GetCurrentCheckpoint(Actor* actor, PlayState* play, s32* upcomingChe
     checkpointIterator = sStartingCheckpointPerSceneExitIndex[sceneExitIndex];
 
     // Iterates through all possible checkpoints that are associated with this sceneExitIndex.
+    //! @bug - sStartingCheckpointPerSceneExitIndex is indexed out of bounds when sceneExitIndex is 18, due to the
```
hensldm, line 156: Hmm, I'm wondering if the bug tag actually makes more sense with the bounds checking above:
`if ((sceneExitIndex < 4) || (sceneExitIndex >= 19))`

inspectredc: I feel like you can argue that it makes sense for the sStartingCheckpointPerSceneExitIndex too (and claim that there should be an extra value of e.g. 46 at the end) so I personally prefer putting it by where the 'illegal' access takes place. But can make this change if you'd prefer

hensldm: I think I prefer with the bounds check, as the whole algorithm probably shouldn't even be running on the final point (kind of curious what happens if you do update the bounds check to `>= 18` but you don't have to check). However it does seem fine where it is anyways.

hensldm (approved):

src/overlays/actors/ovl_En_Mt_tag/z_en_mt_tag.c
AngheloAlf: Bump

AngheloAlf: Actually, I see @hensldm approved this. What's the state of this convo?

hensldm: My preference is to move the bug tag up, but am fine with leaving it where it is so approved it.

inspectredc: Moved it up now 👍🏼 let me know if the wording should change at all

AngheloAlf (approved):

PR #1623, opened 2024-05-05
Consolidate text array

louist103:
IDO will never reorder these arrays but some compilers will depending on optimization settings. This causes a problem when a sub array is indexed or the pointer is incremented more than the capacity of the sub array.


AngheloAlf (approved):
This fixes the issue at `func_80BA4C44`. That function does increment `this->text` by an arbitrary amount, so having individual arrays break the assumption of the data being next to each other.

engineer124 (approved):

hensldm (approved):

PR #1631, opened 2024-05-15
Remove unnecessary stuff from undefined_syms.txt

tom-overton:
The recent Dexihand PR made me notice that there was a lot of stuff in here that didn't need to be in here, so I removed what I could.


AngheloAlf (approved):

hensldm (commented):
Should be good to remove the `ovl_En_Po_Composer` section as well

hensldm (approved):

PR #1632, opened 2024-05-15
Change tooling for better elf correctness

AngheloAlf:
Currently the repository has some oversights regarding some elf quirks, like some asm functions not being marked as a "function" (`STT_FUNC`) on the final elf.

This doesn't really means anything for matching a ROM, but I'm a big fan of correctness for this kind of stuff :sweat_smile: 

This PR also adds some fixes that are tangentially related.

This PR aims to accomplish:
- Make sure every function is marked as a function in the final elf (the symbol type being `STT_FUNC`) and ensure the functions have the size properly set.
  - asm-processor already handles this for GLOBAL_ASM functions in C files. This PR fixes the handwritten asm ones.
  - A remarkable case is `sNonKanjiIndices`, which is an asm data symbol that happens to be in the middle of the `.text` section. This symbol is not marked as a function.
- Mark the `entrypoint` function as the elf's entry.
- Add some linker script asserts.
  - Assert to ensure the `boot` segment is not larger than 1 MiB. If that were the case then IPL3 wouldn't be able to load it completely.
  - Assert to ensure the `framebuffer_hi` has not shifted.
  - Assert to try to check the non-overlay segments are not overflowing the RAM. The check is far from ideal, but I don't know what would be an actually good value.
- Add vram addresses to `makerom` and `boot` segments.
  - `makerom` having a proper address means we could actually use a symbol on the rom header instead of hardcoding the address.
  - I had to move `framebuffer_lo` to be above the `makerom` segment to avoid `dot moved backwards` warnings, since the addresses were not in ascending order.

Shoutouts to @Mr-Wiseguy for letting me know about the original "issue" and telling me where to update some stuff.


tools/asm-processor/.gitrepo
```diff
@@ -6,7 +6,7 @@
 [subrepo]
 	remote = git@github.com:simonlindholm/asm-processor.git
 	branch = main
-	commit = fed1e3ddb9967bd427f4599b1b8ededa296bd9f9
-	parent = 4fa13e4132cf8adb2f07bb34927e62e525c8cc79
+	commit = 92b9ec72f40fde329bd91a83046b6c74bdc00199
+	parent = 1f6e7c5006bcc603714310b9fb891f4fa546c896
 	method = merge
-	cmdver = 0.4.6
+	cmdver = 0.4.3
```
jpburnett, line 12: Any reason for the minor version downgrade?

jpburnett (approved):
Looks good, just a comment on version downgrade.

tools/asm-processor/.gitrepo
AngheloAlf: That's the version of my `git subrepo`. Mine is a bit older than whoever updated asm-processor last time, and I haven't worried to update it since it works either way.
It shouldn't affect anything

undefined_syms.txt
```diff
@@ -457,3 +457,11 @@ D_0E000000 = 0x0E000000;
...
 
 D_0F000000 = 0x0F000000;
+
+ENTRY(entrypoint);
+
+ASSERT ((_bootSegmentRomEnd <= 0x101000), "Error: _bootSegmentRomEnd is larger than 1 MB");
+
+// TODO: This should be checking for an address a lot earlier than this one.
+ASSERT ((_system_heapSegmentStart < 0x80780000), "Error: The game is overflowing the RAM limit for non-overlay segments.");
+ASSERT ((_framebuffer_hiSegmentStart == 0x80780000), "Error: framebuffer_hi shifted?");
```
hensldm, line 467: Instead of putting this in `undefined_syms.txt`, could we perhpas create a linker_scripts folder and add a new linker script file (not quite sure what to call it though)? I would also say move `undefined_syms.txt` to this new folder and rename it to `undefined_syms.ld`.

AngheloAlf: I named the new file `linker_scripts/linker_script_extra.ld`. I'm happy to hear a better name

linker_scripts/linker_script_extra.ld
```diff
@@ -0,0 +1,7 @@
+ENTRY(entrypoint);
```
hensldm, line 1: I'd probably just say `extra.ld` for the name, though perhaps `assert.ld`, `entry.ld`, or `check.ld`?

hensldm (commented):
A couple more thoughts

hensldm (approved):

PR #1633, opened 2024-05-15
Fix a fake match on `EnSGoro`

AngheloAlf:
This code
```c
if ((s32)this != -0x190) {
```

Was actually a `NULL` check for the member at offset `0x190`, in this case `collider`. The only way for `&this->collider` could be `NULL` is that `this` has value `-0x190`.

Shoutouts to @cadmic for figuring this out for an AF scratch.
Convo: https://discord.com/channels/688807550715560050/1139317895491506197/1240420736095490239


hensldm (approved):
Pretty sure `&this->collider` can't even be `NULL` anyways since is the address of a struct member.

tom-overton (approved):
I checked MM3D and it turns out that armcc compiled this in exactly the same way as IDO did, funnily enough

PR #1637, opened 2024-05-18
Msgevent OK

hensldm:
Thanks @Thar0 for most the decompilation work and documentation.


include/z64player.h
```diff
@@ -1388,7 +1388,7 @@ void func_80123E90(struct PlayState* play, Actor* actor);
 s32 func_80123F2C(struct PlayState* play, s32 ammo);
 bool Player_IsBurningStickInRange(struct PlayState* play, Vec3f* pos, f32 xzRange, f32 yRange);
 u8 Player_GetStrength(void);
-u8 Player_GetMask(struct PlayState* play);
+s32 Player_GetMask(struct PlayState* play);
```
AngheloAlf, line 1391: ohhh, surprising. How did we not figure this out earlier?

Since this matches as an `s32`, would it match using the enum as the return type too?
```suggestion
PlayerMask Player_GetMask(struct PlayState* play);
```

src/code/z_msgevent.c
```diff
@@ -1,105 +1,1229 @@
...
+
+/**
+ * Pause the message system
+ *
+ * Command structure:
+ *  0:(u8)  cmd
+ * Command size: 1
+ */
+s32 MsgEvent_Pause(Actor* actor, PlayState* play, u8** scriptPtr, MsgEventCallback callback, s32* endScript) {
+    play->msgCtx.msgMode = 0x44;
```
AngheloAlf, line 472:
```suggestion
    play->msgCtx.msgMode = MSGMODE_PAUSED;
```

src/code/z_msgevent.c
```diff
@@ -1,105 +1,1229 @@
...
+ *  1:(u16) event_inf flag
+ *  3:(s16) skip
+ * Command size: 5
+ */
+s32 MsgEvent_BranchOnEventInf(Actor* actor, PlayState* play, u8** scriptPtr, MsgEventCallback callback,
+                              s32* endScript) {
+    u8* script = *scriptPtr;
+    s16 skip = MSCRIPT_GET_16(script, 3);
+
+    if (MSCRIPT_GET_8(script, 2) & gSaveContext.eventInf[MSCRIPT_GET_8(script, 1)]) {
```
AngheloAlf, line 567: I suppose there's no way to use the `CHECK_EVENTINF` macro here, right?

src/code/z_msgevent.c
```diff
@@ -1,105 +1,1229 @@
...
+ *
+ * Command structure:
+ *  0:(u8)  cmd
+ *  1:(u16) event_inf flag
+ * Command size: 3
+ */
+s32 MsgEvent_SetEventInf(Actor* actor, PlayState* play, u8** scriptPtr, MsgEventCallback callback, s32* endScript) {
+    u8* script = *scriptPtr;
+
+    gSaveContext.eventInf[MSCRIPT_GET_8(script, 1)] |= MSCRIPT_GET_8(script, 2);
```
AngheloAlf, line 584: Same here with `SET_EVENTINF`

src/code/z_msgevent.c
```diff
@@ -1,105 +1,1229 @@
...
+ *
+ * Command structure:
+ *  0:(u8)  cmd
+ *  1:(u16) event_inf flag
+ * Command size: 3
+ */
+s32 MsgEvent_UnsetEventInf(Actor* actor, PlayState* play, u8** scriptPtr, MsgEventCallback callback, s32* endScript) {
+    u8* script = *scriptPtr;
+
+    gSaveContext.eventInf[MSCRIPT_GET_8(script, 1)] &= MSCRIPT_GET_8(script, 2) ^ 0xFF;
```
AngheloAlf, line 599: ohhh. That's a new way of unsetting eventInf.

Would it match like this? This is how the `CLEAR_EVENTINF` macro looks like
```suggestion
    gSaveContext.eventInf[MSCRIPT_GET_8(script, 1)] &= (u8)~MSCRIPT_GET_8(script, 2);
```

src/code/z_msgevent.c
```diff
@@ -1,105 +1,1229 @@
...
+ *  3:(s16) skip
+ * Command size: 9
+ */
+s32 MsgEvent_BranchOnSongObtained(Actor* actor, PlayState* play, u8** scriptPtr, MsgEventCallback callback,
+                                  s32* endScript) {
+    u8* script = *scriptPtr;
+    u16 song = MSCRIPT_GET_16(script, 1);
+    s16 skip = MSCRIPT_GET_16(script, 3);
+
+    if (CHECK_QUEST_ITEM(QUEST_SONG_SONATA + song)) {
```
AngheloAlf, line 662: I wonder if we should add markers for this inside the enum in the future

src/code/z_msgevent.c
```diff
@@ -1,105 +1,1229 @@
...
+        *scriptPtr += skip;
+    }
+    return MSCRIPT_CONTINUE;
+}
+
+#define MSCRIPT_GET_TIME(hour, minute, dest, temp) \
+    (temp) = (hour)*60.0f;                         \
+    (temp) += (minute);                            \
+    (dest) = (temp) * (0x10000 / 60 / 24.0f);      \
+    (dest) = ((dest)-0x10000 / 360 * 90)
```
AngheloAlf, line 693: This is the same macro as `SCHEDULE_CALC_TIME`. Maybe we could merge them?

src/code/z_msgevent.c
```diff
@@ -1,105 +1,1229 @@
...
+
+#define MSCRIPT_TIME_NOW ((gSaveContext.save.time) - 0x10000 / 360 * 90)
```
AngheloAlf, line 695: This is `SCHEDULE_TIME_NOW`

src/code/z_msgevent.c
AngheloAlf, line 1: The schedule system uses structs for each command instead of macros for accesses. Would that match here? I feel like that approach is somewhat cleaner. It would also avoid the need for those `_SIZE` defines

tools/disasm/functions.txt
```diff
@@ -2009,58 +2009,58 @@
...
-    0x8010BC7C:("func_8010BC7C",),
-    0x8010BD48:("func_8010BD48",),
-    0x8010BD90:("func_8010BD90",),
-    0x8010BDDC:("func_8010BDDC",),
-    0x8010BE78:("func_8010BE78",),
-    0x8010BEBC:("func_8010BEBC",),
-    0x8010BEF0:("func_8010BEF0",),
-    0x8010BF24:("func_8010BF24",),
-    0x8010BF58:("func_8010BF58",),
+    0x8010A760:("MsgEvent_Cmd00",),
```
AngheloAlf, line 2012: Would be nice to keep those updated. Specially since those are being touched anyways

src/code/z_msgevent.c
hensldm: I don't think so. The main difference here is how the message script stores the eventinf vs how we do in the defines. The script stores it as index and the raw flag, while for the define we store it as index and the flag position. So get the flag the macro expects we would need to put it back together with something like `(MSCRIPT_GET_8(script, 1) << 4) | BIT_FLAG_TO_SHIFT(MSCRIPT_GET_8(script, 2))` which doesn't match. Note this is fine for weekevent regs as we store them based on index and raw flag as well.

src/code/z_msgevent.c
hensldm: Unfortunately it does not. Get a `nor` instead of an `xori`.

src/code/z_msgevent.c
hensldm: Merged the time macros for schedule and msgevent, placed them in macros.h (at least for now), and prefixed them with `SCRIPT` since they seem to deal with scripting languages.

src/code/z_msgevent.c
AngheloAlf: Oh, I see.

In the future we could consider changing how the evenInf macro works then. Thanks for looking at this

src/code/z_msgevent.c
AngheloAlf: Thanks. A bit annoying to have to put them in macros.h, but it should work for now

src/code/z_msgevent.c
hensldm: Initial test says yes, though will take me a bit to try them all.

AngheloAlf (approved):

jpburnett (approved):
Nice work!

PR #1638, opened 2024-05-22
More General Cleanup

hensldm:
1. Find all linker scripts instead of having to hardcode each one in the Makefile
2. A few instances where we could use COLPOLY_GET_NORMAL
3. `M_PI` -> `F_PI` (https://discord.com/channels/688807550715560050/740615251011174401/1242638083367895101)
4. Add libc to auto includes. OoT already has this and works nicer with ultralib (tested my idea from previous clean PR, but it didn't work :( )
5. ~~`FLT_MAX` -> `MAXFLOAT` (matches what OoT did)~~
6. Remove `SHT_MINV` and rename `SHT_MAX` -> `SHRT_MAX`


include/libc/math.h
```diff
@@ -1,10 +1,11 @@
...
 
-#define M_PI 3.14159265358979323846f
-#define M_SQRT2 1.41421356237309504880f
-#define M_SQRT1_2 0.70710678118654752440f	/* 1/sqrt(2) */
-#define FLT_MAX 340282346638528859811704183484516925440.0f
+#define F_PI 3.14159265358979323846f
+#define F_SQRT2 1.41421356237309504880f
+#define F_SQRT1_2 0.70710678118654752440f /* 1/sqrt(2) */
+
+#define MAXFLOAT 3.40282347e+38f
```
AngheloAlf, line 8: I think we already had the `FLT_MAX` vs `MAXFLOAT` discussion before.
iirc we decided to use `FLT_MAX` because it is an standard macro while `MAXFLOAT` isn't
https://en.cppreference.com/w/c/types/limits

src/code/z_camera.c
```diff
@@ -1536,9 +1536,9 @@ s32 Camera_CalcAtForFriendlyLockOn(Camera* camera, VecGeo* eyeAtDir, Vec3f* targ
             deltaY = focalActorPosRot->pos.y - *yPosOffset;
             temp_f0_6 = Math_FAtan2F(deltaY, OLib_Vec3fDistXZ(at, &camera->eye));
 
-            if (temp_f0_6 > 0.34906584f) { // (M_PI / 9)
+            if (temp_f0_6 > 0.34906584f) { // (F_PI / 9)
```
AngheloAlf, line 1539: Huh, I wonder if `(f32)(M_PI / 9)` (the double) would match here, probably not tho

include/libc/math.h
hensldm: Ahh, didn't know we had already discussed it in MM already. I can revert it back later.

hensldm: Out of curiosity, should I also remove `SHT_MINV` and also rename the Max to `SHRT_MAX` to match the page you linked?

src/code/z_camera.c
hensldm: It actually does :)

include/libc/math.h
AngheloAlf: Actually I'm trying to dig for the discussion and I can't find it. We may have discussed something else related to the C standard.

I kinda prefer using the C standard names, but I'm open to renaming those constants

AngheloAlf (approved):

jpburnett (approved):

PR #1641, opened 2024-06-01
EnOkuta (Octorok and its projectiles) OK and documented

tom-overton:
Thanks to @Synray for the original decomp of this actor, and thanks to @hensldm for helping me clean it up and offering me suggestions.

While documenting this, I discovered there's a totally unused blue Octorok variant with some weird behavior. It's invincible to every attack the player can use, can't be frozen, and can even dive underwater to track the player if the player dives as well:

https://github.com/zeldaret/mm/assets/12245827/29a6f6c4-20c3-4353-9db3-ee53985a858a


src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +54,1086 @@ ActorInit En_Okuta_InitVars = {
...
+        this->drawDmgEffAlpha = 0.0f;
+        Actor_SpawnIceEffects(play, &this->actor, this->bodyPartsPos, EN_OKUTA_BODYPART_MAX, 2, 0.3f, 0.2f);
+        this->collider.base.colType = COLTYPE_HIT0;
+    }
+}
+
+void EnOkuta_SpawnBubbles(EnOkuta* this, PlayState* play) {
+    s32 i;
+
+    for (i = 0; i != 10; i++) {
```
AngheloAlf, line 236: Huh, this has to be `!=`  instead of `<`?

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +54,1086 @@ ActorInit En_Okuta_InitVars = {
...
+    if (this->timer < 11) {
+        Actor_SetScale(&this->actor, (((this->timer - 5) * 0.04f) + 0.8f) * 0.01f);
+        return;
+    }
+
+    if (Math_StepToF(&this->actor.scale.x, 0.0f, 0.0005f)) {
+        SoundSource_PlaySfxAtFixedWorldPos(play, &this->actor.world.pos, 30, NA_SE_EN_COMMON_WATER_MID);
+        Item_DropCollectibleRandom(play, &this->actor, &this->actor.world.pos, 0xA0);
+
+        for (i = 0; i != 20; i++) {
```
AngheloAlf, line 637: Same here

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +54,1086 @@ ActorInit En_Okuta_InitVars = {
...
+    if ((this->actor.bgCheckFlags & BGCHECKFLAG_WALL) || (this->actor.bgCheckFlags & BGCHECKFLAG_GROUND) ||
+        (this->actor.bgCheckFlags & BGCHECKFLAG_CEILING) || (this->collider.base.atFlags & AT_HIT) ||
+        (this->collider.base.acFlags & AC_HIT) || (this->collider.base.ocFlags1 & OC1_HIT) ||
+        (this->actor.floorHeight == BGCHECK_Y_MIN)) {
+        if (player->currentShield == PLAYER_SHIELD_HEROS_SHIELD) {
+            if ((this->collider.base.atFlags & AT_HIT) && (this->collider.base.atFlags & AT_TYPE_ENEMY)) {
+                if (this->collider.base.atFlags & AT_BOUNCED) {
+                    this->collider.base.atFlags &= ~(AT_HIT | AT_BOUNCED | AT_TYPE_ENEMY);
+                    this->collider.base.atFlags |= AT_TYPE_PLAYER;
+                    this->collider.info.toucher.dmgFlags = 0x400000;
```
AngheloAlf, line 776:
```suggestion
                    this->collider.info.toucher.dmgFlags = DMG_THROWN_OBJECT;
```

AngheloAlf (approved):
Great job!

Just a few nits

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +54,1086 @@ ActorInit En_Okuta_InitVars = {
...
+    this->actor.draw = EnOkuta_Draw;
+    this->actor.shape.rot.y = this->actor.yawTowardsPlayer;
+    this->actor.flags |= ACTOR_FLAG_TARGETABLE;
+    Animation_PlayOnce(&this->skelAnime, &gOctorokAppearAnim);
+    EnOkuta_SpawnBubbles(this, play);
+    this->actionFunc = EnOkuta_Appear;
+}
+
+/**
+ * Plays the appear animation to completion while scaling the Octorok to its full size, then transitions the the idle
```
jpburnett, line 347:
```suggestion
 * Plays the appear animation to completion while scaling the Octorok to its full size, then transitions the idle
```
Duplicate word

jpburnett (commented):
Great work, just a typo nit

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
tom-overton: No, it's just something I missed. I fixed it up.

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +55,1086 @@ ActorInit En_Okuta_InitVars = {
...
+        EnOkuta_SpawnSplash(this, play);
+    }
+}
+
+void EnOkuta_SetupFloat(EnOkuta* this) {
+    Animation_PlayLoop(&this->skelAnime, &gOctorokFloatAnim);
+
+    // `EnOkuta_Float` uses the `timer` variable in an unconventional way. It is used to track how many times the
+    // Octorok's float animation needs to loop before the Octorok can start shooting at the player. Note that the unused
+    // blue Octorok variant can ignore this variable and start shooting the player anyway if the player gets too close.
```
mzxrules, line 419:
```suggestion
    // `EnOkuta_Float` uses `timer` to store the number of times the float animation loops before firing at the player.
```

Can cut the fat here to get to the point quicker. Perhaps add a comment where `timer` is defined as well. 
Comments about Blue Octorok's behavior are better off left in `EnOkuta_Float` where the described behavior is actually implemented, since both types of Octoroks are using the value of `timer` in the same way.

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +55,1086 @@ ActorInit En_Okuta_InitVars = {
...
+    }
+}
+
+void EnOkuta_SetupShoot(EnOkuta* this, PlayState* play) {
+    Animation_PlayOnce(&this->skelAnime, &gOctorokShootAnim);
+
+    if (this->actionFunc != EnOkuta_Shoot) {
+        // `EnOkuta_Shoot` uses the `timer` variable in an unconventional way. It is used to track how many projectiles
+        // are remaining in the current "volley". The Octorok will shoot at the player repeatedly, decrementing the
+        // `timer` variable after each shot, until it reaches 0.
```
mzxrules, line 472:
```suggestion
    // `EnOkuta_Float` uses `timer` to store the number of projectiles are remaining in the current "volley". The Octorok will shoot at the player
    // repeatedly, decrementing the `timer` variable after each shot, until it reaches 0.
```

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +54,1086 @@ ActorInit En_Okuta_InitVars = {
...
+        this->actor.flags &= ~ACTOR_FLAG_TARGETABLE;
+        this->actor.flags |= ACTOR_FLAG_10;
+        this->actor.child->csId = this->actor.csId;
+        this->actionFunc = EnOkuta_FrozenInIceBlock;
+        return;
+    }
+
+    // The only way to reach this code is if our attempt to spawn the ObjIceblock actor failed for some reason. In the
+    // final game, this basically never happens during normal play, so the below code is effectively unused and never
+    // seen by most players, even if it *can* be run under certain extreme circumstances.
```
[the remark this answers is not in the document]
AngheloAlf: I think the current explanation is better than just saying this code is unreachable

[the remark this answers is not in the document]
mzxrules: But the code already speaks for itself aside from just the code appearing to be reachable due to it not being in a conditional code block.

With even the slightest bit of familiarity to the code base, it should be obvious to the reader that Actor_SpawnAsChild normally returns a non-null value (like so much so to the point where it's just assumed to always return an actor instance in some spots). It should also be obvious to the reader that Actor_SpawnAsChild has a rare potential failure case because we're checking if the return value is null.

We don't need a comment to explain that Actor_SpawnAsChild can fail in "extreme circumstances", because the code is already doing this for us. There's no important information to be gained from reading the comment that cannot just as easily be summed up by saying "This block is usually unreachable." I don't think we need a sentence to explain the meaning and consequences of unreachable code, which is what all of it is.

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +55,1086 @@ ActorInit En_Okuta_InitVars = {
...
+void EnOkuta_Destroy(Actor* thisx, PlayState* play) {
+    EnOkuta* this = THIS;
+    Collider_DestroyCylinder(play, &this->collider);
+}
+
+/**
+ * Covers the Octorok's body parts with ice and turns the Octorok red. Note that this is different from being frozen
+ * inside a block of ice; see `EnOkuta_FrozenInIceBlock` for how that is handled.
+ */
+void EnOkuta_Freeze(EnOkuta* this) {
```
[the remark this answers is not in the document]
hensldm: Most (all?) actors that have a similar function already name them `Freeze` so seems better to keep the pattern.

hensldm (approved):
Looks good to me. Don't really have an opinion on the comment suggestions.

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
```diff
@@ -40,153 +54,1086 @@ ActorInit En_Okuta_InitVars = {
...
+        CollisionCheck_SetAC(play, &play->colChkCtx, &this->collider.base);
+        CollisionCheck_SetOC(play, &play->colChkCtx, &this->collider.base);
+    }
+}
+
+/**
+ * Returns true if the snout scale should be updated, false otherwise. The snout scale is returned via the scale
+ * parameter.
+ */
+s32 EnOkuta_GetSnoutScale(EnOkuta* this, f32 curFrame, Vec3f* scale) {
```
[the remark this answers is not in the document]
tom-overton: I didn't really see the need for `TryGet`, and there really isn't anything else in the codebase currently named `TryGet` like that, so I just kept it as `Get`. I did update the comment, though.

src/overlays/actors/ovl_En_Okuta/z_en_okuta.c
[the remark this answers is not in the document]
tom-overton: Not everyone is familiar with the codebase, and not everyone who even looks at `Actor_SpawnAsChild` will be aware of its various failure methods. For example, this can fail if `OBJECT_ICE_BLOCK` isn't in the scene, which is something that is reasonably likely to come up in a modding context. However, I really have zero motivation to argue this further, so I just deleted the comment entirely.

PR #1424, opened 2023-10-06
obj_b_heart

Wimmerrad:
Unused Heart Piece object

PR #1642, opened 2024-06-02
obj_b_heart documentation continuation

jpburnett:
This change continues the work done by obj_b_heart #1424 

I agree to my contributions in this PR to be under a CC0 license.

### DIFF
--- a/assets/xml/objects/object_b_heart.xml
+++ b/assets/xml/objects/object_b_heart.xml
@@ -1,9 +1,9 @@
-﻿<Root>
-    <!-- Unused Heart Piece -->
+<Root>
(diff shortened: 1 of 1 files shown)


AngheloAlf (approved):
Not the biggest fan of naming the stuff "unused", but I can't think on anything better, so it should be fine

jpburnett:
Instead of "unused", there could be "alt"? Or type "A" and type "B"?

hensldm:
Personally I still prefer `unused` over the `alt` or type `B` (mainly because there is not really a type `A`, just the ones in gameplay keep I think)

PR #1646, opened 2024-06-19
Add a `z64play.h` header

AngheloAlf:
This PR moves the `PlayState` struct and `z_play.c` functions, variables and macros to the new `z64play.h` header.

This leaves `z64.h` basically empty, just including other headers. I decided to do this in this way because it will make easier to untangle the "include everything" headers by it not being tied to a specific system (Play in this case).

I also made some other cleanups.
- I changed the Play functions that take `GameState*` to take an `PlayState*` argument.
- Move `FrameAdvance_IsEnabled` to `z64frameadvance.h`
  - This is needed to avoid including `z64play.h` or `z64.h` in `z_collision_check.c` (explained below)
- Change all the `PlayState` references in `z_collision_check.c` to use `struct PlayState` instead.
  - This file was using a member of PlayState only to pass the `state` member to `FrameAdvance_IsEnabled` (which was upcasting anyways), this was silly so I changed it to avoid requiring to know the members of this struct, allowing to not include the struct at all here.
  - This allows us to reduce the included headers on this file significantly.
  - Hopefully this would make the bss less volatile.
- Reduced the headers needed in `z_lib.c`
  - Inspired by the ichain discussion in OoT (zeldaret/oot#1956)
  - Also moved `Actor_ProcessInitChain` to `ichain.h`. (I swear we already did this, but oh well)


include/z64play.h
```diff
@@ -0,0 +1,197 @@
...
+#include "z64skybox.h"
+#include "z64sound_source.h"
+#include "z64transition.h"
+#include "z64view.h"
+
+// TODO: Move to its own header?
+typedef struct SequenceContext {
+    /* 0x0 */ u8 seqId;
+    /* 0x1 */ u8 ambienceId;
+} SequenceContext; // size = 0x2
```
hensldm, line 36: Personally I say just remove this struct. It isn't actually needed (and currently isn't even used everywhere it could/should be if we were going to keep it).

src/code/z_play.c
```diff
@@ -1941,19 +1936,15 @@ void func_80169F78(GameState* thisx) {
...
-
+s32 Play_CamIsNotFixed(PlayState* this) {
     return this->roomCtx.curRoom.roomShape->base.type != ROOM_SHAPE_TYPE_IMAGE;
 }
 
-s32 FrameAdvance_IsEnabled(GameState* thisx) {
-    PlayState* this = (PlayState*)thisx;
-
+s32 FrameAdvance_IsEnabled(PlayState* this) {
     return this->frameAdvCtx.enabled != 0;
```
hensldm, line 1948:
```suggestion
    return this->frameAdvCtx.enabled != false;
```
Should we do this?

include/z64play.h
AngheloAlf: I kinda like those members being grouped into a struct, but as you say there's zero code proof that this struct actually exists besides a leftovers string on OoT. I guess it could go

hensldm: Ah I missed that. Should probably keep in then and eventually in its own header yeah.

hensldm (commented):
1 last thing. Since macros.h doesn't really include much anymore, I think we should just include it in sys_math3d.c instead of redefining `ARRAY_COUNT`.

include/z64play.h
mzxrules: don't think it really needs it's own header

hensldm (approved):

include/z64play.h
AngheloAlf: I ended up renaming this struct to `SceneSequences` given Fig's suggestion on https://github.com/zeldaret/oot/pull/1966#discussion_r1651701491

fig02: sorry I missed this conversation before starting the other one over in oot. What oot debug string is being referred to? 
I also kinda dont feel like it necessarily has to be a struct, but am also okay with the SceneSequences suggestion I made

fig02: nevermind I found it `BGM設定game_play->sound_info.BGM=[%d]`

AngheloAlf: This printf on z_kankyo.c

```c

        PRINTF("\n\n\nBGM設定game_play->sound_info.BGM=[%d] old_bgm=[%d]\n\n", play->sceneSequences.seqId,
               ((void)0, gSaveContext.seqId));
```
https://github.com/zeldaret/oot/blob/2ce4742bf69faebb16ea1710c4d4e7902b1be0f1/src/code/z_kankyo.c#L2081

fig02: right okay. well, `SceneSequences` or something like that is def an improvement over "sound info". if others are fine with it

hensldm: `SceneSequences` seems fine, and yeah after further thought it doesn't need its own header.

fig02: Yeah, really it is something internal to play and how it is organizing its internal data extracted from the scene. So keeping it in play just makes sense

fig02 (approved):

PR #1647, opened 2024-06-20
Cleanup the point lighting gbi implementation

Thar0:
This PR updates the gbi.h point lighting implementation to be more complete.
- Wraps all of the point lighting features in a macro to clearly identify it.
- Adds comments to the `PointLight_t` structure describing the fields.
- Adds `G_LIGHTING_POSITIONAL_H`, for completeness.
- Cleans up the `gdSPDefLights*` macros
- Adds `gdSPDefPointLights` macros, for completeness.

### DIFF
--- a/include/PR/gbi.h
+++ b/include/PR/gbi.h
@@ -16,6 +16,9 @@
 
 #define F3DEX_GBI_2
 
+/* F3DEX2 with Point Lighting */
+#define F3DEX_GBI_PL
+
 #ifdef    F3DEX_GBI_2
 # ifndef  F3DEX_GBI
 #  define F3DEX_GBI
@@ -288,7 +291,9 @@
 #define G_TEXTURE_GEN		0x00040000
 #define G_TEXTURE_GEN_LINEAR	0x00080000
 #define G_LOD			0x00100000	/* NOT IMPLEMENTED */
-#define G_LIGHTING_POSITIONAL   0x00400000
+#ifdef F3DEX_GBI_PL
+# define G_LIGHTING_POSITIONAL  0x00400000
+#endif
 #if	(defined(F3DEX_GBI)||defined(F3DLP_GBI))
 # define G_CLIPPING		0x00800000
 #else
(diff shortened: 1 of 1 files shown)


include/PR/gbi.h
```diff
@@ -1277,15 +1285,16 @@ typedef struct {
...
-  unsigned char colc[3];
-  unsigned char kl;
-  short pos[3];
-  unsigned char kq;
+  unsigned char col[3];		/* point light value (rgba) */
+  unsigned char kc;         /* constant attenuation (> 0 indicates point light) */
+  unsigned char colc[3];	/* copy of point light value (rgba) */
+  unsigned char kl;         /* linear attenuation */
+  short pos[3];             /* world-space position of light */
+  unsigned char kq;         /* quadratic attenuation */
```
AngheloAlf, line 1295: tabs :despair:

AngheloAlf (approved):

include/PR/gbi.h
```diff
@@ -16,6 +16,9 @@
 
 #define F3DEX_GBI_2
 
+/* F3DEX2 with Point Lighting */
+#define F3DEX_GBI_PL
```
AngheloAlf, line 20: Actually I wonder if we should pass this define (and `F3DEX_GBI_2`) in the Makefile instead of just defining it here

hensldm: Would we want to auto define `F3DEX_GBI_2` if `F3DEX_GBI_PL` is passed, so we only had to pass `F3DEX_GBI_PL` on the makefile? Something like

```c
#ifdef F3DEX_GBI_PL
#define F3DEX_GBI_2
#endif
```

include/PR/gbi.h
Thar0: The entire file needs reformatting so I didn't bother with it yet. I plan to follow-up with a formatting pass on gbi.h that makes it the same as OoT.

include/PR/gbi.h
Thar0: I think this is a good idea and is how it's supposed to work, but I'd like to do this separately. As mentioned above, I have plans for a formatting pass on this file and it seems like a good time to do this then.

hensldm (approved):

PR #1648, opened 2024-06-20
Document garbage data on the on-screen clock.

AngheloAlf:
Documents the on screen clock bug where the current hour shows garbage data for a few frames when changing from 23 to 24.

This was originally discussed on #1487 . This PR takes a different approach by preserving the garbage data on the `sThreeDayClockHourTextures` array.

I also wanted to rename the `AVOID_UB` define to `PRESERVE_UB`, in an attempt to make clear that we want to _preserve_ the issues caused by the undefined behavior in the code instead of trying to patch them. If we don't agree on this change I can revert it.


hensldm:
Personally I think I would still prefer `AVOID_UB` though perhaps there is someplace we could document the define, and how it is meant to preserve what the UB does in IDO.

mzxrules:
not particularly fond of the idea of "preserving" broken behavior; i think it will just cause far more trouble than it's worth in the long run when you think of all the different potential OOBs reads you could preserve. Perhaps a better approach would be a new directive which generates a warning/error when a file has known UB that we can't fix with AVOID_UB

AngheloAlf:
> not particularly fond of the idea of "preserving" broken behavior; i think it will just cause far more trouble than it's worth in the long run when you think of all the different potential OOBs reads you could preserve. Perhaps a better approach would be a new directive which generates a warning/error when a file has known UB that we can't fix with AVOID_UB

The idea is to preserve behavior in a way that would still work when built with non IDO compilers, like modern GCC. This kind of change has been done before in OoT too. This issue is being addressed here because it literally caused issues when changing the compiler.

tom-overton (approved):

hensldm:
Looks like a format is needed :)

PR #1649, opened 2024-06-21
Introduce a bunch of small new headers

AngheloAlf:
- New headers:
  - `gfxalloc.h`
  - ~~`graph.h`~~ Removed, I moved those symbols to `gfx.h`
  - `z64elf_message.h`
    - I'm not totally convinced by this header. I'm open to better suggestions.
  - `z64inventory.h`
  - `z64olib.h`
- Move `Gfx_DrawDListOpa` and `Gfx_DrawDListXlu` to `gfx.h`
- Move `Regs_InitData` to `regs.h`

`z_olib.c` needs to include `z64actor.h` because of `PosRot`, which feels kinda dumb, we may want to move that struct somewhere else, since it isn't actor specific.


hensldm:
>z_olib.c needs to include z64actor.h because of PosRot, which feels kinda dumb, we may want to move that struct somewhere else, since it isn't actor specific.

Perhaps  just in `z64math.h` since it just combines two of the structs from there already?

hensldm (approved):
Though, looks like sys_math3d bss is having issues again :( .

tools/check_reordering.py
```diff
@@ -18,7 +18,7 @@ def mapPathToSource(origName: Path) -> Path:
     # Try to map built path to the source path
     parts = origName.parts
     if parts[0] == "build":
-        parts = parts[1:]
+        parts = parts[2:]
```
hensldm, line 21: ? . What is this change for?

src/overlays/actors/ovl_En_Kusa/z_en_kusa.c
```diff
@@ -4,7 +4,7 @@
  * Description: Grass / Bush
  */
 
-#include "prevent_bss_reordering.h"
+//#include "prevent_bss_reordering.h"
```
hensldm, line 7:
```suggestion
```

tools/check_reordering.py
AngheloAlf: The output path of this script is meant to point to the C file, so you can ctrl-click it on vscode and open the corresponding file. It currently it has the version as prefix so you can't do that right now.

```
Symbol Name,Build Address,Build File,Expected Address,Expected File,Difference,GOOD/BAD/MISSING
D_80936CDA,80936AD0,n64-us/src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,80936CDA,n64-us/src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,-20A,BAD
D_80936CDC,80936AD2,n64-us/src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,80936CDC,n64-us/src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,-20A,BAD
```

Notice the `n64-us/src/overlays/actors/ovl_En_Kusa/z_en_kusa.c`

By changing this line the output doesn't have the version as prefix on the path anymore

```
Symbol Name,Build Address,Build File,Expected Address,Expected File,Difference,GOOD/BAD/MISSING
D_80936CDA,80936AD0,src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,80936CDA,src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,-20A,BAD
D_80936CDC,80936AD2,src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,80936CDC,src/overlays/actors/ovl_En_Kusa/z_en_kusa.c,-20A,BAD
```

src/code/sys_math3d.c
```diff
@@ -29,6 +29,9 @@ struct Dummy224 { int x; };
 struct Dummy225 { int x; };
 struct Dummy226 { int x; };
 struct Dummy227 { int x; };
+struct Dummy228 { int x; };
+typedef int Dummy229;
```
Thar0, line 33: The comment above here
> bumps the bss index by 56

Should now be 59

src/code/sys_math3d.c
```diff
@@ -29,6 +29,9 @@ struct Dummy224 { int x; };
...
+
 // clang-format on
 
 // Headers are currently valued at 63 mod 256 (./tools/calc_bss.sh <headers>)
```
Thar0, line 37:
```suggestion
// Headers are currently valued at 60 mod 256 (./tools/calc_bss.sh <headers>)
```
(maybe double-check this quickly by running calc_bss.sh on the headers, but this is what it should be to get the same bss index)

Thar0 (commented):
In the gbi.h PR I added some comments to `sys_math3d.c` while resolving bss reordering. They should probably be updated to reflect the new values or removed if we don't want to maintain it like this. I think they're useful though

tools/check_reordering.py
hensldm: Could you add a comment that it is stripping off build and version from the path. Perhaps something like `# Strip off build and version`?

include/z64olib.h
```diff
@@ -0,0 +1,42 @@
+#ifndef Z64OLIB_H
+#define Z64OLIB_H
+
+#include "ultra64.h"
+
+#include "z64math.h"
+
+struct PosRot;
```
hensldm, line 8: Shouldn't need this anymore right?

include/z64olib.h
```diff
@@ -0,0 +1,42 @@
...
+VecSph OLib_Vec3fToVecSph(Vec3f* vec);
+VecGeo OLib_Vec3fToVecGeo(Vec3f* vec);
+VecSph OLib_Vec3fDiffToVecSph(Vec3f* a, Vec3f* b);
+VecGeo OLib_Vec3fDiffToVecGeo(Vec3f* a, Vec3f* b);
+Vec3f OLib_AddVecGeoToVec3f(Vec3f* a, VecGeo* geo);
+Vec3f OLib_Vec3fDiffRad(Vec3f* a, Vec3f* b);
+Vec3f OLib_Vec3fDiffDegF(Vec3f* a, Vec3f* b);
+Vec3s OLib_Vec3fDiffBinAng(Vec3f* a, Vec3f* b);
+void OLib_Vec3fDiff(struct PosRot* a, Vec3f* b, Vec3f* dest, s16 mode);
+void OLib_Vec3fAdd(struct PosRot* a, Vec3f* b, Vec3f* dest, s16 mode);
```
hensldm, line 40: Shouldn't need the `struct`s anymore.

AngheloAlf:
As a small fyi I ended up removing  `graph.h` since only actually two symbols were being used outside of `graph.c`
I moved those two declarations to `gfx.h`

hensldm:
> As a small fyi I ended up removing `graph.h` since only actually two symbols were being used outside of `graph.c` I moved those two declarations to `gfx.h`

Looking at this again, would we also want to just move the 3 `gfxalloc` functions there as well?

jpburnett (approved):

PR #1650, opened 2024-06-21
Format gbi.h, move ucode defines to makefile

Thar0:
This PR syncs gbi.h with OoT, except for the addition of the custom macro `gDPSetdither` and moving the ucode defines into the makefile rather than setting them at the top of the header.

Also corrects usage of `gDPSetDither`.

This will exactly match OoT's gbi.h once https://github.com/zeldaret/oot/pull/1968 is merged

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,12 +122,14 @@ else
   RM_MDEBUG = @:
 endif
 
+UCODE_DEFS := -DF3DEX_GBI_2 -DF3DEX_GBI_PL
+
 # Check code syntax with host compiler
 CC_CHECK_WARNINGS := -Wall -Wextra -Wno-unknown-pragmas -Wno-unused-parameter -Wno-unused-variable -Wno-missing-braces -Wno-unused-but-set-variable -Wno-unused-label -Wno-sign-compare -Wno-tautological-compare
 # Have CC_CHECK pretend to be a MIPS compiler
 MIPS_BUILTIN_DEFS := -DMIPSEB -D_MIPS_FPSET=16 -D_MIPS_ISA=2 -D_ABIO32=1 -D_MIPS_SIM=_ABIO32 -D_MIPS_SZINT=32 -D_MIPS_SZPTR=32
 # The -MMD flags additionaly creates a .d file with the same name as the .o file.
-CC_CHECK_FLAGS   := -MMD -MP -fno-builtin -fsyntax-only -funsigned-char -fdiagnostics-color -std=gnu89 -m32 -DNON_MATCHING -DAVOID_UB -DCC_CHECK=1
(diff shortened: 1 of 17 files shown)


src/code/z_lights.c
```diff
@@ -428,7 +428,7 @@ void Lights_DrawGlow(PlayState* play) {
 
         gfx = Gfx_SetupDL65_NoCD(POLY_XLU_DISP);
 
-        gDPSetDither(gfx++, G_CD_NOISE);
+        gDPSetDither(gfx++, G_AD_PATTERN | G_CD_NOISE);
```
AngheloAlf, line 431: Would be nice to have an automated way to check for missing arguments on gfx macros

include/PR/gbi.h
```diff
@@ -1,92 +1,87 @@
...
 
-#ifndef PR_GBI_H
-#define PR_GBI_H
+#ifndef ULTRA64_GBI_H
+#define ULTRA64_GBI_H
 
-#include "ultratypes.h"
-
-/* Not original to libultra: Point Lights, DW, and Mtx intPart/FracPart union */
+#ifdef GBI_DOWHILE
```
AngheloAlf, line 6: Oh, nice idea

AngheloAlf (approved):

Makefile
```diff
@@ -122,12 +122,14 @@ else
   RM_MDEBUG = @:
 endif
 
+GBI_DEFINES := -DF3DEX_GBI_2 -DF3DEX_GBI_PL -DGBI_DOWHILE
```
hensldm, line 125: Instead of having `GBI_DEFINES` up here, could you put it with with the other defines (just under `COMMON_DEFINES`) and then add `$(GBI_DEFINES)` to both `AS_DEFINES` and `C_DEFINES`?

tools/calc_bss.sh
```diff
@@ -34,7 +34,8 @@ echo "char measurement;" >> $TEMPC
 
 $(pwd)/tools/ido_recomp/linux/7.1/cc -G 0 -non_shared \
     -Xfullwarn -Xcpluscomm -O2 -g3 -Xcpluscomm -mips2 \
-    -I $(pwd)/ -I $(pwd)/include/ -I $(pwd)/src/ -I $(pwd)/assets/ -I $(pwd)/build/n64-us/ \
+    -D_LANGUAGE_C -DF3DEX_GBI_2 -DF3DEX_GBI_PL \
```
hensldm, line 37:
```suggestion
    -D_LANGUAGE_C -DF3DEX_GBI_2 -DF3DEX_GBI_PL -DGBI_DOWHILE \
```
Shouldn't affect bss calcs, but seems better imo to match as much as possible.

Makefile
Thar0: It's more natural up here since `CC_CHECK_FLAGS` needs it too

hensldm: the `CC_CHECK` compilation also uses `C_DEFINES`, so if you add it there, `CC_CHECK_FLAGS` doesn't need it directly.

hensldm (approved):

PR #1656, opened 2024-07-04
Document object_bob

hensldm:
Documents object_bob which is the object file for an unused NPC leftover from OoT.
Also gives better prefixes to object_cne and object_bba.


isghj5:
This object is left over but no actor file uses these assets? Do you have a screenshot of applying a valid enhy animation to the skeleton or a screenshot from oot depicting this creature?
Screenshot_2024-07-04_09-14-49

AngheloAlf (approved):

isghj5 (approved):

PR #1658, opened 2024-07-10
More General Cleanup

hensldm:
Main thing identify more unaccounted asset data (mainly empty TexAnims)
Try to standardized on large blocks of zeros as `EmptyBlob`s
Remove some more entries from undefined_syms.ld
Add an `s` prefix to static data in collision check
Standardize on `IsUnderwater` functions in actors.


AngheloAlf (approved):

isghj5:
All of these textanims are empty? if you tried to apply them nothing would happen?

Kinda feel like that deserves a different tag than just `unused`. There are plenty of unused assets that can still be aplied like unused animations or unused models, but this is empty?

hensldm:
> All of these textanims are empty? if you tried to apply them nothing would happen?

Not all but most are indeed empty, and yes nothing would happen. Based on the ZTextureAnimation.cpp header comment in ZAPD, the system just ignores these empty texanims if they were to be used.

> Kinda feel like that deserves a different tag than just `unused`. There are plenty of unused assets that can still be aplied like unused animations or unused models, but this is empty?

What do you mean a different tag for unused? For the most part I think using the word `Empty` can signify that the texture animation is empty and so doesn't do anything.

isghj5 (approved):
My bad, for some reason earlier I misread all of them were being named to unused from empty, not the other way around.

tom-overton (approved):

PR #1661, opened 2024-07-16
z_en_dinofos documented

TrulyNaN:
Documentation of an already matching actor. Object file was already decompiled. The timers were converted to unions to clarify their multiple roles. Two static variables for drawing the knife collider were left unnamed.


src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1423,15 +1453,15 @@ s32 EnDinofos_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3
     EnDinofos* this = THIS;
 
     if (limbIndex == DINOLFOS_LIMB_HEAD) {
-        rot->y -= this->unk_28E;
+        rot->y -= this->headRotY;
     }
 
     return 0;
```
hensldm, line 1459:
```suggestion
    return false;
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1459,30 +1489,30 @@ static s8 sLimbToBodyParts[DINOLFOS_LIMB_MAX] = {
...
-        Collider_SetQuadVertices(&this->colliderQuad, &sp5C, &sp68, &sp74, &sp80);
-        if (this->colliderQuad.base.atFlags & AT_ON) {
-            EffectBlure_AddVertex(Effect_GetByIndex(this->unk_2A0), &sp68, &sp5C);
+    if ((limbIndex == DINOLFOS_LIMB_RIGHT_HAND) && (this->timer2 != this->timer1) &&
+        ((this->actionFunc == EnDinofos_SlashFromGround) || (this->actionFunc == EnDinofos_JumpSlash))) {
+        Math_Vec3f_Copy(&vtx2, &this->knifeCollider.dim.quad[0]);
+        Math_Vec3f_Copy(&vtx3, &this->knifeCollider.dim.quad[1]);
+        Matrix_MultVec3f(&D_8089E38C, &vtx1);
+        Matrix_MultVec3f(&D_8089E398, &vtx0);
+        Collider_SetQuadVertices(&this->knifeCollider, &vtx0, &vtx1, &vtx2, &vtx3);
```
hensldm, line 1511:
```suggestion
        Math_Vec3f_Copy(&knifeBaseQuadOldPos, &this->knifeCollider.dim.quad[0]);
        Math_Vec3f_Copy(&knifeTipQuadOldPos, &this->knifeCollider.dim.quad[1]);
        Matrix_MultVec3f(&sKnifeTipQuadOffset, &knifeTipQuadPos);
        Matrix_MultVec3f(&sKnifeBaseQuadOffset, &knifeBaseQuadPos);
        Collider_SetQuadVertices(&this->knifeCollider, &knifeBaseQuadPos, &knifeTipQuadPos, &knifeBaseQuadOldPos, &knifeTipQuadOldPos);
```
Some naming suggestions. Note I confirmed which quad vertex corresponded to the tip by using `func_800B10C0` to spawn dust at what you called `vtx1` and it spawned tip of the knife when they swung.

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1230,47 +1260,47 @@ void func_8089D318(EnDinofos* this, PlayState* play) {
...
+        Math_ScaledStepToS(&this->headRotY, Math_CosF(M_PIf) * 0x2C00, 0x2C00 / 20);
+    } else if (this->actionFunc == EnDinofos_BreatheFire) {
+        this->headRotY = Math_CosF(this->headTimer * (M_PIf / 20)) * 0x2C00;
     } else if (!Play_InCsMode(play)) {
-        temp_v0_2 = this->unk_28E + this->actor.shape.rot.y;
-        temp_v0_2 = BINANG_SUB(this->actor.yawTowardsPlayer, temp_v0_2);
-        temp_v0_2 = CLAMP(temp_v0_2, -0x300, 0x300);
-        this->unk_28E += temp_v0_2;
-        this->unk_28E = CLAMP(this->unk_28E, -0x2C00, 0x2C00);
+        headRotYDelta = this->headRotY + this->actor.shape.rot.y;
```
hensldm, line 1282:
```suggestion
        headRotYDiff = this->headRotY + this->actor.shape.rot.y;
```
I believe diff is much more common through out the codebase

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1068,153 +1095,156 @@ void func_8089CB10(EnDinofos* this, PlayState* play) {
...
     this->actor.world.rot.y = this->actor.shape.rot.y;
-    this->actionFunc = func_8089D018;
+    this->actionFunc = EnDinofos_Die;
 }
 
-void func_8089D018(EnDinofos* this, PlayState* play) {
+void EnDinofos_Die(EnDinofos* this, PlayState* play) {
     if (SkelAnime_Update(&this->skelAnime)) {
-        s32 temp_v0 = this->unk_288 - 10;
+        s32 temp_v0 = this->envColorAlpha - 10;
```
hensldm, line 1184:
```suggestion
        s32 envColorAlpha = this->envColorAlpha - 10;
```
Name suggestion

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1068,153 +1095,156 @@ void func_8089CB10(EnDinofos* this, PlayState* play) {
...
 
     SkelAnime_Update(&this->skelAnime);
     Math_ScaledStepToS(&this->actor.shape.rot.y, this->actor.yawTowardsPlayer, 0x800);
-    temp_f20 = Math_SinS(temp_s3);
-    temp_f22 = Math_CosS(temp_s3);
+    sin = Math_SinS(fireRotY);
+    cos = Math_CosS(fireRotY);
 
-    temp_s0 = 10 - this->unk_290;
+    temp_s0 = 10 - this->attackTimer;
```
hensldm, line 1117: Same name suggestion as below `dimAlphaStep`

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1068,153 +1095,156 @@ void func_8089CB10(EnDinofos* this, PlayState* play) {
...
+    accel.y = Rand_CenteredFloat(0.6f) + 1.4f;
+    accel.z = 0.9f * cos;
     Actor_PlaySfx_Flagged(&this->actor, NA_SE_EN_DODO_J_FIRE - SFX_FLAG);
-    EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &sp88, &sp7C, 30, 22, 255 - (temp_s0 * 20), 20,
-                        3, 8);
+    EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &velocity, &accel, 30, 22,
+                        255 - (temp_s0 * 20), 20, 3, 8);
 
+    // This lets the fire balls go away from Dinolfos' jaw one after the other.
     for (end = 6, i = 3; i > 0; i--) {
```
hensldm, line 1131:
```suggestion
    end = 6;
    for (i = 3; i > 0; i--) {
```
Does this work as well?

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -803,261 +825,266 @@ void func_8089BD28(EnDinofos* this, PlayState* play) {
...
     } else if (Animation_OnFrame(&this->skelAnime, 11.0f)) {
         Actor_PlaySfx(&this->actor, NA_SE_EN_RIZA_ATTACK);
     } else if (this->skelAnime.curFrame < 7.0f) {
-        func_8089AE00(this, play);
+        EnDinofos_Dodge(this, play);
     }
 }
 
-void func_8089C690(EnDinofos* this) {
+void EnDinofos_SetupRecoilFromBlockedSlash(EnDinofos* this) {
```
hensldm, line 975:
```suggestion
void EnDinofos_SetupRecoil(EnDinofos* this) {
```
There is only one type of recoil so not sure we really need to add `FromBlockedSlash` here.

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -803,261 +825,266 @@ void func_8089BD28(EnDinofos* this, PlayState* play) {
...
+            EnDinofos_SetupStartBreatheFire(this);
         } else {
-            func_8089A9B0(this, play);
+            EnDinofos_ChooseAction(this, play);
         }
     }
 }
 
-void func_8089C4F8(EnDinofos* this) {
+void EnDinofos_SetupSlashFromGround(EnDinofos* this) {
```
hensldm, line 944:
```suggestion
void EnDinofos_SetupSlash(EnDinofos* this) {
```
We can already distinguish from the other type of slash becuase it has `Jump`, so again not sure we really need the `FromGround` here.

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -803,261 +825,266 @@ void func_8089BD28(EnDinofos* this, PlayState* play) {
...
         } else {
-            func_8089C164(this);
+            EnDinofos_SetupJumpWithoutSlashing(this);
         }
     }
 }
 
-void func_8089C164(EnDinofos* this) {
-    if (this->unk_290 == 2) {
+void EnDinofos_SetupJumpWithoutSlashing(EnDinofos* this) {
```
hensldm, line 859:
```suggestion
void EnDinofos_SetupJump(EnDinofos* this) {
```
I'll admit I'm kind of on the fence here, but the `JumpSlash` is already distinguished by `Slash` so not sure we need to add `WithoutSlashing` here.

hensldm (commented):
Some thoughts and suggestions

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
TrulyNaN: That is because of global renaming. When I use `EnDinofos_Jump`, it clashes with `EnDinofos_JumpExtraStuff` functions.

hensldm: Yeah, but we shouldn't need to worry about global renaming anymore right, now that it has a proper name?

TrulyNaN: True! Will update that. Thanks.

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -295,84 +306,86 @@ void EnDinofos_Init(Actor* thisx, PlayState* play) {
         sTexturesDesegmented = true;
     }
 
-    this->unk_288 = 255;
-    for (i = 6; i < ARRAY_COUNT(this->colliderJntSphElement); i++) {
-        dim = &this->colliderJntSph.elements[i].dim;
+    this->envColorAlpha = 255;
+
+    for (i = 6; i < ARRAY_COUNT(this->bodyAndFireColliderElements); i++) {
```
hensldm, line 311:
```suggestion
    for (i = DINOFOS_COLLIDER_FIRE_START_INDEX; i < ARRAY_COUNT(this->bodyAndFireColliderElements); i++) {
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -385,143 +398,149 @@ void func_8089ABF4(EnDinofos* this, PlayState* play) {
...
+ * After this function is run, Dinolfos resumes being vulnerable to goron pound if it had dodged one.
+ */
+void EnDinofos_EnableBumperCollision(EnDinofos* this) {
     s32 i;
 
-    if (this->unk_28B != 0) {
-        for (i = 0; i < ARRAY_COUNT(this->colliderJntSphElement) - 3; i++) {
-            this->colliderJntSph.elements[i].info.bumper.dmgFlags |= 0x400;
+    if (this->isDodgingGoronPound) {
+        for (i = 0; i < ARRAY_COUNT(this->bodyAndFireColliderElements) - 3; i++) {
```
hensldm, line 429:
```suggestion
        for (i = 0; i < DINOFOS_COLLIDER_FIRE_START_INDEX; i++) {
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -385,143 +398,149 @@ void func_8089ABF4(EnDinofos* this, PlayState* play) {
...
     }
 
     if ((GET_PLAYER_FORM == PLAYER_FORM_GORON) && (player->actor.velocity.y < -5.0f) && (player->av1.actionVar1 == 1) &&
-        (this->unk_28B == 0)) {
-        this->unk_28B = 1;
-        for (i = 0; i < ARRAY_COUNT(this->colliderJntSphElement) - 3; i++) {
-            this->colliderJntSph.elements[i].info.bumper.dmgFlags &= ~0x400;
+        (!this->isDodgingGoronPound)) {
+        this->isDodgingGoronPound = true;
+        for (i = 0; i < ARRAY_COUNT(this->bodyAndFireColliderElements) - 3; i++) {
```
hensldm, line 484:
```suggestion
        for (i = 0; i < DINOFOS_COLLIDER_FIRE_START_INDEX; i++) {
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -803,262 +826,267 @@ void func_8089BD28(EnDinofos* this, PlayState* play) {
...
         this->actor.world.rot.y = this->actor.shape.rot.y;
     } else {
         this->actor.speed = 3.0f;
     }
-    this->actionFunc = func_8089C724;
+    this->actionFunc = EnDinofos_RecoilFromBlockedSlash;
 }
 
-void func_8089C724(EnDinofos* this, PlayState* play) {
+void EnDinofos_RecoilFromBlockedSlash(EnDinofos* this, PlayState* play) {
```
hensldm, line 988:
```suggestion
void EnDinofos_Recoil(EnDinofos* this, PlayState* play) {
```
This should also be updated :) .

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1068,153 +1096,157 @@ void func_8089CB10(EnDinofos* this, PlayState* play) {
...
     Actor_PlaySfx_Flagged(&this->actor, NA_SE_EN_DODO_J_FIRE - SFX_FLAG);
-    EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &sp88, &sp7C, 30, 22, 255 - (temp_s0 * 20), 20,
-                        3, 8);
+    EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &velocity, &accel, 30, 22,
+                        255 - (dimAlphaStep * 20), 20, 3, 8);
 
-    for (end = 6, i = 3; i > 0; i--) {
-        if (this->unk_290 < (20 + -(i * 2))) {
+    // This lets the fire balls go away from Dinolfos' jaw one after the other.
+    end = 6;
```
hensldm, line 1132:
```suggestion
    end = DINOFOS_COLLIDER_FIRE_START_INDEX;
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1068,153 +1096,157 @@ void func_8089CB10(EnDinofos* this, PlayState* play) {
...
+    for (i = 3; i > 0; i--) {
+        if (this->attackTimer < (20 + -(i * 2))) {
             end = i + 6;
```
hensldm, line 1135:
```suggestion
            end = i + DINOFOS_COLLIDER_FIRE_START_INDEX;
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1068,153 +1096,157 @@ void func_8089CB10(EnDinofos* this, PlayState* play) {
...
             break;
         }
     }
 
+    // The fire balls sweep in front of Dinolfos.
     for (i = 6; i < end; i++) {
```
hensldm, line 1141:
```suggestion
    for (i = DINOFOS_COLLIDER_FIRE_START_INDEX; i < end; i++) {
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1068,153 +1096,157 @@ void func_8089CB10(EnDinofos* this, PlayState* play) {
...
-    EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &sp88, &sp7C, 30, 22, 255 - (temp_s0 * 20), 20,
-                        3, 8);
+    EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &velocity, &accel, 30, 22,
+                        255 - (dimAlphaStep * 20), 20, 3, 8);
 
-    for (end = 6, i = 3; i > 0; i--) {
-        if (this->unk_290 < (20 + -(i * 2))) {
+    // This lets the fire balls go away from Dinolfos' jaw one after the other.
+    end = 6;
+    for (i = 3; i > 0; i--) {
```
hensldm, line 1133:
```suggestion
    for (i = ARRAY_COUNT(bodyAndFireColliderElements) - DINOFOS_COLLIDER_FIRE_START_INDEX; i > 0; i--) {
```
I think this makes sense?

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1491,19 +1521,19 @@ void EnDinofos_PostLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3s*
...
-            sp4C.z -= sp48->mf[3][2];
-            EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &sp4C, &gZeroVec3f, 30, 22,
+            Matrix_MultVec3f(&sBaseCsFireVelocity, &fireVelocity);
+            fireVelocity.x -= matrix->mf[3][0];
+            fireVelocity.y -= matrix->mf[3][1];
+            fireVelocity.z -= matrix->mf[3][2];
+            EffectSsDFire_Spawn(play, &this->bodyPartsPos[DINOFOS_BODYPART_JAW], &fireVelocity, &gZeroVec3f, 30, 22,
                                 255 - (sp58 * 20), 20, 3, 8);
-            this->unk_292 = this->unk_290;
+            this->timer2 = this->timer1;
```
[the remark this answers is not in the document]
hensldm: Did you mean to mark this resolved with no changes?

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.h
```diff
@@ -24,36 +24,73 @@ typedef enum EnDinofosBodyPart {
...
+        s16 stunTimer;
+        s16 jumpType;
+        s16 isJumpingBackward;
+    };
+    /* 0x292 */ union { // second multi-use timer
+        s16 timer2;
+        s16 sidestepTimer;
+        s16 walkTimer;
+        s16 idleTimer;
+    };
```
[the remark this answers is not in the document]
hensldm: Even if we don't do anything in this PR can you leave this unresolved so other people can see it and possibly comment.

hensldm (commented):
Mostly comments in usage around the new `DINOFOS_COLLIDER_FIRE_START_INDEX`. Also I think you might have missed some of my previous review as Github marked them hidden.

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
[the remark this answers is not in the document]
TrulyNaN: Nope, my bad. Probably got interrupted by something else. Should be in soon.

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
TrulyNaN: What? This one I'm sure I did. Very odd. Will be in next commit.

hensldm (approved):

AngheloAlf (approved):
Great work!
Just one very small nit, but I'll approve either way

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1459,30 +1490,31 @@ static s8 sLimbToBodyParts[DINOLFOS_LIMB_MAX] = {
...
-    if ((limbIndex == DINOLFOS_LIMB_RIGHT_HAND) && (this->unk_292 != this->unk_290) &&
-        ((this->actionFunc == func_8089C56C) || (this->actionFunc == func_8089C2A8))) {
-        Math_Vec3f_Copy(&sp74, &this->colliderQuad.dim.quad[0]);
-        Math_Vec3f_Copy(&sp80, &this->colliderQuad.dim.quad[1]);
-        Matrix_MultVec3f(&D_8089E38C, &sp68);
-        Matrix_MultVec3f(&D_8089E398, &sp5C);
-        Collider_SetQuadVertices(&this->colliderQuad, &sp5C, &sp68, &sp74, &sp80);
-        if (this->colliderQuad.base.atFlags & AT_ON) {
-            EffectBlure_AddVertex(Effect_GetByIndex(this->unk_2A0), &sp68, &sp5C);
+    if ((limbIndex == DINOLFOS_LIMB_RIGHT_HAND) && (this->timer2 != this->timer1) &&
```
tom-overton, line 1506:
```suggestion
    if ((limbIndex == DINOLFOS_LIMB_RIGHT_HAND) && (this->timer2 != this->attackTimer) &&
```

Below, you set `timer2` to `attackTimer`:
```c
this->timer2 = this->attackTimer;
```

So you should do the same thing up here too.

I did some research into what the whole point of this check is, and I think I have this one figured out. This whole `timer2` stuff is a bit of a hacky way to ensure that this code only gets run once per Update step. It's possible for a Draw function to be called multiple times for any given Update call (for example, by repeatedly pausing quickly so the game can't advance to the next frame), so if you put gameplay logic in the Draw function like this one does, you have to be careful. This is the cause of a lot of glitches like Epona Push, Sword Hitbox Displacement, etc. I've seen developers do a few things to work around this, but never anything like this. The way it works is that `attackTimer` is updated in an Update step, then `timer2` is updated in the Draw step. If they're matching, that means there hasn't been a new Update step since the last time we drew something, so we shouldn't do anything here.

I have no idea what you would even call `timer2` in this scenario, though. But maybe that gets the creative juices flowing.

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.h
```diff
@@ -24,36 +24,79 @@ typedef enum EnDinofosBodyPart {
...
+        s16 headTimer;
+        s16 stunTimer;
+        s16 jumpType;
+        s16 isJumpingBackward;
+    };
+    /* 0x292 */ union { // second multi-use timer
+        s16 timer2;
+        s16 sidestepTimer;
+        s16 walkTimer;
+        s16 idleTimer;
```
tom-overton, line 81: This is not something I'm asking or demanding you to do; this is just something I was thinking about. But one thing I've noticed is that the developers sometimes like to give certain kinds of movement their own timer. I first noticed it with Garo Master; you can see a variable called `attackMovementTimer` because it's used for dash attacks and teleport attacks. Looking at this actor, I can sort of see something similar here. These timers all seem to be related to movement of some sort. Even `idleTimer` is sort of movement related in the sense that it's measuring how long it needs to wait before it makes a move.

The other timer doesn't really seem to be tied to movement, at least directly. For example, `attackTimer` is used for the jump attack, but that's a jumping *attack*; there's more to it than just the movement. So maybe we could collapse all of these down into a single `movementTimer`? You would still likely want the union for `timer2`, which as I stated before, I have *no* clue how to name that one. But perhaps that would simplify things by making the functions a little more consistent. It might *complicate* things, though, by using "movement" timer for idling, which is sort of the opposite of moving. 

What do you think? Again, if you prefer it as-is, that's fine too. I just saw this as potentially similar to something I had seen before.

tom-overton (commented):
Excellent work! I have a few minor things, but overall this is great. Github also won't let me comment on it, but you can put in some enum values for `sJntSphElementsInit`. For example:
```c
    {
        {
            ELEMTYPE_UNK0,
            { 0x00000000, 0x00, 0x00 },
            { 0xF7CFFFFF, 0x00, 0x00 },
            TOUCH_NONE | TOUCH_SFX_NORMAL,
            BUMP_ON | BUMP_HOOKABLE,
            OCELEM_ON,
        },
        { 1, { { 200, 300, 0 }, 19 }, 100 },
    },
```
can become:
```c
    {
        {
            ELEMTYPE_UNK0,
            { 0x00000000, 0x00, 0x00 },
            { 0xF7CFFFFF, 0x00, 0x00 },
            TOUCH_NONE | TOUCH_SFX_NORMAL,
            BUMP_ON | BUMP_HOOKABLE,
            OCELEM_ON,
        },
        { DINOLFOS_LIMB_LOWER_BODY, { { 200, 300, 0 }, 19 }, 100 },
    },
```

src/overlays/actors/ovl_En_Dinofos/z_en_dinofos.c
```diff
@@ -1459,30 +1490,31 @@ static s8 sLimbToBodyParts[DINOLFOS_LIMB_MAX] = {
...
-    Vec3f sp74;
-    Vec3f sp68;
-    Vec3f sp5C;
-    s32 sp58;
-    Vec3f sp4C;
-    MtxF* sp48;
-
-    Collider_UpdateSpheres(limbIndex, &this->colliderJntSph);
+    Vec3f knifeTipQuadOldPos;
+    Vec3f knifeBaseQuadOldPos;
```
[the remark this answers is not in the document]
hensldm: My bad, I was the one that suggested `old`. Personally I would actually say prefix with `prev`.

tom-overton (approved):

PR #1665, opened 2024-07-27
Misc Cleanup

hensldm:
See commits.

For the removed blob comments, they are just things that were already accounted for, or things zapd already implicitly extracted.

For the TLUTs, they are unusually sized so just decided to make them height 1 so they fit nicely.


AngheloAlf (approved):

hensldm:
Small note. I noticed a few initialize misspelling typos that I just pushed up a fix.

jpburnett (approved):

PR #1669, opened 2024-08-07
Cleanup: Pass all paths to tools rather than tools constructing them

Dragorn421:
As a not-knower of MM I didn't expect this port of oot's https://github.com/zeldaret/oot/pull/2017 to be so complicated 8)

For example, `extract_assets.py` now is called by make with `$(EXTRACTED_DIR)/baserom` rather than constructing that path like `extracted/mm_version/baserom` based on the `--version`

This PR also normalizes the CLI arguments for passing in paths (all required arguments besides oot version are positional) and the oot version (now `-v`,`--version`)

Note: this is a joint PR with OoT, wait for both approved before any merge
https://github.com/zeldaret/oot/pull/2017


Makefile
```diff
@@ -372,17 +372,17 @@ venv:
 ## Extraction step
 setup:
 	$(MAKE) -C tools
-	$(PYTHON) tools/buildtools/decompress_baserom.py $(VERSION)
+	$(PYTHON) tools/buildtools/decompress_baserom.py --mm-version $(VERSION)
```
AngheloAlf, line 375: I'm not sure if I like `--mm-version` for the name of the flag, just `--version` should be fine.

We could even omit the flag entirely since those scripts aren't meant to be invoked by the user directly. Seeing that this script is being passed `$(VERSION)` feels obvious enough, doesn't it?

Makefile
```diff
@@ -372,17 +372,17 @@ venv:
...
 	$(PYTHON) tools/buildtools/extract_baserom.py $(BASEROM_DIR)/baserom-decompressed.z64 -o $(EXTRACTED_DIR)/baserom --dmadata-start `cat $(BASEROM_DIR)/dmadata_start.txt` --dmadata-names $(BASEROM_DIR)/dmadata_names.txt
-	$(PYTHON) tools/buildtools/extract_yars.py $(VERSION)
+	$(PYTHON) tools/buildtools/extract_yars.py --baserom-segments $(EXTRACTED_DIR)/baserom --mm-version $(VERSION)
```
AngheloAlf, line 377: Same here and for the other scripts. I'm not a big fan of using explicit flags for required parameters. Feels counterintuitive

extract_assets.py
```diff
@@ -68,25 +69,47 @@ def ExtractFunc(fullPath):
...
         globalExtractedAssetsTracker[fullPath]["timestamp"] = currentTimeStamp
 
-def initializeWorker(abort, unaccounted: bool, extractedAssetsTracker: dict, manager):
+def initializeWorker(abort, unaccounted: bool, extractedAssetsTracker: dict, manager, baseromSegmentsDir: Path, outputDir: Path):
     global globalAbort
     global globalUnaccounted
     global globalExtractedAssetsTracker
     global globalManager
+    global globalBaseromSegmentsDir
+    global globalOutputDir
```
AngheloAlf, line 78: Using globals for passing flags :despair:

I guess its fine since the script was already doing this

tools/disasm/disasm.py
```diff
@@ -2099,6 +2108,8 @@ def find_late_rodata_start(rodata):
 files_spec = None
 with open("tools/disasm/files.txt", "r") as infile:
     files_spec = ast.literal_eval(infile.read())
+for segment in files_spec:
+    segment[1] = segment[1].replace("$(BASEROM)", str(baserom_segments_dir))
```
AngheloAlf, line 2112: Does doing this has any measurable impact on running the disassembler? Hopefully it doesn't but who knows

Dragorn421: I didn't measure but I bet it does not have any measurable impact

extract_assets.py
Dragorn421: this script is on a suspended yeet sentence in OoT (see new assets system tm) so I really don't care about it yeah, I don't want to clean it up completely (would be basically a rewrite)

Makefile
Dragorn421: tools in OoT take the version as --oot-version/-v and it's optional, defaulting to gc-eu-mq-dbg
For consistency with those I also used it even if mandatory
not sure how to reconcile

tools/disasm/files.txt
```diff
@@ -8,7 +8,7 @@
 #
 
 [
-    ['makerom', 'extracted/n64-us/baserom/', 'makerom',
+    ['makerom', '$(BASEROM)', 'makerom',
```
mzxrules, line 11: this change doesn't seem appropriate given that the data within this file is version specific, and providing a `$(BASEROM)` of a different version string would be incorrect.

Makefile
hensldm: I know there was quite a big disucssion about this in OoT when it was discussed. `--oot-version` was the compromise. Kind of thinking for the sake of keeping things easier to port between games, curious if we should just do something like `--rom-version` or `--game-version` so they could be the same in both.

tools/disasm/files.txt
Dragorn421: well yes it's implied the baserom folder from arguments should contain the segments for the right version

Makefile
AngheloAlf: I think I like `--game-version` over `--mm-version`.
@Dragorn421 thoughts?

Dragorn421: My thoughts are I wanted to avoid -v and --version in the first place as they could be confused with a verbose flag and/or the tool version rather than the game version

However we now have -v so really I have no reason to oppose to anything if my only point is already defeated. We could have --version too

hensldm: I would think it would still be worth it to keep `--version` reserved, but I'm not sure we would even add version info for the script.

tools/disasm/files.txt
mzxrules: The way it looks to me makes it seem like `input path` isn't only specifying the rom path, it also associates the data to a specific version (n64-us). It makes no sense to replace `$(BASEROM) with `extracted/n64-jp/baserom/` for example because the data in this file won't be applicable to that version.

In fact, I don't understand why we're even storing the `$(BASEROM)` path here to begin with. Shouldn't disasm.py just be handling that internally now?

AngheloAlf: I guess this is a fair point. The paths of every entry is the same so it would be fine to yeet it.
I'm not sure how many changes this would require tho

mzxrules: If you do the bare minimum it should just be a one line change (replace the only `segment[1]` reference with `baserom_segments_dir`). I wouldn't do much more than that in this PR (unless you feel motivated to fix everything right now) since disasm.py looks like it needs to be cleaned up a bit to support multi-version anyway

hensldm: This really depends on how we decide to deal with disassembling in multi-version, which will require discussion when we get to it. I would think we would use spimdisasm (which uses a different config system anyways). Tbh I would say my preferred option would be just revert the disasm changes, however doing the bare minimum mzx suggests or even just leaving as is I would be fine with.

Makefile
hensldm: I'll admit I don't think I have too strong opinion on using explicit flags for required arguments. I think I like the consistency just a little bit more so would probably prefer going that route. 
However I would say if the argument really is required, it should probably not have a default and actually require the user to specify it (`extract_baserom.py` in OoT for example doesn't have a default option).

Makefile
hensldm: Personally I would say go with `--game-version` for now.

Dragorn421: Changed to `-v`,`--version`, not required, and defaults to n64-us

Makefile
Dragorn421: Made optional arguments positional

tools/disasm/files.txt
Dragorn421: Replaced `'$(BASEROM)'` with None and it'sn ow unused.
I didn't remove it because disasm.py is very spaghetti in its usage of this data, would need to offset dozens of indexing by 1

AngheloAlf (approved):

hensldm (approved):

PR #1670, opened 2024-08-08
[Audio 2/?] Extract audio samples to wav

Thar0:
This PR extracts all audio samples from all samplebanks to wav files which will match on round-trip. Note this PR only adds extraction, rebuilding is planned for a future PR.

The wav files have a single custom section named `zzbk` to store the prediction codebook. Unfortunately getting matching codebooks is much more difficult than getting matching vadpcm frames, so we elect to save the codebook into the wav file and if the custom section is present when re-encoding it will use the saved book rather than generating a new one. Thankfully loop states are fully matching with no tricks, so no extra data for loops needed to be saved in the wav files.

Special thanks to @zelda2774 for the brute-force-matching vadpcm decoder that is incorporated into the `sampleconv` tool (parts of `src/codec/vadpcm.c`).


AngheloAlf (approved):
Works on my machine.

I think my only complain is how the C functions have the return type on a different line than the function name, but I don't plan to actually fight that

hensldm (approved):
See OOT review

PR #1671, opened 2024-08-09
[Audio 3/?] Set up building samples

Thar0:
This PR adds Makefile machinery for building samples, laying the groundwork for building samplebanks and soundfonts. Note that samples are not yet built, as nothing currently depends on these rules. Using these rules will begin in the next PR where samplebanks will be built.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,7 @@ SCHC_FLAGS  :=
 
 # Audio tools
 AUDIO_EXTRACT := $(PYTHON) tools/audio_extraction.py
+SAMPLECONV    := tools/audio/sampleconv/sampleconv
 
 # Command to replace path variables in the spec file. We can't use the C
 # preprocessor for this because it won't substitute inside string literals.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -213,6 +214,24 @@ $(shell mkdir -p asm data extracted)
...
+ifneq ($(wildcard assets/audio/samples),)
+  SAMPLE_DIRS := $(shell find assets/audio/samples -type d)
+else
+  SAMPLE_DIRS :=
+endif
+
+SAMPLE_FILES         := $(foreach dir,$(SAMPLE_DIRS),$(wildcard $(dir)/*.wav))
+SAMPLE_EXTRACT_FILES := $(foreach dir,$(SAMPLE_EXTRACT_DIRS),$(wildcard $(dir)/*.wav))
+AIFC_FILES           := $(foreach f,$(SAMPLE_FILES),$(BUILD_DIR)/$(f:.wav=.aifc)) $(foreach f,$(SAMPLE_EXTRACT_FILES:.wav=.aifc),$(f:$(EXTRACTED_DIR)/%=$(BUILD_DIR)/%))
+SAMPLE_BLOBS_IN      := $(foreach dir,$(SAMPLE_EXTRACT_DIRS),$(wildcard $(dir)/*.bin))
```
hensldm, line 232: What does the `IN` mean? Perhaps `BIN` is better since it seems to find `bin` files?

hensldm (commented):
1 MM specific question.

Makefile
Thar0: It's meant to be short for "input" since it's the input to the build that's written during extraction

hensldm (approved):

Makefile
```diff
@@ -505,6 +524,30 @@ $(BUILD_DIR)/assets/%.jpg.inc.c: assets/%.jpg
 $(BUILD_DIR)/%.schl.inc: %.schl
 	$(SCHC) $(SCHC_FLAGS) -o $@ $<
 
+# Audio
+
+AUDIO_BUILD_DEBUG ?= 0
```
AngheloAlf, line 529: I wouldn't mind moving this to the top of the Makefile with the other settings we have, but since this is kinda niche I guess is fine to have it here

AngheloAlf (approved):

Makefile
```diff
@@ -505,6 +524,30 @@ $(BUILD_DIR)/assets/%.jpg.inc.c: assets/%.jpg
...
+
+# first build samples...
+
+$(BUILD_DIR)/assets/audio/samples/%.half.aifc: assets/audio/samples/%.half.wav
+	$(SAMPLECONV) vadpcm-half $< $@
+
+$(BUILD_DIR)/assets/audio/samples/%.half.aifc: $(EXTRACTED_DIR)/assets/audio/samples/%.half.wav
+	$(SAMPLECONV) vadpcm-half $< $@
+ifeq ($(AUDIO_BUILD_DEBUG),1)
```
AngheloAlf, line 538: Most of the makefile does this kind of checks against 0 instead, so the consistency would be nice
```suggestion
ifneq ($(AUDIO_BUILD_DEBUG),0)
```

Makefile
Thar0: I think I'd like to leave it here for now at least, it can always be moved later

Makefile
Thar0: I think this is less readable and there's already `ifeq ($(ORIG_COMPILER),1)`, and in oot's makefile there's `ifeq ($(DEBUG),1)`

hensldm: Fwiw, the main reason to use `ifneq` 0 is so that `0` means OFF while everything else means ON instead of `ifeq` 1 means `1` is ON and everything else is off.

AngheloAlf: Currently, out of the 10 boolean-like checks we have on this makefile, and only 1 is checked against `1` instead of against `0`.

In OoT there are 14 boolean checks, and 7 of them are checks against `1`.

So MM has a majority and OoT is half and half. I do think comparing against `0` is better since it is more boolean-like.

fig02: double negative in general is really hard to parse. Im not a fan of `ifndef` in general, even for things like NDEBUG.
I have no horse in this race but think that how tharo has it is much more readable

PR #1673, opened 2024-08-12
Extract sFaultDrawerFont as blob

hensldm:
This extracts sFaultDrawerFont as a blob using ZAPD (at least for a temporary stop gap for now). This allows removing the `GLOBAL_ASM` `#pragma` from `fault_drawer.c`.

### DIFF
--- a/assets/xml/boot/fault_drawer.xml
+++ b/assets/xml/boot/fault_drawer.xml
@@ -1,0 +1,6 @@
+<Root>
+    <File Name="boot" OutName="fault_drawer_font" BaseAddress="0x80080060" RangeStart="0x18BF0" RangeEnd="0x18FF0" >
+        <!-- The real version is const and found in fault_drawer.c, though shares the same name -->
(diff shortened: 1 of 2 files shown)


assets/xml/boot/fault_drawer.xml
```diff
@@ -0,0 +1,6 @@
+<Root>
+    <File Name="boot" OutName="fault_drawer_font" BaseAddress="0x80080060" RangeStart="0x18BF0" RangeEnd="0x18FF0" >
+        <!-- The real version is const and found in fault_drawer.c, though shares the same name -->
```
hensldm, line 3: Would like opinions on this. Is this acceptable, any ideas what to name the unused extracted variable instead?

AngheloAlf: I think it is fine this way

AngheloAlf (approved):

tom-overton (approved):

PR #1674, opened 2024-08-15
[Audio 5/?] Build the Samplebank Table

Thar0:
This PR builds the Samplebank Table `gSampleBankTable` automatically during the build. The tool `atblgen` does this by collecting all the samplebank xmls, checking that their indices are unique and all accounted for, then emitting a header file that looks like
```c
/**
 * DEFINE_SAMPLE_BANK(name, medium, cachePolicy)
 * DEFINE_SAMPLE_BANK_PTR(name, medium, cachePolicy)
 */
DEFINE_SAMPLE_BANK    (SampleBank_0, MEDIUM_CART, CACHE_LOAD_EITHER_NOSYNC)
DEFINE_SAMPLE_BANK_PTR(0, MEDIUM_CART, CACHE_LOAD_EITHER_NOSYNC)
DEFINE_SAMPLE_BANK    (SampleBank_2, MEDIUM_CART, CACHE_LOAD_EITHER_NOSYNC)
```
This generated header file is then included in `src/audio/tables/samplebank_table.c` to be compiled.

Since the table is structured as a header followed by a dynamically sized array, some tricks needed to be employed:
- The table is split into two declarations, the header and entries array, which are marked with the `NO_REORDER` attribute so modern compilers do not reorder them.
- The table symbol is declared as a weak alias to the header symbol, `#pragma weak` is supported by both IDO and GCC.

The table has to end up in `.rodata`, but the contents of the C file ends up in `.data`. To resolve this there's an incremental link (`ld -r`) step that moves `.data` to `.rodata` via the linker script `include/audio/atbl_rdata.ld`. Initially, `objcopy --rename-section` was tried but this broke relocations.

Finally, since the table header contains the number of table entries, we count them with the preprocessor via:
```c
#define DEFINE_SAMPLE_BANK(name, medium, cachePolicy) 1 +
#define DEFINE_SAMPLE_BANK_PTR(name, medium, cachePolicy) 1 +

#include "assets/audio/samplebank_table.h"

#undef DEFINE_SAMPLE_BANK
#undef DEFINE_SAMPLE_BANK_PTR
    0,
```
This expands to `1 + 1 + ... + 1 + `, the 0 is present to complete the final addition. This evaluates to the number of entries, as desired.


linker_scripts/audio_table_rodata.ld
```diff
@@ -0,0 +1,17 @@
+OUTPUT_ARCH (mips)
```
hensldm, line 1: With this in linker_scripts it is actually getting used in the final link as well which is probably not ideal.
Not sure if we should move it to a different folder or add a subfolder that gets excluded from the find command?
```bash
mips-linux-gnu-ld -T build/n64-us/mm-n64-us.ld -T build/n64-us/linker_scripts/audio_table_rodata.ld build/n64-us/linker_scripts/extra.ld build/n64-us/linker_scripts/hardware_regs.ld build/n64-us/linker_scripts/undefined_syms.ld --no-check-sections --accept-unknown-input-arch --emit-relocs -Map build/n64-us/mm-n64-us.map -o build/n64-us/mm-n64-us.elf
```

hensldm (approved):

Makefile
```diff
@@ -609,6 +610,20 @@ ifeq ($(AUDIO_BUILD_DEBUG),1)
...
+
+$(BUILD_DIR)/assets/audio/samplebank_table.h: $(SAMPLEBANK_BUILD_XMLS)
+	$(ATBLGEN) --banks $@ $^
+
+# build the tables into objects, move data -> rodata
+
+$(BUILD_DIR)/src/audio/tables/samplebank_table.o: src/audio/tables/samplebank_table.c $(BUILD_DIR)/assets/audio/samplebank_table.h
+	$(CC_CHECK_COMP) $(CC_CHECK_FLAGS) $(IINC) $(CC_CHECK_WARNINGS) $(C_DEFINES) $(MIPS_BUILTIN_DEFS) -o $(@:.o=.tmp) $<
+	$(CC) -c $(CFLAGS) $(IINC) $(WARNINGS) $(C_DEFINES) $(MIPS_VERSION) $(ENDIAN) $(OPTFLAGS) -o $(@:.o=.tmp) $<
+	@$(LD) -r -T linker_scripts/audio_table_rodata.ld $(@:.o=.tmp) -o $@
```
AngheloAlf, line 623: I kinda prefer to not hide this command, since it is the command that actually generates the target file we want

AngheloAlf (commented):
1 nit

Makefile
Thar0: Done

AngheloAlf (approved):

PR #1675, opened 2024-08-19
[Audio 6/?] Build Soundfonts and the Soundfont Table

Thar0:
This PR builds Soundfonts and the Soundfont table. Matches the `Audiobank` file.

Soundfonts are structured as:
- Pointer to Drums pointer list
- Pointer to Sfx list
- List of Instrument pointers
- Sample Definitions; consisting of the `Sample` structure, ADPCM Book (may be deduplicated if identical to another sample), ADPCM Loop (the header always exists even if the sample doesn't loop, if it does loop an extra two frames worth of decoded state is included)
- Envelope Definitions, referenced by Instruments and Drums
- Instrument Definitions
- Drum Definitions
- Drums Pointer List (usually this is 64 entries long if it exists at all, but some "oddball" soundfonts can have it shorter)
- Sfx Array

Soundfont build process:
- The Soundfont Compiler (`sfc`) reads a soundfont xml, the samplebank xml(s) it references and the AIFC file(s) the samplebank xml(s) reference and produces:
  - A .c source file for the Soundfont data, structured according to above. The intermediate C representation ensures that if a structure changes in the source (e.g. `Sample`) the soundfonts will fail to compile without account for such changes.
  - A .h file containing the number of instruments/drums/sfx in this Soundfont, instrument/drum/sfx names to use in sequence files, and an assembly snippet that declares the Soundfont id that will be used for sequences.
  - A .name file containing the Soundfont name. This lets Soundfont xml filenames be distinct from the symbolic name used to refer to them in code. This is used when adding the start and length symbols later.
- The output .c file is compiled to a .o file.
- The .o file is patched by `elfpatch` to make all the defined symbols ABS symbols, so that when linked their values do not update. This is necessary as pointers to data in Soundfonts are file-relative, not segment-relative.
- The .o file is partially linked to move .data -> .rodata and add a length symbol, a (non-ABS) start symbol is added by objcopy. The start and length symbols are used in the Soundfont table.

The Soundfont table is structured in the same way as the Samplebank table seen previously, but contains a little more information. In addition to the start and length of each file, it contains the number of instruments/drums/sfx. The Soundfont headers emitted by `sfc` are used to source these quantities.

Note that `atblgen` contains an assumption that Soundfont headers in build are adjacent to their headers, such that a replacement of `.xml` -> `.h` results in a valid header path. This will always be true for our build system.

The generated header looks like:
```c
/**
 * DEFINE_SOUNDFONT(name, medium, cachePolicy, sampleBankNormal, sampleBankDD, nInstruments, nDrums, nSfx)
 */
#include "build/n64-us/assets/audio/soundfonts/Soundfont_0.h"
DEFINE_SOUNDFONT(Soundfont_0, MEDIUM_CART, CACHE_LOAD_PERMANENT, 0, 255, SF0_NUM_INSTRUMENTS, SF0_NUM_DRUMS, SF0_NUM_SFX)
#include "build/n64-us/assets/audio/soundfonts/Soundfont_1.h"
DEFINE_SOUNDFONT(Soundfont_1, MEDIUM_CART, CACHE_LOAD_PERMANENT, 0, 255, SF1_NUM_INSTRUMENTS, SF1_NUM_DRUMS, SF1_NUM_SFX)
#include "build/n64-us/assets/audio/soundfonts/Soundfont_2.h"
DEFINE_SOUNDFONT(Soundfont_2, MEDIUM_CART, CACHE_LOAD_PERMANENT, 0, 255, SF2_NUM_INSTRUMENTS, SF2_NUM_DRUMS, SF2_NUM_SFX)
[...]
```


include/audio/soundfont.h
```diff
@@ -5,22 +5,34 @@
...
+
+typedef struct AdpcmLoop {
+    /* 0x00 */ AdpcmLoopHeader header;
     /* 0x10 */ s16 predictorState[16]; // only exists if count != 0. 8-byte aligned
 } AdpcmLoop; // size = 0x30 (or 0x10)
 
+typedef struct AdpcmBookHeader {
+    /* 0x00 */ s32 order;
+    /* 0x04 */ s32 numPredictors;
+} AdpcmBookHeader;
```
hensldm, line 23:
```suggestion
typedef struct AdpcmBookHeader {
    /* 0x0 */ s32 order;
    /* 0x4 */ s32 numPredictors;
} AdpcmBookHeader; // size = 0x8
```

hensldm (commented):
1 MM specific review. Will leave more general thoughts on the OoT one.

Makefile
```diff
@@ -624,6 +685,13 @@ $(BUILD_DIR)/src/audio/tables/samplebank_table.o: src/audio/tables/samplebank_ta
 	@$(RM) $(@:.o=.tmp)
 	$(RM_MDEBUG)
 
+$(BUILD_DIR)/src/audio/tables/soundfont_table.o: src/audio/tables/soundfont_table.c $(BUILD_DIR)/assets/audio/soundfont_table.h $(SOUNDFONT_HEADERS)
+	$(CC_CHECK_COMP) $(CC_CHECK_FLAGS) $(IINC) $(CC_CHECK_WARNINGS) $(C_DEFINES) $(MIPS_BUILTIN_DEFS) -o $(@:.o=.tmp) $<
+	$(CC) -c $(CFLAGS) $(IINC) $(WARNINGS) $(C_DEFINES) $(MIPS_VERSION) $(ENDIAN) $(OPTFLAGS) -o $(@:.o=.tmp) $<
+	@$(LD) -r -T linker_scripts/audio_table_rodata.ld $(@:.o=.tmp) -o $@
```
hensldm, line 691:
```suggestion
	$(LD) -r -T linker_scripts/audio_table_rodata.ld $(@:.o=.tmp) -o $@
```
Note this already seems addressed in the OoT version.

hensldm (commented):
Missed something, sorry.

tools/warnings_count/warnings_build_current.txt
```diff
@@ -1,2 +1,3 @@
+[95mWarning: [97mInvalid pointer indirect: 0 for samplebank SampleBank_2[0m
```
AngheloAlf, line 1: I'm curious. What produces this warning?

Makefile
```diff
@@ -163,17 +163,20 @@ SCHC_FLAGS  :=
...
+SFCFLAGS := --matching
 
 # Command to replace $(BUILD_DIR) in some files with the build path.
 # We can't use the C preprocessor for this because it won't substitute inside string literals.
 BUILD_DIR_REPLACE := sed -e 's|$$(BUILD_DIR)|$(BUILD_DIR)|g'
 
 CFLAGS           += -G 0 -non_shared -Xcpluscomm -nostdinc -Wab,-r4300_mul
 
-WARNINGS         := -fullwarn -verbose -woff 624,649,838,712,516,513,596,564,594
+WARNINGS         := -fullwarn -verbose -woff 624,649,838,712,516,513,596,564,594,807
```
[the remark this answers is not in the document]
Thar0: This warning is due to the incomplete type struct member in `AdpcmBook`:
```
cfe: Warning 807: include/audio/soundfont.h, line 35: member cannot be of function or incomplete type.
       AdpcmBookData codeBook;
 --------------------^
```
It doesn't actually do anything for codegen and gcc is ok with it so I disabled the IDO warning

include/attributes.h
```diff
@@ -8,9 +8,10 @@
...
 
-#define UNUSED      __attribute__((unused))
-#define FALLTHROUGH __attribute__((fallthrough))
-#define NORETURN    __attribute__((noreturn))
-#define NO_REORDER  __attribute__((no_reorder))
+#define UNUSED       __attribute__((unused))
+#define FALLTHROUGH  __attribute__((fallthrough))
+#define NORETURN     __attribute__((noreturn))
+#define NO_REORDER   __attribute__((no_reorder))
+#define SECTION_DATA __attribute__((section(".data")))
```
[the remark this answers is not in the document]
Thar0: I'm a little reluctant to do this now since it isn't needed anywhere else but if we ever need to use a section attribute elsewhere for a different section it could be changed in the future pretty easily (just replace `SECTION_DATA` usages in `soundfont_compiler.c` with `SECTION(.data)` or similar)

tools/warnings_count/warnings_build_current.txt
Thar0: This warning comes from the bugged soundfont 40. Soundfont 40 reports a samplebank index of 0 in the soundfont table but it actually uses samplebank 2 (if you look at the sample pointers in the soundfont, they're nonsense when applied to samplebank 0 but make sense for samplebank 2).

Samplebank indices can either refer to a "real" samplebank or a pointer to a real samplebank. I have no idea why this system exists, but unfortunately it does.

The build system treats soundfont 40 as if it is trying to use samplebank 2 through a pointer index of 0. But 0 is not a pointer index, it's a real samplebank, hence the warning.

Makefile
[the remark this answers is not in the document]
hensldm: Hmm, definitely not in this PR, but this makes me wonder if for other dynamically size arrays we should actually not give it an actual size (thinking relocs but perhaps there are others) with this warning now off.

hensldm (approved):

PR #1676, opened 2024-08-24
Import data + bss

hensldm:
All data, rodata, bss should now basically be imported expect for microcode related, audio tables related, handwritten asm related, Boss07 related, and whatever `code_801D1E80` is.


src/audio/lib/heap.c
```diff
@@ -1396,6 +1396,9 @@ typedef struct {
     /* 0xC */ u8 newMedium;
 } StorageChange; // size = 0x10
 
+s32 D_801FD120;
+static s32 sBssPad[3];
```
AngheloAlf, line 1400: I don't see any other bss besides this, so I think this pad is not really required, it should be generated automatically.

src/audio/session_config.c
```diff
@@ -1,6 +1,14 @@
 #include "global.h"
 #include "buffers.h"
 
+static s32 sBssPad[36];
```
AngheloAlf, line 4: huh, weird, that's a lot of padding

src/boot/O2/gfxprint.c
```diff
@@ -7,11 +7,7 @@
 #define GFXP_FLAG_ENLARGE (1 << 6)
 #define GFXP_FLAG_OPEN (1 << 7)
 
-//! TODO: Need to extract
-extern u16 sGfxPrintFontTLUT[64];
-extern u16 sGfxPrintRainbowTLUT[16];
-extern u8 sGfxPrintRainbowData[8];
-extern u8 sGfxPrintFontData[2048];
+#include "assets/boot/gfxprint/gfxprint.c"
```
AngheloAlf, line 10: I would kinda prefer if we defined the C symbols on this file (using the corresponding includes) instead of including this generated file

src/audio/lib/heap.c
hensldm: Not quite. Without the pad bss ends up at `0x10` size, but it needs to be `0x20` size.

src/audio/session_config.c
hensldm: OoT has even more :)

AngheloAlf (approved):

engineer124 (approved):

PR #1679, opened 2024-09-04
[Audio 8/?] Check-in handwritten sequences, build sequences, automate various sfx arrays

Thar0:
This PR completes various sequence-related tasks:
- Checks-in the "handwritten" sequences that are not extracted during setup, so that they can be carefully documented. Much of these sequences are not yet documented, only a cleaned-up disassembly is added here in view of adding more documentation later. Sequence 0 in particular is full of esoteric constructs that need untangling.
- Builds all sequences into object files. The sequence assembly language is implemented inside the gnu assembler using both the C preprocessor and assembler macros. See `include/audio/aseq.h` for the language implementation.
- Automates various sfx-related arrays in code and sequence 0.

With this, `Audioseq` is matched.


Makefile
```diff
@@ -668,6 +690,22 @@ ifeq ($(AUDIO_BUILD_DEBUG),1)
...
 endif
 
+# then assemble the sequences...
+
+$(BUILD_DIR)/assets/audio/sequences/%.o: assets/audio/sequences/%.seq include/audio/aseq.h $(SEQUENCE_TABLE) | $(SOUNDFONT_HEADERS)
+	$(SEQ_CPP) $(SEQ_CPPFLAGS) $< -o $(@:.o=.S) -MMD -MT $@
+	$(AS) $(ASFLAGS) -I $(BUILD_DIR)/assets/audio/soundfonts -I include/audio $(@:.o=.S) -o $@
+
+$(BUILD_DIR)/assets/audio/sequences/%.o: $(EXTRACTED_DIR)/assets/audio/sequences/%.seq include/audio/aseq.h $(SEQUENCE_TABLE) | $(SOUNDFONT_HEADERS)
+	$(SEQ_CPP) $(SEQ_CPPFLAGS) $< -o $(@:.o=.S) -MMD -MT $@
```
hensldm, line 700: More so out of curiosity than anything else, but any reason for the capital `S` for the file suffix vs just a lowercase `s`?

include/audio/aseq.h
```diff
@@ -0,0 +1,1979 @@
...
+#define PITCH_E0       123
+#define PITCH_F0       124
+#define PITCH_GF0      125
+#define PITCH_G0       126
+#define PITCH_AF0      127
+
+// Hardcoded Instruments
+#define FONTANY_INSTR_SFX          126
+#define FONTANY_INSTR_DRUM         127
+// gWaveSamples[]
```
hensldm, line 161: What is this comment for?

hensldm (commented):
Couple curiosity questions

Makefile
cadmic: in GNU convention .S files are preprocessed by CPP while .s files aren't. Although in this project we preprocess all .s files anyway so I think it would be more consistent to use .s here too

src/audio/code_8019AF00.c
```diff
@@ -179,12 +179,37 @@ u16 sCustomSequencePc;
...
+#include "tables/sfx/systembank_table.h"
+        0) > UINT8_MAX,
+    (
+#include "tables/sfx/ocarinabank_table.h"
+        0) > UINT8_MAX,
+    (
+#include "tables/sfx/voicebank_table.h"
+        0) > UINT8_MAX,
 };
+#undef DEFINE_SFX
```
AngheloAlf, line 206: Oh god, this is so funny.

AngheloAlf (approved):

include/audio/aseq.h
Thar0: These instruments are implemented by the waveforms pointed to by `gWaveSamples`

include/tables/sequence_table.h
```diff
@@ -0,0 +1,133 @@
+/**
+ * DEFINE_SEQUENCE(name, seqId, storageMedium, cachePolicy, seqFlags)
+ * DEFINE_SEQUENCE_PTR(seqIdReal, seqId, storageMediumReal, cachePolicyReal, seqFlags)
+ */
```
hensldm, line 4: Don't think we really need to do it now in this PR, but it would be nice if this header was in the same format as the other table headers.

include/audio/aseq.h
hensldm: Wonder if this would be better as maybe like `gWaveSamples` Instruments, but not probably fine as is.

hensldm (approved):
Couple more thoughts, but looks good overall to me :) .

PR #1681, opened 2024-09-06
[headers] Move padmgr, scheduler, irqmgr symbols to respective headers; add segmented_address.h

Dragorn421:
Companion PR to https://github.com/zeldaret/oot/pull/2151

### DIFF
--- a/include/segmented_address.h
+++ b/include/segmented_address.h
@@ -1,0 +1,12 @@
+#ifndef SEGMENTED_ADDRESS_H
+#define SEGMENTED_ADDRESS_H
+
+#include "ultra64.h"
+
+#include "stdint.h"
+
+extern uintptr_t gSegments[NUM_SEGMENTS];
+
+#define SEGMENTED_TO_K0(addr) (void*)((gSegments[SEGMENT_NUMBER(addr)] + K0BASE) + SEGMENT_OFFSET(addr))
+
+#endif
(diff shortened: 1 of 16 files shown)


Dragorn421:
this builds not OK, probably bss memes
I'm not familiar enough with MM to figure it out, any helper?

src/boot/idle.c
```diff
@@ -1,3 +1,5 @@
+#include "stdbool.h"
```
hensldm, line 1: Can you move this below the `prevent_bss_reordering`? We try to keep `prevent` as the first so it is easier to find and remove as needed.

hensldm (commented):
Also as mentioned in discord, could you move the `gIrqMgr` extern to `irqmgr.h`? If you have more bss issues feel free to reach out.

AngheloAlf (approved):
Hopefully bss won't be that terrible to fix

Dragorn421:
also added segmented_address.h , keeping up with the oot counterpart PR

include/segmented_address.h
AngheloAlf, line 1: I wonder if we should merge this header with `segment_symbols.h`, but arguably we should get rid of the latter so it is fine this way

AngheloAlf (approved):

hensldm (approved):

include/segmented_address.h
cadmic: segmented addresses and ROM segments are fairly separate concepts though. It's true that some ROM segments are placed at a segmented address but do not be deceived

PR #1683, opened 2024-09-07
Remove assets from include paths

hensldm:
Removes the assets folder from `IINC` to match OoT. Also in preparation for moving asset extraction to `extracted` with committed assets in `assets`.

Sorry for the rather large PR, though most were just find and replace.


Dragorn421 (approved):
The list of directories where the compiler searches for files to include is called the "include path", "automatic includes" is a weird way to refer to it (and would misguide me to think of the [cpp `-include` flag](https://gcc.gnu.org/onlinedocs/gcc-14.2.0/cpp/Invocation.html#index-include) for including a file)

AngheloAlf (approved):

PR #1685, opened 2024-09-07
Port over OoT's new text pipeline

hensldm:
Ported over OoT's newish msgdis.py and msgenc.py. The extracted data is much nicer imo with a new HEADER macro for the message headers, ability to use sfxIDs directly, decimal for timers, and imporved formatting.

Main differences in order to get it to work with MM are:
1. Add the ability to decode message headers
2. Add an additional encoding to msgenc/charmap for the credits as the newline character is different between the 2.
3. Had to add an additional check for sfxIds since 5 messages seem to use ids that don't exist

One thing to note is I decided to follow OoT and not use the `CMD` prefix for message commands.

This also splits off text extraction out of extract_assets.py and just calls msgdis.py directly

Tested myself and everything seems to be working fine, including adding additional text. Additional testing would be appreciated though.
Note: their is no check anymore for if the data has already been extracted, so running the script will overwrite the message data (including anything added). This will be better when we switch to actually extracting assets to extracted.


tools/text/msgdis.py
```diff
@@ -0,0 +1,3190 @@
...
+        return (u << 8) | l
+
+    def format_sfx_id(self, c : int) -> str:
+        try:
+            if c & 0x800:
+                return sfx_ids[c & ~0x800]
+            else:
+                return f"{sfx_ids[c]} - SFX_FLAG"
+        except KeyError:
+            return f"0x{c:04X}"
```
Thar0, line 2596: You mentioned this case in the PR description, could you list the sfx ids that this applies to?

tools/text/msgdis.py
```diff
@@ -0,0 +1,3190 @@
...
+        return f"0x{c:02X}"
+    
+    def format_2byte_hex(self, c: int) -> str:
+        return f"0x{c:04X}"
+
+    def format_item_id(self, c : int) -> str:
+        try:
+            return item_ids[c]
+        except KeyError:
+            return f"0x{c:04X}"
```
Thar0, line 2608: Is this try needed? In which cases? Item ids are also encoded as bytes so the format should be `:02X`?

Thar0 (changes requested):
2 things

tools/text/msgdis.py
hensldm: MessageID: SfxID
0x08E8: 0x2693
0x0963: 0x6394
0x0CEA: 0x3AE8
0x0CEF: 0x3AE8
0x0E14: 0x3AFD

tools/text/msgdis.py
hensldm: It is not. Nothing actually uses this yet. Credits have a ctrl code that allow its use, but nothing in the credits use it.
Normal messages have something in the Header that is almost the itemId, but has a layer of indirection.  I think I added it just because I saw it under the sfx which I needed to add it. I'll remove the try block.

AngheloAlf (approved):

assets/text/charmap.txt
```diff
@@ -0,0 +1,21 @@
+# Determines how certain text sequences should be encoded. The text sequence is
+# converted to either the first or second tuple element based on whether the
+# target encoding is the "wide" encoding. The first element is for the non-wide
+# encoding, used for all languages besides JP, while the second element is for
+# the wide encoding, used for JP, the third element is for encoding credits.
```
Thar0, line 5:
```suggestion
# Determines how certain text sequences should be encoded. The text sequence is
# converted to one of the tuple elements based on the chosen encoding:
#   Element 1: Non-wide encoding, for text other than Japanese text and the credits.
#   Element 2: For Japanese text.
#   Element 3: For credits text.
```

Thar0 (approved):

PR #1686, opened 2024-09-07
func_80183070 to Sys_Freeze

Ozero4:


src/code/code_80183070.c
```diff
@@ -1,6 +1,6 @@
 #include "libc64/sleep.h"
 
-void func_80183070(void) {
```
hensldm, line 3: With the function named I think we can probably name the file now as well. I would say just `sleep_forever.c` but I think it would be a good idea to let others give opinions before making the change.

Ozero4: Yes I hesitated

AngheloAlf: I think renaming the file should be fine

hensldm:
Looks like main needs to be merged in and conflicts resolved.

hensldm (approved):

src/code/sleep_forever.c
```diff
@@ -1,6 +1,6 @@
 #include "libc64/sleep.h"
 
-void func_80183070(void) {
+void Sleep_Forever(void) {
```
hensldm, line 3: So taking another look at OoT, they have an additional log message that uses the word freeze which I'm wondering if should actually use in the function name (`func_800D31A0` is the function in OoT).

hensldm: For reference, OoT (https://github.com/zeldaret/oot/pull/2315) looks to rename the file to `sys_freeze` (no word on the function yet, though I think just `Freeze` or `Freeze_Forever` works) which I agree with.

mzxrules: I disagree with OoT's decision to call it `sys_freeze`. The word freeze in the context most appropriate is a system hang, or an unintended fault of software/hardware causing unresponsiveness, and not a deliberate action. Additionally, calling `Sleep_Forever` does not necessarily result in an unresponsive system.

The "freeze!" comment in OoT more seems like a joke than a hint at the actual source name.

hensldm: Sorry it has been awhile. Personally, I still think `sys_freeze` is good (and better to match OoT anyways).
While the comment may be a joke, the intention of it freezing the CPU still seems accurate as it gets the CPU stuck in an infinite loop.

> Additionally, calling Sleep_Forever does not necessarily result in an unresponsive system.

Perhaps I'm missing something, but what could be done to kick it out of this state?

hensldm (approved):

AngheloAlf (approved):

PR #1688, opened 2024-09-08
[headers] Add kaleido_manager.h, clean up z64jpeg.h protos

Dragorn421:
Sync with oot:
- https://github.com/zeldaret/oot/pull/2171
- https://github.com/zeldaret/oot/pull/2174

### DIFF
--- a/src/code/jpegdecoder.c
+++ b/src/code/jpegdecoder.c
@@ -2,6 +2,10 @@
 #include "stdbool.h"
 #include "macros.h"
 
+s32 JpegDecoder_ProcessMcu(JpegHuffmanTable* hTable0, JpegHuffmanTable* hTable1, u16* mcu, s16* unk);
+s32 JpegDecoder_ParseNextSymbol(JpegHuffmanTable* hTable, s16* outCoeff, s8* outZeroCount);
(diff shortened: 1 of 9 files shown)


src/code/jpegdecoder.c
```diff
@@ -2,6 +2,10 @@
 #include "stdbool.h"
 #include "macros.h"
 
+s32 JpegDecoder_ProcessMcu(JpegHuffmanTable* hTable0, JpegHuffmanTable* hTable1, u16* mcu, s16* unk);
+s32 JpegDecoder_ParseNextSymbol(JpegHuffmanTable* hTable, s16* outCoeff, s8* outZeroCount);
+u16 JpegDecoder_ReadBits(u8 len);
```
AngheloAlf, line 7: I'm not a huge fan of declaring prototypes in C files instead of on a header since you lose the compiler checking the declaration actually matches the function definition (like the OoT, where the crash debugger is called with wrong parameters)

Dragorn421: I agree, these are forward declarations for functions in this same file though :)

AngheloAlf (approved):

hensldm (approved):

PR #1701, opened 2024-09-28
Divy out mkdir some more in the makefile

hensldm:
Was starting to get some argument list too long errors again that would cause `make init` fail. Changes the mkdir commands slightly to help address this.


AngheloAlf (approved):

Dragorn421 (approved):
yes yes

PR #1705, opened 2024-10-10
libu64

Dragorn421:
Companion PR to OoT PR https://github.com/zeldaret/oot/pull/2267
Not sure how to document this similarly to OoT, wait for OoT's to be merged and then link to that doc? ( https://github.com/Dragorn421/oot/blob/meta_libu64/docs/libu64.md )


hensldm (approved):
Don't have too much opinion on the doc page. I kind of think we could just duplicate the full markdown file but the link does avoid the duplication which is nice.

hensldm:
Held for documentation and https://github.com/zeldaret/oot/pull/2267

fig02:
oot is ready to merge this fwiw, but dragorn marked it as a draft over there for some reason
@Dragorn421 was this just to hold it until mm was ready?

Dragorn421:
> Don't have too much opinion on the doc page. I kind of think we could just duplicate the full markdown file but the link does avoid the duplication which is nice.

I think I'd rather link to it, because the document aggregates knowledge from several repos but is written from the point of view of OoT, if copied to MM it would need rewritten from MM's point of view. Or it could be rewritten to not be from OoT's point of view

in any case I'll wait for OoT's to be merged so we can work from there

Dragorn421:
Added libu64.md linking to OoT's, comments welcome

engineer124 (approved):

PR #1715, opened 2024-10-16
Clean up gitignores

hensldm:
Cleans up some gitignore files.
Highly recommended to checkout the PR and see what it no longer ignores for you to see if we need to add something back.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,56 +7,28 @@ __pycache__/
 .vscode/
 .vs/
 .idea/
-CMakeLists.txt
-cmake-build-debug
-venv/
-.venv/
-tags
+.cache/
+compile_commands.json
 
 # Project-specific ignores
+.make_options.mk
+extracted/
+build/
+expected/
+asm/
+data/
+docs/doxygen/
 *.z64
 *.n64
 *.v64
-*.sra
-*.bin
-*.elf
-*.flash
-archive/
-build/
-baserom/
-decomp/
-asm/
-data/
-extracted/
-expected/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -7,56 +7,28 @@ __pycache__/
...
-tools/ido5.3_compiler/
-tools/ido7.1_compiler/
-tools/qemu-mips
-tools/ido_recomp/* binary
+# Tools
+.venv/
 ctx.c
+tools/*dSYM/
 graphs/
-*.c.m2c
```
AngheloAlf, line 41: `.c.m2c` is a cache file that m2c generates, so I think it is fine to keep it, even if there's little to none decomp work to do. Specially because we are keeping `ctx.c` on the gitignore

.gitignore
```diff
@@ -7,56 +7,28 @@ __pycache__/
...
 
-# Tool artifacts
-tools/ido5.3_compiler/
-tools/ido7.1_compiler/
-tools/qemu-mips
-tools/ido_recomp/* binary
+# Tools
+.venv/
 ctx.c
+tools/*dSYM/
```
AngheloAlf, line 28: What is a dSYM file?

hensldm: Looks like it helps with OSX. Not sure if we have anyone that still uses a Mac that could check if they are still needed.

AngheloAlf (approved):

Dragorn421 (approved):

engineer124 (approved):

PR #1742, opened 2024-11-16
Fix ocarina start UB for Japas jam session

Archez:
When playing the zora guitar jam session with Japas, this expression would eval as `1 << (18 + 65)` leading to a left shift of 83, which is UB for u32. Rewriting this to use enums for `EVAN_PART` and matching the surround expressions for `AudioOcarina_Start` produces a matching rom without UB.

The new expression will eval as `1 << ((65 - 65) + 19)` leading to a left shift of 19. I'm guessing the previous shift of 83 was rolling over twice (`83-32-32 == 19`)


engineer124 (approved):
Oh lmao, nice find. I'm amazed that matched as it was before

hensldm (approved):

AngheloAlf (approved):

PR #1756, opened 2024-12-11
Remove THIS macro

tom-overton:
When OoT hit 100%, a PR to remove the `THIS` macro followed shortly thereafter: https://github.com/zeldaret/oot/pull/1047
This is just the MM equivalent of that change. Like Fig said in the original PR, this was useful to have around when we were decompiling things, but it's not really necessary now, and it hides exactly what is going on with that upcast.

Some stray thoughts I had:
- ~~Should we change the documentation? I did as part of this PR, but I can undo that change if it's desired. I really wasn't sure if we should leave it as-is for historical purposes or modify it to better suit the current repo.~~
- There were a few instances where `THIS` was being used inline (which were probably fake matches), particularly in the `z_fbdemo_X` files. I replaced them too, but I wasn't sure if that was the best approach.

I wrote a really crappy script to do this automatically for me, so hopefully nothing slipped through or broke.


hensldm:
> Should we change the documentation? I did as part of this PR, but I can undo that change if it's desired. I really wasn't sure if we should leave it as-is for historical purposes or modify it to better suit the current repo.

Tbh, my thoughts would be leave it as is for historical purposes. My guess is that we will eventually do what OoT did and just remove it (though keeping the object tutorial is probably still worth it).

fig02 (approved):
no opinions on tutorial

tom-overton:
Ended up reverting the changes to the documentation; let's just leave it as-is for now unless someone feels super strongly about it.

AngheloAlf (approved):

hensldm:
Could you add it back to 'style.md'? The style guide should continue to be updated as needed imo.

tom-overton:
Oops, I just reverted everything in `docs` and missed that `style.md` was supposed to remain the way it was. I re-did those changes.

PR #1758, opened 2024-12-11
Check in makerom files

Thar0:


hensldm:
Marking as held until the discussion can be finished.

hensldm (approved):
So far no negative opinions to including handwritten asm, and so the PR seems fine to me :).

AngheloAlf (approved):

PR #1760, opened 2024-12-14
Remove asm-processor in favor of preprocess.sh

hensldm:
Copies over preprocess.sh and related tools from OoT to remove the use of asm-processor.
This also allows for the use of `#pragma increment_block_number` instead of `#include prevent_bss_reordering.h` and `tools/fix_bss.py`.
Note I've also changed extract_assets.py so zapd now uses the CS_FLOATS macro.


include/z64cutscene_commands.h
```diff
@@ -6,7 +6,33 @@
...
+
+/**
+ * CMD_F expects an (IEEE 754) encoded float (colloquially "in hex", such as `0x42280000`),
+ * rather than a C float literal (such as `42.0f`).
+ * Float literals cannot be used because cutscenes are arrays of union type CutsceneData, which may contain integers and floats.
+ * Regardless of CutsceneData having a float member, initializing with a float will cast the float to s32.
+ * Designated initializers (added in C99) would solve this problem but are not supported by IDO (C89 and some extensions).
+ */
+#ifdef __GNUC__
+#define CS_FLOAT(ieee754bin, f) (f)
```
AngheloAlf, line 32: Shouldn't this macro be using the designated initializer? Otherwise this value will be casted to an int
```suggestion
#define CS_FLOAT(ieee754bin, f_val) ( {.f = f_val} )
```

hensldm: The designated initializer happens in `CMD_F`, and AFAIK CS_FLOAT should just be used to determine whether `CMD_F` gets a hex value or float value.

cadmic: does IDO support these anyway? Designated initializers are a C99 feature

fig02: they do not. but I think anghelo is asking because it is under `#ifdef __GNUC__`

cadmic: oh apparently I can't read

hensldm:
Looks like this might cause issues for Macs (https://discord.com/channels/688807550715560050/688851337085190255/1317981637086154783). Converted the PR to a draft until I can take a look at it.

hensldm:
Okay, ported over the mac iconv fixes from OoT, so I believe this should work for them as well (don't know for sure as I don't have a Mac).

hensldm:
Seems to work fine on an M1 mac (https://discord.com/channels/688807550715560050/688851976746041391/1318678178642006027)

AngheloAlf (approved):

Dragorn421 (approved):
I've held off from reviewing since I wrote some of this stuff and it feels like giving a medal to myself :1st_place_medal: but it looks like you need me :sunglasses:

PR #1763, opened 2024-12-15
Port over macOS fix

AngheloAlf:
Ports over the fix from https://github.com/zeldaret/oot/pull/2370

I haven't checked if there are other issues when building on macOS tho, since I don't have one.


hensldm (approved):

jpburnett (approved):
I also don't have a mac to verify the fixes, but looks good in theory.

PR #1764, opened 2024-12-16
Updated Docker Image

ptweezy:
- increased parity with OOO project
- updated to latest Ubuntu LTS version
- removed redundant build layers
- fixed Docker console warnings
- tested and confirmed OK with local Docker build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,7 @@ ENV LANG=C.UTF-8
 
 # Install Required Dependencies
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
-    apt-get update && \
-    apt-get install -y \
+    apt-get update && apt-get install -y \
     curl \
     build-essential \
     binutils-mips-linux-gnu \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
-FROM ubuntu:22.04 as build
+FROM ubuntu:24.04 AS build
+
 ENV TZ=UTC
+ENV LANG=C.UTF-8
 
 # Install Required Dependencies
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
@@ -10,6 +12,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     pkg-config \
     python3 \
     python3-pip \
+    python3-venv \
     git \
     wget \
     unzip \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -18,25 +21,15 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     clang-tidy-14 \
     clang-format-14 \
     libpng-dev && \
+    # Intall practicerom
+    curl https://practicerom.com/public/packages/debian/pgp.pub | apt-key add - && \ 
+    echo deb http://practicerom.com/public/packages/debian staging main >/etc/apt/sources.list.d/practicerom.list &&  \ 
+    apt update && \
```
AngheloAlf, line 27: I think we should use `apt-get` over plain `apt` on automated scripts, since `apt` is meant for user interaction.
I know this isn't something you introduced since the old docker script was doing using `apt`, but could you fix it?

AngheloAlf: Actually, it may be better to move the practice rom setup a bit above so we can put the `practicerom-dev` dependency with the other packages

ptweezy: Definitely agree with you. Changed to apt-get.

ptweezy: Also took your feedback and moved it up a bit. Let me know what you think.

Dockerfile
```diff
@@ -18,25 +25,14 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
...
-
 WORKDIR /mm
-
 RUN git config --global --add safe.directory /mm
 
-ENTRYPOINT ["/bin/bash", "-c"]
+CMD ["/bin/sh", "-c", \
+    "make -j $(nproc) init && \
+    echo Completed build. Open another terminal and run 'docker-compose exec mm sh' to shell into this running container. && \
+    tail -f /dev/null"]
```
AngheloAlf, line 38: I don't think setting up the docker image should build the rom too. Opening an interactive terminal would be better and more in line with the the workflow of the other build/usage methods

ptweezy: I actually did not intend to commit that change, that was something I was working on locally. Accidentally forgot to change it back to address your feedback :D

Dockerfile
```diff
@@ -1,15 +1,22 @@
...
+ENV LANG=C.UTF-8
 
 # Install Required Dependencies
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
-    apt-get update && apt-get install -y \
+    # Add source for practicerom-dev install
+    curl https://practicerom.com/public/packages/debian/pgp.pub | apt-key add - && \ 
+    echo deb http://practicerom.com/public/packages/debian staging main >/etc/apt/sources.list.d/practicerom.list &&  \ 
+    apt-get update && \
+    apt-get install -y \
```
jpburnett, line 12: Not sure on the dependencies of practicerom stuff, but update and install should be at the top.

ptweezy: Absolutely agree on the importance of updating ASAP. The lines above this add to the 'list' of things for apt-update to update, so we can have one apt-get update step instead of two as it is currently.

hensldm (approved):
Seems fine to me.

jpburnett (approved):
Nice, thanks for the improvement! :)

PR #1765, opened 2024-12-16
Add original asset names to all `object_y*` and `object_z*` files

tom-overton:
This wraps up this type of work for every `object_*` file. There are still the assets in overlays and in the `*_keep` files remaining. Some notes on this PR:
- I saw a note in `object_yb` stating that a particular animation doesn't work in Z64Utils. This was a little strange to me, since it worked fine for me, so I just removed the comment. You can see it load just fine here:
kihon
- Similarly, there was a comment in `z_en_yb.c` saying that loading that animation in-game crashes. Again, this was strange to me, since it worked fine for me, so I again just removed the comment. I just changed `EnYb_ChangeAnim(play, this, 2, ANIMMODE_LOOP, 0.0f)` to `EnYb_ChangeAnim(play, this, 0, ANIMMODE_LOOP, 0.0f)` and it worked as expected. Here's a video of it in action:

https://github.com/user-attachments/assets/dbf58fd5-57ef-493c-9e6c-8c12158039e3
- While I was in `object_yb`, I took a look at the blobs and figured them out. They're just vertices. I copied the comment from `object_bigpo` and slightly tweaked it, because similar to that file, the unused vertices were probably just combined with the rest of Kamaro's vertices.
- For `object_zl1` and `object_zl4`, MM3D combined them both into a single `zelda2_zl.gar.lzs`. I needed to use OoT3D's versions of the same assets to figure out a lot of these.
  - For `object_zl1`, it was easy enough to use OoT3D's `zelda_zl1.zar` to get all the animation names. 
  - For `object_zl4`, a lot of these animations were new to MM; the ones carried over from OoT could be grabbed from OoT3D's `zelda_zl4.zar`. For the ones new to MM, however, we have to use MM3D's object, and unfortunately all the unused animations are just T-poses in `zelda2_zl.gar.lzs`, making it a bit tricky to figure out which animation is which. However, we can be fairly confident that I got the names right, as not only are the animations arranged in alphabetic order (besides `demo_onegai2_wait` coming before `demo_onegai_wait`, not sure what happened there), but the Japanese names are fairly unambiguous ("bakutyu" means "backflip" for example; it's pretty easy to tell which animation that one is associated with).
- While I was in `object_zl1`, I fixed a typo in one of the animation names.
- While I was in `object_zl4`, I noticed one of the animations was out of order, so I moved it back to be in the correct order (it also makes the original animation names alphabetical, which is a secondary but still nice thing).

### DIFF
--- a/assets/xml/objects/object_yado_obj.xml
+++ b/assets/xml/objects/object_yado_obj.xml
@@ -1,7 +1,7 @@
 ﻿<Root>
     <File Name="object_yado_obj" Segment="6">
         <DList Name="object_yado_obj_DL_000320" Offset="0x320" />
-        <DList Name="object_yado_obj_DL_000430" Offset="0x430" />
+        <DList Name="object_yado_obj_DL_000430" Offset="0x430" /> <!-- Original name might be "w2_yad_win_model" -->
         <Texture Name="object_yado_obj_Tex_0005D8" OutName="tex_0005D8" Format="ia16" Width="16" Height="32" Offset="0x5D8" />
         <Texture Name="object_yado_obj_Tex_0009D8" OutName="tex_0009D8" Format="rgba16" Width="32" Height="16" Offset="0x9D8" />
         <Texture Name="object_yado_obj_Tex_000DD8" OutName="tex_000DD8" Format="i8" Width="16" Height="16" Offset="0xDD8" />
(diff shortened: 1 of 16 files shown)


hensldm (approved):

AngheloAlf (approved):

isghj5 (approved):
Interesting that mikau's animations have different names based on in-game and cutscene, like two different animators worked on the different sections..

edit: oh it was merged already

PR #1768, opened 2024-12-17
Apply the NORETURN attribute to functions that do not return

Thar0:
Some rationale in the corresponding OoT PR
https://github.com/zeldaret/oot/pull/1532

### DIFF
--- a/src/boot/fault.c
+++ b/src/boot/fault.c
@@ -1114,19 +1114,23 @@ void Fault_HangupFaultClient(const char* exp1, const char* exp2) {
  * error occurs. The parameters specify two messages detailing the error, one
  * or both may be NULL.
  */
-void Fault_AddHungupAndCrashImpl(const char* exp1, const char* exp2) {
+NORETURN void Fault_AddHungupAndCrashImpl(const char* exp1, const char* exp2) {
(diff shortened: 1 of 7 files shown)


engineer124 (approved):

hensldm (approved):

src/boot/fault.c
```diff
@@ -1114,19 +1114,23 @@ void Fault_HangupFaultClient(const char* exp1, const char* exp2) {
...
+    // this function does not return and the rest of the function is unreachable.
+    UNREACHABLE();
 }
 
 /**
  * Like `Fault_AddHungupAndCrashImpl`, however provides a fixed message containing
  * file and line number
  */
-void Fault_AddHungupAndCrash(const char* file, s32 line) {
+NORETURN void Fault_AddHungupAndCrash(const char* file, s32 line) {
```
AngheloAlf, line 1133: I think `_dbg_hungup` and `Reset` from `debug.c` should have the `NORETURN` attribute since they unconditionally call `Fault_AddHungupAndCrash` too

AngheloAlf (changes requested):

AngheloAlf (approved):

PR #1770, opened 2024-12-17
Add missing ExternalFile tags to object_dmask and object_osn

Thar0:
Skeletons in these objects have references to display lists in gameplay_keep, so they should rely on gameplay_keep as an ExternalFile.

Unfortunately though, adding these to the xmls do not appear to actually replace the hardcoded gameplay_keep pointers with symbols.. a ZAPD bug?

### DIFF
--- a/assets/xml/objects/object_dmask.xml
+++ b/assets/xml/objects/object_dmask.xml
@@ -1,4 +1,7 @@
 ﻿<Root>
+    <!-- Dependencies -->
+    <ExternalFile XmlPath="objects/gameplay_keep.xml" OutPath="assets/objects/gameplay_keep/"/>
+
     <File Name="object_dmask" Segment="6">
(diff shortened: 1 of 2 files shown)


hensldm (approved):

AngheloAlf (commented):
I don't think this PR actually does anything at all since we list `gameplay_keep` as a dependency for every other object by listing it on the global config (https://github.com/zeldaret/mm/blob/main/tools/ZAPDConfigs/MM/Config.xml#L7). This needs to be fixed on ZAPD's side

Thar0:
Lots of other objects explicitly list gameplay_keep as an ExternalFile, so if nothing else it makes these objects consistent with the rest. But yes, a ZAPD fix is needed.

AngheloAlf (approved):
In that case I think we should probably clean up the xmls. I guess we could remove gameplay_keep from the global config and just be explicit about it. Either way, that decision doesn't affect this PR.

PR #1771, opened 2024-12-17
Move gAudioTatumInit and gAudioHeapInitSizes to their own file, using data-with-rodata instead of const qualifiers

Thar0:
~~`gAudioHeapInitSizes` isn't really const since it is written to, but it seems necessary for matching. When `AVOID_UB` is defined, do not qualify `gAudioHeapInitSizes` with const so we can use a regular assignment where it is written to.~~

New strategy:
Split `gAudioTatumInit` and `gAudioHeapInitSizes` into `session_init.c`, which incrementally links .data into .rodata instead of using any const qualifiers.

### DIFF
--- a/src/audio/session_init.c
+++ b/src/audio/session_init.c
@@ -2,6 +2,7 @@
 #include "buffers.h"
 #include "assets/audio/sequence_sizes.h"
 #include "assets/audio/soundfont_sizes.h"
+
 #define SFX_SEQ_SIZE Sequence_0_SIZE
 #define AMBIENCE_SEQ_SIZE Sequence_1_SIZE
 #define SFX_SOUNDFONTS_SIZE (Soundfont_0_SIZE + Soundfont_1_SIZE + Soundfont_2_SIZE)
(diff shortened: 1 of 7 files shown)


AngheloAlf (approved):

AngheloAlf:
Marking as held until we have a bit of discussion on the corresponding OoT PR

src/audio/session_init.c
```diff
@@ -0,0 +1,25 @@
+#include "z64audio.h"
+#include "buffers.h"
+#include "assets/audio/sequence_sizes.h"
+#include "assets/audio/soundfont_sizes.h"
+#define SFX_SEQ_SIZE Sequence_0_SIZE
```
AngheloAlf, line 5:
```suggestion
#include "assets/audio/soundfont_sizes.h"

#define SFX_SEQ_SIZE Sequence_0_SIZE
```

hensldm (approved):

engineer124 (approved):

PR #1774, opened 2024-12-21
Add the bgcheck macros from OoT

Yanis002:
I checked if the surface types elements are placed the same as OoT and it is afaict, note that there's minor inconsistencies (like the "wall damage" thing that is still named `unk27`), I wasn't really sure so I thought this could be solved during review (or in the future)

I placed the macros at the beginning of the file since this is where most macros are located in this file, let me know if I should move the waterbox and surface types ones around the corresponding structs like it is on OoT (though for the waterbox macros there was already one at the top of the file so that's basically the whole reason I did it that way)


AngheloAlf (commented):
I see you added those macros but you aren't using them anywhere. Are they meant to be used by assets tools like ZAPD/fast64/any other tool (if any?)

Yanis002:
afaik it's for modding yeah, I'm already using those with Fast64 on OoT

Yanis002:
(on OoT it was added in https://github.com/zeldaret/oot/pull/1478 btw)

AngheloAlf (approved):

hensldm (approved):

PR #1783, opened 2025-01-07
Misc FAKE fixes

hensldm:
Tried going through and fixing things marked as `FAKE`.
Note there are some that I couldn't fully fix, but was able to find a way to match with an `if (1)` instead. In those cases I did change them to imo the better fake match.

### DIFF
--- a/src/code/z_player_lib.c
+++ b/src/code/z_player_lib.c
@@ -1687,9 +1687,8 @@ PlayerExplosive Player_GetExplosiveHeld(Player* player) {
 PlayerSword Player_SwordFromIA(Player* player, PlayerItemAction itemAction) {
     PlayerSword sword = PLAYER_SWORD_KOKIRI;
 
-    //! FAKE:
     if ((itemAction == PLAYER_IA_LAST_USED) ||
-        ((sword = GET_SWORD_FROM_IA(itemAction), (sword > PLAYER_SWORD_NONE)) && (sword < PLAYER_SWORD_MAX))) {
+        (sword = GET_SWORD_FROM_IA(itemAction), ((sword > PLAYER_SWORD_NONE) && (sword < PLAYER_SWORD_MAX)))) {
         return sword;
     }
 
--- a/src/code/z_player_lib.c
+++ b/src/code/z_player_lib.c
@@ -3802,29 +3802,28 @@ void Player_PostLimbDrawGameplay(PlayState* play, s32 limbIndex, Gfx** dList1, G
             }
         }
     } else if (limbIndex == PLAYER_LIMB_HEAD) {
-        //! FAKE:
-        if (((*dList1 != NULL) && ((((void)0, player->currentMask)) != (((void)0, PLAYER_MASK_NONE)))) &&
+        if (((*dList1 != NULL) && ((u32)player->currentMask != PLAYER_MASK_NONE)) &&
             (((player->transformation == PLAYER_FORM_HUMAN) &&
               ((player->skelAnime.animation != &gPlayerAnim_cl_setmask) || (player->skelAnime.curFrame >= 12.0f))) ||
              ((((player->transformation != PLAYER_FORM_HUMAN) && (player->currentMask >= PLAYER_MASK_FIERCE_DEITY)) &&
                ((player->transformation + PLAYER_MASK_FIERCE_DEITY) != player->currentMask)) &&
               (player->skelAnime.curFrame >= 10.0f)))) {
             if (func_80127438(play, player, player->currentMask)) {
-                s32 maskMinusOne = ((void)0, player->currentMask) - 1;
+                s32 maskMinusOne = player->currentMask - 1;
 
                 OPEN_DISPS(play->state.gfxCtx);
 
-                if (((void)0, player->currentMask) == PLAYER_MASK_COUPLE) {
+                if (player->currentMask == PLAYER_MASK_COUPLE) {
                     Player_DrawCouplesMask(play, player);
-                } else if (((void)0, player->currentMask) == PLAYER_MASK_CIRCUS_LEADER) {
+                } else if (player->currentMask == PLAYER_MASK_CIRCUS_LEADER) {
                     Player_DrawCircusLeadersMask(play, player);
-                } else if (((void)0, player->currentMask) == PLAYER_MASK_BLAST) {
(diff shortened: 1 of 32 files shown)


src/code/z_player_lib.c
```diff
@@ -1687,9 +1687,8 @@ PlayerExplosive Player_GetExplosiveHeld(Player* player) {
 PlayerSword Player_SwordFromIA(Player* player, PlayerItemAction itemAction) {
     PlayerSword sword = PLAYER_SWORD_KOKIRI;
 
-    //! FAKE:
     if ((itemAction == PLAYER_IA_LAST_USED) ||
```
hensldm, line 1690: Note I just removed the fake label here as I actually don't think this is a fake.

src/overlays/actors/ovl_Bg_Astr_Bombwall/z_bg_astr_bombwall.c
```diff
@@ -86,9 +86,7 @@ void BgAstrBombwall_InitCollider(ColliderTrisInit* init, Vec3f* pos, Vec3s* rot,
 
     for (i = 0; i < init->count; i++) {
         for (j = 0; j < 3; j++) {
-            //! FAKE:
-            // https://decomp.me/scratch/JrEnl
-            Matrix_MultVec3f(&(init->elements + i)->dim.vtx[j], &sp54[j]);
+            Matrix_MultVec3f(init->elements[i].dim.vtx + j, &sp54[j]);
```
hensldm, line 89: Looks better (though might still be considered a fake). Note this matches what was already in `func_809CE068` in `func_809CE068`

src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
```diff
@@ -188,8 +188,10 @@ void EnKakasi_Init(Actor* thisx, PlayState* play) {
     i = 0;
     csId = this->picto.actor.csId;
     while (csId != CS_ID_NONE) {
-        //! FAKE:
-        csId = CutsceneManager_GetAdditionalCsId(this->csIdList[i] = csId);
+        // clang-format off
+        this->csIdList[i] = csId; \
```
hensldm, line 192: Perhaps still a fake, but I prefer this over what it was. Possibly a macro of some kind?

src/code/z_player_lib.c
```diff
@@ -3803,29 +3802,28 @@ void Player_PostLimbDrawGameplay(PlayState* play, s32 limbIndex, Gfx** dList1, G
             }
         }
     } else if (limbIndex == PLAYER_LIMB_HEAD) {
-        //! FAKE:
-        if (((*dList1 != NULL) && ((((void)0, player->currentMask)) != (((void)0, PLAYER_MASK_NONE)))) &&
+        if (((*dList1 != NULL) && ((u32)player->currentMask != PLAYER_MASK_NONE)) &&
```
AngheloAlf, line 3805: Would this match too?
```suggestion
        if (((*dList1 != NULL) && ((PlayerMask)player->currentMask != PLAYER_MASK_NONE)) &&
```

src/overlays/actors/ovl_En_Kusa2/z_en_kusa2.c
```diff
@@ -404,12 +404,12 @@ EnKusa2UnkBssSubStruct* func_80A5C074(EnKusa2UnkBssStruct* arg0) {
...
     s32 i;
     EnKusa2UnkBssSubStruct2* phi_v1 = &arg0->unk_0480[0];
+    EnKusa2UnkBssSubStruct2* iter;
 
     for (i = 1; i < ARRAY_COUNT(D_80A5F1C0.unk_0480); i++) {
-        if (phi_v1->unk_2C > arg0->unk_0480[i].unk_2C) {
-            phi_v1 = &arg0->unk_0480[i];
-            //! FAKE:
-            if (1) {}
+        iter = &arg0->unk_0480[i];
```
AngheloAlf, line 410: Would it match if you moved the temp here?
```suggestion
        EnKusa2UnkBssSubStruct2* iter = &arg0->unk_0480[i];
```

src/overlays/actors/ovl_En_Po_Sisters/z_en_po_sisters.c
```diff
@@ -263,10 +263,8 @@ void EnPoSisters_MatchPlayerY(EnPoSisters* this, PlayState* play) {
     // equalize to player height
     Math_ApproachF(&this->actor.world.pos.y, player->actor.world.pos.y + 5.0f, 0.5f, 3.0f);
 
-    if (this->floatingBobbingTimer == 0) {
+    if ((u32)this->floatingBobbingTimer == 0) {
```
AngheloAlf, line 266: this is still kinda fake tho

src/overlays/actors/ovl_En_Trt/z_en_trt.c
```diff
@@ -1053,10 +1053,7 @@ void EnTrt_ItemGiven(EnTrt* this, PlayState* play) {
                 break;
 
             default:
-                //! FAKE:
-                if (1) {}
                 this->textId = 0x849;
```
AngheloAlf, line 1056: It may be a good idea to add a comment noting this is lacking the final `break;`

src/overlays/actors/ovl_En_Po_Sisters/z_en_po_sisters.c
AngheloAlf: Did you try `!this->floatingBobbingTimer`?

hensldm: `!this->floatingBobbingTimer` is probably the correct solution, yeah. However I thought we decided it was fine to use `u32` casts to still keep the explicit comparison to 0 (we at least have multiple examples of it in the repo already).

src/code/z_player_lib.c
hensldm: Does not work. The real solution would be `!player->currentMask`, but to keep the comparison to the enum we need the `u32` cast (see https://github.com/zeldaret/mm/pull/1783#discussion_r1906226979)

AngheloAlf (approved):

Dragorn421 (approved):

PR #1784, opened 2025-01-07
Sync soundfont_compiler.c with recent OoT change, give automatic names to sequence 0 channels that implement sfx

Thar0:
sfc source was change in oot as part of https://github.com/zeldaret/oot/pull/2396


include/audio/soundfont.h
```diff
@@ -28,11 +28,9 @@ typedef struct AdpcmBookHeader {
  * The procedure used to design the codeBook is based on an adaptive clustering algorithm.
  * The size of the codeBook is (8 * order * numPredictors) and is 8-byte aligned
  */
-typedef s16 AdpcmBookData[];
-
 typedef struct AdpcmBook {
     /* 0x0 */ AdpcmBookHeader header;
-    /* 0x8 */ AdpcmBookData codeBook;
+    /* 0x8 */ s16 codeBook[1];
```
hensldm, line 33: Is this causing the non matching build?

hensldm (approved):
Overall changes look good, though looks like we are getting an incorrect checksum on Jenkins.

include/audio/soundfont.h
Thar0: It was bss

AngheloAlf (approved):

PR #1785, opened 2025-01-07
Provide AVOID_UB for some bugs found in GCC compiler testing

Thar0:
Special thanks to @fig02 for the GCC playthrough that uncovered these, and for everyone else who looked into and discussed the problems.

The `z_message_nes.c` AVOID_UB is particularly nasty.. suggestions welcome

### DIFF
--- a/src/overlays/actors/ovl_En_Dnq/z_en_dnq.c
+++ b/src/overlays/actors/ovl_En_Dnq/z_en_dnq.c
@@ -465,6 +465,9 @@ void EnDnq_HandleCutscene(EnDnq* this, PlayState* play) {
 #ifdef AVOID_UB
         //! @bug Z2_DEKU_KINGCutsceneData_008258 provides a cue id of 6, reading out-of-bounds
         //! of this array that happens to be 0 padding.
+        //! Note that adding another idle anim entry still matches, since what follows this array is
+        //! section padding full of 0s. However in MM3D the data in this file is ordered differently,
+        //! proving that the array was originally missing an entry.
         // Avoid UB: Add an explicit array member rather than relying on 0 padding.
         DEKU_KING_ANIM_IDLE,
 #endif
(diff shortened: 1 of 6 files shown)


hensldm (approved):
Can't think of anything better for Message_DecodeNes unfortunately.

AngheloAlf (approved):

src/overlays/actors/ovl_En_Dnq/z_en_dnq.c
```diff
@@ -459,6 +459,12 @@ void EnDnq_HandleCutscene(EnDnq* this, PlayState* play) {
         DEKU_KING_ANIM_IDLE,          DEKU_KING_ANIM_IDLE_MORPH,
         DEKU_KING_ANIM_SURPRISE,      DEKU_KING_ANIM_JUMPED_ON_START,
         DEKU_KING_ANIM_JUMPED_ON_END, DEKU_KING_ANIM_JUMPED_ON_END_MORPH,
+#ifdef AVOID_UB
+        //! @bug Z2_DEKU_KINGCutsceneData_008258 provides a cue id of 6, reading out-of-bounds
+        //! of this array that happens to be 0 padding.
+        // Avoid UB: Add an explicit array member rather than relying on 0 padding.
+        DEKU_KING_ANIM_IDLE,
+#endif
```
fig02, line 470: should this mention mm3d and how it proves that it was infact an original bug?

fig02 (commented):
the bug in the dog actor would be good to include, dunno if thats planned for another time though

src/overlays/actors/ovl_En_Dnq/z_en_dnq.c
mzxrules: i would be interested to see how MM3D proves this to be padding, because it does currently match without the AVOID_UB tag

fig02: see discord (gcc thread)

Thar0:
I'm not sure where to put a comment about the dog actor bug, but since it doesn't have an associated AVOID_UB it could probably do with being done separately anyway

fig02 (approved):
fair enough, just dont want it to be forgotten. It will certainly leave my brain after this PR is merged

PR #1788, opened 2025-01-12
`En_Kaizoku` and `object_kz` documented: the pirate fighter miniboss

isghj5:
I haven't documented an actor in a long time, so I'm rusty and this may need more scrutiny.

A good amount of this documentation was taken from OOTs version of this actor. All of the object is named, all of the functions are named, however there are still several struct values and states undocumented.

Screenshot_2025-01-12_14-50-57


assets/xml/objects/object_kz.xml
```diff
@@ -1,93 +1,105 @@
...
-        <DList Name="object_kz_DL_00CF20" Offset="0xCF20" />
+        <DList Name="gKaizokuLeftFootDl" Offset="0xC318" />
+        <DList Name="gKaizokuRightFootDl" Offset="0xC4D8" />
+        <DList Name="gKaizokuLeftHandDl" Offset="0xC698" />
+        <DList Name="gKaizokuLeftSwordDl" Offset="0xC830" />
+        <DList Name="gKaizokuRightHandDl" Offset="0xCA08" />
+        <DList Name="gKaizokuRightSwordDl" Offset="0xCBA0" />
+        <DList Name="gKaizokuVeilDl" Offset="0xCD78" />
+        <DList Name="gKaizokuPonytailDl" Offset="0xCE30" />
+        <DList Name="gKaizokuHeadDl" Offset="0xCF20" />
```
AngheloAlf, line 65: Uppercase `DL` here too

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -12,31 +12,36 @@ typedef void (*EnKaizokuActionFunc)(struct EnKaizoku*, PlayState*);
 #define KAIZOKU_GET_EXIT_INDEX(thisx) ((thisx)->params & 0x3F)
 #define KAIZOKU_GET_SWITCH_FLAG(thisx) (((thisx)->params >> 6) & 0x7F)
 
+// used for text and for outfit color
+#define KAIZOKU_GET_TYPE(this) ((this)->picto.actor.world.rot.z)
+
 #define KAIZOKU_SWITCH_FLAG_NONE 0x7F
 
+// todo name these
```
AngheloAlf, line 20: The enum seems named, doesn't it?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -12,31 +12,36 @@ typedef void (*EnKaizokuActionFunc)(struct EnKaizoku*, PlayState*);
...
-    /* 13 */ EN_KAIZOKU_ANIM_13,
-    /* 14 */ EN_KAIZOKU_ANIM_14,
-    /* 15 */ EN_KAIZOKU_ANIM_15,
-    /* 16 */ EN_KAIZOKU_ANIM_16,
-    /* 17 */ EN_KAIZOKU_ANIM_17,
-    /* 18 */ EN_KAIZOKU_ANIM_18,
-    /* 19 */ EN_KAIZOKU_ANIM_MAX
+    /*  0 */ KAIZOKU_ANIM_FIGHTING_IDLE,
+    /*  1 */ KAIZOKU_ANIM_OOT_CONVERSATION, // giving player membership card?
+    /*  2 */ KAIZOKU_ANIM_OOT_JUMP, // replaced with _LAND
```
AngheloAlf, line 24: I think we can drop the `OOT` part of the name and just add a comment that they are unused remnants from that game instead

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -12,31 +12,36 @@ typedef void (*EnKaizokuActionFunc)(struct EnKaizoku*, PlayState*);
...
+    /*  3 */ KAIZOKU_ANIM_SIDESTEP,
+    /*  4 */ KAIZOKU_ANIM_WALK,
+    /*  5 */ KAIZOKU_ANIM_DAMAGE,
+    /*  6 */ KAIZOKU_ANIM_OOT_DEFEAT, // replaced with _DEFEAT
```
AngheloAlf, line 28: What's the difference between this one and the other DEFEAT?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -12,31 +12,36 @@ typedef void (*EnKaizokuActionFunc)(struct EnKaizoku*, PlayState*);
...
+    /* 13 */ KAIZOKU_ANIM_UNSHEATHE,
+    /* 14 */ KAIZOKU_ANIM_LOWER_WEAPONS,
+    /* 15 */ KAIZOKU_ANIM_DEMONSTRATIVE_SWORD_SWING,
+    /* 16 */ KAIZOKU_ANIM_DEFEAT,
+    /* 17 */ KAIZOKU_ANIM_DEFEAT_IDLE,
+    /* 18 */ KAIZOKU_ANIM_THROW_FLASH,
+    /* 19 */ KAIZOKU_ANIM_MAX
 } EnKaizokuAnimation;
 
+// this is a different size than limb enum in object..?
```
AngheloAlf, line 44: Bodyparts enums are almost always always smaller than limb enums.
You can check which one is which enum value corresponds to which limb value by looking at the enumerated limbs on the `if` just before the `this->bodyPartIndex++;`.
For example, `KAIZOKU_BODYPART_0` and `KAIZOKU_BODYPART_1` correspond to the torso and head respectively.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +61,50 @@ typedef enum EnKaizokuBodyPart {
...
     /* 0x2B4 */ s16 lookTimer;
-    /* 0x2B6 */ s16 unk_2B6;
-    /* 0x2B8 */ s16 unk_2B8;
+    /* 0x2B6 */ s16 iceTimer;
+    /* 0x2B8 */ s16 colorFilterTimer; // name taken from oot, damaged flashing red
     /* 0x2BA */ s16 drawDmgEffType;
     /* 0x2BC */ s16 switchFlag;
-    /* 0x2BE */ s16 unk_2BE;
+    /* 0x2BE */ s16 unused2BE;
     /* 0x2C0 */ f32 drawDmgEffScale;
```
AngheloAlf, line 85: This looks like padding
```suggestion
    /* 0x2C0 */ f32 drawDmgEffScale;
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +61,50 @@ typedef enum EnKaizokuBodyPart {
...
+    /* 0x2E0 */ f32 approachRate;
     /* 0x2E4 */ EnKaizokuAnimation animIndex;
     /* 0x2E8 */ s32 bodyPartIndex;
-    /* 0x2EC */ s32 unk_2EC;
-    /* 0x2F0 */ f32 unk_2F0;
-    /* 0x2F4 */ s16 unk_2F4;
-    /* 0x2F8 */ Vec3f unk_2F8;
-    /* 0x304 */ Vec3f unk_304;
+    /* 0x2EC */ s32 colorType;
+    /* 0x2F0 */ f32 unk_2F0; // modfified in Update, but not read/used?
```
AngheloAlf, line 102: It is used tho, it is passed to `Matrix_MultVecZ`

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +61,50 @@ typedef enum EnKaizokuBodyPart {
...
+    /* 0x2F4 */ s16 unk_2F4; // read but never set?
```
AngheloAlf, line 103: This and `unk_2F0` are used on https://github.com/zeldaret/mm/blob/74266850b09ac865cb2a9e5f208a903f4b5ec44c/src/overlays/actors/ovl_Obj_Boyo/z_obj_boyo.c#L100-L105

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -101,20 +113,20 @@ typedef struct EnKaizoku {
     /* 0x578 */ Vec3s headRot;
     /* 0x580 */ Vec3f rightFootPos;
     /* 0x58C */ Vec3f leftFootPos;
-    /* 0x598 */ s32 unk_598;
-    /* 0x59C */ s16 unk_59C;
+    /* 0x598 */ s32 cutsceneTimer; // why is it double len?
```
AngheloAlf, line 116: What do you mean by double len?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -101,20 +113,20 @@ typedef struct EnKaizoku {
...
+    /* 0x59C */ s16 cutsceneState; 
     /* 0x59E */ s16 subCamId;
-    /* 0x5A0 */ s16 unk_5A0;
-    /* 0x5A2 */ s16 unk_5A2;
+    /* 0x5A0 */ s16 flashTimer;
+    /* 0x5A2 */ s16 unused5A2;
```
AngheloAlf, line 120: This looks like padding too, better remove it

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -258,22 +269,24 @@ void EnKaizoku_Init(Actor* thisx, PlayState* play) {
...
+    if (this->textType >= 2) {
+        this->textType = 0;
+    } else if (this->textType == 2) { // @Bug: '== 2' also matches '>= 2'
+        this->textType = 0;
     }
 
-    this->unk_2EC = this->picto.actor.world.rot.z;
-    this->picto.actor.world.rot.z = 0;
+    this->colorType = KAIZOKU_GET_TYPE(this);
+    KAIZOKU_GET_TYPE(this) = 0; // clear param, which was rot.z, resets skew
```
AngheloAlf, line 286: It may be better to have a `SET_TYPE` instead of abusing the getter in this way

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -396,44 +409,45 @@ s32 func_80B85A00(EnKaizoku* this, PlayState* play, s16 arg2) {
...
-            sp2A = player->actor.shape.rot.y - this->picto.actor.shape.rot.y;
+            angleToFacingLink = player->actor.shape.rot.y - this->picto.actor.shape.rot.y;
             if ((this->picto.actor.xzDistToPlayer <= 65.0f) && !Actor_OtherIsTargeted(play, &this->picto.actor) &&
-                (ABS_ALT(sp2A) < 0x5000)) {
-                if (this->action != KAIZOKU_ACTION_9) {
-                    func_80B87F70(this);
+                (ABS_ALT(angleToFacingLink) < 0x5000)) {
+                if (this->action != KAIZOKU_ACTION_SLASH) {
+                    EnKaizoku_SetupSlash(this);
                     return 1;
```
AngheloAlf, line 447: Just noticed this function shuold be returning a boolean
```suggestion
                    return true;
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -750,122 +766,126 @@ void func_80B868B8(EnKaizoku* this, PlayState* play) {
...
+                    dekuNut->gravity = -10.0f;
                     play->envCtx.fillScreen = true;
                 }
             }
 
-            if (this->unk_5A0 >= 0x12) {
-                Math_ApproachF(&this->unk_5EC, 60.0f, 1.0f, 20.0f);
-                this->unk_5E8 = this->unk_5EC / 60.0f;
-                play->envCtx.screenFillColor[3] = this->unk_5E8 * 255.0f;
+            if (this->flashTimer >= 0x12) {
```
AngheloAlf, line 860: iirc we write timers in dec instead of hex
```suggestion
            if (this->flashTimer >= 18) {
```
Could you do the same for the other uses too?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1023,112 +1044,114 @@ void func_80B8760C(EnKaizoku* this, PlayState* play) {
...
+        this->dontUpdateSkel = false;
+        angleToLink = this->picto.actor.yawTowardsPlayer - this->picto.actor.shape.rot.y;
+        if ((ABS_ALT(angleToLink) <= 0x4000) && (this->picto.actor.xzDistToPlayer < 40.0f) &&
             (fabsf(this->picto.actor.playerHeightRel) < 50.0f)) {
             if (func_800BE184(play, &this->picto.actor, 100.0f, 10000, 0x4000, this->picto.actor.shape.rot.y)) {
                 if (player->meleeWeaponAnimation == PLAYER_MWA_JUMPSLASH_START) {
                     this->bodyCollider.base.acFlags &= ~AC_HARD;
-                    func_80B874D8(this, play);
+                    EnKaizoku_SetupSpinDodge(this, play);
                 } else if (!(play->gameplayFrames & 1)) {
```
AngheloAlf, line 1079: I know it wasn't you, but would using a `% 2` instead matches here?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
isghj5: The old one is cut off, the first few frames of both look to be the same or inspired upon the previous, but then they hop back onto their feet, so that the animation blends into DEFEAT_IDLE

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
isghj5: Most timers are only s16, and even that is far larger than required, where most timers could be one unsigned byte without issues. Why would this one be double the bitsize?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
isghj5: Hmm, What should "KAIZOKU_ANIM_OOT_CONVERSATION" be renamed to then? There are 4 other animations used in the game that actually involve talking, and this one does not in this game.

AngheloAlf (approved):

assets/xml/objects/object_kz.xml
```diff
@@ -1,93 +1,106 @@
...
-        <Texture Name="object_kz_Tex_005AD0" OutName="tex_005AD0" Format="ci8" Width="8" Height="8" Offset="0x5AD0" />
-        <Texture Name="object_kz_Tex_005B10" OutName="tex_005B10" Format="ci8" Width="8" Height="8" Offset="0x5B10" />
-        <Texture Name="object_kz_Tex_005B50" OutName="tex_005B50" Format="ci8" Width="16" Height="16" Offset="0x5B50" />
-        <Texture Name="object_kz_Tex_005C50" OutName="tex_005C50" Format="i8" Width="16" Height="16" Offset="0x5C50" />
-        <Texture Name="object_kz_Tex_005D50" OutName="tex_005D50" Format="ci8" Width="16" Height="16" Offset="0x5D50" />
-        <Texture Name="object_kz_Tex_005E50" OutName="tex_005E50" Format="i8" Width="16" Height="16" Offset="0x5E50" />
-        <Texture Name="object_kz_Tex_005F50" OutName="tex_005F50" Format="i8" Width="8" Height="16" Offset="0x5F50" />
+        <Texture Name="gKaizoku1TLUT" OutName="kaizoku_tlut1" Format="rgba16" Width="16" Height="16" Offset="0x58D0" />
+        <!-- This is the same skin texture as below, but this one is her arms and back -->
+        <Texture Name="gKaizokuSkinGradient1Tex" OutName="skin1" Format="ci8" Width="8" Height="8" Offset="0x5AD0" TlutOffset="0x58D0" />
```
hensldm, line 20: tbh, I've never been a big fan of SkinGradient since I think `Skin` by itself is sufficient.

assets/xml/objects/object_kz.xml
```diff
@@ -1,93 +1,106 @@
...
+        <Texture Name="gKaizokuShirtTopRedGradientTex" OutName="shirt_burgundy" Format="ci8" Width="8" Height="8" Offset="0x5B10" />
```
hensldm, line 21: Outname differs from texture name?

assets/xml/objects/object_kz.xml
```diff
@@ -1,93 +1,106 @@
...
+        <DList Name="gKaizokuWaistDL" Offset="0x87F0" />
+        <DList Name="gKaizokuRightThighDL" Offset="0x8A10" />
+        <DList Name="gKaizokuRightShinDL" Offset="0x8B10" />
+        <DList Name="gKaizokuLeftThighDL" Offset="0x8D80" />
+        <DList Name="gKaizokuLeftShinDL" Offset="0x8E80" />
 
-        <Texture Name="object_kz_TLUT_009100" OutName="tlut_009100" Format="rgba16" Width="252" Height="1" Offset="0x9100" />
-        <Texture Name="object_kz_Tex_0092F8" OutName="tex_0092F8" Format="ci8" Width="8" Height="8" Offset="0x92F8" />
-        <Texture Name="object_kz_Tex_009338" OutName="tex_009338" Format="ci8" Width="8" Height="16" Offset="0x9338" />
+        <Texture Name="gKaizoku2TLUT" OutName="kaizoku_tlut2" Format="rgba16" Width="4" Height="63" Offset="0x9100" />
```
hensldm, line 41: Would probably just remove `kaizoku` from the outname since everything else you've done doesn't include it (note there are other outnames this would apply to).

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +59,49 @@ typedef enum EnKaizokuBodyPart {
     /* 15 */ KAIZOKU_BODYPART_MAX
 } EnKaizokuBodyPart;
 
+typedef enum KaizokuTextOffset{
```
hensldm, line 62:
```suggestion
typedef enum KaizokuTextOffset {
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
hensldm: Felt like it was just preference of the developer. I would say the comment isn't really necessary though.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +59,49 @@ typedef enum EnKaizokuBodyPart {
...
+    /* 0x2B6 */ s16 iceTimer;
+    /* 0x2B8 */ s16 colorFilterTimer; // name taken from oot, damaged flashing red
     /* 0x2BA */ s16 drawDmgEffType;
     /* 0x2BC */ s16 switchFlag;
-    /* 0x2BE */ s16 unk_2BE;
     /* 0x2C0 */ f32 drawDmgEffScale;
     /* 0x2C4 */ f32 drawDmgEffFrozenSteamScale;
-    /* 0x2C8 */ s16 unk_2C8;
-    /* 0x2CA */ s16 unk_2CA;
+    /* 0x2C8 */ s16 textidOffset; // which of the 4 text states during interaction
```
hensldm, line 84:
```suggestion
    /* 0x2C8 */ s16 textIdOffset; // which of the 4 text states during interaction 
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -258,22 +269,24 @@ void EnKaizoku_Init(Actor* thisx, PlayState* play) {
...
 
-    if (this->unk_2CA >= 2) {
-        this->unk_2CA = 0;
-    } else if (this->unk_2CA == 2) {
-        this->unk_2CA = 0;
+    // There are three spawns (0,1,2)
+    // they decided to re-align 2 into 0 instead of making a third text combination
+    if (this->textType >= 2) {
+        this->textType = 0;
+    } else if (this->textType == 2) { // @Bug: '== 2' also matches '>= 2'
```
hensldm, line 281:
```suggestion
    } else if (this->textType == 2) { //! @Bug: '== 2' also matches '>= 2'
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -356,38 +369,38 @@ void func_80B85900(EnKaizoku* this) {
     }
 }
 
-s32 func_80B85A00(EnKaizoku* this, PlayState* play, s16 arg2) {
+s32 EnKaizoku_ReactToPlayer(EnKaizoku* this, PlayState* play, s16 arg2) {
     Player* player = GET_PLAYER(play);
-    s16 sp32 = ABS_ALT(this->picto.actor.wallYaw - this->picto.actor.shape.rot.y);
-    s16 sp30 = ABS_ALT(this->picto.actor.yawTowardsPlayer - this->picto.actor.shape.rot.y);
+    s16 angleToWall = ABS_ALT(this->picto.actor.wallYaw - this->picto.actor.shape.rot.y);
+    s16 angleToLink = ABS_ALT(this->picto.actor.yawTowardsPlayer - this->picto.actor.shape.rot.y);
```
hensldm, line 375:
```suggestion
    s16 angleToPlayer = ABS_ALT(this->picto.actor.yawTowardsPlayer - this->picto.actor.shape.rot.y);
```

There are a few of things could you change them all?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -93,63 +104,63 @@ Vec3f D_80B8AA04[] = {
...
 };
 
-Color_RGBA8 D_80B8AB30[] = {
-    { 255, 130, 10, 255 },
-    { 185, 130, 210, 255 },
-    { 135, 195, 80, 255 },
+static Color_RGBA8 sKaizokuOutfitColors[] = {
+    { 255, 130, 10, 255 },  // orange
+    { 185, 130, 210, 255 }, // lavender
+    { 135, 195, 80, 255 },  // pale green
```
hensldm, line 116: Hmm, thoughts on perhaps having enums for these?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -527,17 +541,17 @@ void func_80B85FA8(EnKaizoku* this, PlayState* play) {
...
                 this->picto.actor.gravity = -2.0f;
             }
             break;
 
-        case 2:
+        case 2: // waiting for fall to land
             if (this->picto.actor.bgCheckFlags & BGCHECKFLAG_GROUND) {
-                if (this->unk_2D8 != 0) {
-                    this->unk_2D8 = 0;
+                if (this->dontUpdateSkel != false) {
```
hensldm, line 553:
```suggestion
                if (this->dontUpdateSkel) {
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -667,7 +682,8 @@ void func_80B85FA8(EnKaizoku* this, PlayState* play) {
     }
 }
 
-void func_80B86804(EnKaizoku* this, PlayState* play) {
+// Player has lost the fight
+void EnKaizoku_SetupLoss(EnKaizoku* this, PlayState* play) {
```
hensldm, line 686: I would probably suggest `SetupWin` since we try to keep things in perspective of the actor (another suggestion is `SetupPlayerLoss`).

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1023,112 +1044,114 @@ void func_80B8760C(EnKaizoku* this, PlayState* play) {
...
+        if ((ABS_ALT(angleToLink) <= 0x4000) && (this->picto.actor.xzDistToPlayer < 40.0f) &&
             (fabsf(this->picto.actor.playerHeightRel) < 50.0f)) {
             if (func_800BE184(play, &this->picto.actor, 100.0f, 10000, 0x4000, this->picto.actor.shape.rot.y)) {
                 if (player->meleeWeaponAnimation == PLAYER_MWA_JUMPSLASH_START) {
                     this->bodyCollider.base.acFlags &= ~AC_HARD;
-                    func_80B874D8(this, play);
-                } else if (!(play->gameplayFrames & 1)) {
-                    func_80B87900(this);
+                    EnKaizoku_SetupSpinDodge(this, play);
+                } else if (!(play->gameplayFrames % 2)) {
```
hensldm, line 1079:
```suggestion
                } else if ((play->gameplayFrames % 2) == 0) {
```
Much more of a preference when use modulos.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
isghj5: Yeah I was worried about the perspective, can put player in the function names.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +59,49 @@ typedef enum EnKaizokuBodyPart {
...
     /* 0x2D2 */ s16 swordState;
     /* 0x2D4 */ s16 exitIndex;
     /* 0x2D6 */ s16 csId;
-    /* 0x2D8 */ u8 unk_2D8;
-    /* 0x2D9 */ u8 unk_2D9;
+    /* 0x2D8 */ u8 dontUpdateSkel;
+    /* 0x2D9 */ u8 defeatBreathingStarted;
     /* 0x2DC */ f32 animEndFrame;
-    /* 0x2E0 */ f32 unk_2E0;
+    /* 0x2E0 */ f32 approachRate;
```
tom-overton, line 95: Doesn't seem to be used for approaching in `EnKaizoku_SpinDodge`. I see an equivalent variable is called `circlingRate` in EnDinolfos; maybe that's a better name for this one?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -101,20 +110,19 @@ typedef struct EnKaizoku {
...
-    /* 0x5C8 */ Vec3f unk_5C8;
-    /* 0x5D4 */ Vec3f unk_5D4;
-    /* 0x5E0 */ f32 unk_5E0;
-    /* 0x5E4 */ f32 unk_5E4;
-    /* 0x5E8 */ f32 unk_5E8;
-    /* 0x5EC */ f32 unk_5EC;
+    /* 0x5C8 */ Vec3f subCamEyeTarget;
+    /* 0x5D4 */ Vec3f subCamAtTarget;
+    /* 0x5E0 */ f32 subCamVelocity;
+    /* 0x5E4 */ f32 unused5E4;
```
tom-overton, line 123: Would prefer to just leave this as `unk_5E4` to be perfectly honest. That tends to be how we've been handling unused variables that we can't remove in other actors.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +59,49 @@ typedef enum EnKaizokuBodyPart {
...
     /* 0x2CC */ s16 eyeIndex;
     /* 0x2CE */ s16 blinkTimer;
-    /* 0x2D0 */ s16 unk_2D0;
+    /* 0x2D0 */ s16 spinAttackState;
     /* 0x2D2 */ s16 swordState;
     /* 0x2D4 */ s16 exitIndex;
     /* 0x2D6 */ s16 csId;
-    /* 0x2D8 */ u8 unk_2D8;
-    /* 0x2D9 */ u8 unk_2D9;
+    /* 0x2D8 */ u8 dontUpdateSkel;
```
tom-overton, line 92: Maybe something like `disableAnimations`/`pauseAnimations`/`freezeAnimations` might work a bit cleaner for this one? I think those make it a little more obvious what will happen when you set this to `true`.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +59,49 @@ typedef enum EnKaizokuBodyPart {
     /* 15 */ KAIZOKU_BODYPART_MAX
 } EnKaizokuBodyPart;
 
+typedef enum KaizokuTextOffset {
+    /*  0 */ KAIZOKU_COVERSATION_INTRO_1, // shout before landing
+    /*  1 */ KAIZOKU_COVERSATION_INTRO_2, // after landing
+    /*  2 */ KAIZOKU_COVERSATION_WIN,  // after losing to player
+    /*  3 */ KAIZOKU_COVERSATION_LOSS  // after defeating player 
+} KaizokuTextOffset;
```
tom-overton, line 67: Seems like you could just name these in a way such that the comments (and the disambiguation between `INTRO_1` and `INTRO_2`) are unnecessary, e.g.:
```c
typedef enum KaizokuTextOffset {
    /*  0 */ KAIZOKU_COVERSATION_SHOUT_FROM_ABOVE,
    /*  1 */ KAIZOKU_COVERSATION_INTRO,
    /*  2 */ KAIZOKU_COVERSATION_DEFEAT,
    /*  3 */ KAIZOKU_COVERSATION_VICTORY
} KaizokuTextOffset;
```
I renamed the latter two to have them from the actor's perspective. If you *really* wanted to be explicit about who is defeating who, you could name the latter two like:
```c
    /*  2 */ KAIZOKU_COVERSATION_DEFEATED_BY_PLAYER,
    /*  3 */ KAIZOKU_COVERSATION_DEFEATED_PLAYER
```

Also, the enum is called `KaizokuTextOffset`, but then the values are prefixed with `KAIZOKU_CONVERSATION`. Should the enum be `KaizokuConversation`, or should these be prefixed with `KAIZOKU_TEXT_OFFSET_`? I *think* it's probably the latter, but I'm not sure.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -12,29 +12,32 @@ typedef void (*EnKaizokuActionFunc)(struct EnKaizoku*, PlayState*);
...
-    /* 13 */ EN_KAIZOKU_ANIM_13,
-    /* 14 */ EN_KAIZOKU_ANIM_14,
-    /* 15 */ EN_KAIZOKU_ANIM_15,
-    /* 16 */ EN_KAIZOKU_ANIM_16,
-    /* 17 */ EN_KAIZOKU_ANIM_17,
-    /* 18 */ EN_KAIZOKU_ANIM_18,
-    /* 19 */ EN_KAIZOKU_ANIM_MAX
+    /*  0 */ KAIZOKU_ANIM_FIGHTING_IDLE,
+    /*  1 */ KAIZOKU_ANIM_UNUSED_CONVERSATION, // from OOT: giving player membership card?
+    /*  2 */ KAIZOKU_ANIM_UNUSED_JUMP, // from OOT: replaced with _LAND
```
tom-overton, line 23: Are you sure about that? It looks like `KAIZOKU_ANIM_LAND` is used for when the pirate lands from above in the intro cutscene, whereas in OoT, the jump animation was used when the fighter flees after being defeated.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -12,29 +12,32 @@ typedef void (*EnKaizokuActionFunc)(struct EnKaizoku*, PlayState*);
...
+    /*  3 */ KAIZOKU_ANIM_SIDESTEP,
+    /*  4 */ KAIZOKU_ANIM_WALK,
+    /*  5 */ KAIZOKU_ANIM_DAMAGE,
+    /*  6 */ KAIZOKU_ANIM_UNUSED_DEFEAT, // from OOT: replaced with _DEFEAT
```
tom-overton, line 27:
```suggestion
    /*  6 */ KAIZOKU_ANIM_UNUSED_DEFEAT, // from OOT: replaced with `KAIZOKU_ANIM_DEFEAT`
```
I would prefer to write out the whole thing, especially since this animation *already* contains the string `_DEFEAT` in it.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -12,29 +12,32 @@ typedef void (*EnKaizokuActionFunc)(struct EnKaizoku*, PlayState*);
...
-    /* 12 */ EN_KAIZOKU_ANIM_12,
-    /* 13 */ EN_KAIZOKU_ANIM_13,
-    /* 14 */ EN_KAIZOKU_ANIM_14,
-    /* 15 */ EN_KAIZOKU_ANIM_15,
-    /* 16 */ EN_KAIZOKU_ANIM_16,
-    /* 17 */ EN_KAIZOKU_ANIM_17,
-    /* 18 */ EN_KAIZOKU_ANIM_18,
-    /* 19 */ EN_KAIZOKU_ANIM_MAX
+    /*  0 */ KAIZOKU_ANIM_FIGHTING_IDLE,
+    /*  1 */ KAIZOKU_ANIM_UNUSED_CONVERSATION, // from OOT: giving player membership card?
```
tom-overton, line 22: You can delete the question mark here; the animation (which is called `gGerudoRedStandAnim` in OoT) is used by EnGe3, an actor whose only purpose is to give you the Gerudo Membership Card.

assets/xml/objects/object_kz.xml
```diff
@@ -1,93 +1,106 @@
...
-        <Animation Name="object_kz_Anim_00ED1C" Offset="0xED1C" /> <!-- Original name is "kz_syutawait" -->
-        <Animation Name="object_kz_Anim_00EF9C" Offset="0xEF9C" /> <!-- Original name is "kz_talk" -->
-        <Animation Name="object_kz_Anim_00F288" Offset="0xF288" /> <!-- Original name is "kz_togetout" -->
-        <Animation Name="object_kz_Anim_00F8E4" Offset="0xF8E4" /> <!-- Original name is "kz_wait" -->
+        <Animation Name="gKaizokuSidestepAnim" Offset="0xDBE4" /> <!-- Original name is "kz_side_walk" -->
+        <!-- This is the old version from OOT, back when she would jump to escape instead of talking and flashing to escape -->
+        <Animation Name="gKaizokuUnusedJumpAnim" Offset="0xE1C8" /> <!-- Original name is "kz_start" -->
+        <Animation Name="gKaizokuLandAnim" Offset="0xE8BC" /> <!-- Original name is "kz_syuta" -->
+        <Animation Name="gKaizokuChallengeTalkAnim" Offset="0xED1C" /> <!-- Original name is "kz_syutawait" -->
+        <Animation Name="gKaizokuUnusedConversationAnim" Offset="0xEF9C" /> <!-- Original name is "kz_talk" -->
```
tom-overton, line 100: For the first one of these, maybe something like `LandTalkAnim` or `LandIdleAnim` would work? The reason I bring that up is because `gKaizokuLandAnim` perfectly transitions into it, and because the original animation names pretty clearly show that these two animations are related. For the second one of these, I would prefer to use `Talk` instead of `Conversation` (e.g. `gKaizokuUnusedTalkAnim`) because there isn't a single animation currently named `Conversation` in the whole repo and for consistency with the previous animation.

tom-overton (commented):
I didn't get time to look at the C file today, but I'll look at it tomorrow.

assets/xml/objects/object_kz.xml
```diff
@@ -1,93 +1,106 @@
...
+        <Limb Name="gKaizokuLeftForearmLimb" Type="Standard" EnumName="KAIZOKU_LIMB_L_FOREARM" Offset="0xD748" />
+        <Limb Name="gKaizokuLeftWristLimb" Type="Standard" EnumName="KAIZOKU_LIMB_L_WRIST" Offset="0xD754" />
+        <Limb Name="gKaizokuLeftHandLimb" Type="Standard" EnumName="KAIZOKU_LIMB_L_HAND" Offset="0xD760" />
+        <Limb Name="gKaizokuLeftSwordLimb" Type="Standard" EnumName="KAIZOKU_LIMB_L_SWORD" Offset="0xD76C" />
+        <Limb Name="gKaizokuLeftThighLimb" Type="Standard" EnumName="KAIZOKU_LIMB_L_THIGH" Offset="0xD778" />
+        <Limb Name="gKaizokuLeftShinLimb" Type="Standard" EnumName="KAIZOKU_LIMB_L_SHIN" Offset="0xD784" />
+        <Limb Name="gKaizokuLeftFootLimb" Type="Standard" EnumName="KAIZOKU_LIMB_L_FOOT" Offset="0xD790" />
+        <Limb Name="gKaizokuRightThighLimb" Type="Standard" EnumName="KAIZOKU_LIMB_R_THIGH" Offset="0xD79C" />
+        <Limb Name="gKaizokuRightShinLimb" Type="Standard" EnumName="KAIZOKU_LIMB_R_SHIN" Offset="0xD7A8" />
+        <Limb Name="gKaizokuRightFootLimb" Type="Standard" EnumName="KAIZOKU_LIMB_R_FOOT" Offset="0xD7B4" />
```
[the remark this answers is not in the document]
isghj5: I kinda agree with oot here, its shorter and still understandable, but will change to match the project.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
isghj5: These sound like actions/commands rather than states tho, animationsDisabled?

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
isghj5: I want to avoid both being called "defeat" with a small not "by" that confuses them at a glance

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
isghj5: for some reason when I skimmed OOT's actor, I thought I saw it use the jump animation for the flee function but also for the intro, when they land from the ceiling.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
isghj5: I can return it, but I change every variable as part of documentation so I can keep track of what variables I have inspected, leaving a variable completely auto-generated gives a sense that it's completely unfinished

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
hensldm: `animationsDisabled` seems good to me.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1604,99 +1641,99 @@ void func_80B89280(EnKaizoku* this, PlayState* play) {
...
+            this->drawDmgEffTimer = 0;
             this->drawDmgEffType = ACTOR_DRAW_DMGEFF_FIRE;
             this->picto.actor.flags |= ACTOR_FLAG_HOOKSHOT_PULLS_PLAYER;
         }
     }
 }
 
-void func_80B893CC(EnKaizoku* this, PlayState* play) {
+void EnKaizoku_SetupDamaged(EnKaizoku* this, PlayState* play) {
     Vec3f sp34;
```
hensldm, line 1661: This you could just name this `velocity` since it gets directly copied into `this->velocity`

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1418,76 +1457,76 @@ void func_80B88964(EnKaizoku* this, PlayState* play) {
...
+    this->circlingRate = 0.0f;
 }
 
-void func_80B88D6C(EnKaizoku* this, PlayState* play) {
+void EnKaizoku_Circle(EnKaizoku* this, PlayState* play) {
     Player* player = GET_PLAYER(play);
-    s16 sp2A;
-    s32 temp_ft4;
-    s32 temp_ft5;
+    s16 angleBehindLink;
```
hensldm, line 1515:
```suggestion
    s16 angleBehindPlayer;
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1380,28 +1418,29 @@ void func_80B887AC(EnKaizoku* this, PlayState* play) {
...
     this->picto.actor.speed = 0.0f;
 }
 
-void func_80B88964(EnKaizoku* this, PlayState* play) {
+void EnKaizoku_SpinAttack(EnKaizoku* this, PlayState* play) {
     Player* player = GET_PLAYER(play);
     f32 curFrame = this->skelAnime.curFrame;
-    s16 v1;
-    s16 a0;
+    s16 angleFacingLink;
```
hensldm, line 1433:
```suggestion
    s16 yawDiff;
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1204,83 +1240,83 @@ void func_80B87FDC(EnKaizoku* this, PlayState* play2) {
...
 
-void func_80B88378(EnKaizoku* this, PlayState* play) {
+void EnKaizoku_Advance(EnKaizoku* this, PlayState* play) {
     Player* player = GET_PLAYER(play);
-    s32 sp30;
-    s32 sp2C;
-    s16 sp2A;
+    s32 beforeCurFrame;
+    s32 afterCurFrame;
+    s16 facingAngleToLink;
```
hensldm, line 1317:
```suggestion
    s16 yawDiff;
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -396,44 +421,45 @@ s32 func_80B85A00(EnKaizoku* this, PlayState* play, s16 arg2) {
...
 
     if (arg2) {
-        s16 sp2A;
+        s16 angleToFacingLink;
 
         //! FAKE:
     label:;
-        if (sp30 >= 10000) {
-            func_80B87900(this);
+        if (angleToPlayer >= 10000) {
```
hensldm, line 450:
```suggestion
        if (angleToPlayer >= 0x2710) {
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -396,44 +421,45 @@ s32 func_80B85A00(EnKaizoku* this, PlayState* play, s16 arg2) {
...
 
-        if (this->action != KAIZOKU_ACTION_7) {
-            func_80B87E28(this);
+        if (this->action != KAIZOKU_ACTION_ROLL_BACK) {
+            EnKaizoku_SetupRollBack(this);
         }
         return true;
     }
 
     if (arg2) {
```
hensldm, line 445: Hmm I wonder if we can come up with a name for this arg2. Not quite sure what yet though

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.h
```diff
@@ -56,43 +59,49 @@ typedef enum EnKaizokuBodyPart {
...
     /* 0x2CC */ s16 eyeIndex;
     /* 0x2CE */ s16 blinkTimer;
-    /* 0x2D0 */ s16 unk_2D0;
+    /* 0x2D0 */ s16 spinAttackState;
     /* 0x2D2 */ s16 swordState;
     /* 0x2D4 */ s16 exitIndex;
     /* 0x2D6 */ s16 csId;
-    /* 0x2D8 */ u8 unk_2D8;
-    /* 0x2D9 */ u8 unk_2D9;
+    /* 0x2D8 */ u8 dontUpdateSkel;
```
hensldm, line 92: Did you mean to update this name when you resolved the other conversation?

isghj5: Oh weird, yeah I meant to change that.

hensldm (approved):

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -93,63 +104,75 @@ Vec3f D_80B8AA04[] = {
...
+    { 185, 130, 210, 255 }, // KAIZOKU_OUTFIT_LAVENDER
+    { 135, 195, 80, 255 },  // KAIZOKU_OUTFIT_GREEN
 };
 
-typedef enum EnKaizokuDamageEffect {
+typedef enum KaizokuDamageEffect {
     /* 0x0 */ KAIZOKU_DMGEFF_NONE,
-    /* 0x1 */ KAIZOKU_DMGEFF_1,
-    /* 0x2 */ KAIZOKU_DMGEFF_2,
+    /* 0x1 */ KAIZOKU_DMGEFF_STUNNED,
```
tom-overton, line 133:
```suggestion
    /* 0x1 */ KAIZOKU_DMGEFF_STUN,
```
I would prefer this for a few reasons:
1. It's more consistent with other enemy actors
2. It differentiates it a little more from `KAIZOKU_DMGEFF_STUNNED_ONLY`

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -93,63 +104,75 @@ Vec3f D_80B8AA04[] = {
...
     /* 0x3 */ KAIZOKU_DMGEFF_FREEZE,
-    /* 0x4 */ KAIZOKU_DMGEFF_4,
-    /* 0x5 */ KAIZOKU_DMGEFF_5,
-    /* 0xD */ KAIZOKU_DMGEFF_D = 0xD,
-    /* 0xE */ KAIZOKU_DMGEFF_E,
-    /* 0xF */ KAIZOKU_DMGEFF_F
-} EnKaizokuDamageEffect;
+    /* 0x4 */ KAIZOKU_DMGEFF_LIGHT_ARROW,
+    /* 0x5 */ KAIZOKU_DMGEFF_ZORA_SHIELD,
+    /* 0xD */ KAIZOKU_DMGEFF_STUNNED_ONLY = 0xD, // smashed? it checks stun first?
```
tom-overton, line 138:
```suggestion
    /* 0xD */ KAIZOKU_DMGEFF_STUNNED_ONLY = 0xD, // deals damage only if the pirate is stunned; she jumps away otherwise
```
I feel like this comment could be a little better at describing what this effect does. I'm not picky about the wording, so feel free to reword it if you'd like; anything that describes this piece of logic in one sentence will do:
```c
            case KAIZOKU_DMGEFF_STUNNED_ONLY:
                if (this->action == KAIZOKU_ACTION_STUNNED) {
                    wasHit = true;
                } else if (this->action != KAIZOKU_ACTION_JUMP) {
                    EnKaizoku_SetupJump(this);
                }
                break;
```

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -750,122 +778,126 @@ void func_80B868B8(EnKaizoku* this, PlayState* play) {
...
 }
 
-void func_80B86B58(EnKaizoku* this) {
-    this->unk_2D8 = 0;
-    this->action = KAIZOKU_ACTION_0;
-    this->actionFunc = func_80B86B74;
+// Player has won the fight
+void EnKaizoku_SetupPlayerWinCutscene(EnKaizoku* this) {
+    this->animationsDisabled = false;
+    this->action = KAIZOKU_ACTION_HIDDEN;
```
tom-overton, line 784: I noticed this was set in a couple of different places, namely:
- `EnKaizoku_SetupWaitForApproach`
- `EnKaizoku_SetupPlayerLoss`
- `EnKaizoku_SetupPlayerWinCutscene`

However, the pirate is really only "hidden" for the first one. It seems like this action is doing a few different things, like disabling head turning, preventing `EnKaizoku_UpdateDamage` and `Actor_SetFocus` from being called, and preventing the AC collider on the body from being turned on. I'm having some trouble coming up with a name for this one; maybe something like `KAIZOKU_ACTION_NOT_FIGHTING`? It seems like there's some common code they want to run while the fight is going on, but they *don't* want to run that code before the fight starts or after it's over. I'm totally open for suggestions on this one because I'm not super confident in this name myself.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1418,76 +1456,76 @@ void func_80B88964(EnKaizoku* this, PlayState* play) {
...
-            this->unk_2B2 = Rand_ZeroOne() * 5.0f + 5.0f;
+            EnKaizoku_SetupReady(this);
+            this->combatTimer = Rand_ZeroOne() * 5.0f + 5.0f;
             return;
         }
 
         this->picto.actor.world.rot.y = this->picto.actor.yawTowardsPlayer;
         if (BREG(12) * 0.1f + 0.01f * 40.0f < Rand_ZeroOne()) {
-            func_80B87900(this);
+            // dev seemed unsatisfied with randomized behavior
```
tom-overton, line 1481: Is this in response to them changing it from sidestepping to blocking? This is more of a philosophical debate than anything else, but while I think it's fine to comment that it's different from OoT, and while I think it's fine to even speculate on certain things (I am not innocent of this; see my massive comment above `MAJORAS_INCARNATION_COLLIDER_BODYPART_ROOM_ORIGIN` as an example), I don't think we can really tell whether or not they changed this from sidestepping to blocking because of dissatisfaction or because of some other reason (a designer told them to change it, it fixed a weird bug, etc.). They even changed the odds of this behavior happening; in OoT, it was:
```c
            if (Rand_ZeroOne() > 0.7f) {
                EnGeldB_SetupSidestep(this, play);
            }
```
So it's clearly more than just a quick swap.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
```diff
@@ -1418,76 +1456,76 @@ void func_80B88964(EnKaizoku* this, PlayState* play) {
...
+    EnKaizoku_ChangeAnim(this, KAIZOKU_ANIM_SIDESTEP);
     this->picto.actor.speed = Rand_CenteredFloat(12.0f);
     this->skelAnime.playSpeed = 1.0f;
     this->picto.actor.world.rot.y = this->picto.actor.shape.rot.y;
-    this->unk_2B2 = Rand_ZeroOne() * 30.0f + 30.0f;
-    this->action = KAIZOKU_ACTION_3;
-    this->actionFunc = func_80B88D6C;
-    this->unk_2E0 = 0.0f;
+    this->combatTimer = Rand_ZeroOne() * 30.0f + 30.0f;
+    this->action = KAIZOKU_ACTION_SIDESTEP;
```
tom-overton, line 1507: Shouldn't the action be `KAIZOKU_ACTION_CIRCLE` then? I'm fine with the animation being `SIDESTEP` (since there's nothing about the animation that suggests that it's being used to circle the player), but for the action, it makes sense to name it similarly to the action function

tom-overton (commented):
Sorry for the absolutely gigantic delay in my response; I have been swamped with work for most of 2025, though I know that's no excuse. Most things look good to me, so I just had a few comments here and there.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
isghj5: I'm tired, and its been awhile since I made this bad comment which should be removed, but I think I was referring to the chain expression
```
BREG(12) * 0.1f + 0.01f * 40.0f < Rand_ZeroOne()
```
which suggests the dev wasn't happy with the results they were getting and kept tweaking it. It's not worth keeping as a comment either way.

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
isghj5: For some reason I thought cirlc was already used, but yeah it looks like CIRCLE is correct

src/overlays/actors/ovl_En_Kaizoku/z_en_kaizoku.c
isghj5: Can't think of anything better, "Out of combat" is probably too weird

tom-overton (approved):
Sorry again for taking so long to review this! Looks great, thanks for doing this!

PR #1791, opened 2025-01-21
GCC Compiler Support

Thar0:
This PR adds support for building the project with a modern MIPS GCC compiler.

Thanks to @fig02 for testing a build on hardware, and everyone who engaged in related discussions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,8 @@ COMPARE ?= 1
 NON_MATCHING ?= 0
 # If ORIG_COMPILER is 1, compile with QEMU_IRIX and the original compiler
 ORIG_COMPILER ?= 0
+# If COMPILER is "gcc", compile with GCC instead of IDO.
+COMPILER ?= ido
 # if WERROR is 1, pass -Werror to CC_CHECK, so warnings would be treated as errors
 WERROR ?= 0
 # Keep .mdebug section in build
@@ -69,9 +71,18 @@ N64_EMULATOR ?=
 # Ensure the map file being created using English localization
 export LANG := C
 
+CFLAGS :=
+CPPFLAGS :=
+
+ifeq ($(COMPILER),gcc)
+  CPPFLAGS += -DCOMPILER_GCC
+  NON_MATCHING := 1
+  RUN_CC_CHECK := 0
+endif
+
 ifneq ($(NON_MATCHING),0)
-  CFLAGS := -DNON_MATCHING
-  CPPFLAGS := -DNON_MATCHING
+  CFLAGS += -DNON_MATCHING -DAVOID_UB
+  CPPFLAGS += -DNON_MATCHING -DAVOID_UB
   COMPARE := 0
 endif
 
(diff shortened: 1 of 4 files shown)


hensldm (approved):
Thoughts on bringing mkldscript.c over from OoT (though does require adding the else if for `STMT_increment` on lines 40-41) to have parity between the two (mainly looks like just formatting is the difference between the 2). If you don't want to that is fine, I would probably do it after this gets merged.

Makefile
```diff
@@ -177,18 +195,36 @@ SFCFLAGS := --matching
...
-WARNINGS         := -fullwarn -verbose -woff 624,649,838,712,516,513,596,564,594,807
-ASFLAGS          := -march=vr4300 -32 -G0
-GBI_DEFINES 	 := -DF3DEX_GBI_2 -DF3DEX_GBI_PL -DGBI_DOWHILE
-COMMON_DEFINES   := -D_MIPS_SZLONG=32 $(GBI_DEFINES)
-AS_DEFINES       := $(COMMON_DEFINES) -DMIPSEB -D_LANGUAGE_ASSEMBLY -D_ULTRA64
-C_DEFINES        := $(COMMON_DEFINES) -DLANGUAGE_C -D_LANGUAGE_C
-ENDIAN           := -EB
+ifeq ($(COMPILER),gcc)
+  # MIPS options
+  CFLAGS           += -G 0 -march=vr4300 -mtune=vr4300 -mfix4300 -mabi=32 -mno-abicalls -mdivide-breaks
```
AngheloAlf, line 202: This reminds me, why do we use `-mdivide-breaks`? The n64 supports trap instructions

Makefile
```diff
@@ -177,18 +195,36 @@ SFCFLAGS := --matching
...
+  CFLAGS           += -nostdinc -fno-PIC -fno-common -ffreestanding -fbuiltin -fno-builtin-sinf -fno-builtin-cosf -funsigned-char
 
-OPTFLAGS := -O2 -g3
-MIPS_VERSION := -mips2
+  WARNINGS         := $(CC_CHECK_WARNINGS)
+  ASFLAGS          := -march=vr4300 -32 -G0
+  COMMON_DEFINES   := $(GBI_DEFINES)
+  AS_DEFINES       := $(COMMON_DEFINES) -DMIPSEB -D_LANGUAGE_ASSEMBLY -D_ULTRA64
+  C_DEFINES        := $(COMMON_DEFINES) -D_LANGUAGE_C
+  ENDIAN           :=
```
AngheloAlf, line 211: Why omit this?
```suggestion
  ENDIAN           := -EB
```

Makefile
Thar0: It's mostly for compatibility. I think it's a reasonable default that people can change themselves depending on their tolerance for what platforms they want to support

Makefile
Thar0: It's an architecture default, it doesn't really need to be explicit. We already rely on `-mhard-float` being implicit from the arch

AngheloAlf (approved):

PR #1795, opened 2025-02-19
Small sys_flashrom clean up

hensldm:
Perhaps these are just my thoughts, but feel like this is a little cleaner for sys flashrom. PR does
1. `requestType` -> `type` as request seems redundant as the struct is `FlashromRequest`
2. mv the msgBuf into `FlashromRequest` since it already has the queue
3. Remove Data from function names as I don't think it really adds much
4. Other misc clean up


fig02 (approved):

include/sys_flashrom.h
```diff
@@ -4,24 +4,25 @@
...
     /* 0x04 */ s32 response;
     /* 0x08 */ void* addr;
-    /* 0x0C */ s32 pageNum;
-    /* 0x10 */ s32 pageCount;
-    /* 0x14 */ OSMesgQueue messageQueue;
-} FlashromRequest; // size = 0x2C
+    /* 0x0C */ u32 pageNum;
+    /* 0x10 */ u32 pageCount;
+    /* 0x14 */ OSMesgQueue queue;
+    /* 0x2C */ OSMesg msgBuf[1];
```
AngheloAlf, line 13: Pretty cool that you can put the buffer here and still match

AngheloAlf (approved):

PR #1796, opened 2025-02-19
Document credits dog OOB access

hensldm:


src/overlays/actors/ovl_En_Dg/z_en_dg.c
```diff
@@ -414,6 +414,10 @@ void EnDg_SetupIdleMove(EnDg* this, PlayState* play) {
         } else if (play->sceneId == SCENE_CLOCKTOWER) {
             EnDg_ChangeAnim(&this->skelAnime, sAnimationInfo, DOG_ANIM_RUN);
         } else if (sRacetrackDogInfo[this->index].textId & 0x11) {
+            //! @bug: There is no bounds check on sRacetrackDogInfo access.
+            //! The dog in the Romani Ranch credits uses params of 0x03FF which means an index equal to
+            //! `ENDG_INDEX_SOUTH_CLOCK_TOWN`. Since the if above just checks the scene not the index, this results in
+            //! an OOB access of `sRacetrackDogInfo` in this if.
```
fig02, line 420: I would add a concluding sentence explaining what happens (upside down dog)
Also maybe "in this if" -> "in the above condition"

fig02 (approved):

src/overlays/actors/ovl_En_Dg/z_en_dg.c
```diff
@@ -414,6 +414,12 @@ void EnDg_SetupIdleMove(EnDg* this, PlayState* play) {
         } else if (play->sceneId == SCENE_CLOCKTOWER) {
             EnDg_ChangeAnim(&this->skelAnime, sAnimationInfo, DOG_ANIM_RUN);
         } else if (sRacetrackDogInfo[this->index].textId & 0x11) {
+            //! @bug: There is no bounds check on sRacetrackDogInfo access.
+            //! The dog in the Romani Ranch credits uses params of 0x03FF which means an index equal to
+            //! `ENDG_INDEX_SOUTH_CLOCK_TOWN`. Since the above condition just checks the scene not the index, this
+            //! results in an OOB access of `sRacetrackDogInfo` in this condition. With IDO, the OOB access results in
+            //! this condition evaluating as true and the dog uses the walking animation. Due to this, and since the dog
+            //! doesn't update in the credits due to being considered an enemy, it ends up being upside down.
```
AngheloAlf, line 422: The "it ends up being upside down" part comes a bit out of nowhere. Does it happen because the actor's default position is upside down? Or maybe is it the animation's default position? Or something else?

hensldm: So I did a little investigation. I believe the reason (I'll admit animation stuff goes a bit over my head) is that this particular walking animation uses morph frames (in this case -6 though anything but 0 gives the same result). With the morphing from the NULL animation set in Init, `Animation_ChangeImpl` doesn't get the frameData for the `jointTable`. Instead it is just filled with 0's. I think this would mean it just draws the default pose, which I guess is upside down for the dog (this is where I am the most hazy).

AngheloAlf: I see, Thanks for looking into it.
For some reason I was under the impression that we knew why this happens. I guess we could add a brief explanation like "it is speculated it ends up like this due to its default pose being upside down too, but it isn't completely certain."

tom-overton (approved):

AngheloAlf (approved):

PR #1800, opened 2025-03-03
Misc fixes

hensldm:
Some more misc fixes mostly revolving around disassembly, though a few other things.


src/boot/rspboot.s
```diff
@@ -1,7 +1,9 @@
...
-.section .data
+.section .text
 
-dlabel rspbootTextStart
+.balign 16
+
+glabel rspbootTextStart
     .incbin "incbin/rspbootText"
-dlabel rspbootTextEnd
+glabel rspbootTextEnd
```
AngheloAlf, line 9: Our `glabel` macro uses the `.ent` directive which expects a corresponding `.end` directive, so gas complains about the missing `.end`. You could write it like this instead:

```mips
glabel rspbootTextStart
    .incbin "incbin/rspbootText"
endlabel rspbootTextStart
dlabel rspbootTextEnd
```

Note I used a `dlabel` for the `rspbootTextEnd` symbol because, as far as I know, the "end" ucode symbol shouldn't be marked as a function (and to avoid dealing with the `.ent`/`.end` shenanigans)

AngheloAlf (commented):
CI is angry because of the new warnings

AngheloAlf (approved):

mzxrules (approved):

PR #1801, opened 2025-03-17
Remove us specific dmadata table

hensldm:
Generating the table from the spec is enough for matching, so decided to remove the redundant us specific dmadata table


AngheloAlf (approved):
Ohh, nice finding.

Hopefully this will be true for other versions too

mzxrules (approved):

PR #1803, opened 2025-03-20
split micrcode into their own files

hensldm:
Has a little overlap with #1800 but I will deal with that when one of them gets in.

Noticed the map file would interleave the End symbols with the Start symbols of the next microcode, so the End symbol would end up with a size using mapfile parser instead of the Start symbol.

```
Symbol 'gspS2DEX2_fifoTextStart' (VRAM: 0x801ABAB0, VROM: 0xC41FF0, SIZE: 0x0, build/n64-us/src/code/rspcode.o)
Symbol 'aspMainTextEnd' (VRAM: 0x801ABAB0, VROM: 0xC41FF0, SIZE: 0x18C0, build/n64-us/src/code/rspcode.o)
```

Note: `gspF3DZEX2_NoN_PosLight_fifoTextEnd` and `gspF3DZEX2_NoN_PosLight_fifoDataStart` are still interleaved so it looks like `gspF3DZEX2_NoN_PosLight_fifoDataStart` has a size, but not sure there is a good way to handle that since they are both in the same section.


AngheloAlf (approved):

hensldm:
Superseeded by #1806

PR #1807, opened 2025-04-18
Handwritten asm build prep, add handwritten asm for the code segment

Thar0:


hensldm (commented):
Couple thoughts.
Also looks like this breaks the partial disassembly since makerom/ipl3 and audio/lib/stack no longer have actual source files.

Makefile
```diff
@@ -607,8 +627,44 @@ $(BUILD_DIR)/dmadata/dmadata_table_spec.h $(BUILD_DIR)/dmadata/compress_ranges.t
...
+$(BUILD_DIR)/src/makerom/rom_header.o: src/makerom/rom_header.s
+ifeq ($(COMPILER),ido)
+	$(CPP) $(CPPFLAGS) $(MIPS_BUILTIN_DEFS) $(IINC) $< | $(AS) $(ASFLAGS) $(IINC) $(ENDIAN) -o $@
+else
+	$(CCAS) -c $(CCASFLAGS) $(IINC) $(MIPS_VERSION) $(ASOPTFLAGS) -o $@ $<
+endif
+	$(OBJDUMP_CMD)
+
+$(BUILD_DIR)/src/makerom/ipl3.o: $(EXTRACTED_DIR)/incbin/ipl3
+	$(OBJCOPY) -I binary -O elf32-big --rename-section .data=.text $< $@
```
AngheloAlf, line 649: What's the advantage of using objcopy over incbin assembly files?

Thar0: incbin isn't supported by IDO, so we'd need custom makefile rule anyway. At that point, objcopy is simpler

src/makerom/entry.s
```diff
@@ -1,34 +1,49 @@
 #include "PR/asm.h"
+#include "PR/regdef.h"
 #include "boot.h"
 
-.set noreorder
+.text
 
-.section .text
+#define PAD_TO 0x60
+#define LA(reg, sym) la reg, sym
```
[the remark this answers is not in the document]
AngheloAlf: Probably to be synced with OoT's entry
https://github.com/zeldaret/oot/blob/f7073a7837de95ac72a0189c74ffa96335280f12/src/makerom/entry.s#L7-L18

include/PR/asm.h
```diff
@@ -1,72 +1,124 @@
...
 
 #ifdef __sgi
 #define _MIPS_ISA_MIPS1 1
 #define _MIPS_ISA_MIPS2 2
 #define _MIPS_ISA_MIPS3 3
 #define _MIPS_ISA_MIPS4 4
+
+#define _MIPS_SIM_ABI32     1   /* MIPS MSIG calling convention */
+#define _MIPS_SIM_NABI32    2   /* MIPS new 32-bit abi */
+                                /* NABI32 is 64bit calling convention but 32bit type sizes) */
```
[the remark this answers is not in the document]
hensldm: Bump (~tbh it feels like the `)` should just be removed~), nevermind. On a reread the second comment could all be wrapped in a `()`.

include/PR/asm.h
```diff
@@ -1,72 +1,124 @@
-#ifndef PR_ASM_H
-#define PR_ASM_H
+#ifndef ASM_H
```
hensldm: Not a big deal but could you add the `PR_` prefix back. Its what we have been doing for all header files in PR.

include/PR/regdef.h
```diff
@@ -0,0 +1,164 @@
+#ifndef REGDEF_H
```
hensldm: Not a big deal but could you add a `PR_` prefix.

hensldm (approved):

PR #1808, opened 2025-04-19
Fix OOB in bom bowl man

louist103:
This fixes an issue in 2ship where the camera would not pan correctly during the code reveal.
I also fixed an OOB write in bomjima though that seems more harmless since those values are never read.


src/overlays/actors/ovl_En_Bom_Bowl_Man/z_en_bom_bowl_man.h
```diff
@@ -44,9 +44,7 @@ typedef struct EnBomBowlMan {
     /* 0x2C2 */ s16 unk_2C2;
     /* 0x2C4 */ f32 animEndFrame;
     /* 0x2C8 */ f32 unk_2C8;
-    /* 0x2CC */ s16 csIdList[2];
-    /* 0x2D0 */ s16 csId1;
-    /* 0x2D2 */ s16 csId2;
+    /* 0x2CC */ s16 csIdList[4];
     /* 0x2D4 */ s16 unk_2D4;
     /* 0x2D6 */ s16 csId3;
```
hensldm, line 49:
```suggestion
    /* 0x2D6 */ s16 csId;
```
Now that there is only 1 we should remove the number 3.

hensldm (approved):

AngheloAlf (approved):

PR #1810, opened 2025-04-20
Document inconsistent empty C-button disabling bug

t0mtee:
In Interface_UpdateButtonsPart2, there is a bug causing empty C-buttons to be disabled inconsistently (for most forms, they are enabled, but when Deku Link, they are disabled): [Example](https://tcrf.net/The_Legend_of_Zelda:_Majora%27s_Mask/Program_Revision_Differences#C-Buttons)

This is because the function checks the 255th index of the second dimension of gPlayerFormItemRestrictions - a dimension which has only been allocated 114 elements. This leads to inconsistent behaviour.

### DIFF
--- a/src/code/z_parameter.c
+++ b/src/code/z_parameter.c
@@ -2044,9 +2044,10 @@ void Interface_UpdateButtonsPart2(PlayState* play) {
         if (GET_PLAYER_FORM == player->transformation) {
             for (i = EQUIP_SLOT_C_LEFT; i <= EQUIP_SLOT_C_RIGHT; i++) {
                 // Individual C button
-                //! @bug When C-buttons are empty, their item code is 255. However, gPlayerFormItemRestrictions's second dimension has
-                //! only been allocated 114 elements. This leads to inconsistent behaviour when checking the status of empty
-                //! C-buttons - for most forms, the C-buttons are enabled when empty, however for Deku Link only, empty C-buttons are disabled.
+                //! @bug When C-buttons are empty, their item code is 255. However, gPlayerFormItemRestrictions's second
+                //! dimension has only been allocated 114 elements. This leads to inconsistent behaviour when checking
+                //! the status of empty C-buttons - for most forms, the C-buttons are enabled when empty, however for
+                //! Deku Link only, empty C-buttons are disabled.
                 if (!gPlayerFormItemRestrictions[GET_PLAYER_FORM][GET_CUR_FORM_BTN_ITEM(i)]) {
                     // Item not usable in current playerForm
                     if (gSaveContext.buttonStatus[i] != BTN_DISABLED) {
--- a/src/code/z_parameter.c
+++ b/src/code/z_parameter.c
@@ -2044,6 +2044,9 @@ void Interface_UpdateButtonsPart2(PlayState* play) {
         if (GET_PLAYER_FORM == player->transformation) {
             for (i = EQUIP_SLOT_C_LEFT; i <= EQUIP_SLOT_C_RIGHT; i++) {
                 // Individual C button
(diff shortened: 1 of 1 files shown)


t0mtee:
Formatting seems to be fixed! I've agreed to CC0 as well :)

hensldm (approved):

src/code/z_parameter.c
```diff
@@ -2044,6 +2044,10 @@ void Interface_UpdateButtonsPart2(PlayState* play) {
         if (GET_PLAYER_FORM == player->transformation) {
             for (i = EQUIP_SLOT_C_LEFT; i <= EQUIP_SLOT_C_RIGHT; i++) {
                 // Individual C button
+                //! @bug When C-buttons are empty, their item code is 255. However, gPlayerFormItemRestrictions's second
+                //! dimension has only been allocated 114 elements. This leads to inconsistent behaviour when checking
+                //! the status of empty C-buttons - for most forms, the C-buttons are enabled when empty, however for
+                //! Deku Link only, empty C-buttons are disabled.
                 if (!gPlayerFormItemRestrictions[GET_PLAYER_FORM][GET_CUR_FORM_BTN_ITEM(i)]) {
```
AngheloAlf, line 2051: Huh, interesting. So every form is doing an OoB read when there are no items on each C button (when `GET_CUR_FORM_BTN_ITEM(i)` returns `ITEM_NONE`), but some forms do an OoB inside `gPlayerFormItemRestrictions` (by reading the data of another form) while others do read data from outside this variable.

If I didn't mess up, Deku and Human are the only forms that read outside `gPlayerFormItemRestrictions` by 0x1B and 0x8D bytes each. So they read a byte from address `0x801C2665` and `0x801C26D7` respectively.
The first address corresponds to a zero byte from the first entry of `sSceneTextureFiles`, an in-function-static symbol from `Scene_LoadAreaTextures` from `z_scene.c`.
```c
    static RomFile sSceneTextureFiles[9] = {
        ROM_FILE_UNSET, // Default
```
<details>
<summary><b>Video changing the byte pointed by that address on an emu (click to see)</b></summary>

https://github.com/user-attachments/assets/cdfb147e-c6a3-470e-ba68-b2ab04c1842f

</details>


The second one access the index "11.75" from the `sSceneCmdHandlers` array of functions, which corresponds to the last byte of the address of `Scene_CommandObjectList`. On the US rom this byte is 0x18. If the function's address happened to end at a 0 byte then this issue would have been a lot more notorious since it would have happened to Human.
```c
void (*sSceneCmdHandlers[SCENE_CMD_MAX])(PlayState*, SceneCmd*) = {
    // ...
    Scene_CommandObjectList,     // SCENE_CMD_ID_OBJECT_LIST
```
<details>
<summary><b>Video changing the byte pointed by that address on an emu (click to see)</b></summary>

https://github.com/user-attachments/assets/e76ab9db-cc62-4ec1-91e5-9478995ca25c

</details>


Deku and Zora are simpler, since they "just" access the element 27 of the `gPlayerFormItemRestrictions` array for forms Deku and Human respectively, which funnily corresponds to `ITEM_BUG`. Those forms do not present the issue because every form is allowed to use that item.

AngheloAlf: Either way, I'm not telling to you to include all this information in the bug comment. I just wanted to know exactly why this OoB read was an issue only for Deku and not for the others. It can be summarized in "the other forms were luckier". I can see this being a bit annoying on shifted/modded builds

AngheloAlf (approved):

src/code/z_parameter.c
t0mtee: Thats really interesting, thank you for sharing. I was wondering what values were being accessed.

PR #1812, opened 2025-07-03
Replace erroneous CsCmdStartSeq with CsCmdStopSeq

Eblo:
The cutscene command for stopping a sequence advances the script by `sizeof(CsCmdStartSeq)` instead of `sizeof(CsCmdStopSeq)`. The values are the same.


mzxrules (approved):

AngheloAlf:
Nice find! Thanks!

Something else to take into consideration: We are trying to get everybody who has contributed / wants to contribute to MM decomp to agree to licensing the repo before merging PRs from people who have not yet agreed to it. It is better explained here: https://github.com/louist103/mm-licensing/issues/1.

Eblo:
No problem, done.

PR #1813, opened 2025-07-30
adding mcp template for AI assist

Axle-Bucamp:
i fix some line on your dockerfile and also added an mcp compatibility, which is easy on bash like project. 
But well still struggle using your dock to finalyse the make init. got some result but feel like something is missing.
Is it rom version ?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,36 +1,41 @@
-FROM ubuntu:24.04 AS build
+# Base image
+FROM ubuntu:22.04
 
 ENV TZ=UTC
 ENV LANG=C.UTF-8
 
-# Install Required Dependencies
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
-    apt-get update && apt-get install -y \
-    curl \
+# Set timezone
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# Install build dependencies
+RUN apt-get update && apt-get install -y \
+    make \
+    git \
     build-essential \
     binutils-mips-linux-gnu \
-    pkg-config \
+    curl \
     python3 \
     python3-pip \
     python3-venv \
-    git \
-    wget \
-    unzip \
-    vbindiff \
-    vim \
-    clang-tidy-14 \
-    clang-format-14 \
-    libpng-dev && \
-    # Add source for practicerom-dev install
-    curl https://practicerom.com/public/packages/debian/pgp.pub | apt-key add - && \ 
-    echo deb http://practicerom.com/public/packages/debian staging main >/etc/apt/sources.list.d/practicerom.list && \
-    apt-get update && \
-    apt-get install -y \
-    practicerom-dev && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+    libpng-dev \
+    libxml2-dev && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Set working directory to project
+WORKDIR /
+
+# Optional: mark directory safe for git is that needed
+RUN git config --global --add safe.directory /mm
+# future after wget rom
+RUN git clone https://github.com/zeldaret/mm.git
+
+# remove dependancies
+# add volume for rom endpoint
+WORKDIR /mm/baseroms/n64-us
+COPY ./baseroms/n64-us/baserom.z64 .
 
 WORKDIR /mm
-RUN git config --global --add safe.directory /mm
+RUN make init
 
-ENTRYPOINT ["/usr/bin/env", "bash", "-c"]
+# Default entry (you may override in mcp config)
+ENTRYPOINT ["/usr/bin/env", "bash"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,10 @@ RUN git clone https://github.com/zeldaret/mm.git
 # remove dependancies
 # add volume for rom endpoint
 WORKDIR /mm/baseroms/n64-us
-COPY ./baseroms/n64-us/baserom.z64 .
+COPY ./baseroms/n64-us .
+
+WORKDIR /mm/build/n64-us
+COPY ./build/n64-us/mm-n64-us.z64 .
 
 WORKDIR /mm
 RUN pip install -r requirements.txt
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ WORKDIR /mm/baseroms/n64-us
 COPY ./baseroms/n64-us/baserom.z64 .
 
 WORKDIR /mm
+RUN pip install -r requirements.txt
 RUN make init
 
 # Default entry (you may override in mcp config)
--- a/README.md
+++ b/README.md
@@ -156,6 +156,159 @@ Running `make init` will also make the `./expected` directory and copy all of th
 The disadvantage that the ordering of the terminal output is scrambled, so for debugging it is best to stick to one thread (i.e. not pass `-jN`).
 (`-j` also exists, which uses unlimited jobs, but is generally slower.)
 
+#### work with AI agent :
+
+---
+
+##### 📁 `mcp-config.json` exemple
+
+```json
+{
+  "name": "zeldaret_mm_tools",
+  "description": "Zelda: Majora’s Mask decompilation tools using Docker. Integrates with AI agents such as Camel Owl, Claude Desktop, and MCP Proxy.",
+  "dockerfile_path": "./Dockerfile",
+  "image_name": "mm",
+  "entrypoint": "/usr/bin/env bash",
+  "mount": {
+    "type": "bind",
+    "source": "$(pwd)",
+    "destination": "/mm"
+  },
+  "commands": [
+    {
+      "name": "launch_terminal",
+      "description": "Launch an interactive shell inside the mm Docker container",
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -1,36 +1,45 @@
-FROM ubuntu:24.04 AS build
+# Base image
+FROM ubuntu:22.04
```
AngheloAlf, line 2: What is fixed by removing the `AS build` part?

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
-# Install Required Dependencies
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
-    apt-get update && apt-get install -y \
-    curl \
+# Set timezone
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# Install build dependencies
+RUN apt-get update && apt-get install -y \
+    make \
```
AngheloAlf, line 12: Listing `make` explicitly is a bit redundant since `build-essential` should already take care of it.

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
-    vim \
-    clang-tidy-14 \
-    clang-format-14 \
-    libpng-dev && \
-    # Add source for practicerom-dev install
-    curl https://practicerom.com/public/packages/debian/pgp.pub | apt-key add - && \ 
-    echo deb http://practicerom.com/public/packages/debian staging main >/etc/apt/sources.list.d/practicerom.list && \
-    apt-get update && \
-    apt-get install -y \
-    practicerom-dev && \
```
AngheloAlf, line 29: Could you tell me why you removed all those utilities?

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+    libpng-dev \
+    libxml2-dev && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
 
-WORKDIR /mm
+# Set working directory to project
+WORKDIR /
```
AngheloAlf, line 25: Why did you change this?

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
+
+# Optional: mark directory safe for git is that needed
 RUN git config --global --add safe.directory /mm
+# future after wget rom
```
AngheloAlf, line 29: What do you mean by `wget`ing the rom? we don't provide the rom to anybody, the user has to provide their own.

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
+RUN git clone https://github.com/zeldaret/mm.git
```
AngheloAlf, line 30: Errr, what?
The repo should be already cloned, otherwise you couldn't access the Dockerfile at all. With this you end up with a clone inside a clone.

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
+
+# remove dependancies
+# add volume for rom endpoint
+WORKDIR /mm/baseroms/n64-us
+COPY ./baseroms/n64-us .
```
AngheloAlf, line 35: This doesn't make sense at all. This would copy some configuration files to the root folder and do nothing with them.

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
+
+WORKDIR /mm/build/n64-us
+COPY ./build/n64-us/mm-n64-us.z64 .
```
AngheloAlf, line 38: There's no build folder at this point yet, so this would probably error out or do nothing at all.

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
+
+WORKDIR /mm
+RUN pip install -r requirements.txt
+RUN make init
```
AngheloAlf, line 42: We expect the user to build the project _after_ the docker image is up and running, because that way they can customize their builds, like specifying number of jobs, version to build, compiler to use, etc.

Also installing the python requirements globally via `pip` here doesn't help the build process, because the Makefile will automatically setup a venv, managing it and using it automatically, so it won't use the globally installed packages at all.

Dockerfile
```diff
@@ -1,36 +1,45 @@
...
 
-ENTRYPOINT ["/usr/bin/env", "bash", "-c"]
+# Default entry (you may override in mcp config)
+ENTRYPOINT ["/usr/bin/env", "bash"]
```
AngheloAlf, line 45: What's the point of removing the `-c` flag?

README.md
```diff
@@ -98,7 +98,7 @@ cd mm
 
 Place a copy of the US ROM inside the `baseroms/n64-us/` folder.
 
-Rename the file to `baserom.z64`, `baserom.n64` or `baserom.v64`, depending on the original extension.
+Rename the file to `baserom.z64`, `mm-n64-us.z64` or `mm-n64-us.z64`, depending on the original extension.
```
AngheloAlf, line 101: This is plain wrong. The build system won't look up for roms named this way.

README.md
AngheloAlf, line 1: We don't really use AI nor we enforce its use on this project, so having such a large wall of text explaining how to use them feels out of place for us.
Those instructions also seem wrong, since they refer to stuff that doesn't exist anywhere, like `/mm/tools/natural.sh`.

mcp-config.json
```diff
@@ -0,0 +1,15 @@
+{
+  "mcpServers": {
+    "zeldaret-mm": {
+      "command": "docker",
+      "args": [
+        "run",
+        "--rm",
+        "-v", "${PWD}:/mm",
+        "-w", "/mm",
+        "ghcr.io/zeldaret/mm",
```
AngheloAlf, line 10: What is this supposed to link to? It redirects me to a Github 404 error.

toolchains/zeldaret-mm.json
```diff
@@ -0,0 +1,155 @@
+{
+  "name": "zeldaret_mm_tools",
+  "description": "Run Zelda decompilation tools for Majora's Mask using the Docker environment provided by zeldaret/mm. Supports extracting assets, compiling, and verifying against the original ROM. Designed for automation via Camel Owl Agent.",
+  "dockerfile_path": "C:\Users\dying\Documents\mm\Dockerfile",
```
AngheloAlf, line 4: Hardcoded paths are never a good idea, even less in collaborative projects that are meant to be used by multiple people and different configurations.
Also this project always uses Unix paths, not Windows ones, since the tooling is designed more around Unix environments.

toolchains/zeldaret-mm.json
AngheloAlf, line 1: I don't know what's the point of this file. Just to have a list of programs to run?
People is expected to run these programs in their terminals and learn to use them. Some of these tools are interactive (i.e. diff.py), so I'm not sure how they would integrate with whatever this file is doing.
Finally, some paths seems to be wrong (again, diff.py), so I don't think this will work as expected.

Dockerfile
Axle-Bucamp: did not checkout lot of things, but was not compiling like at all. so maybe there still some hints. i re-base from a first GPT call just in case beeing lazy could work. Then started using the new file and doing command inside. But as i wanted a one line builder i needed the rom so. and i realise the rom got multiple different name and usage .... which made me lazy again. maybe you know more so

Dockerfile
Axle-Bucamp: mhmm probably indeed

Dockerfile
Axle-Bucamp: ho lazyness wanted straight building of rom + everything so i just wanted to test out some possibilities. can be simplify ofc you right

Dockerfile
Axle-Bucamp: wget, curl or i don t know. i agree we should not and just use a volume ... but i feel like everyone will just want to download the base mm rom instead of adding it to from the repo or downloading manualing and maybe misisng out MD5 hash check. also i think it s ok now to just let it in repo, nintendo should not come with rocket launcher for OOT

README.md
Axle-Bucamp: don t know, it did on my side. was really weird so

Axle-Bucamp: made me bug way to much time so i agree

mcp-config.json
Axle-Bucamp: plan to make a docker hub later with a stable build. Use a prebuild name mm for now

toolchains/zeldaret-mm.json
Axle-Bucamp: agree, sorry lazy

Axle-Bucamp: but for toolchain, it s decent enougth. it s just hint for AI, i will fix later on so

toolchains/zeldaret-mm.json
Axle-Bucamp: AI way to understand how to use the terminal for some MCP / AI framework. imagine the GPT text prompt printing command base on this open api config file sort of

Axle-Bucamp: like typing, command name, etc. I wonder so, using toolchains instead of a standard MCP builder is new to me, in theory work

Axle-Bucamp: just need a proper build

Axle-Bucamp: by the way do you need anything on this repo ? it seems weird but mostly working when i tested it

PR #1814, opened 2025-07-31
Fix docker builds

AngheloAlf:
I got curious about PR #1813, because it claimed Docker builds to be broken, but the PR did not provide any reason why Docker doesn't work, so I tried it out.

The issue was caused by not being able to setup the practicerom stuff while creating the Docker image.
- The gpg key was moved on the host, so the old link leads to a 404 error.
- The old contents of the `practicerom.list` apt list are no longer valid either.
  - This actually causes issues when trying to `apt update` on a machine that still uses the old `practicerom.list`. I had to `sudo rm /etc/apt/sources.list.d/practicerom.list` on my main pc to get rid of the errors.

The changes to the practicerom repository seems to have happened 2 days ago by judging by [this commit on glankk's repo](https://github.com/glankk/n64/commit/e7bec93b0b11b37a63abcf5ac28728f2fcd4045b#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

I changed the `Dockerfile` to use the updated instructions instead.

I also updated some updated references on the markdown files.

Besides this, I could use updated Docker image to build MM without any other problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,15 @@ FROM ubuntu:24.04 AS build
 ENV TZ=UTC
 ENV LANG=C.UTF-8
 
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# Add source for practicerom-dev install
+RUN apt-get update && apt-get install -y curl && \
+    curl -o /etc/apt/trusted.gpg.d/practicerom.gpg https://practicerom.com/public/packages/practicerom.gpg && \
+    echo 'deb [arch=amd64] http://practicerom.com/public/packages/debian unstable main' > /etc/apt/sources.list.d/practicerom.list
+
 # Install Required Dependencies
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
-    apt-get update && apt-get install -y \
-    curl \
+RUN apt-get update && apt-get install -y \
     build-essential \
     binutils-mips-linux-gnu \
     pkg-config \
@@ -20,12 +25,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     vim \
     clang-tidy-14 \
     clang-format-14 \
-    libpng-dev && \
-    # Add source for practicerom-dev install
-    curl https://practicerom.com/public/packages/debian/pgp.pub | apt-key add - && \ 
-    echo deb http://practicerom.com/public/packages/debian staging main >/etc/apt/sources.list.d/practicerom.list && \
-    apt-get update && \
-    apt-get install -y \
+    libpng-dev \
     practicerom-dev && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ Preparation is covered in [Building Docker](docs/BUILDING_DOCKER.md).
 
 The build process has the following package requirements:
 
-* make
 * git
 * build-essential
 * binutils-mips-linux-gnu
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -3,10 +3,15 @@ FROM ubuntu:24.04 AS build
...
+# Add source for practicerom-dev install
+RUN apt-get update && apt-get install -y curl && \
+    curl -o /etc/apt/trusted.gpg.d/practicerom.gpg https://practicerom.com/public/packages/practicerom.gpg && \
+    echo 'deb [arch=amd64] http://practicerom.com/public/packages/debian unstable main' > /etc/apt/sources.list.d/practicerom.list
+
 # Install Required Dependencies
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
-    apt-get update && apt-get install -y \
-    curl \
+RUN apt-get update && apt-get install -y \
```
AngheloAlf: I broke these into multiple steps because I think it is better organized this way.

README.md
```diff
@@ -62,7 +62,6 @@ Preparation is covered in [Building Docker](docs/BUILDING_DOCKER.md).
 
 The build process has the following package requirements:
 
-* make
```
AngheloAlf, line 65: Removed `make` from the list of requirements because `build-essential` takes care of it already so it is redundant.

README.md
```diff
@@ -102,12 +101,14 @@ Rename the file to `baserom.z64`, `baserom.n64` or `baserom.v64`, depending on t
...
-To start the extraction/build process, run the following command:
+For a first time build, run the following command, where `N` is the number of cores your processor has (see the note at the bottom for more information):
 
 ```bash
-make init
+make init -j N
 ```
 
+After the very first build you'll want to run `make -j N` instead, since you likely won't need all the setup the `init` command provides.
+
```
AngheloAlf, line 111: I wanted to mention about using `-jN` because most people will just run this command without reading the footnotes.

I also wanted to mention about not needing to run `make init` after the first run.

docs/BUILDING_DOCKER.md
```diff
@@ -38,10 +38,4 @@ docker run -it --rm --mount type=bind,source="$(pwd)",destination=/mm mm "/usr/b
 
 ## 4. Setup and Build the ROM
 
-Once inside the container, you will need to set the repository as a trusted repository with the following command:
-
-```bash
-git config --global --add safe.directory /mm
```
AngheloAlf, line 44: The `Dockerfile` does this already, so I thought it was redundant, correct me if I'm wrong because I'm no Docker expert.

Dockerfile
```diff
@@ -20,12 +26,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
...
-    libpng-dev && \
-    # Add source for practicerom-dev install
-    curl https://practicerom.com/public/packages/debian/pgp.pub | apt-key add - && \ 
-    echo deb http://practicerom.com/public/packages/debian staging main >/etc/apt/sources.list.d/practicerom.list && \
-    apt-get update && \
-    apt-get install -y \
+    libpng-dev \
     practicerom-dev && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
```
hensldm: I would say we should split these out as well so this just becomes a list of the packages we are installing.

hensldm (approved):
1 suggestion but it shouldn't stop a merge

Dockerfile
```diff
@@ -3,12 +3,18 @@ FROM ubuntu:24.04 AS build
 ENV TZ=UTC
 ENV LANG=C.UTF-8
 
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+# Add source for practicerom-dev install
+RUN apt-get update && apt-get install -y curl && \
+    curl -o /etc/apt/trusted.gpg.d/practicerom.gpg https://practicerom.com/public/packages/practicerom.gpg && \
+    echo 'deb [arch=amd64] http://practicerom.com/public/packages/debian unstable main' > /etc/apt/sources.list.d/practicerom.list
```
hensldm: Looks like the script on the practice rom repo has been updated. We should probably update here as well.

AngheloAlf:
The docker image seems to work properly once again, so I'm undrafting the PR.

hensldm (approved):

PR #1815, opened 2025-08-03
Handwritten asm: Register getter/setter functions

Thar0:


mzxrules (approved):
just curious if there's a reason why comments were changed to /* */

hensldm (approved):

PR #1816, opened 2025-08-05
Handwritten asm: gu and mgu

Thar0:

PR #1817, opened 2025-08-05
Handwritten asm: bcmp, bcopy, bzero

Thar0:


src/libultra/libc/bcopy.s
```diff
@@ -0,0 +1,211 @@
...
+    b       forwards_32
+
+forw_copy3:
+    lb      v0, (a0)
+    lh      v1, 1(a0)
+    addiu   a0, a0, 3
+    sb      v0, (a1)
+    sh      v1, 1(a1)
+    addiu   a1, a1, 3
+    addiu   a2, a2, -3
```
AngheloAlf: Same as #1816
Both `addu a0, a0, 2` and `addiu a0, a0, 3` are essentially the same (besides the 2 vs 3 ofc).

PR #1819, opened 2025-08-05
Handwritten asm: exceptasm, get/setintmask

Thar0:


include/ultra64.h
```diff
@@ -14,6 +14,7 @@
 #include "PR/sptask.h"
 #include "PR/rcp.h"
 #include "PR/rdp.h"
+#include "PR/os_internal_rsp.h"
```
hensldm: Should we really put this in ultra64.h? Feels better to keep it in os_internal.h

src/libultra/os/exceptasm.s
```diff
@@ -0,0 +1,879 @@
...
+#include "PR/version.h"
+
+#define MESG(x) ((x) << 3)
+
+.data
+.align 2
+
+DATA(__osHwIntTable)
+    .word 0, 0
+    .word 0, 0 /* cart */
```
hensldm, line 21: Any reason why this is the only one with a comment?

Thar0: The others are unknown, this is the only one that's used

include/PR/os_internal.h
```diff
@@ -1,14 +1,24 @@
...
 } __osHwInt; // size = 0x8
 
+#else
+
+/* __osHwInt struct member offsets */
+#define HWINT_CALLBACK 0x00
+#define HWINT_SP       0x04
+
+/* __osHwInt struct size */
+#define HWINT_SIZE     0x8
```
AngheloAlf: Could be worth having `static_assert`s to make sure those values don't get desynced?
Also having an extra zero in some and not in the other is a bit weird (`0x04` vs `0x8`)

AngheloAlf (approved):

include/PR/os_internal.h
Thar0: I thought about static asserts a bit, while they would be good there's two issues that I'm not sure make it worth it
- `assert.h` needs to be included in these headers or in a common header like `ultratypes.h` when it didn't need to be there otherwise
- To not duplicate anything these defines need to be available to C as well which involves a lot of new defines being exposed to the rest of the source files, unless we undef them on the C path but it feels like a mess

AngheloAlf: Ohh, I see. In that case I think it is fine to leave it like this.

PR #1822, opened 2025-08-20
Switch to spimdisasm

hensldm:
With the handwritten asm committed, I think we can replace the custom disassembler by Tharo (thank you for the work you put into it) with spimdisasm. This is just a module people can pip install and has precedent used in OoT for multiversion support.
The major draw back is it will no longer disassemble specific files and will be all or nothing. But since we don't need to disassemble for n64-us really I think it is worth it so we can make use of existing tools in multiversion support, and don't have to worry about updating the current dissasembler to work with multiple versions.


Makefile
```diff
@@ -590,13 +575,13 @@ assets:
 
 ## Assembly generation
 disasm:
-	$(RM) -r asm data
-	$(PYTHON) tools/disasm/disasm.py $(EXTRACTED_DIR)/baserom -j $(N_THREADS) $(DISASM_FLAGS)
+	$(RM) -r $(EXTRACTED_DIR)/asm
+	VERSION=$(VERSION) DISASM_BASEROM=$(BASEROM_DIR)/baserom-decompressed.z64 DISASM_DIR=$(EXTRACTED_DIR)/asm PYTHON=$(PYTHON) ./tools/disasm/do_disasm.sh
```
Dragorn421, line 568: note OoT disassembles (disassembled) into expected/, not extracted/
This is fine too I guess, are you going towards matching other MM versions by keeping an OK build using asm-processor instead of doing it like OoT did with starting from another version and approaching OKing a target?

hensldm: > note OoT disassembles (disassembled) into expected/, not extracted/

Do you know if there was any particular reason? `extracted` is just a personal preference as it seems to make more sense to me (also this comes from a time when I was going to use hasm in extracted to build, but with them committed I now longer need that).

> are you going towards matching other MM versions by keeping an OK build using asm-processor instead of doing it like OoT did with starting from another version and approaching OKing a target

Same as OoT. Approaching OKing a target. Going to start with code as well first, than deal with assets at a later time.

Jenkinsfile
```diff
@@ -52,16 +52,6 @@ pipeline {
...
-        stage('Disasm') {
-            steps {
-                sh 'bash -c "make -j disasm 2> >(tee tools/warnings_count/warnings_disasm_new.txt)"'
-            }
-        }
-        stage('Check disasm warnings') {
-            steps {
-                sh 'bash -c "./tools/warnings_count/compare_warnings.sh disasm"'
-            }
-        }
```
AngheloAlf, line 64: Would we add this back in the future when other versions are being worked on? I think checking if other versions "build" fine is important, even if the resulting rom isn't OK.

tools/disasm/disasm.py
```diff
@@ -1,2553 +1,188 @@
...
-                file_syms.append(
-                    {"name": file_name, "first_sym": syms[0], "syms": syms}
-                )
-                syms = []
-                file_name = new_file
-        syms.append(sym)
-    if len(syms) > 0:
-        file_syms.append({"name": file_name, "first_sym": syms[0], "syms": syms})
+    if spimdisasm.__version_info__ < (1, 28, 1):
+        print(f"Error: spimdisasm>=1.28.1 is required (you have {spimdisasm.__version__})")
```
AngheloAlf: This seems outdated. `requirements.txt` says `1.34.2`.
Maybe we should also add a comment in both this file and requirements.txt reminding to update the other file when the version gets bumped?

tools/disasm/disasm.py
```diff
@@ -1,2553 +1,188 @@
...
-    reloc = [
-        (
-            (rel >> 30) & ((1 << 2) - 1),
-            (rel >> 24) & ((1 << 6) - 1),
-            (rel) & ((1 << 24) - 1),
+    print("Analyzing...")
+    for i, file_splits in enumerate(all_file_splits):
+        f = i / len(all_file_splits)
+        spimdisasm.common.Utils.printQuietless(
+            f"{f*100:3.0f}%", "Analyzing", file_splits.name, end="    \r"
```
AngheloAlf: This is nice and fancy, but it leaves some ugly traces when the new lines are shorter than the previous ones.
Would be cool if it were fixed, but no big deal.
(Same to the other prints using `\r`)

tools/disasm/sym_info.py
```diff
@@ -0,0 +1,230 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Print information on a symbol/address"
+        " (and possibly surrounding symbols)"
+        " from the spimdisasm disassembly context saved in"
+        " expected/build/VERSION/context.csv"
+    )
+    parser.add_argument("sym_or_vma")
+    default_version = "ntsc-1.2"
```
AngheloAlf: I suppose we want to change this default.

Jenkinsfile
hensldm: We shouldn't need to no. No rom should need to be disassembled to build since all the handwritten ASM is committed.

Makefile
Dragorn421: I guess OoT put it in expected/ because that's where the built .o from the .s was put, but .s in extracted/ is ok too

AngheloAlf (approved):

AngheloAlf (commented):
as a fyi, spimdisasm just had a new release, and it now uses new extra labels in the generated assembly.
So it would be good to update our `macro.inc` file in this PR too.

This is the recommended macro definitions. Feel free to take them and adapt them as you see.

```mips
# A function symbol.
.macro glabel label, visibility=global
    .\visibility \label
    .type \label, @function
    \label:
        .ent \label
.endm

# The end of a function symbol.
.macro endlabel label
    .size \label, . - \label
    .end \label
.endm

# An alternative entry to a function.
.macro alabel label, visibility=global
    .\visibility \label
    .type \label, @function
    \label:
        .aent \label
.endm

# A label referenced by an error handler table.
.macro ehlabel label, visibility=global
    .\visibility \label
    \label:
.endm


# A label referenced by a jumptable.
.macro jlabel label, visibility=global
    \label:
.endm


# A data symbol.
.macro dlabel label, visibility=global
    .\visibility \label
    .type \label, @object
    \label:
.endm

# End of a data symbol.
.macro enddlabel label
    .size \label, . - \label
.endm


# Label to signal the symbol haven't been matched yet.
.macro nonmatching label, size=1
    .global \label\().NON_MATCHING
    .type \label\().NON_MATCHING, @object
    .size \label\().NON_MATCHING, \size
    \label\().NON_MATCHING:
.endm

```

hensldm:
I just used the definitions you gave since I don't really see a need to change them.

PR #1845, opened 2025-12-09
[n64-jp-1.1] __osMalloc.c matched

AngheloAlf:
Matches __osMalloc.c. The differences are mainly debug stuff.

I took a lot of inspiration from OoT, since a lot is very similar.

This introduces a bunch of duplicated functions in their "debug" variant. It is a bit annoying since the code is pretty much the same between the debug and non debug variants, but whatever (`__osReallocDebug` didn't even try to copy/paste).


src/boot/libc64/__osMalloc.c
```diff
@@ -275,16 +416,81 @@ void __osFree(Arena* arena, void* ptr) {
...
+    (void)"__osFree:不正解放(%08x)\n";
+    // __osFree:Double release(%08x)\n
+    (void)"__osFree:二重解放(%08x)\n";
+    // __osFree:arena(%08x) and __osMallocのarena(%08x) do not match\n
+    (void)"__osFree:arena(%08x)が__osMallocのarena(%08x)と一致しない\n";
+#endif
+
+    node = (ArenaNode*)((uintptr_t)ptr - sizeof(ArenaNode));
+
+    if ((ptr != NULL) && (node->magic == NODE_MAGIC) && !node->isFree && CHECK_CORRECT_ARENA(node, arena)) {
```
hensldm: Looking at OoT's version, these should actually be split up with the strings being associated with the different conditions which I think we should do here as well.

AngheloAlf: I can't do an early return here like OoT does because there's an `ArenaImpl_Unlock(arena);` at the bottom of the function.

It does work with a goto tho. I'll push the change, lmk if you prefer it this way or not

hensldm: I am fine with the gotos

hensldm (approved):

hensldm:
Just noting there may be a little conflict with #1871 , though it would also be nice if this could use those macros. If this one gets in first though I don't mind going through and updating my PR.

PR #1848, opened 2025-12-16
Bombers Part 2: Guard

JordanLongstaff:
There are a few fields that are still named `unk_***` because they are either never modified, never read, or neither. I might convert them to `pad` fields if asked.

This PR also renames a `WEEKEVENTREG`.


src/overlays/actors/ovl_En_Bombers2/z_en_bombers2.c
```diff
@@ -144,17 +150,17 @@ void EnBombers2_ChangeAnim(EnBombers2* this, s32 animIndex, f32 playSpeed) {
                      sAnimationModes[this->animIndex], -10.0f);
 }
 
-void func_80C04B40(EnBombers2* this) {
-    if (!this->unk_2AC) {
+void EnBombers2_SetupIdle(EnBombers2* this) {
```
AngheloAlf: The way we name Setup functions is by reusing the same name of the action that was set up.
So ideally this function would be called `EnBombers2_SetupAwaitTalk` (since it sets `EnBombers2_AwaitTalk`) or the other function would be named `EnBombers2_Idle`.
Same comment for the other setup functions

JordanLongstaff: I'm choosing the former rename in this case because, in accordance with #1847, the animation being set up is referred to as either "idle" or "frolic" and I don't want to create confusion.

isghj5 (approved):

AngheloAlf (approved):

PR #1849, opened 2025-12-16
Bombers Part 3: `EnBombers`

JordanLongstaff:
There are two situations covered by this actor type: Bombers just wandering around town outside of the hide-and-seek minigame (except Jim), and Bombers sitting in North Clock Town after being caught during the minigame.

More `WEEKEVENTREG` flags are also renamed in this PR.


include/z64save.h
```diff
@@ -1326,7 +1326,7 @@ typedef enum {
 #define WEEKEVENTREG_73_10 PACK_WEEKEVENTREG_FLAG(73, 0x10)
 
 #define WEEKEVENTREG_73_20 PACK_WEEKEVENTREG_FLAG(73, 0x20)
-#define WEEKEVENTREG_73_40 PACK_WEEKEVENTREG_FLAG(73, 0x40)
+#define WEEKEVENTREG_DEKU_LEARNED_WHERE_JIM_IS PACK_WEEKEVENTREG_FLAG(73, 0x40)
```
AngheloAlf: I may be wrong, but I don't see a hard restriction on player being deku for this flag to be set, so I think we could drop the `DEKU` part of the name.
Also, we should mention somewhere that Jim is a bomber, because most people may not remember who Jim is, I feel like most people refer to them as the bombers over individual names. So either add `BOMBER_JIM` in the name or mention it in a comment.

JordanLongstaff: This flag gets set in only one place, with a text ID that only gets seen in Deku form.
"Oh! Don't go telling Jim in North Clock Town I told you that!"
That message only gets displayed if Link is a Deku Scrub. I'm fairly certain it's intended specifically for that first cycle.

JordanLongstaff: To drive that point home, I'm now renaming four other `WEEKEVENTREG` flags that serve a similar purpose, except they're specific to each Bomber, so if one Bomber tells Deku Link where Jim is and then Link goes and talks to another Bomber, he'll say something different.

AngheloAlf (approved):

hensldm (approved):

PR #1860, opened 2026-02-17
Documentation pass on WaterBox code

mzxrules:
As I was attempting to document the newly found waterbox bug, I found a few mistakes and improvements to make in my old documentation. 

I've implemented a similar pr in Ocarina of Time, and would like there to be a consensus reached between both games before merging.

The sister PR can be found here:

https://github.com/zeldaret/oot/pull/2695


include/z64bgcheck.h
```diff
@@ -30,15 +30,19 @@ struct DynaPolyActor;
...
-#define WATERBOX_PROPERTIES(bgCamIndex, lightIndex, room, setFlag19) \
+#define WATERBOX_PROPERTIES(bgCamIndex, lightIndex, roomIndex, isDisabled) \
     ((((bgCamIndex) & 0xFF) <<  0) | \
      (((lightIndex) & 0x1F) <<  8) | \
-     (((room)       & 0x3F) << 13) | \
-     (((setFlag19)  &    1) << 19))
+     (((roomIndex)  & 0x3F) << 13) | \
+     (((isDisabled) &    1) << 19))
+
+#define WATERBOX_LIGHT_INDEX_NONE 0x1F // Generates warning when built for debug and defaults to 0
```
AngheloAlf: What warning??

mzxrules: In Environment_EnableUnderwaterLights in z_kankyo.c, there is a PRINTF in Ocarina that generates an error message when `DEBUG_FEATURES` is enabled. It is unlikely this message has been removed in og source, but it currently does not appear in MM decomp

AngheloAlf: Ah, you mean a printf. I thought you meant a compilation warning.
I think we should word this part a bit different to avoid confusion, specially because that printf is not in this repo (yet?)

mzxrules:
I noticed some stuff missing related to CollisionPoly vertex flags, so I also ported that over here.

Dragorn421 (approved):

hensldm (approved):

PR #1868, opened 2026-04-13
Document a few symbols regarding the Ikana Canyon guard

JordanLongstaff:
Started as two `WEEKEVENTREG` flags, but also includes a function and a field in `EnGb2`, which is the actor used for the guard.

Is "guard" the right term? He does gatekeep your entry into the canyon.


AngheloAlf (approved):

isghj5:
More of a "guide" maybe? I might need to read his lines again. He does help you, trying to first tell you where to find a mask to talk to the garo and then, if he's the same guy in secret shrine, gives you an item after "helping" the ghosts of the minibosses.

JordanLongstaff:
I wouldn't consider him a "guide," though. I wanted a better word to describe him than "hermit," but maybe that's good enough?

src/overlays/actors/ovl_En_Gb2/z_en_gb2.c
hensldm: The temp can actually be removed with the use of a switch, which is cleaner in my opinion.
```c
            this->textId = func_80B0FB24(this);
            Message_StartTextbox(play, this->textId, &this->actor);
            switch (this->textId) {
                case 0x14E7:
                case 0x14E9:
                case 0x14EC:
                case 0x14F0:
                    CutsceneManager_Stop(this->csIdList[this->csIdIndex]);
                    this->csIdIndex = 1;
                    CutsceneManager_Queue(this->csIdList[this->csIdIndex]);
                    this->actionFunc = func_80B10DAC;
                    break;
                
                default:
                    break;

            }
```

hensldm (commented):
One suggestion to remove a temp.

hensldm (approved):

PR #1871, opened 2026-04-18
String macros

hensldm:
Adds some string helper macros:
PRINTF: which is designed like OoT's though are also designed to keep strings around when versions are before n64-us since the Japanese versions seem to keep strings around.
T: Just like OoT's

Also went through all the Japanese strings currently in the code base (except loadfragment since those are just comments at the moment) to use the T macro. I did skip the debug things (z_select.c and z_kaleido_debug) since they are large.

I also edited stackcheck a little bit to use its own PRINTF function since its Japanese versions use osSyncPrintf instead of just leaving the strings behind.


src/boot/libu64/stackcheck.c
```diff
@@ -104,10 +111,10 @@ StackStatus StackCheck_GetState(StackEntry* entry) {
...
-#if MM_VERSION < N64_US
-    osSyncPrintf("head=%08x tail=%08x last=%08x used=%08x free=%08x [%s]\n", entry->head, entry->tail, last, used, free,
-                 (entry->name != NULL) ? entry->name : "(null)");
-#else
+    STACKCHECK_PRINTF("head=%08x tail=%08x last=%08x used=%08x free=%08x [%s]\n", entry->head, entry->tail, last, used,
+                      free, (entry->name != NULL) ? entry->name : "(null)");
+
+#if MM_VERSION >= N64_US
     (void)"(null)";
 #endif
```
AngheloAlf, line 119: This `"(null)"` seems like the result of getting rid of that printf.
Could the macro be made to include this somehow?

hensldm: Yeah, though it does require specifically having the correct number of args in the macro (see below), which seems fine since this is an isolated macro. Though would be curious on your thoughts.

```
#if MM_VERSION < N64_US || DEBUG_FEATURES
#define STACKCHECK_PRINTF osSyncPrintf
#elif IDO_PRINTF_WORKAROUND
#define STACKCHECK_PRINTF(format, arg0, arg1, arg2, arg3, arg4, arg5) arg5
#else
#define STACKCHECK_PRINTF(format, ...) (void)0
#endif
```

AngheloAlf: Ohh, I see
That sucks.
I guess it is fine to just leave it as it is now

AngheloAlf (approved):

Dragorn421 (approved):

PR #1872, opened 2026-04-22
[n64-jp-1.1] Boss03

hensldm:


src/overlays/actors/ovl_Boss_03/z_boss_03.c
```diff
@@ -122,6 +123,7 @@ GyorgEffect sGyorgEffects[GYORG_EFFECT_COUNT];
 Boss03* sGyorgBossInstance;
 
 void Boss03_PlayUnderwaterSfx(Vec3f* projectedPos, u16 sfxId) {
+    PRINTF(VT_FGCOL(CYAN) T(" ☆ ボス１発生 ☆ \n", " ☆ Boss 1 appears ☆ \n") VT_RST);
```
AngheloAlf: This would make more sense in the init function wouldn't it?
It could help with that `if (1)`

src/overlays/actors/ovl_Boss_03/z_boss_03.c
```diff
@@ -131,6 +133,8 @@ void Boss03_SpawnEffectWetSpot(PlayState* play, Vec3f* pos) {
     s16 i;
     GyorgEffect* eff = play->specialEffects;
 
+    PRINTF("FISH COUNT %d\n");
```
AngheloAlf: And by that logic this could be moved to the update function.
There we got access to the actor instance, so it would be nice to pass `this->numSpawnedSmallFish` too, to avoid issues with the missing parameter

src/overlays/actors/ovl_Boss_03/z_boss_03.c
AngheloAlf, line 645: This if else looks awkward, what do you think about changing them to early returns instead?
```c
        if ((this->waterHeight < player->actor.world.pos.y) && (player->actor.bgCheckFlags & BGCHECKFLAG_GROUND)) {
            Boss03_SetupPrepareCharge(this, play);
            return;
        }

#if MM_VERSION >= N64_US
        if ((player->transformation == PLAYER_FORM_GORON) || (player->transformation == PLAYER_FORM_DEKU)) {
            return;
        }
#endif

        if (KREG(70) == 0) {
            Boss03_SetupChasePlayer(this, play);
        } else if (this->numSpawnedSmallFish <= 0) {
            Boss03_SetupChasePlayer(this, play);
        }
```

src/overlays/actors/ovl_Boss_03/z_boss_03.c
hensldm: Yeah. The floats came after the strings in the rodata, so thought I couldn't move them, but turns out they could be moved with the `if (1)`s

AngheloAlf (approved):

PR #1874, opened 2026-04-22
[n64-jp-1.1] libu64/gfxprint

hensldm:


src/boot/libu64/gfxprint.c
```diff
@@ -8,6 +9,14 @@
 #define GFXP_FLAG_ENLARGE (1 << 6)
 #define GFXP_FLAG_OPEN (1 << 7)
 
+#if MM_VERSION < N64_US || DEBUG_FEATURES
+#define GFXPRINT_PRINTF osSyncPrintf
+#elif IDO_PRINTF_WORKAROUND
+#define GFXPRINT_PRINTF(args) (void)0
+#else
+#define GFXPRINT_PRINTF(format, ...) (void)0
+#endif
```
AngheloAlf, line 18: This seems to be becoming a pattern. Maybe we should move this to a header?

hensldm: I wouldn't be opposed, though will admit not quite sure what to call it so as to not conflict with the original PRINTF that keeps the strings, but not the osSyncPrintf call. Would you have any suggestion?

AngheloAlf (approved):
I don't have any.
I guess we can worry about it later

PR #1876, opened 2026-04-22
[n64-jp-1.1] irqmgr

hensldm:


src/boot/irqmgr.c
```diff
@@ -53,6 +53,30 @@ s32 sIrqMgrRetraceCount = 0;
...
 
+#if DEBUG_FEATURES
+#define IRQMGR_PRINTF osSyncPrintf
+#elif IDO_PRINTF_WORKAROUND
+#if MM_VERSION < N64_US
+#define IRQMGR_PRINTF(args) \
+    do {                    \
+        if (1)              \
+            (void)args;     \
+    } while (0)
```
AngheloAlf: huh??
that's ugly, but whatever

AngheloAlf (approved):

src/boot/irqmgr.c
hensldm: So working on other files, it seems a scope around the leftover strings is neccessary in many other places. So I've gone ahead and edited the print macro, and was able to remove the specific IRQMGR one. Though I still needed to keep an extra 3 `if (1) {}` at the end of `IrqMgr_ThreadEntry`.

AngheloAlf (approved):
Nice catch with the do while inside the PRINTF macro

PR #1878, opened 2026-04-22
[n64-jp-1.1] yaz0

hensldm:


src/boot/yaz0.c
```diff
@@ -83,6 +85,8 @@ s32 Yaz0_DecompressImpl(u8* src, u8* dst) {
     u32 magic = ((Yaz0Header*)src)->magic;
 
     if (magic != YAZ0_MAGIC) {
+        PRINTF(T("slidstart_szs IDが違います (%02x %02x %02x %02x %08x)\n",
+                 "The slidstart_szs ID is incorrect (%02x %02x %02x %02x %08x)\n"));
```
AngheloAlf:
```suggestion
        PRINTF(T("slidstart_szs IDが違います (%02x %02x %02x %02x %08x)\n",
                 "The slidstart_szs ID is incorrect (%02x %02x %02x %02x %08x)\n"), src[0], src[1], src[2], src[3], magic);
```

src/boot/yaz0.c
```diff
@@ -11,7 +11,9 @@ uintptr_t sYaz0CurRomStart;
 size_t sYaz0CurSize;
 u8* sYaz0MaxPtr;
 
+#if MM_VERSION >= N64_US
 void* gYaz0DecompressDstEnd;
+#endif
```
AngheloAlf, line 16: Could you add this `#if` on the header too?

AngheloAlf (approved):

PR #1882, opened 2026-06-05
Fixed Dockerfile.

Zeri-Ai:
Dockerfile updated per the file on https://practicerom.com/public/packages/debian/howto.txt and Debian/Ubuntu's apt-key command being deprecated.


Dockerfile
```diff
@@ -27,7 +22,13 @@ RUN apt-get update && apt-get install -y \
     clang-tidy-14 \
     clang-format-14 \
     libpng-dev \
-    practicerom-dev
+    sudo \
+    curl
+
+# Add source for practicerom-dev install
+
+RUN sudo sh -c "$(curl -fsSL https://raw.githubusercontent.com/PracticeROM/packages/HEAD/scripts/install-debian_amd64.sh)" && \
```
hensldm: Looks like the instructions to download and install the practice rom packages have changed again. (that URL provides a 404). (see https://github.com/PracticeROM/packages)

Zeri-Ai: Updated!

Dockerfile
```diff
@@ -27,7 +22,13 @@ RUN apt-get update && apt-get install -y \
...
     clang-format-14 \
     libpng-dev \
-    practicerom-dev
+    curl
+
+# Add source for practicerom-dev install
+
+RUN curl -O https://practicerom.com/public/packages/debian/dists/stable/practicerom-repository_latest_$(dpkg --print-architecture).deb && \
+dpkg -i practicerom-repository_latest_$(dpkg --print-architecture).deb && \
+apt update
```
hensldm: This needs to back up before the Install Required Dependencies.
Basically all you need to do is replace the
```sh
curl -o /usr/share/keyrings/practicerom-archive-keyring.gpg https://practicerom.com/public/packages/debian/practicerom-archive-keyring.gpg && \
    echo 'deb [arch=all,amd64 signed-by=/usr/share/keyrings/practicerom-archive-keyring.gpg] http://practicerom.com/public/packages/debian unstable main' > /etc/apt/sources.list.d/practicerom.list
```
with the new
```sh
curl -O https://practicerom.com/public/packages/debian/dists/stable/practicerom-repository_latest_$(dpkg --print-architecture).deb && \
dpkg -i practicerom-repository_latest_$(dpkg --print-architecture).deb
```

You will still need to sudo apt install practicerom-dev like it was doing before.

Also could you add a comment with a link to the PracticeRom packages Readme for future reference of the instructions.

Zeri-Ai: Thank you for your guidance!
I modified the Dockerfile per your indications.

Dockerfile
```diff
@@ -27,6 +29,7 @@ RUN apt-get update && apt-get install -y \
     clang-tidy-14 \
     clang-format-14 \
     libpng-dev \
+    sudo \
```
hensldm: Any reason to really add sudo? Not sure it is necessary if all you are doing is fixing the practicerom download.

Zeri-Ai: Back when I made the original pull request, since the installing script required sudo to run I added it to the dockerfile.
Since the installation changed once again to how it was, it might not be necessary anymore. I can update the branch and remove sudo if it's not necessary but it would have to be in at least 20 hours from now.

hensldm: I would say remove it.

AngheloAlf: It shouldn't be required, so yeah remove it

AngheloAlf (commented):
We are trying to get everybody who has contributed / wants to contribute to MM decomp to agree to licensing the repo before merging PRs from people who have not yet agreed to it. It is better explained here: https://github.com/louist103/mm-licensing/issues/1.

Zeri-Ai:
Removed sudo from the Dockerfile (as it is indeed not required) and agreed to the aforementioned license.

hensldm (approved):

AngheloAlf (approved):

PR #1883, opened 2026-07-21
z_player `Player_Action_DismountLadder`

djevangelia:
Function is identical with OoT, PR here https://github.com/zeldaret/oot/pull/2771
Some small changes to make them match up better.
Didn't bring the bug comments as I haven't tested that they are still valid.


hensldm (approved):
Something else to take into consideration: We are trying to get everybody who has contributed / wants to contribute to MM decomp to agree to licensing the repo before merging PRs from people who have not yet agreed to it. It is better explained here: https://github.com/louist103/mm-licensing/issues/1.

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -16420,8 +16421,9 @@ void Player_Action_50(Player* this, PlayState* play) {
                             this->skelAnime.prevTransl = this->ageProperties->unk_44;
                         }
 
-                        func_8083DCC4(this, this->ageProperties->unk_CC[this->av2.actionVar2], play);
-                        this->av2.actionVar2 = 1;
+                        Player_SetupDismountLadder(
+                            this, this->ageProperties->dismountLadderDownAnim[this->av2.actionVar2], play);
+                        this->av2.dismountDown = true;
```
AngheloAlf, line 16426: I see this action var is set to true here, but I don't see anywhere where it is set to false, so maybe there's a missing reference for this enum that wasn't updated yet?

AngheloAlf: Actually, this doesn't make sense.
The rest of the function treats this actionvar as a plain number instead of a boolean, for example the code is using it as an index in the line right above. Also `Player_SetupDismountLadder` does other negative checks on this variable, which doesn't make sense for a bool.

djevangelia: - In `Player_Action_50` (`Player_Action_Climbing`), actionVar2 is used initially as a flag for whether the start climbing animation is done (the negative values are to signal starting from below/above and wall/ladder).
- It then switches function to flag that the left foot (in OoT, haven't verified it in MM but should be same) is above right for get animation purposes.
-  `Player_SetupDismountLadder` sets it to zero through `Player_SetupAction` which always sets av1 and av2 to zero.
- actionVar2 is then set to true if player is dismounting down.
- `dismountDown` is thus how it is used in `Player_Action_DismountLadder`, the `Player_Action_50` action returns here and is irrelevant now.

djevangelia: Is there still anything unclear? I'm happy to answer further questions.

AngheloAlf (approved):
Sorry for the delay

PR #1885, opened 2026-08-01
Migrate CI to Github Actions

AngheloAlf:
Changes the CI system from Jenkins to Github Actions (GHA) like we did for OoT

This is an adaptation of the Jenkinsfile we have and the solution made by @Dragorn421 for OoT

Relevant OoT PRs:
- https://github.com/zeldaret/oot/pull/2740
- https://github.com/zeldaret/oot/pull/2742
- https://github.com/zeldaret/oot/pull/2754

New features include:
- Not needing to selfhost a runner
  - Required private stuff for building is hosted on a private repo
- Building multiple versions in parallel
  - Our old jenkinsfile wasn't building `n64-jp-1.1`. Even if it doesn't match right now, I think it is important checking a PR doesn't completely break it.
- Run bss fixer in CI
- Upload mapfiles as build artifacts
- Split compiler download as a two step process to be able to reuse the IDO compiler included in the docker image.

This specific approach to handle GHA for decomp projects was adapted from the one used by the GC/Wii community. It is documented here https://github.com/encounter/dtk-template/blob/main/docs/github_actions.md There's a writeup about this adaptation for N64 projects [here](https://github.com/AngheloAlf/drmario64/pull/19).


Dragorn421 (approved):
obama gives obama a medal meme

hensldm:
I think we should add something similar to https://github.com/zeldaret/oot/pull/2788 as well so these actions don't run on forks.

Also a little more out of scope but since you are messing with bss stuff, perhaps we could also bring in the new bss script tiebreaking algorithm fix (https://github.com/zeldaret/oot/pull/2779). Understandable if that should be its own PR.

AngheloAlf:
I included both PRs, thanks for the reminder

hensldm (approved):

PR #1886, opened 2026-08-09
New assets system

Dragorn421:
Currently WIP but OK!

Help welcome on implementing the MM-specific data structures:
- `AnimatedMaterial[]` (partially started for matching reasons in `tools/assets/extract/extase_oot64/animated_materials.py`)
- `KeyFrameAnimation`
- `KeyFrameSkel`
- various maps stuff: animated material list, actor cutscene list, map data, cutscene script list, actor cutscene cam list, map data chests, light list

Also need a csdis.py for MM like OoT has for disassembling cutscene data.


include/array_count.h
```diff
@@ -0,0 +1,8 @@
+#ifndef ARRAY_COUNT_H
+#define ARRAY_COUNT_H
+
+#define ARRAY_COUNT(arr) (s32)(sizeof(arr) / sizeof(arr[0]))
+#define ARRAY_COUNTU(arr) (u32)(sizeof(arr) / sizeof(arr[0]))
+#define ARRAY_COUNT_2D(arr) (s32)(sizeof(arr) / sizeof(arr[0][0]))
```
AngheloAlf, line 6: I see you added this file, but I don't see you removed these macros from `macros.h`

I guess this would touch up way too many files, so it is fine for now. Alternatively you could include this file in `macros.h`

src/overlays/actors/ovl_En_Bigslime/z_en_bigslime.c
```diff
@@ -130,22 +130,22 @@ void EnBigslime_DrawShatteringEffects(EnBigslime* this, PlayState* play);
 
 // Reference data: used to store the original vertices
 static Vtx sBigslimeStaticVtx[BIGSLIME_NUM_VTX] = {
-#include "assets/overlays/ovl_En_Bigslime/sBigslimeStaticVtx.vtx.inc"
+#include "assets/overlays/ovl_En_Bigslime/sBigslimeStaticVtx.inc.c"
```
AngheloAlf, line 133: I kinda prefer the `.inc` extension over the `.c` extension for these generated assets.
They aren't C files, they can't be compiled on their own, they have to be included this way to have meaning, so an extension reflecting this sounds better to me

Dragorn421: Hmm but the .c extension has the advantage that IDEs will at least syntax-highlight the contents like it's C which is mostly correct

AngheloAlf: IDEs usually let you configure this, so people can decide to opt in instead of forcing everyone into this.
For example, you can add this to your `.vscode/settings.json`
```json
{
    "files.associations": {
        "*.inc": "c"
    }
}
```

but whatever, i guess it is fine

PR #1893, opened 2026-08-16
gSetupActorDLs

hensldm:
Name D_801AEF88-> gActorSetupXluDL and D_801AEFA0 -> gActorSetupOpaDL.


src/overlays/actors/ovl_En_Ishi/z_en_ishi.c
```diff
@@ -779,7 +779,7 @@ void EnIshi_DrawGameplayKeepBoulder(EnIshi* this, PlayState* play) {
 
         Gfx_SetupDL25_Opa(play->state.gfxCtx);
 
-        gSPSegment(POLY_OPA_DISP++, 0x08, D_801AEFA0); // empty displaylist
+        gSPSegment(POLY_OPA_DISP++, 0x08, gActorEmptyDL); // empty displaylist
```
Dragorn421:
```suggestion
        gSPSegment(POLY_OPA_DISP++, 0x08, gActorEmptyDL);
```
while you're at it

src/overlays/actors/ovl_En_Po_Composer/z_en_po_composer.c
```diff
@@ -691,7 +691,7 @@ void EnPoComposer_Draw(Actor* thisx, PlayState* play) {
         gSPSegment(POLY_OPA_DISP++, 0x0B,
                    Gfx_EnvColor(play->state.gfxCtx, clothingColor2->r, clothingColor2->g, clothingColor2->b,
                                 this->mainAlpha.a));
-        gSPSegment(POLY_OPA_DISP++, 0x0C, D_801AEFA0); // empty DL
+        gSPSegment(POLY_OPA_DISP++, 0x0C, gActorEmptyDL); // empty DL
```
Dragorn421:
```suggestion
        gSPSegment(POLY_OPA_DISP++, 0x0C, gActorEmptyDL);
```

src/overlays/actors/ovl_En_Po_Sisters/z_en_po_sisters.c
```diff
@@ -1207,7 +1207,7 @@ void EnPoSisters_Draw(Actor* thisx, PlayState* play) {
 
     if ((this->color.a == 255) || (this->color.a == 0)) {
         gDPSetEnvColor(POLY_OPA_DISP++, this->color.r, this->color.g, this->color.b, this->color.a);
-        gSPSegment(POLY_OPA_DISP++, 0x09, D_801AEFA0); // Empty DL
+        gSPSegment(POLY_OPA_DISP++, 0x09, gActorEmptyDL); // Empty DL
```
Dragorn421:
```suggestion
        gSPSegment(POLY_OPA_DISP++, 0x09, gActorEmptyDL);
```

AngheloAlf (approved):
LGTM (as far as you apply dragorn's suggestions)

mzxrules:
This is not a good rename, all things considered.

In Zelda64, you have some cases where display lists are either rendered opaque or translucent, such as the Kokiri, the sages, the poes and poe sisters, etc. `z_actor.c` provides some of the code to handle drawing these actors, which includes the D_801AEF88 and D_801AEFA0 (which is the one being renamed) display lists. 

The purpose of these display list, at least to my very limited understanding, is to configure the color combiner for rendering for the translucent display list and opaque display list respectively. It's just that for the opaque version, there is nothing to update so it is a simple SpEndDisplayList() command.

The pattern is nearly the same in Ocarina of Time, where MM's `D_801AEF88` display list is identical to OoT's `D_80116280`. The difference is that in OoT, the opaque display list is stored within the translucent, at `D_80116280[2]`, whereas MM refactored them to no longer be coupled together.

Incidently, the two calls to DPSetRenderMode should be modified to use OoT's simplified arguments.

Dragorn421:
yeah; what names would you suggest?

hensldm:
Hmm, what about `gActorSetupOpaDL` and `gActorSetupXluDL`? For what it is worth, OoT has them separate as well with `D_80116280` (xlu) and `D_80115FF0` (opa). Though, they never used `D_80115FF0` instead doing `D_80116280[2]` as you mentioned.

Dragorn421 (approved):

PR #1894, opened 2026-08-16
Xml fixes

hensldm:
Quite a few xml fixes:
1. More TlutOffsets
2. Dependencies on gameplay_keep
3. Identified some more tluts in object_in
4. Named some gameplay_keep textures from OoT
5. Merged gEffFireMask1Tex and gEffFireMask2Tex into one and just named it gEffFireMaskTex (similar to what OoT has)

### DIFF
--- a/assets/xml/code/eff_footmark.xml
+++ b/assets/xml/code/eff_footmark.xml
@@ -1,4 +1,7 @@
 <Root>
+    <!-- Dependencies -->
+    <ExternalFile XmlPath="objects/gameplay_keep.xml" OutPath="assets/objects/gameplay_keep/"/>
+
     <File Name="code" BaseAddress="0x800A5AC0" RangeStart="0x116750" RangeEnd="0x1167E0" OutName="eff_footmark">
         <Array Name="gEffFootprintVtx" Count="3" Offset="0x116750">
             <Vtx/>
--- a/assets/xml/objects/gameplay_keep.xml
+++ b/assets/xml/objects/gameplay_keep.xml
@@ -1,5 +1,7 @@
 ﻿<Root>
+    <!-- Dependencies -->
     <ExternalFile XmlPath="misc/link_animetion.xml" OutPath="assets/misc/link_animetion/"/>
+
     <File Name="gameplay_keep" Segment="4">
         <!-- Bottles -->
         <Texture Name="gBottleTex" OutName="bottle" Format="i4" Width="16" Height="16" Offset="0x0" />
--- a/assets/xml/objects/gameplay_keep.xml
+++ b/assets/xml/objects/gameplay_keep.xml
@@ -1093,7 +1093,7 @@
         <Texture Name="gEffHitmarkRed8Tex" OutName="eff_hitmark_red_8" Format="i4" Width="32" Height="24" Offset="0x35590" />
         <DList Name="gEffHitmarkDL" Offset="0x35710" />
         <Texture Name="gameplay_keep_Tex_0357B0" OutName="tex_0357B0" Format="ia8" Width="32" Height="64" Offset="0x357B0" />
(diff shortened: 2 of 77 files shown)


assets/xml/objects/gameplay_keep.xml
```diff
@@ -1426,8 +1428,7 @@
 
         <DList Name="gEffFire1DL" Offset="0x7D590" />
         <DList Name="gEffFire2DL" Offset="0x7D650" />
-        <Texture Name="gEffFireMask1Tex" OutName="eff_fire_mask_1" Format="i4" Width="32" Height="64" Offset="0x7D710" />
-        <Texture Name="gEffFireMask2Tex" OutName="eff_fire_mask_2" Format="i4" Width="32" Height="64" Offset="0x7DB10" />
+        <Texture Name="gEffFireMaskTex" OutName="eff_fire_mask" Format="i4" Width="32" Height="128" Offset="0x7D710" />
         <Texture Name="gEffFireTex" OutName="eff_fire" Format="i8" Width="32" Height="64" Offset="0x7DF10" />
```
Dragorn421, line 1432: oot currently names these two `gEffFire1Tex` and `gEffFire2Tex`
the names were changed for cc0 reasons (they are Elliptic's), idk if MM wants to care though since realistically no one is going to do the same cc0 stuff I did on oot for mm

assets/xml/code/eff_footmark.xml
```diff
@@ -1,4 +1,7 @@
 <Root>
+    <!-- Dependencies -->
+    <ExternalFile XmlPath="objects/gameplay_keep.xml" OutPath="assets/objects/gameplay_keep/"/>
```
AngheloAlf, line 3: iirc gameplay keep was kinda special cased to always be available to all xmls
https://github.com/zeldaret/mm/blob/a1c6253cd7d2627a1fe7876663b74cabc1a80692/tools/ZAPDConfigs/MM/Config.xml#L7

This will no longer be the case for the new asset extraction system? CC @Dragorn421

AngheloAlf (approved):

assets/xml/objects/gameplay_keep.xml
hensldm: I'd rather not worry about it now. If someone does end up doing it for MM they can deal with it then.

assets/xml/code/eff_footmark.xml
hensldm: From what I can tell, it is not but Dragorn can answer more definitively.

Dragorn421: no it will no longer be the case in new assets tm

the reason is that for multiprocessed extraction (wanted for speed reasons), you can't split extraction of dependent assets across processes
For example if eff_footmark references gkeep, eff_footmark needs access to the gkeep symbols and also provide back any information on gkeep to the gkeep extraction. So dependent assets (termed a "pool" of assets) need to be extracted in the same process
So for multiprocessing to be usable, everything cannot depend on gkeep as that would put everything in the same pool and force extraction to happen in a single process rather than several

Dragorn421 (approved):

assets/xml/code/eff_footmark.xml
AngheloAlf: Why does eff_footmark need to provide back information to gameplay_keep? As far as I remember there aren't any loops like this.
Also, you can just read the gameplay_keep xml without needing any info on the extracted assets, unless you are reading tluts from gameplay_keep maybe(?)

Dragorn421: In this case it doesn't because the gkeep xml is well made / was fixed recently but in general it allows checking resources are correctly identified (like texture format and dimensions) as well as discovering resources by cross files references

Eventually we could indeed consider dropping this pool logic partly (?, external tluts indeed need data access) and just assume the xmls are correct but for now I like it very much since it's very systematic and reveals errors

AngheloAlf (approved):

PR #1898, opened 2026-08-29
`gActorSetup{Opa,Xlu}DL` docs

Dragorn421:
fuller explanation here https://discord.com/channels/688807550715560050/873211000122384415/1538758944376492083
I left a lot of details out because I don't think random comments in the codebase should be n64 graphics tutorials unfortunately

also while researching I noticed one odd usage of gActorSetupOpaDL, the octoroks EnOkuta use it as a "empty dlist" rather than with opa semantics which suggests gActorSetupOpaDL may not have been the original name, but w/e

### DIFF
--- a/src/code/z_actor.c
+++ b/src/code/z_actor.c
@@ -4724,12 +4724,24 @@ void Npc_TrackPoint(Actor* actor, NpcInteractInfo* interactInfo, s16 presetIndex
                              rotLimits.rotateYaw);
 }
 
+/**
+ * A display list setting up for translucent drawing when a fading actor is partially faded,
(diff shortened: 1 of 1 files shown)


hensldm (approved):

AngheloAlf (approved):
> also while researching I noticed one odd usage of gActorSetupOpaDL, the octoroks EnOkuta use it as a "empty dlist" rather than with opa semantics which suggests gActorSetupOpaDL may not have been the original name, but w/e

Or maybe the dev just noticed they already had an empty dl and decided to reuse it even if it wasn't intended as one. We'll never know

PR #1899, opened 2026-08-30
Overwrite PRINTF instead of whole new macros

hensldm:
Instead of having multiple seperate PRINTF macros for the files that leave osSyncPrintf behind, I thought of an idea where where just overwrite the PRINTF macro.

### DIFF
--- a/src/boot/libu64/debug.c
+++ b/src/boot/libu64/debug.c
@@ -1,12 +1,9 @@
 #include "global.h"
 #include "fault.h"
 
-#if MM_VERSION < N64_US || DEBUG_FEATURES
-#define DEBUG_PRINTF osSyncPrintf
-#elif IDO_PRINTF_WORKAROUND
-#define DEBUG_PRINTF(args) (void)0
-#else
-#define DEBUG_PRINTF(format, ...) (void)0
+#if MM_VERSION < N64_US
+#undef PRINTF
+#define PRINTF osSyncPrintf
 #endif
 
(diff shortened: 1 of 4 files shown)


Dragorn421 (approved):

src/boot/libu64/debug.c
```diff
@@ -1,12 +1,9 @@
 #include "global.h"
 #include "fault.h"
 
-#if MM_VERSION < N64_US || DEBUG_FEATURES
-#define DEBUG_PRINTF osSyncPrintf
-#elif IDO_PRINTF_WORKAROUND
-#define DEBUG_PRINTF(args) (void)0
-#else
-#define DEBUG_PRINTF(format, ...) (void)0
+#if MM_VERSION < N64_US
```
AngheloAlf, line 4: I think it would be nice to have if debug features are enabled, but whatever is fine tbh
```suggestion
#if MM_VERSION < N64_US || DEBUG_FEATURES
```

AngheloAlf (approved):
I wonder if we should introduce something like a general `PRINTF2`/`PRINTF_ALT` since all those macros are pretty much the same, but we can discuss it later

src/boot/libu64/debug.c
hensldm: That is what is nice here. The original PRINTF macro already has it, so we actually don't need it here.

AngheloAlf: Oh, you are right, neat.

PR #1900, opened 2026-09-03
[n64-jp-1.1] code files

hensldm:
Various code files OK for n64-jp-1.1
Also implements setting the gBuildCreator and gBuildDate through defines like OoT to match src/boot/build.c

Note graph.c bss is too large, but this is due to bss reordering of `gGfxMasterDL`. This may be able to be addressed with fix_bss.py but the ordering is weird and not quite sure how to solve it.


src/code/padmgr.c
```diff
@@ -58,7 +59,7 @@ typedef enum {
 
 PadMgr* sPadMgrInstance = &gPadMgr;
 
-s32 sPadMgrRetraceCount = 0;
+s32 sRumbleUpdateCounter = 0;
```
mzxrules:
```suggestion
s32 sPadMgrRetraceCount = 0;
```

The older name is much clearer in purpose and timing.

Edit: Oh yea now I see. OoT/MM J update it in `PadMgr_UpdateRumble` which is only not run during PreNMI. Wonder if it matters too much, since the thread is shutting down anyway.

src/code/game.c
```diff
@@ -191,21 +202,23 @@ void GameState_Realloc(GameState* gameState, size_t size) {
     THA_Destroy(&gameState->tha);
     GameAlloc_Free(alloc, heapStart);
     GetFreeArena(&systemMaxFree, &bytesFree, &bytesAllocated);
-    size = ((systemMaxFree - sizeof(ArenaNode)) < size) ? 0 : size;
+    size = ((systemMaxFree - 0x10) < size) ? 0 : size;
     if (size == 0) {
-        size = systemMaxFree - sizeof(ArenaNode);
+        size = systemMaxFree - 0x10;
```
AngheloAlf: wait, really? that sucks
I wonder if we should split the `ArenaNode` struct or something.
Either way, could you add a macro define here to avoid the magic number?

src/code/padmgr.c
```diff
@@ -282,20 +283,22 @@ void PadMgr_UpdateRumble(void) {
...
     PadMgr_ReleaseSerialEventQueue(serialEventQueue);
 }
 
+#if MM_VERSION >= N64_US
 /**
  * Immediately stops rumble on all controllers
  */
 void PadMgr_RumbleStop(void) {
     OSMesgQueue* serialEventQueue = PadMgr_AcquireSerialEventQueue();
     s32 i;
```
AngheloAlf: I feel like we already have this discussion, but...
What do you think about putting the `#if version` inside the function definition? The function signature is the same.

```suggestion
/**
* Immediately stops rumble on all controllers
*/
void PadMgr_RumbleStop(void) {
#if MM_VERSION >= N64_US
    OSMesgQueue* serialEventQueue = PadMgr_AcquireSerialEventQueue();
    s32 i;
```

src/code/padmgr.c
```diff
@@ -583,18 +605,30 @@ void PadMgr_UpdateConnections(void) {
...
                     }
                     // Missing break required for matching
             }
-        } else if (sPadMgrInstance->ctrlrType[i] != PADMGR_CONT_NONE) {
-            // Plugged controller errored
-            sPadMgrInstance->ctrlrType[i] = PADMGR_CONT_NONE;
-            sPadMgrInstance->pakType[i] = CONT_PAK_NONE;
-            sPadMgrInstance->rumbleTimer[i] = 0xFF;
+        } else {
+            if (sPadMgrInstance->ctrlrType[i] != PADMGR_CONT_NONE) {
```
AngheloAlf: Does this need to break the `else if` to match?
```suggestion
        } else if (sPadMgrInstance->ctrlrType[i] != PADMGR_CONT_NONE) {
```

src/code/z_DLF.c
```diff
@@ -20,39 +21,90 @@ void Overlay_LoadGameState(GameStateOverlay* overlayEntry) {
...
+        (void*)(uintptr_t)((overlayEntry->unk_20 != NULL) ? (void*)((uintptr_t)overlayEntry->unk_20 -
+                                                                    (intptr_t)((uintptr_t)overlayEntry->vramStart -
+                                                                               (uintptr_t)overlayEntry->loadedRamAddr))
+                                                          : NULL);
+
+    overlayEntry->unk_24 =
+        (void*)(uintptr_t)((overlayEntry->unk_24 != NULL) ? (void*)((uintptr_t)overlayEntry->unk_24 -
+                                                                    (intptr_t)((uintptr_t)overlayEntry->vramStart -
+                                                                               (uintptr_t)overlayEntry->loadedRamAddr))
+                                                          : NULL);
```
AngheloAlf, line 64: we should probably macroify those conversions

AngheloAlf (approved):
Looks good.
I left some comments, but nothing that prevents me from approving

src/code/padmgr.c
hensldm: So looking at it again, I do agree `sPadMgrRetraceCount` is better. When comparing with OoT I didn't notice that they actually got incremented in different functions. In this case they moved it so that they could use it with `PadMgr_RumbleStop` which only stopped rumble on one controller at a time based on this value, so depending on the retrace.

src/code/game.c
hensldm: Do you have a suggestion for the name?

src/code/padmgr.c
hensldm: So the reason I didn't was the function fundamentally changed, so the function comment actually doesn't apply to both versions. Instead of stopping Rumble on all controllers, it only stops it one at a time based on the retrace count. If you would prefer them together I can still make that work, though we would probably want to extend the comment to explain the version difference.

src/code/game.c
AngheloAlf: First thing that comes to mind is `NODE_SIZE`. Maybe `OFFSET_`something.
Sorry for not having a better suggestion :sweat_smile:

src/code/padmgr.c
AngheloAlf: That's fair

I think in this case the function is somewhat similar enough that we can merge them together and adjust the comment like this
```
/**
 * Immediately stops controller rumbling.
 *
 * On N64_US onward this function stops rumbling on all controllers,
 * while on the older implementation it only handled a single controller
 * based on `sRumbleUpdateCounter`
 */
```
Or something among those lines.

I have to agree this is outside the scope of this PR tho, so we can leave this bit out and worry later

PR #1906, opened 2026-09-12
Another small misc Cleanup

hensldm:
See commits to see what has been cleanedup.

Will note a couple places that may be a little more controversial.

### DIFF
--- a/src/code/z_kaleido_setup.c
+++ b/src/code/z_kaleido_setup.c
@@ -60,8 +60,8 @@ void func_800F4A10(PlayState* play) {
             }
         }
 
-        if ((gSaveContext.save.saveInfo.playerData.owlActivationFlags >> 4) & 1) {
-            pauseCtx->cursorPoint[PAUSE_WORLD_MAP] = 4;
+        if ((gSaveContext.save.saveInfo.playerData.owlActivationFlags >> OWL_WARP_CLOCK_TOWN) & 1) {
+            pauseCtx->cursorPoint[PAUSE_WORLD_MAP] = OWL_WARP_CLOCK_TOWN;
         }
     }
 
--- a/src/code/z_kaleido_setup.c
+++ b/src/code/z_kaleido_setup.c
@@ -93,53 +93,33 @@ void KaleidoSetup_Update(PlayState* play) {
         if (msgCtx && msgCtx) {}
     }
 
-    if (IS_PAUSED(pauseCtx) || (play->gameOverCtx.state != GAMEOVER_INACTIVE)) {
+    if ((IS_PAUSED(pauseCtx) || (play->gameOverCtx.state != GAMEOVER_INACTIVE)) ||
+        ((play->transitionTrigger != TRANS_TRIGGER_OFF) || (play->transitionMode != TRANS_MODE_OFF)) ||
+        ((gSaveContext.save.cutsceneIndex >= 0xFFF0) || (gSaveContext.nextCutsceneIndex >= 0xFFF0)) ||
+        (Play_InCsMode(play) && ((msgCtx->msgMode == MSGMODE_NONE) || (msgCtx->currentTextId != 0xFF))) ||
+        (play->bButtonAmmoPlusOne >= 2) ||
+        ((gSaveContext.magicState == MAGIC_STATE_STEP_CAPACITY) || (gSaveContext.magicState == MAGIC_STATE_FILL)) ||
+        (CHECK_EVENTINF(EVENTINF_17) || (player->stateFlags1 & PLAYER_STATE1_20)) ||
+        (play->actorCtx.flags & ACTORCTX_FLAG_TELESCOPE_ON) || (play->actorCtx.flags & ACTORCTX_FLAG_PICTO_BOX_ON)) {
         return;
     }
 
-    if ((play->transitionTrigger != TRANS_TRIGGER_OFF) || (play->transitionMode != TRANS_MODE_OFF)) {
-        return;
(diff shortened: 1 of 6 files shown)


src/code/z_kaleido_setup.c
```diff
@@ -93,53 +93,33 @@ void KaleidoSetup_Update(PlayState* play) {
         if (msgCtx && msgCtx) {}
     }
 
-    if (IS_PAUSED(pauseCtx) || (play->gameOverCtx.state != GAMEOVER_INACTIVE)) {
+    if ((IS_PAUSED(pauseCtx) || (play->gameOverCtx.state != GAMEOVER_INACTIVE)) ||
```
hensldm, line 96: I personally like all the conditions in one place. It also removes the fake match of `!Play_InCsMode(play) || ((msgCtx->msgMode != MSGMODE_NONE) && (msgCtx->currentTextId == 0xFF))` not being an early return like the rest. Note I could also change that to an early return, but it would require an additional `if(1) {}` somewhere.

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -2989,20 +2991,18 @@ void Player_SetUpperAction(PlayState* play, Player* this, PlayerUpperActionFunc
     func_8082E00C(this);
 }
 
-#define GET_PLAYER_ANIM(group, type) ((PlayerAnimationHeader**)D_8085BE84)[group * PLAYER_ANIMTYPE_MAX + type]
-
 void Player_InitItemActionWithAnim(PlayState* play, Player* this, PlayerItemAction itemAction) {
     PlayerAnimationHeader* curAnim = this->skelAnime.animation;
-    PlayerAnimationHeader*(*iter)[PLAYER_ANIMTYPE_MAX] = (void*)&D_8085BE84[0][this->modelAnimType];
+    PlayerAnimationHeader** iter = &GET_PLAYER_ANIM(0, this->modelAnimType);
```
hensldm, line 2996: Personally I like this syntax more, though it does mean the `iter += PLAYER_ANIMTYPE_MAX` down below.

isghj5: That does seem weird but not enough for pushback imo, it's still understandable.

isghj5 (approved):

src/code/z_kaleido_setup.c
mzxrules: Man this condition is a mess to read.

This is a section where I think you really need comments breaking down what each check is accomplishing, and what the early return is ultimately doing, because I'm not sure if all the checks are written in a sensible way.

src/overlays/actors/ovl_En_Kusa2/z_en_kusa2.c
```diff
@@ -743,7 +743,7 @@ void func_80A5CD0C(EnKusa2* this) {
     s32 pad;
 
     for (i = 0; i < 2; i++) {
-        temp_s0 = Rand_Next() & 0xFFFF;
+        temp_s0 = Rand_Next() % 0x10000;
```
mzxrules, line 746: Aren't all of these & 0xFFFF actually casts?

hensldm: Yes they, good catch :). Unfortunately they all need the & 0xFFFF or % 0x10000 to match, not even explicit s16 casts work. With that in mind I like `& 0xFFFF` better for the casts so switched those back.

src/code/z_kaleido_setup.c
hensldm: Yeah, definitely worth more documentation, though that would be out of scope for just this clean up PR.

mzxrules (approved):

src/code/z_kaleido_setup.c
AngheloAlf: oof, this doesn't seem an improvement to me, it is pretty hard to read, specially with a random `&&` in the middle of this whole check.
I feel like it would be better to go back and actually try to change the other check as an early return with that additional `if (1) {}`

src/code/z_scene.c
```diff
@@ -137,21 +137,17 @@ void Object_LoadAll(ObjectContext* objectCtx) {
...
-    RomFile* fileTableEntry;
-
-    objectCtx->slots[slot].id = -id;
-    objectCtx->slots[slot].dmaReq.vromAddr = 0;
-
-    fileTableEntry = &gObjectTable[id];
-    vromSize = fileTableEntry->vromEnd - fileTableEntry->vromStart;
+    ObjectEntry* entry = &objectCtx->slots[slot];
+    RomFile* objectFile = &gObjectTable[id];
+    u32 size;
```
AngheloAlf:
```suggestion
    size_t size;
```

src/overlays/actors/ovl_player_actor/z_player.c
AngheloAlf: This kinda feels going backwards, back to pointer arithmetic again.
idk, this array is such a weird case

AngheloAlf (commented):
btw there are merge conflicts

src/code/z_kaleido_setup.c
hensldm: So I switched to separate conditions with the extra at if(1) `{}`.

src/overlays/actors/ovl_player_actor/z_player.c
hensldm: Since there was more push back, I reverted whole change. Tbh, I think if `D_8085BE84` is a 2d array, the `GET_PLAYER_ANIM` macro isn't really necessary and probably should be removed entirely. That is a bigger change so don't want to deal with that in this PR.

PR #1907, opened 2026-09-16
Enum for boss remains in blue warp

JordanLongstaff:
There were a lot of ones flying around. I wanted to get rid of them. I think this enum will add some clarity.


hensldm (approved):
Seems fine to me

src/overlays/actors/ovl_Door_Warp1/z_door_warp1.c
```diff
@@ -553,7 +561,8 @@ void func_808B9CE8(DoorWarp1* this, PlayState* play) {
     }
 
     if (!Actor_HasParent(&this->dyna.actor, play)) {
-        Actor_OfferGetItem(&this->dyna.actor, play, DoorWarp1_GetRemains(this, play) + (GI_REMAINS_ODOLWA - 1), 30.0f,
+        Actor_OfferGetItem(&this->dyna.actor, play,
+                           DoorWarp1_GetRemains(this, play) + (GI_REMAINS_ODOLWA - DOORWARP1_REMAINS_ODOLWA), 30.0f,
```
AngheloAlf: I don't think subtracting two different enums holds much meaning

JordanLongstaff: Would it help to use a macro? The meaning is that the difference is not zero.

hensldm: Instead of subtracting from the other enum, I would say subtract from the return of `DoorWarp1_GetRemains`.
Do something like `(DoorWarp1_GetRemains(this, play) - DOORWARP1_REMAINS_ODOLWA) + GI_REMAINS_ODOLWA`. The idea here is that the DoorWarp1 enum needs to be offset back to zero.
Personally I don't like the macro idea.

AngheloAlf (approved):

src/overlays/actors/ovl_Door_Warp1/z_door_warp1.c
```diff
@@ -510,11 +518,11 @@ void func_808B98A8(DoorWarp1* this, PlayState* play) {
 
 void func_808B9B30(DoorWarp1* this, PlayState* play) {
     if (fabsf(this->dyna.actor.xzDistToPlayer) >= 60.0f) {
-        if (DoorWarp1_GetRemains(this, play) != 0) {
+        if (DoorWarp1_GetRemains(this, play) != DOORWARP1_REMAINS_NONE) {
             this->unk_1A0 = (DmHina*)Actor_SpawnAsChild(&play->actorCtx, &this->dyna.actor, play, ACTOR_DM_HINA,
                                                         this->dyna.actor.world.pos.x, this->dyna.actor.world.pos.y,
                                                         this->dyna.actor.world.pos.z, 0, 0, 0,
-                                                        DoorWarp1_GetRemains(this, play) - 1);
+                                                        DoorWarp1_GetRemains(this, play) - DOORWARP1_REMAINS_ODOLWA);
```
hensldm: Not sure why I didn't think of it until now, but we should add a `DmHina` enum as well, then do something like 
`(DoorWarp1_GetRemains(this, play) - DOORWARP1_REMAINS_ODOLWA) + DMHINA_REMAINS_ODOLWA'. `DMHINA_REMAINS_ODOLWA` would just be zero, however it more explicitly show what is going on here and mirror the other usages of `DoorWarp1_GetRemains`

hensldm (changes requested):

hensldm (approved):

mzxrules (approved):

PR #1908, opened 2026-09-19
OBJECT_UNSET_0 -> OBJECT_INVALID

hensldm:
Rename `OBJECT_UNSET_0`  to `OBJECT_INVALID` like OoT, since it is used the same way.


mzxrules:
I question whether `OBJECT_INVALID` is appropriate to use in all contexts for both games.

As I understand it, the biggest reason we use `OBJECT_INVALID` over `OBJECT_NONE` is because for an actor to spawn, an `ActorProfile`'s object dependency MUST be assigned a slot in the object exchange, even if that actor doesn't require a dependency. There are a couple of places where the game would just break if objectId of 0 was allowed.

 However, there are these cases where `OBJECT_INVALID` is passed when a value of 0 is expected.

 * scene initialization (`SCENE_CMD_SPECIAL_FILES`).
 * GET_ITEM macro (only applicable to MM)

AngheloAlf (approved):

Dragorn421 (approved):

hensldm:
> I question whether `OBJECT_INVALID` is appropriate to use in all contexts for both games.
> 
> As I understand it, the biggest reason we use `OBJECT_INVALID` over `OBJECT_NONE` is because for an actor to spawn, an `ActorProfile`'s object dependency MUST be assigned a slot in the object exchange, even if that actor doesn't require a dependency. There are a couple of places where the game would just break if objectId of 0 was allowed.
> 
> However, there are these cases where `OBJECT_INVALID` is passed when a value of 0 is expected.
> 
>     * scene initialization (`SCENE_CMD_SPECIAL_FILES`).
> 
>     * GET_ITEM macro (only applicable to MM)

OoT uses the same GET_ITEM macro, it just has a single GET_ITEM_NONE that uses `OBJECT_INVALID`, while MM needs separate entries for different text ids.
Fwiw, I wouldn't be opposed to `OBJECT_NONE` used in all cases, that seems like it would fit as well.

Dragorn421:
I think `OBJECT_INVALID` is perfectly fine to use as is to indicate "none"

AngheloAlf:
I also think `OBJECT_INVALID` is okay for those contexts too

mzxrules (approved):
I don't necessarily agree, but I won't block.

